PR #124, opened 2021-10-23
Change: support more architectures in Dockerfile

zv0n:
Hi, I wanted to run AudioBookShelf on my raspberry pi, but since `jrottenberg/ffmpeg` only offers x64 images it wasn't possible. 

I've changed the Dockerfile to install `ffmpeg` rather than copy it from an existing image, this way the image can be built for any platform that is supported by `node:12-alpine` (which is a lot).

I've tested this and the image is running fine on my RPi4. The only problem I've encountered was with ziplib not being available during server build phase, but the same problem is present when building the x64 version, so I guess it's fine ¯\\_(ツ)_/¯

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,15 @@
-### STAGE 0: FFMPEG ###
-FROM jrottenberg/ffmpeg:4.1-alpine AS ffmpeg
-
-### STAGE 1: Build client ###
+### STAGE 0: Build client ###
 FROM node:12-alpine AS build
 WORKDIR /client
 COPY /client /client
 RUN npm install
 RUN npm run generate
 
-### STAGE 2: Build server ###
+### STAGE 1: Build server ###
 FROM node:12-alpine
+RUN apk update && apk add --no-cache --update ffmpeg
 ENV NODE_ENV=production
 COPY --from=build /client/dist /client/dist
-COPY --from=ffmpeg / /
 COPY index.js index.js
 COPY package.json package.json
 COPY server server


advplyr:
Thanks for sharing that great article, I was able to build for arm64. I think it will run okay even if node-zipfile doesn't work.

You can see both builds in docker hub: https://hub.docker.com/repository/docker/advplyr/audiobookshelf/tags

According to the article, you just need to change the docker run command to include the sha256 value. I haven't tested this yet, hoping that you could do the honors.

```
docker run --rm -d -p 13378:80 \
  -v /your/audiobooks:/audiobooks -v /path/to/metadata:/metadata -v /path/to/config:/config \
  advplyr/audiobookshelf:latest@sha256:0ef990b5b67e102c6515ef126e9065f900df1da29edf8aa2340efff57af2fa53
```

The only library using node-zipfile is [epub.js](https://github.com/futurepress/epub.js/), and seeing as most users are only interested in serving audiobooks then this is no problem. May need to set an env variable or something to disable epub for arm64 images.

advplyr:
Read back through your comments and now I get the impression that you weren't having issues with node-zipfile once you built the docker image for arm64. Hopefully that is the case! And I didn't realize you already had this up and running on your rpi. That is awesome.
I would like to update the readme and install guide once this is confirmed good to go.

p1xelll:
I just launched the official image and everything works. I ran it using docker-compose and didn't use the sha256 value.

advplyr:
Nice! Is it pulling the correct image based on your current architecture?
I didn't see that in the docs or that article.

p1xelll:
Pulls the correct image and architecture.

zv0n:
I've just tried the official image and everything seems to be working fine, I tried listening to a book and downloading a zipped book and both are fine. Don't know if there's anything else that needs testing. I started my container like this:
`docker run -d -p 8011:80 -v $audiobooks_path:/audiobooks -v $config_path:/config -v $metadata_path:/metadata --name audiobookshelf advplyr/audiobookshelf:latest`
I think that docker pulls the same architecture it's running on if available, so the sha256 doesn't seem necessary. 

Thanks for incorporating arm64 into your process! Much appreciated :)

advplyr:
If you have an epub file and can put it inside one your audiobook folders it should give you a Read button and you can see if you can open the ebook.

I guess if it pulls the correct image automatically then there is nothing to update in the docs.

p1xelll:
I added the epub and it opened fine.

p1xelll:
The last Docker push has no support for arm64. Would you add it or release version 1.6?

advplyr:
Just pushed the arm64 build

PR #190, opened 2021-11-21
Added support for audible metadata

keaganhilliard:
I figured that since I've already built one plugin for using Audible's API, I might as well help out here. Audible gives a bit more contextual information than the other APIs so I added the ability to map Series, Volume Number and Narrator to the match UI.

Let me know what you think.


advplyr:
This is great! I really appreciate your contribution, a lot of requests for this one.

I just released this as `v1.6.24`.

PR #192, opened 2021-11-23
Better handle books with multiple series

keaganhilliard:
Currently Audiobookshelf only supports books being part of a single series, not sure if that should change, but Audible returns more than one for a large number of books. This should hopefully help pick the series consistently.


advplyr:
Included in `v1.6.27`!

PR #218, opened 2021-12-04
Get more data from opf

igorkaldowski:
When there is html in the description on a opf file it will strip it rather than not returning. It also gets series and volume number from the opf.

### DIFF
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -179,7 +179,7 @@ class LibraryController {
       series = series.slice(startIndex, startIndex + payload.limit)
     }
 
-    payload.results = series
+    payload.results = sort(series).asc(s => s.name)
     res.json(payload)
   }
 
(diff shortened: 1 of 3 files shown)


Helico-Pter:
I was just thinking it would be great if 'series' could be read from the file.

advplyr:
Thanks, I had been meaning to eventually use the strip html package there.

Can you move this meta check https://github.com/advplyr/audiobookshelf/pull/218/commits/6e621ecce163247b395f7af4bcbc0ca6250c174a#diff-3ddb151b8ce6aa69c5d2527e033c0aac28a2c4873111ffb0948fa37bb951fbe1R59
into the parseOpfMetadataXML function instead of xmlToJson. xmlToJson might eventualy be used for something other than Opf files.

igorkaldowski:
Done. I also change the series page to sort alphabetically and on the series pages to sort by volume number as a number not alphabetically.

advplyr:
This is released in `v1.6.35`

PR #219, opened 2021-12-05
Add: Force re-scan library

igorkaldowski:
Added a force rescan button to the library settings page.

### DIFF
--- a/client/components/app/Appbar.vue
+++ b/client/components/app/Appbar.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="w-full h-16 bg-primary relative">
-    <div id="appbar" class="absolute top-0 bottom-0 left-0 w-full h-full px-2 md:px-6 py-1 z-40">
+    <div id="appbar" class="absolute top-0 bottom-0 left-0 w-full h-full px-2 md:px-6 py-1 z-50">
       <div class="flex h-full items-center">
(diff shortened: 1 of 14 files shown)


advplyr:
In testing I found some other bugs. There is another one which is the force rescan is always detecting an update when there isn't one. I'm looking into that now.
Also, I want to think of a way a different place for the force re-scan button. For large libraries over an SMB share the force re-scan could take a long time.

advplyr:
The `rescanAudiobook` method was assuming that the audiobook has updates. I changed that function to return null if no updates were made so they don't get recorded as updates.
Existing audio files were not getting re-scanned either. My understanding of the forceRescan flag is that it will probe every audio file again, which would pickup on changed ID3 tags for example. I updated the method to first re-scan existing audio files.

igorkaldowski:
I wanted it to also rescan other files like opf files in case of changes because the current scan ignores them.

igorkaldowski:
That's perfect. I mainly wanted it to rescan opf files for changes. 
One way to make it faster for large libraries would be to store the date of when the file was last scanned then compare it to the date modified before doing anything. 
The only thing I can think to add is a way to stop manually edited fields in books from being changed. Similar to how you can lock fields in plex.

advplyr:
This is released in `v1.6.36`. I added a note explaining the force re-scan under the libraries table until there is a better design for that button.
Agree on the last modified date on file should be checked, it should also be stored on the audiobook so it can be used as a sort.

PR #226, opened 2021-12-07
Added ASIN support in the title field

keaganhilliard:
Stop gap measure until the Audible API is back to full functionality, but it's probably worth it to keep in there for those times that the Audible search has a hard time finding the right book.


advplyr:
Good call. It appears the API is still down.
The unfortunate part is they are returning an empty result instead of an error code, so we can't detect the service is down.

advplyr:
This is released in `v1.6.38`. I also added the provider selection in the cover search and have the provider persisted in the browser so you don't need to select Audible every time.

PR #233, opened 2021-12-12
Added an endpoint to serve different cover sizes

keaganhilliard:
I've been playing around with the api and I thought it would be useful to be able to serve cover images at whatever size the client might want. This should accomplish that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "podcast": "^1.3.0",
     "read-chunk": "^3.1.0",
     "recursive-readdir-async": "^1.1.8",
+    "sharp": "^0.29.3",
     "socket.io": "^4.1.3",
     "string-strip-html": "^8.3.0",
     "watcher": "^1.2.0",
--- a/server/ApiController.js
+++ b/server/ApiController.js
@@ -5,6 +5,7 @@ const date = require('date-and-time')
 
 const Logger = require('./Logger')
 const { isObject } = require('./utils/index')
+const resize = require('./utils/resizeImage')
 const audioFileScanner = require('./utils/audioFileScanner')
 
 const BookController = require('./controllers/BookController')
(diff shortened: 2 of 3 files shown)


advplyr:
I think sharp is a good choice. I think we should make this the main endpoint for cover images and implement caching.
We would also need to reset the cache for a cover if it is updated.
Also, I think we should generate webp images. I'm not sure if it is still worth checking if the browser supports webp.
Thoughts?

keaganhilliard:
Hmmm... Maybe we could add in a format parameter so that the client could specify. I don't know a ton about webp. Caching is a good idea especially if the images are smaller it shouldn't take up a ton of space. And yeah I thought about adding it as a general endpoint, but I wasn't sure of the best way to determine if the ID was a book/author, etc.

advplyr:
I think that endpoint is fine, `/books/:id/cover`.
If it was used on an image element the src would be `{server url}:13378/api/books/:id/cover?token={api token}`.
The raw image could still be fetched directly with the `/s/book/:id/*` and `/metadata` routes/static paths.

advplyr:
I pushed `v1.6.40` replacing the cover source with that endpoint and caching, but I could not get sharp to install in arm7.
Spent about 2 hours and couldn't get it so I didn't include arm7 in that release. May need to find a different library.

keaganhilliard:
@advplyr Awesome! Two things I noticed:
The format needs to be jpeg not jpg otherwise the response is a garbled mess for some reason. 

The other thing is that it's caching a file regardless of size parameters. So say you have a client that wants a higher quality image for a detail screen, you would pass a bigger width say 600, if that image is cached the endpoint will always return the cached file, which is not quite what I think it should do. Maybe we could cache multiple sizes or purge the cache if the size is different?

PR #240, opened 2021-12-13
Remove sharp in favor of ffmpeg, fallback to cover

keaganhilliard:
Since you were having issues getting sharp to work... I made ffmpeg work instead. Also fixed a couple of issues that I noticed in testing with my library. Namely embedded covers that have been extracted don't have a fullPath so I'm falling back to use the cover attribute if that fullPath is null.


advplyr:
Awesome! I'm looking at it now and wondering why you would have covers that don't have a "coverFullPath" set?

In old versions of ABS there was only the "cover" attribute, but now there should always be both set. The "cover" attribute won't be a path in the filesystem. This must be a bug if you have this happening.

keaganhilliard:
It's happening on a significant portion of my library. I think it's every one that is extracted with Ffmpeg.

keaganhilliard:
I even saw it happen on newly imported files with this version so I'm not sure what it happening. If I match the file and download a cover then everything is fine.

advplyr:
Oh I see why.. you're right it is not getting set properly on extraction

keaganhilliard:
Don't know if this is also related, but any time I rescan a book the cover disappears. Like I can either download a new one or type in the right path and it will come back.

advplyr:
My guess is the scan is trying to fix the full cover path and failing for some reason, there should be logs.
When requesting a cover I added a fix for the full cover path. There are a lot of bandaids on the cover logic now because it has changed so many times, so the entire cover structure is going to need to be re-thought out soon I think.

I'm going to push this as v1.6.41, thanks for implementing ffmpeg also so we can build for arm7 again.

advplyr:
1.6.41 is deployed

PR #249, opened 2021-12-21
Fix audible match if title or author has special characters

mx03:
If the audible match contains special characters like german ä ö ü , the response is a bad request (400).

Example: "Wenn du mir gehörst"
https://www.audible.com/pd/Wenn-du-mir-gehoerst-Audiobook/B09MSG8PJ5


advplyr:
Great, thanks for fixing that. I'm pushing that patch over top of `v1.6.46`, I don't think it requires a version bump.

keaganhilliard:
Might also be worth making some separate providers for each of the countries audible supports, (Germany, UK, etc).

Issue #260, opened 2021-12-25
Bug: Series cover hover works only with 30 books


mx03:
The overlapping coverlist for series works only with 30 books. Because of the z-index( currently: z-30).

I don't know if there is a normal audiobook series with more than 30 books, but i use it also for radio plays and some have 100 or more volumes.

Solution would be a z-index on the parent with the relative position.

![image](https://user-images.githubusercontent.com/1390929/147387683-66397747-dd06-4c02-a6fd-ba00bd266759.png)

PR #262, opened 2021-12-25
Fix Series cover hover works only with 30 books

mx03:
Issue #260


advplyr:
Makes sense. You can remove the `z-2` class. That class isn't defined in tailwind and it shouldn't be necessary to assign a z-index there anymore.

mx03:
Yes the zindex is not needed. Updated the pull request.

Issue #265, opened 2021-12-26
Sort select filter data


mx03:
Current the filter for example genre or authors is not filtered by something useful, most logic would be a-z.

![image](https://user-images.githubusercontent.com/1390929/147420007-b3b45b19-52e0-43d2-9a6d-745ee329d991.png)

PR #266, opened 2021-12-27
Sort filters ascending

mx03:
#265


advplyr:
Nice solution. I just pushed this over the current version.

PR #270, opened 2021-12-27
Bug: Missing path variable

mx03:
` UnhandledPromiseRejectionWarning: ReferenceError: Path is not defined at /server/controllers/FileSystemController.js:8:7 ...`


advplyr:
Just patched the current version

PR #284, opened 2022-01-03
Added isbn fetching for metadata

ISO-B:
If file metadata has isbn tag it will be fetched and used for book


advplyr:
Looks good, thanks. I just patched the current version 1.6.51 for docker and this will be included in the next version bump for debian.

PR #287, opened 2022-01-04
Added isbn search

ISO-B:
Search now also searches books using isbn. ISBN match has to be exact to prevent isbn matches to flood results. 

I was thinking about two other approaches:
1. Only 10 and 13 character searches to be isbn searches since those are isbn lengths. This would limit using isbn field with other type of identify codes. I for one have some podcasts and other stuff that won't have isbn, but have guid instead.
2. Start searching isbn when search string is longer than 4,  because book name might have year in it. This might be good solution since not many books have more than 4 consecutive numbers so there would be no other results anyway. I don't know if there is use case for searching partial isbn. This would also require changes to client side search results list so you could see what part of isbn is matching. Since search only returns 12 books results would need to be ranked or sorted to some order, for example prefer matches from start of isbn.

Ended up with this proposed solution since I think it is good enough for most of people

### DIFF
--- a/client/components/cards/AudiobookSearchCard.vue
+++ b/client/components/cards/AudiobookSearchCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex h-full px-1 overflow-hidden">
+  <div class="flex items-center h-full px-1 overflow-hidden">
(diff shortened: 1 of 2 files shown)


advplyr:
Works well, I just added this:

image

advplyr:
This is in `v1.6.52`

PR #300, opened 2022-01-07
Fixes reading multiline data from file tags

ISO-B:
Swiched fluent-ffmpeg to node-ffprobe. Fluent's ffprobe implementation can't handle multiline tag data due to it's parsing process. There is open issue (https://github.com/fluent-ffmpeg/node-fluent-ffmpeg/issues/457) about problem and even PR(https://github.com/fluent-ffmpeg/node-fluent-ffmpeg/pull/476), but both are over 5 years old so I wouldn't hold my breath for that. 

node-ffprobe supports getting probe data using json format which keeps line changes. Not sure if this is best library for this but it won't have any dependencies and it does the job.

Tested with two mp3 files. One has metadata and one doesn't and it seemed to get same metadata as the old implementation. I advice you to do some testing if there is something that I didn't notice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "jsonwebtoken": "^8.5.1",
     "libgen": "^2.1.0",
     "njodb": "^0.4.27",
+    "node-ffprobe": "^3.0.0",
     "node-cron": "^3.0.0",
     "node-stream-zip": "^1.15.0",
     "podcast": "^1.3.0",
--- a/server/utils/prober.js
+++ b/server/utils/prober.js
@@ -1,4 +1,5 @@
 var Ffmpeg = require('fluent-ffmpeg')
+const ffprobe = require('node-ffprobe')
 const Path = require('path')
 
 const AudioProbeData = require('../scanner/AudioProbeData')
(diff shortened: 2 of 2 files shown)


advplyr:
Do you have a file that fails to parse because of an empty line?

ISO-B:
It won't fail to parse, but it leaves rest of lines out. Any mp3 file that has line change on example description or comment does that.

ISO-B:
> Tested with two mp3 files. One has metadata and one doesn't and it seemed to get same metadata as the old implementation. I advice you to do some testing if there is something that I didn't notice.

What I meant to write was:
Previously if you had file that had line change in comment or description tag then book description content was cut to first line change. After this change description won't be cut. 

My test "process" was to scan file and take probedata from old implementation first and then run scan again with new implementation and take probedata again. Then I compared those and only thing that changed was description. Also tested with file that wont have any id3 metadata and it resulted same result with both implementation that no metadata(tagdata) was fetched because there is none in file.

I don't have any m4b or m4a files to test with those so can't say if those still work. That's reason why I encourage you to do some testing also. If you need I can send you mp3 file with line changes on description later today if you need, but I think that you can edit some file yourself.

advplyr:
Just tested and confirmed multiline description is working. Nice job, thanks.

advplyr:
This is in `v1.6.54`

Issue #321, opened 2022-01-21
Cover of series with lots of episodes looks odd


JKamsker:
![image](https://user-images.githubusercontent.com/11245306/150613100-fdcc4a07-f633-42b0-9f97-25e029dcafc4.png)

There is no advantage in adding all titles to the cover. Maybe consider capping at 10

mx03:
Yes i had the same problem, but forget to make a pull request.
I set the limit to 30, thats a good balance.

Alternative a good solution would be that every series can have an custom cover, (e.g.: a cover.jpg inside the series folder)
Instead
![image](https://user-images.githubusercontent.com/1390929/150635045-a21ae950-5fac-47fe-852a-a7fc264b2ed6.png)
just
![image](https://user-images.githubusercontent.com/1390929/150635074-41d16c92-e79f-45fb-99d1-a943a1b3b788.png)

JKamsker:
Can you also add an indicator how many items are in the series?

Im thinking of smth like the series-title but on the top left or buttom right.

advplyr:
Thoughts on this?

advplyr:
Just pushed `v1.6.60` with pr #325 to limit covers to 30 and added the # of books in series.

jcrmit:
@advplyr - yes updated - I like it - the most I have in a series is 25 - so cannot comment on the that. A custom or auto (first book) view as per @mx03 - or first 5-10 might also look optimal. I think 25 looks busy. Happy either way. Having the number is great!

![image](https://user-images.githubusercontent.com/49168770/150890854-4e32a072-5d8f-439d-96b3-7871fde505df.png)

advplyr:
I believe @mx03 fixed this and his PR's were merged. Re-open if there are still issues here

PR #325, opened 2022-01-22
Limit group cover

mx03:
#321 
Limit is currently 30, i think thats a good balance


advplyr:
Agreed

advplyr:
This is included in v1.6.60

PR #326, opened 2022-01-22
Fix: Series link from search goes straight to series

MartinCa:
Clicking on the series in the search result box now goes straight to the series instead of to the series library view.


advplyr:
Ah yeah missed that one, thanks

advplyr:
This is included in v1.6.60

PR #330, opened 2022-01-25
Fix: Group cover does not use full width

mx03:
#321 

I will edit the amount of showed covers based on #321.


advplyr:
This is included in `v1.6.62`, I haven't actually tested as I don't have a 30+ book series.

PR #331, opened 2022-01-25
Add continue series category to home

igorkaldowski:
Added section to home with next book in any series that has been started.


advplyr:
What about if I'm part-way through a book in the series, like this:
image

Do you think I should be shown volume 2 or 3?

image

igorkaldowski:
It Ignores the series as it is in continue reading

advplyr:
This is included in `v1.6.62`

PR #343, opened 2022-02-01
Fetch Narrators from opf custom column

igorkaldowski:
Added getting narrators from calibre custom column in opf files.

### DIFF
--- a/server/utils/parseOpfMetadata.js
+++ b/server/utils/parseOpfMetadata.js
@@ -80,6 +80,15 @@ function fetchVolumeNumber(metadata) {
   return fetchTagString(metadata.meta, "calibre:series_index")
 }
 
+function fetchNarrators(creators, metadata) {
+  var roleNrt = fetchCreator(creators, 'nrt')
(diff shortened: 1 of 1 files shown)


server/utils/parseOpfMetadata.js
```diff
@@ -80,6 +80,15 @@ function fetchVolumeNumber(metadata) {
   return fetchTagString(metadata.meta, "calibre:series_index")
 }
 
+function fetchNarrators(creators, metadata) {
+  var roleNrt = fetchCreator(creators, 'nrt')
+  if(typeof metadata.meta == "undefined" || roleNrt != null) return roleNrt
+  var narratorsTag = fetchTagString(metadata.meta, "calibre:user_metadata:#narrators")
+  if(narratorsTag == null) return narratorsTag
+  var narratorsJSON = JSON.parse(narratorsTag.replace(/&quot;/g,'"'))
```
advplyr, line 88: I think this should be wrapped in a try catch

PR #351, opened 2022-02-06
Basic SSO support (openid connect)

davidLeimroth:
I started adding sso support via passport js yesterday.
Users are generated as guests on the fly, if they don't exist.
There are a couple of things missing, like:
* ~hiding the sso button in the web ui if sso is not configured. But as it was my first time looking into vue today, it was out of scope for now.~
* ~being able to change the password, if the user was generated via sso access~ create a flow for users logging in for the first time via sso with the ability to choose username and/or password
* add the ability for existing users to link their "sso account"


advplyr:
Great start, I will need to setup openid to test.

I am wondering though about the use-case where someone wants to use openid for the root user. Or for a user they already have been using that has all their data. It seems like we will have to make a function to connect an openid profile to an existing account.

As for the password change you mentioned, wouldn't the user chance their password in their openid client? I think we will need to add some variable on the account for the provider, so we know it is from openid.

davidLeimroth:
I use [Authentik](https://github.com/goauthentik/authentik/) in my setup. Keycloak should work as well.

I agree on linking the profile to an existing account, the user id is not ideal. A new field on the user object could be a solution here. Maybe an array to potentially support multiple sso providers per audiobookstack instance somewhen in the future. That should cover the root user as well, shouldn't it?

As for changing the password, personally I prefer to link accounts while still being able to log into the account using username and password. This has the benefit of being able to use the app, even if the sso provider is down. With the way the password is generated right now, users created through sso are not able to change their bookstack specific password on their own

advplyr:
I see. So maybe adding a new field to accounts and an input in the UI to paste in SSO ids?
Instead of creating a guest account, the login would be rejected since it first needs to be connected from within the web app.

davidLeimroth:
What do you think about making the option for account creation configurable?
That'll make it easier from an administrator point of view.

advplyr:
I'm not sure what you mean. Configurable how?

davidLeimroth:
Configurable like this. I added a sso page to config to make the administration easier

advplyr:
I'm going to move SSO settings into the settings db, instead of creating a new db. Then I'm assuming the SSO settings should be initialized with the env variables if they exist.

advplyr:
I setup Authelia yesterday before realizing their OIDC is in beta. I may still try to get Authelia working, but as of now I haven't tested the updates I made.

davidLeimroth:
Maybe authenik/keycloak are the way the go until  Authelia is out of beta. I had some problems with Keycloak in the past. As long as it was running it was nice, but restarting threw errors sometimes. Authentik worked flawless so far.
All you need is to set up an oauth/openid provider and a corresponding app using that provider

advplyr:
What is your server setup?

davidLeimroth:
[Freeipa](freeipa.org) to store users and integrate with applications that don't support saml/oauth/oidc (such as jellyfin).
Authentik as an identity provider for sso. 

Users and groups are synced from freeipa to authentik, password changes are synced back to freeipa

davidLeimroth:
> I don't understand the user.settings and user.isActive inside the SSOSettings. I think the only thing we need there is user permissions.

I thought about adding more configuration options to the settings page. Not sure about that though.

> I'll be short on time the next few days (...) 

Same, I'll try to add the ability to link an account from an idp to an existing account somewhen soon

advplyr:
I setup Keycloak and couldn't get this to work. Keycloak works well, but the passport system is not making sense to me.

The /oidc/callback is getting called but the sso cookie is never set, that callback function is never called.

A cookie is getting set though after a successful login to keycloak and I am redirected back to audiobookshelf login.
image

I'm guessing passport is setting that and using HttpOnly preventing it from being seen on localhost.

When setting this up though I was wondering why use the redirect when audiobookshelf login form could just send a post request to the OIDC server to authenticate.
What services do you use that support OIDC?

davidLeimroth:
> I setup Keycloak and couldn't get this to work. Keycloak works well, but the passport system is not making sense to me.
> 
> The /oidc/callback is getting called but the sso cookie is never set, that callback function is never called.

Do you know why the callback isn't called? Can you attach a debugger? It works perfectly fine on my system using chrome, so it's hard to tell why it doesn't work for you. The only thing I can imagine right now is keycloak handling things differently compared to authentik. I'll test that in the next couple of days.

> A cookie is getting set though after a successful login to keycloak and I am redirected back to audiobookshelf login. image
>
> I'm guessing passport is setting that and using HttpOnly preventing it from being seen on localhost.

We need to change this from a http only cookie to returning a value for the mobile apps or add a different endpoint for them altogether. If I remember correctly, anything based on a webview has the same httponly restrictions and deletes cookies when the app is closed. This has the disadvantage of us having to store the value securely ourselves instead of relying on a builtin mechanism.

> When setting this up though I was wondering why use the redirect when audiobookshelf login form could just send a post request to the OIDC server to authenticate.

Right now the server is implemented as a confidential client with a client id and client secret.
There's a nice [overview at auth0](https://auth0.com/docs/get-started/applications/confidential-and-public-applications)
As we're having a proper backend, a confidential client is the way to go here in my opinion.

> What services do you use that support OIDC?

Off the top of my head: nextcloud and gitea

alexanderadam:
> I setup Authelia yesterday before realizing their OIDC is in beta

I'm pretty sure all Authelia users are using OIDC anyway :wink: 

Will it be possible to have OpenID Connect configurable via `ENV` variables on the [container image](https://hub.docker.com/r/advplyr/audiobookshelf)?

DDriggs00:
What is this currently waiting on?

advplyr:
This will need to be re-done. It was started pre version 2 and so much has changed.

It was never clear to me the best way to support this. I think we will need to outline a plan before this gets re-built.

couladin:
Just in case this topic gets picked up later: I think the way [Wallabag](https://github.com/wallabag) does OTP integration in the server works well.

AutoGitr:
Are there plans to continue working on this soon? Any form of SSO in Audiobookshelf would be fantastic, as goes for any self hosted app.

DDriggs00:
Since this pull request is incompatible with the current version (and thus closed), Discussion should probably be moved to an issue. I've created one [here](https://github.com/advplyr/audiobookshelf/issues/998)

PR #418, opened 2022-03-16
feat: add auto docker build. fix yaml extension

tvories:
Signed-off-by: tvories <taylor@tmtech.me>

## Requirements

To make this work, you will need to have a couple github repo secrets added.  You'll need:

- [ ] `DOCKERHUB_USERNAME` from dockerhub.com.
- [ ] `DOCKERHUB_PASSWORD` the token for your dockerhub account
- [ ] `GHCR_PASSWORD` a github token with the following permissions: 
image

That will enable the workflow to push to dockerhub and ghcr.

## What this does

This PR enables auto docker builds on the following events:

1. Pushes to `master`
2. Pull requests to `master`
3. A release is published or edited.

It builds with a tag of master when a change is pushed to master.  It builds with a tag of the release tag when a release is cut.  It builds to both ghcr and dockerhub.  It builds for the following environments:

1. linux/amd64
2. inux/arm64
3. linux/arm/v7


rasmuslos:
I think it would be better if the `audiobookshelf` image wouldn't get updated every time the master branch received a new commit. Naming the image `audiobookshelf-dev` would be an alternative

tvories:
> I think it would be better if the `audiobookshelf` image wouldn't get updated every time the master branch received a new commit. Naming the image `audiobookshelf-dev` would be an alternative

Why not just have it auto build on master with the `dev` tag?

advplyr:
What about every time a commit is made to master it builds with the `next` tag. Then when a commit is made with a Version bump it will build to `latest` and the current version.
I might prefer doing it that way then building up new versions in branches, but let me know what you think.

tvories:
what is a "next" tag?  I've never seen that before.  I would think dev would handle most cases for master, no?  Then also build and push any specifically tagged release with the tag of the release (this is already set in this pr)

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: advplyr/audiobookshelf,ghcr.io/${{ github.repository_owner }}/audiobookshelf
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
```
advplyr, line 44: I don't think we need to push a major.minor tag

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: actions/checkout@v2
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: advplyr/audiobookshelf,ghcr.io/${{ github.repository_owner }}/audiobookshelf
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
```
advplyr, line 42: I don't think we need to build and tag for every pull request

advplyr:
Ah yeah this is done already. There are some annoying quirks though so if you are an expert on this stuff I could use some help.

Occasionally I want to manually trigger a build using a specific tag and the latest source code. When you re-run any of the previous runs it won't use the latest source code it will use whatever source was used when it first ran.

bskrtich:
@advplyr Okay, Im pretty sure I can setup a manual docker build for this. Lets close out this PR and I will create an issue to track that work.

PR #423, opened 2022-03-17
Allowed the configuration of a "HOST" parameter to enable ipv6 support.

Quietus:
My local network runs mostly on ipv6, so I would like to be able to adjust the listening host (to "::" for me) to enable the server to listen on both ipv4 and v6.  This also allows people to limit the server to listen on specific IPs if needed.


advplyr:
Just had to add support for HOST in production, your update would only work for local.

PR #426, opened 2022-03-24
Upgrade Node to v16 and update packages

jflattery:
Upgraded socket.io to 4.4.1 to address uncaught Exception in older version of engine.io


jflattery:
Tested and working on both x86_64 and aarch64.

Images available at:
jflattery/abs-dev:node16

jflattery:
Tested and working on both x86_64 and aarch64.

Images available at:
jflattery/abs-dev:n16u

PR #440, opened 2022-04-03
Improve Accessibility: Labels on UI Components

duplaja:
Hello! 

I did some testing with iOS (web view), and noticed a few things that could be improved as far as accessibility. These are a couple of the more immediate ones. I will continue to test and see what else I can find.


jflattery:
Hey, so you were mistakenly editing the name of the material-icons. In the spirit of what you were looking to do I have made the changes and submitted a new pull request #459. If you would like to give it a spin I have a test docker image available listed on my pull request. Keep in mind the dev image is not compatible with current versions of the apps (ios & android) and you should not use it in production yet.

advplyr:
Merged #459

Issue #441, opened 2022-04-03
[Bug]: Accessibility Issues with Items that Expand, and Missing Labels


duplaja:
### Describe the issue

Items that can expand when clicked, such as: 

* filter dropdown
* sort dropdown
* chapter selection from the player

are hard-coded with aria-expanded="true". This should be false when they are not opened up, and true when they are expanded.

In addition, these items are all missing properly accessible labels.

They are using the aria-labeledby attribute which is supposed to reference another element that has the correct label. However, these all are set as "listbox-label", which is not actually an element that exists on the page. If you want to label these directly, these should use aria-label="<Name of Item>".

### Steps to reproduce the issue

1. Appears when using Audiobookshelf site with a screen reader of any sort.

### Audiobookshelf version

v1.7.2

### How are you running audiobookshelf?

Debian/PPA

duplaja:
![library-screen](https://user-images.githubusercontent.com/20245408/161409833-8d8efe6c-4376-429f-bb53-65bb6bce9684.png)

This is the aria view of the library screen, for example. I've submitted a pull request that should address the add / remove at the bottom (those are the adjust the ebook cover size), and partially address the account label (need to probably remove "root", and the pull request replaces "person" with "User Account").

You can see that the Filter just shows as "All", with no other label, and as expanded even when it first loads. Similarly, "Sort" appears as "Title expand_less".

One good resource for (quickly) checking what a screenreader will see, as far as labels, is [this Chrome extension](https://chrome.google.com/webstore/detail/aria-devtools/dneemiigcbbgbdjlcdjjnianlikimpck/related?hl=en)

PR #459, opened 2022-04-13
podcast episode number & accessibility improvements

jflattery:
Add episode number to list group view

### DIFF
--- a/client/components/app/Appbar.vue
+++ b/client/components/app/Appbar.vue
@@ -23,15 +23,15 @@
         </div>
 
         <nuxt-link to="/config/stats" class="outline-none hover:text-gray-200 cursor-pointer w-8 h-8 flex items-center justify-center mx-1">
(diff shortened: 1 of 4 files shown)


jflattery:
Accessibility Labels as suggested by @duplaja

jflattery:
Docker image is currently building and will be available at `jflattery/abs-dev:accessibility`

advplyr:
much obliged

duplaja:
@jflattery  Thank you, I appreciate it! That does look much cleaner / much closer to what I'm used to doing for these.  I wasn't sure if there was anything particular to the framework, so I had initially just changed what it was currently pulling the label from (icon text).

Again, coming from a non-framework background, there were a few other accessibility items I noted in #441, but don't (currently) have the background to tackle. If you feel up to taking a look, or can give me some pointers for where to start looking, I would appreciate it! (if you have the time and inclination, of course)

PR #486, opened 2022-04-23
Set description when quick matching media

selfhost-alt:
All the audiobook media providers seem to provide a `description` field, so we're just failing to apply it to the media item after it is looked up. This PR updates the list of updated properties to include `description`, so now doing a Quick Match or matching an entire library should also update media item descriptions if available.


advplyr:
That's strange. It seems I must have left that out for a reason but I can't think of why.

PR #489, opened 2022-04-23
Make maximum backup size configurable

selfhost-alt:
On sufficiently large libraries, backups can start to fail because they hit the hardcoded 1GB backup limit. This PR makes that limit configurable so that users with large libraries can set a larger limit.

Screenshot of new UI:

image


advplyr:
Nice!

PR #551, opened 2022-05-02
docker compose and run changes

jflattery:


readme.md
```diff
@@ -71,10 +71,11 @@ docker run -d \
   -e AUDIOBOOKSHELF_GID=100 \
   -p 13378:80 \
   -v </path/to/audiobooks>:/audiobooks \
+  -v </path/to/your/podcasts>:/podcasts \
   -v </path/to/config>:/config \
   -v </path/to/metadata>:/metadata \
   --name audiobookshelf \
-  ghcr.io/advplyr/audiobookshelf
+  ghcr.io/advplyr/audiobookshelf:2.0.8
```
advplyr, line 78: Why include a specific version here?

jflattery: It is concidered to be good practice. Example, there major changes going from `1.7.2` to `2.0.*`. When you don't include a tag, the default (often `latest`) is used. This could lead to very upset users when their container updates to a newer version on it own. They should explicitly look to update it.

jflattery: Notes on tagging:
- https://vsupalov.com/docker-latest-tag/
- https://stackoverflow.com/questions/70424052/should-i-set-docker-image-version-in-docker-compose

advplyr: We are using latest to mean the latest stable version. For the normal user this is what they want so they can continue to get the latest stable releases. We are not currently building major and minor versions so with every patch the user would need to go in and update it the tag, and we would need to go in and update the readme.
A user familiar with docker and wants to manually update their version tags probably won't need to see the version tag in the readme.

jflattery: Ok, than disregard the "add version number" commit :D

PR #561, opened 2022-05-04
Add support for seasonal podcasts

jflattery:
Podcasts such as [Command Line Heroes](https://podcasts.apple.com/us/podcast/command-line-heroes/id1319947289) have multiple seasons in which each has it's own "Episode 1, ••• Episode n". This seeks to add support for this class of podcast series.

image

### DIFF
--- a/client/components/modals/podcast/EditEpisode.vue
+++ b/client/components/modals/podcast/EditEpisode.vue
@@ -7,16 +7,16 @@
     </template>
     <div ref="wrapper" class="p-4 w-full text-sm py-2 rounded-lg bg-bg shadow-lg border border-black-300 relative overflow-hidden">
       <div class="flex flex-wrap">
-        <div class="w-1/3 p-1">
+        <div class="w-1/5 p-1">
           <ui-text-input-with-label v-model="newEpisode.season" label="Season" />
         </div>
-        <div class="w-1/3 p-1">
+        <div class="w-1/5 p-1">
           <ui-text-input-with-label v-model="newEpisode.episode" label="Episode" />
         </div>
-        <div class="w-1/3 p-1">
+        <div class="w-1/5 p-1">
           <ui-text-input-with-label v-model="newEpisode.episodeType" label="Episode Type" />
         </div>
-        <div class="w-1/3 p-1">
+        <div class="w-2/5 p-1">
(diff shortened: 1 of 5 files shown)


jflattery:
Tested locally, currently building `jflattery/abs-dev:podcast-seasons` for further testing

server/utils/podcastUtils.js
```diff
@@ -85,7 +85,7 @@ function extractEpisodeData(item) {
     episode.descriptionPlain = stripHtml(episode.description || '').result
   }
 
-  var arrayFields = ['title', 'pubDate', 'itunes:episodeType', 'itunes:episode', 'itunes:author', 'itunes:duration', 'itunes:explicit', 'itunes:subtitle']
+  var arrayFields = ['title', 'pubDate', 'itunes:episodeType', 'season', 'itunes:episode', 'itunes:author', 'itunes:duration', 'itunes:explicit', 'itunes:subtitle']
```
advplyr, line 88: I pulled the feed for command line heros and see that the tag is actually `itunes:season` and not `season`. I'm not sure if you tested this already and it scraped properly but I don't see the tag `season`.
image

advplyr: fyi the itunes part of the string gets stripped out when returning from that function.

client/components/modals/podcast/EditEpisode.vue
```diff
@@ -7,6 +7,9 @@
     </template>
     <div ref="wrapper" class="p-4 w-full text-sm py-2 rounded-lg bg-bg shadow-lg border border-black-300 relative overflow-hidden">
       <div class="flex flex-wrap">
+        <div class="w-1/3 p-1">
```
advplyr, line 10: I might suggest changing the widths of these so that all 4 season, episode, episode type and pub date fit on the same line.
You can try doing `w-1/5`, `w-1/5`, `w-1/5` and `w-2/5` respectively. So the `w-2/5` would go on the pub date input. Let me know if that doesn't make sense.

server/utils/podcastUtils.js
jflattery: Good catch! I'll get this fixed

jflattery: It's verking!

client/components/modals/podcast/EditEpisode.vue
jflattery: Much better!

PR #586, opened 2022-05-12
Fix for unsorted feeds, update npm, add docker tag

jflattery:
Fix for Podcasts such as "Beers with Talos" who don't publish their feed in a chronological order

### DIFF
--- a/client/components/modals/podcast/EpisodeFeed.vue
+++ b/client/components/modals/podcast/EpisodeFeed.vue
@@ -148,6 +148,7 @@ export default {
         })
     },
     init() {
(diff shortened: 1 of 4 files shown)


jflattery:
[Tagging Discussion](https://discord.com/channels/942908292873723984/942914154254176257/973952292506202122)

advplyr:
I'm not sure what all these changes are in the package-lock files. It looks like some dependencies are getting added

jflattery:
This is the result of a `npm update` and `npm dedupe` in both the root directory and `/client`. No new packages should be added.

PR #622, opened 2022-05-23
Updated matching

kaldigo:
Updated matching to include multiple series, multiple authors, genres, tags, language.
Added toggle in settings to force quick match to override existing data.
Quickmatch now sends asin and isbn along with query.
Updated Audible provider to use audnexus and get information for all updated fields.


server/providers/Audible.js
```diff
@@ -6,83 +6,81 @@ class Audible {
...
-            if (!res || !res.data || !res.data.products) return []
-            return res.data.products
-        }).catch(error => {
-            Logger.error('[Audible] search error', error)
-            return []
-        })
-        return items.map(item => this.cleanResult(item))
+        if(!items) {
+            var queryObj = {
+                response_groups: 'rating,series,contributors,product_desc,media,product_extended_attrs',
```
advplyr, line 64: If now we only need to get the ASIN from the audible api I think we can safely remove the response groups and image sizes.

server/providers/Audible.js
```diff
@@ -6,83 +6,81 @@ class Audible {
...
+                image_sizes: '500,1024,2000',
+                num_results: '25',
```
advplyr, line 66: Since each result item will be sending a request to audnexus API, I think we should limit the number of results here to 10. What do you think?

kaldigo: Totaly agree. Realisticly it shouldn't return that many if you are search the full title. I have limited it to 10 and removed the queries that were no longer nessasary.

advplyr:
The multi-series input on the Match tab will need to be editable I think. I'm sure you already saw it, the editable multi-series input used for book details is here: https://github.com/advplyr/audiobookshelf/blob/master/client/components/widgets/BookDetailsEdit.vue#L25
Which has an ugly modal for editing that goes with it.

If it's not editable it won't be long before that gh issue gets submitted. There is more effort here though in that the edit series modal should probably go into a new component instead of being copied as is.

server/scanner/ScanOptions.js
```diff
@@ -32,7 +33,8 @@ class ScanOptions {
       findCovers: this.findCovers,
       storeCoverWithItem: this.storeCoverWithItem,
       preferAudioMetadata: this.preferAudioMetadata,
-      preferOpfMetadata: this.preferOpfMetadata
+      preferOpfMetadata: this.preferOpfMetadata,
+      preferOpfMetadata: this.preferMatchedMetadata
```
advplyr, line 37: Forgot to rename the property here and on line 49

PR #631, opened 2022-05-24
Get book data from dir refactor

cassieesposito:
Minor tweaks to the way scanner handles information about series volumes.

### DIFF
--- a/server/utils/scandir.js
+++ b/server/utils/scandir.js
@@ -215,9 +215,6 @@ function getBookDataFromDir(folderPath, relPath, parseSubtitle = false) {
   // Audio files will always be in the directory named for the title
   var [title, narrators] = getTitleAndNarrator(splitDir.pop())
 
-  const parseNameString = require('./parseNameString')
-  console.log(`\n\n\n${JSON.stringify(parseNameString.parse(narrators), 0, 2)}\n\n\n`)
-
   var series = null
   var author = null
   // If there are at least 2 more directories, next furthest will be the series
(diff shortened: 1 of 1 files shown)


server/utils/scandir.js
```diff
@@ -209,91 +209,23 @@ function getBookDataFromDir(folderPath, relPath, parseSubtitle = false) {
...
-  // Subtitle is everything after " - "
-  var subtitle = null
-  if (parseSubtitle && title.includes(' - ')) {
-    var splitOnSubtitle = title.split(' - ')
-    title = splitOnSubtitle.shift()
-    subtitle = splitOnSubtitle.join(' - ')
-  }
+  // The  may contain various other pieces of metadata, these functions extract it.
+  var [folder, narrators] = getNarrator(folder)
+  if (series) { var [folder, sequence] = getSequence(folder) }
```
advplyr, line 218: Does this work? Folder and sequence are being defined inside the if statement scope.

advplyr: I don't think I ever new that variables defined in if statements are available outside, but it appears to be so. "Function scope" they say.

cassieesposito: I believe this is one of the differences between var and let.

cassieesposito: Yep.

Welcome to Node.js v16.15.0.
Type ".help" for more information.
> if (1) {var w = 'hello'}
undefined
> if (1) {let x = 'world'}
undefined
> w
'hello'
> x
Uncaught ReferenceError: x is not defined
>

advplyr: Probably best to avoid putting this in that scope in the future, might cause some confusion for others too.

PR #651, opened 2022-05-31
Gracefully handle unexpected end of file when listing backup files

selfhost-alt:
A few weeks ago I added some logic to handle one kind of backup file corruption when listing backups in #555. Today when looking at my backups, I found that there's another type of error that should also be handled. Namely, when a backup file is unexpectedly truncated.

### DIFF
--- a/server/managers/BackupManager.js
+++ b/server/managers/BackupManager.js
@@ -141,6 +141,9 @@ class BackupManager {
             if (error.message === "Bad archive") {
               Logger.warn(`[BackupManager] Backup appears to be corrupted: ${fullFilePath}`)
               continue;
+            } else if (error.message === "unexpected end of file") {
+              Logger.warn(`[BackupManager] Backup appears to be corrupted: ${fullFilePath}`)
+              continue;
             } else {
               throw error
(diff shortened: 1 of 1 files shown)


advplyr:
Is that error message all lowercase? Seems strange that would be lowercase but the others aren't.

selfhost-alt:
The reason the messages are capitalized differently is because they come from different codebases.

The `Bad archive` message comes from [inside node-stream-zip](https://github.com/antelle/node-stream-zip/blob/7c5d50393418b261668b0dd4c8d9ccaa9ac913ce/node_stream_zip.js#L198)

The `unexpected end of file` error originates in [Node.js core zlib code](https://github.com/nodejs/node/blob/55f56efbac56ce2a1ad94ee16ebbfc77fed9d82e/src/node_zlib.cc#L900)

If it's less confusing, I'd be happy to coerce all the error messages with `toLower()` or something, but I have tested against backup files that throw both of these errors, so I'm confident that these error strings match correctly.

advplyr:
Got it. No I think it's fine just making sure.

Issue #662, opened 2022-06-01
[Enhancement]: Add healthcheck endpoint


bskrtich:
### Describe the feature/enhancement

We need some sort of health check endpoint to make sure the webapp/server is up and responding correctly. This then could be added to the docker setup or other setups to auto restart if the server dies.

advplyr:
There is a ping endpoint now that returns a JSON object.
`https://yourserver.com/ping`

I think the JSON object is unnecessary to return but if all you need to check is the status code then it will return a 200 if it succeeds.

Issue #661, opened 2022-06-01
[Enhancement]: Ability to trigger manual docker build


bskrtich:
### Describe the feature/enhancement

It would be nice if we can manually trigger a docker build using a specific tag and the latest source code

Jdiesel87:
You should already be able to do this 

https://hub.docker.com/r/advplyr/audiobookshelf/tags

bskrtich:
This is a request from @advplyr 
https://github.com/advplyr/audiobookshelf/pull/418#issuecomment-1144193822

PR #664, opened 2022-06-02
feat: Updates to docker file and gh action

bskrtich:
* Clean up Dockerfile 
* Add health check to Dockerfile [662](https://github.com/advplyr/audiobookshelf/issues/662)
* Update gh action versions
* Update manual trigger to include input [661](https://github.com/advplyr/audiobookshelf/issues/661)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,13 +7,23 @@ RUN npm run generate
 
 ### STAGE 1: Build server ###
 FROM node:16-alpine
-RUN apk update && apk add --no-cache --update ffmpeg
 ENV NODE_ENV=production
+RUN apk update && \
+    apk add --no-cache --update \
+    curl \
+    tzdata \
+    ffmpeg
+
 COPY --from=build /client/dist /client/dist
-COPY index.js index.js
-COPY package-lock.json package-lock.json
-COPY package.json package.json
+COPY index.js package* /
 COPY server server
+
 RUN npm ci --only=production
+
 EXPOSE 80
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -7,13 +7,23 @@ RUN npm run generate
 
 ### STAGE 1: Build server ###
 FROM node:16-alpine
-RUN apk update && apk add --no-cache --update ffmpeg
 ENV NODE_ENV=production
+RUN apk update && \
+    apk add --no-cache --update \
+    curl \
+    tzdata \
```
advplyr, line 14: What does timezone do for us here? Is that for the healthcheck?

bskrtich: It allows me to change the timezone inside the container using an environment var. From what I've heard the server doesnt respect the timezone yet but it can be usefull for running other commands. I also wanted to look into seeing what it would take to use the local timezone in the logs.

bskrtich: I can remove it if you want.

advplyr: I'm not sure how that works. If we wanted to add the timezone as an environment var wouldn't we need to set env using the `ENV` command?

bskrtich: The tzdata package automatically picks up the `TZ` env and uses it to set the timezone inside the container.

advplyr:
Tried out the manually build with tag input and build is failing.
https://github.com/advplyr/audiobookshelf/runs/6740936624?check_suite_focus=true

PR #680, opened 2022-06-04
Update Bookfinder.js : getAudibleResults : remove (unabridged)

sxb1n9:
cleantitle for search in audible as (unabridged) doesn't find anything

### DIFF
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -167,7 +167,9 @@ class BookFinder {
   }
 
   async getAudibleResults(title, author, asin) {
-    var books = await this.audible.search(title, author, asin);
+    // cleantitle for search in audible as (unabridged) doesn't find anything
+    var cleanedTitle = title.replace(/ *\([^)]*\) */g, "")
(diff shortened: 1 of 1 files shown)


advplyr:
If a book had parenthesis in the title then this would filter that out.
This feels like a workaround to your specific issue with how your audiobook was titled.

sxb1n9:
Almost all of my audiobooks from audible have (unabridged) in the title.   I did not edit them at all and downloaded them and converted using https://github.com/audiamus/BookLibConnect

They come directly from that program into a folder for AudioBookShelf.

samosbor:
I think this a good idea and I found that a bunch of my audiobooks had (unabridged) in the title. 

I do agree that the regex is too broad though. It should probably just be a case-insensitive check for (unabridged).

Even better: add a new setting for filtering certain terms in the title match.

samosbor:
Thanks for this good idea @sxb1n9. I was able to incorporate a fix for this here https://github.com/advplyr/audiobookshelf/pull/735

advplyr:
#735 was merged

PR #682, opened 2022-06-05
Support embedding cover art metadata in the embed metadata tool

jmt-gh:
Added support for chapter metadata in #678, but completely missed that coverart wasn't getting embedded in the embed metadata tool. This commit adds that in.

I tested this by putting an m4b in my library, scraping a new cover, embedding the metadata through the embed tool, and then opening in VLC to validate the new cover is there

### DIFF
--- a/server/managers/AudioMetadataManager.js
+++ b/server/managers/AudioMetadataManager.js
@@ -32,7 +32,9 @@ class AudioMetadataMangaer {
     var metadataFilePath = Path.join(outputDir, 'metadata.txt')
     await writeMetadataFile(libraryItem, metadataFilePath)
 
-    var coverPath = libraryItem.media.coverPath.replace(/\\/g, '/')
+    if (libraryItem.media.coverPath != null) {
+      var coverPath = libraryItem.media.coverPath.replace(/\\/g, '/')
+    }
 
     // TODO: Split into batches
     const proms = audioFiles.map(af => {
@@ -53,7 +55,7 @@ class AudioMetadataMangaer {
     this.emitter('audio_metadata_finished', itemAudioMetadataPayload)
   }
 
-  updateAudioFileMetadata(libraryItemId, audioFile, outputDir, metadataFilePath, coverPath) {
+  updateAudioFileMetadata(libraryItemId, audioFile, outputDir, metadataFilePath, coverPath = '') {
     return new Promise((resolve) => {
       const resultPayload = {
         libraryItemId,
@@ -75,10 +77,6 @@ class AudioMetadataMangaer {
         },
         {
           input: metadataFilePath
-        },
-        {
-          input: coverPath,
-          options: ['-f image2pipe']
         }
       ]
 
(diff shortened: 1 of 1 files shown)


server/managers/AudioMetadataManager.js
```diff
@@ -32,9 +32,11 @@ class AudioMetadataMangaer {
     var metadataFilePath = Path.join(outputDir, 'metadata.txt')
     await writeMetadataFile(libraryItem, metadataFilePath)
 
+    var coverPath = libraryItem.media.coverPath.replace(/\\/g, '/')
```
advplyr, line 35: `libraryItem.media.coverPath` could be null here, we need to check if the cover exists

server/managers/AudioMetadataManager.js
```diff
@@ -73,6 +75,10 @@ class AudioMetadataMangaer {
         },
         {
           input: metadataFilePath
+        },
+        {
+          input: coverPath,
+          options: ['-f image2pipe']
```
advplyr, line 81: We should only add this if the cover exists

server/managers/AudioMetadataManager.js
```diff
@@ -84,7 +90,7 @@ class AudioMetadataMangaer {
         Ffmpeg premapped id3 tags: https://wiki.multimedia.cx/index.php/FFmpeg_Metadata
       */
 
-      const ffmpegOptions = ['-c copy', '-map_chapters 1', '-map_metadata 1', `-metadata track=${audioFile.index}`, '-write_id3v2 1', '-movflags use_metadata_tags']
+      const ffmpegOptions = ['-c copy', '-map_chapters 1', '-map_metadata 1', `-metadata track=${audioFile.index}`, '-write_id3v2 1', '-movflags use_metadata_tags', '-c:v copy', '-map 2:v', '-map 0:a']
```
advplyr, line 93: Also should only add the video codec copy and map if the cover exists

server/managers/AudioMetadataManager.js
```diff
@@ -85,6 +89,32 @@ class AudioMetadataMangaer {
...
+        var ffmpegCoverPathOptions = [
+          '-c:v copy',
+          '-map 2:v',
+          '-map 0:a'
+        ]
+
+        ffmpegInputs.push(ffmpegCoverPathInput)
+        Logger.debug(`[AudioFileMetaDataManager] Cover found for "${audioFile.metadata.filename}". Cover will be merged to metadata`)
+      } else {
+        // remove the video stream to account for the user getting rid an existing cover in abs
```
advplyr, line 107: This isn't necessary actually since this is encoding the metadata into a new audio file which then gets moved and overwrites the existing audio file. See https://github.com/advplyr/audiobookshelf/blob/master/server/managers/AudioMetadataManager.js#L107

The reason for this is because ffmpeg doesn't allow the input and output files to be the same, so we need to write to a new file.

advplyr: Disregard, I see what you mean now. If the audio file already has a cover then if it isn't explicitly removed it will remain with the audio file.

PR #697, opened 2022-06-06
Use `show` and `episode_id` tags for audiobook series

jvanbruegge:
FFmpeg only supports a very limited number of tags for m4b files (see https://wiki.multimedia.cx/index.php?title=FFmpeg_Metadata). `series` and `series-part` are not possible to use. To work around that, `show` and `episode_id` are added as second option.

### DIFF
--- a/server/utils/prober.js
+++ b/server/utils/prober.js
@@ -70,11 +70,6 @@ function tryGrabChannelLayout(stream) {
   return String(layout).split('(').shift()
 }
 
-function tryGrabTag(stream, tag) {
-  if (!stream.tags) return null
-  return stream.tags[tag] || stream.tags[tag.toUpperCase()] || null
-}
-
 function tryGrabTags(stream, ...tags) {
   if (!stream.tags) return null
   for (let i = 0; i < tags.length; i++) {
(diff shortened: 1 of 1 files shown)


server/utils/prober.js
```diff
@@ -182,8 +182,8 @@ function parseTags(format, verbose) {
     file_tag_comment: tryGrabTags(format, 'comment', 'comm', 'com'),
     file_tag_description: tryGrabTags(format, 'description', 'desc'),
     file_tag_genre: tryGrabTags(format, 'genre', 'tcon', 'tco'),
-    file_tag_series: tryGrabTag(format, 'series'),
-    file_tag_seriespart: tryGrabTag(format, 'series-part'),
+    file_tag_series: tryGrabTag(format, 'series', 'show'),
```
advplyr, line 185: We need to use the method `tryGrabTags` here to support passing in multiple tags. They should all be using `tryGrabTags` actually and the `tryGrabTag` method isn't necessary. That is just a remnant of when only 1 tag was accepted in for each mapping.

advplyr:
On a similar note there is currently an experimental tool for mapping metadata to audio files. I commented there about the accepted m4b meta tags https://github.com/advplyr/audiobookshelf/blob/master/server/managers/AudioMetadataManager.js#L84

For the first iteration I opted to use the same set of meta tags on both mp3 and m4b and for the meta tags not supported they just get embedded as custom tags. I'm not sure if there is a downside to this yet but with ffmpeg we can both read and write the same set of tags on mp3 and m4b.

jvanbruegge:
Thanks for pointing me to [`-movflags use_metadata_tags`](https://superuser.com/questions/1208273/how-to-add-new-and-non-defined-metadata-to-an-mp4-file/1208277#1208277) I was not aware of that

PR #699, opened 2022-06-06
Update some instances of mkdir to ensureDir (#698)

### LINKED ISSUE #698: [Bug]: Merge does not create /metadata/downloads
### Describe the issue

Tried the merge into m4b feature but it fails with

```[AbMergeManager] Failed to make directory /metadata/downloads/abmerge_qkrbjaje2jmop142vx```

fixed it by manually creating ```/metadata/downloads```

hope this helps

regards

Fabian

1. Try merge multiple files into one

### Audiobookshelf version

v2.0.20

Docker

jmt-gh:
This PR resolves #698.

mkdir was being used instead of [ensureDirs, or mkdirs or mkdirp](https://github.com/jprichardson/node-fs-extra/blob/master/docs/ensureDir.md). AbMergeManager doesn't precreate the Downloads folder structure it needes -- updating it to use ensureDirs handles creating the entire path needed when a merge attempt gets kicked off.

I also did a driveby and updated mkdir for CacheManager to also leverage ensureDirs. CacheManager was a bit safer already, but if the first mkdir had failed for any reason, the resulting ones would have failed as well, resulting in a server crash at startup. While changing these to ensureDir doesn't necessarily guarantee whatever might cause the first creation to fail won't happen again on these subsequent attempts, it will at least mean that we'll try and recreate the entire structure needed a few times.

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -139,7 +139,8 @@ class Server {
     await this.checkUserMediaProgress() // Remove invalid user item progress
     await this.purgeMetadata() // Remove metadata folders without library item
     await this.cacheManager.ensureCachePaths()
-
+    await this.abMergeManager.ensureDownloadDirPath()
+    
     await this.backupManager.init()
     await this.logManager.init()
     this.podcastManager.init()
--- a/server/managers/AbMergeManager.js
+++ b/server/managers/AbMergeManager.js
@@ -73,9 +73,10 @@ class AbMergeManager {
 
 
     try {
-      await fs.mkdir(download.dirpath)
+      await fs.ensureDir(download.dirpath)
     } catch (error) {
       Logger.error(`[AbMergeManager] Failed to make directory ${download.dirpath}`)
+      Logger.debug(`[AbMergeManager] Make directory error: ${error}`)
       var downloadJson = download.toJSON()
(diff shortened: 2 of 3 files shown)


advplyr:
They used to be ensureDir but I changed it to first check if the dir exists then create it if it doesn't that way we can set the correct permissions and owner of the dirs.
If we use ensureDir there then the `/metadata/downloads/` folder won't have the uid/gid passed in by the user.

advplyr:
We just need to make sure that all folders getting created have the permissions set. I wrote a method in `utils/filePerms.js` that will set the permissions and owner for files or folders.

jmt-gh:
Updated.

I changed the approach I was taking to now match that of CacheManager, including updating the file perms, and creating the download directory upon server start. This will make it so that if any other Manager needs the `metadata/downloads` folder in the future, we don't need to duplicate any logic of ensuring the directory is created, etc.

I did notice that a DownloadManager class was a thing, but didn't see it implemented anywhere, so I chose to stick with putting this in AbMergeManager (I elected not to put it in AudioMetadataMangaer, as that class only uses downloads as a transient location that seems to not need to care about permissions on the `metadata/downloads` foder)

PR #709, opened 2022-06-09
Add serverVersion to login response

jmt-gh:
This commit adds the serverVersion to the login response. This will be used to show the current server version of the connected server in the mobile app


jmt-gh:
this actually already exists in serverSettings. woops!

PR #716, opened 2022-06-11
Add support for leveraging chapter data directly from Overdrive mp3s during scanning

jmt-gh:
Overdrive MP3s come in multiple parts when downloaded from Overdrive with each part containing a certain number of chapters. Each part also includes a special metadata tag, called `OverDrive MediaMarkers` that describes the timestamps for those chapters (relative to the part) in an XML format.

This PR adds support to ABS to natively parse the `OverDrive MediaMarkers` metadata, and offer the option to use it to populate the chapter metadata in ABS during scanning. Users can toggle this functionality as a server setting from the server settings UI. 

Once enabled, it behaves similarly to other metadata scanning:
- For existing books, it requires a force re-scan (either individual, or at the library level) to apply
- For new books, it will take place automatically upon a regular scan, a force re-scan, or the automated watcher scan
- For new and existing books that do not have this meta tag, ABS will fall back to the standard chapter logic

 
**Future Work**
There are two things I'd like to look at in future PRs:
1. I'm not super happy with how `Book` is now having to be made aware of server settings. I'd like to continue extracting out the logic for setChapters further, and creating `server/parsers/chapters`, which will hold different ways for us to parse and generate chapter information. This will then also allow us to handle decision logic on which parser further up in one of the scanners, outside of `Book`, and inject it directly as a dependency.
2. Bringing this to the client-side experience. Particularly the edit-chapter page, to allow users to leverage this on a per-book basis without having to enable it globally

### DIFF
--- a/server/controllers/PodcastController.js
+++ b/server/controllers/PodcastController.js
@@ -103,7 +103,7 @@ class PodcastController {
         Logger.error('Invalid podcast feed request response')
         return res.status(500).send('Bad response from feed request')
       }
-      Logger.debug(`[PdocastController] Podcast feed size ${(data.data.length / 1024 / 1024).toFixed(2)}MB`)
+      Logger.debug(`[PodcastController] Podcast feed size ${(data.data.length / 1024 / 1024).toFixed(2)}MB`)
       var payload = await parsePodcastRssFeedXml(data.data, false, includeRaw)
       if (!payload) {
         return res.status(500).send('Invalid podcast RSS feed')
--- a/server/scanner/Scanner.js
+++ b/server/scanner/Scanner.js
@@ -316,7 +316,7 @@ class Scanner {
 
   async scanNewLibraryItemDataChunk(newLibraryItemsData, libraryScan) {
     var newLibraryItems = await Promise.all(newLibraryItemsData.map((lid) => {
-      return this.scanNewLibraryItem(lid, libraryScan.libraryMediaType, libraryScan.preferAudioMetadata, libraryScan.preferOpfMetadata, libraryScan.findCovers, libraryScan.preferOverdriveMediaMarker, libraryScan)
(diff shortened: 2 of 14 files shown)


server/scanner/Scanner.js
```diff
@@ -310,7 +316,7 @@ class Scanner {
 
   async scanNewLibraryItemDataChunk(newLibraryItemsData, libraryScan) {
     var newLibraryItems = await Promise.all(newLibraryItemsData.map((lid) => {
-      return this.scanNewLibraryItem(lid, libraryScan.libraryMediaType, libraryScan.preferAudioMetadata, libraryScan.preferOpfMetadata, libraryScan.findCovers, libraryScan)
+      return this.scanNewLibraryItem(lid, libraryScan.libraryMediaType, libraryScan.preferAudioMetadata, libraryScan.preferOpfMetadata, libraryScan.findCovers, libraryScan.scanOptions.preferOverdriveMediaMarker, libraryScan)
```
jmt-gh, line 319: I'm very confused as to why this requires scanOptions, but the other onces, like preferOpfMetadata, don't?

advplyr: LibraryScan.js has getters set up just to make it easier access.

server/controllers/PodcastController.js
```diff
@@ -103,7 +103,7 @@ class PodcastController {
         Logger.error('Invalid podcast feed request response')
         return res.status(500).send('Bad response from feed request')
       }
-      Logger.debug(`[PdocastController] Podcast feed size ${(data.data.length / 1024 / 1024).toFixed(2)}MB`)
+      Logger.debug(`[PodcastController] Podcast feed size ${(data.data.length / 1024 / 1024).toFixed(2)}MB`)
```
jmt-gh, line 106: drive-by typo fix

jmt-gh:
Decided to remove the client side work from this PR. I'll come back to it in a future one

advplyr:
Awesome, nice job! I agree with your future work ideas. The server settings for the scanner need to be planned out better as this is getting messy. The scanner still needs to get optimized for multi-threading which will probably require a significant rewrite.

PR #728, opened 2022-06-14
[DRAFT] Inject chapter scanner dependency in to Book

jmt-gh:
This commit implements depdency injection for the chapter scanner to be used in Book. In a previous commit when the ovedrive media marker support was added, Book was making it's own decisions about which scanner to use, but that opened it up to needing to know a lot about serverSettings which felt wrong.

Instead, now we will set the chapterScanner property on Book to either OverdriveMediaMarkers or parsing it directly from MediaFile.

This required converting both parses to classe, of which I decided to make them static utility classes, as instanced classes wouldn't provide much value here.

**Todo / Open Questions / still being investigated**
- **Q**: Is Scanner the right place to handle the logic of figuring out which parser to pass in? If not, where?
- **Todo**: support the rest of the scan types (right now it only supports single book re-scan)
- **Todo**: Clean up the rest of the old OverdriveMediaMarker preferences code that was being passed around
- **Todo**: Add some class level documentation to both parsers


server/scanner/Scanner.js
```diff
@@ -64,6 +68,11 @@ class Scanner {
   async scanLibraryItem(libraryMediaType, folder, libraryItem) {
     // TODO: Support for single media item
     var libraryItemData = await getLibraryItemFileData(libraryMediaType, folder, libraryItem.path, false, this.db.serverSettings)
+    var useOverdriveChapters = this.db.serverSettings.scannerPreferOverdriveMediaMarker
+
+    // Depending on the users setting, tell Book to use a specific parser for chapters
+    libraryItem.media.chapterParser = (useOverdriveChapters ? OverdriveMediaMarkersChapterParser : MediaFileChapterParser)
```
jmt-gh, line 74: I think this might actually be too early in the stack to do this? We have other logic around scanMediaFiles:220 that does book explicit changes to Book (adding media files) -- it might make more sense to do there? I'll need to see if bringing in serverSettings to scanMediaFiles feels OK or not though

advplyr: The design of the Scanner class isn't great because there would be 3 places you need to set this. `scanNewLibraryItem` and `rescanLibraryItem`.
I'm thinking now that defining chapterParser in the class constructor of Book.js may be uglier then passing in the `preferOverdriveMediaMarker` to the setChapters. I'm not sure though, neither seem optimal.

jmt-gh: >I'm thinking now that defining chapterParser in the class constructor of Book.js

This doesn't actually define it in the constructor. It just leverages the fact that JS classes can take arbitrary properties at run time. Being new to JS, I have no idea how big of a taboo this is haha

>may be uglier then passing in the preferOverdriveMediaMarker to the setChapters

Using composition to build the right set of functionality for Book feels a bit more OOP-y though, and will longer term let us simplify things, as neither Book, nor the developer will need to understand the state of the application through serverSettings to be able to understand chapter parsing logic. I can see a world where we offer additional chapter parsers for the scanner, such as using audnexus by default, and the non-composition path feels like it gets even more gross at that point

advplyr:
This is outdated now. I recently rewrote the scanner and now overdrive media markers are used if they are found. There is no server setting for it anymore

PR #730, opened 2022-06-16
Properly support multiple narrators when Matching (#719)

### LINKED ISSUE #719: When Matching, Separate Narrators
Ok at the moment when you match a book with say Audible, if you have multiple authors, ABS recognises the commas separate the authors and adds them individually.

It doesn’t do this for narrators, and just adds multiple narrators as one big single entry, would it be possible for it to crate separate entries for each narrator instead of just one big long one with them all.

jmt-gh:
This commit fixes #719 adds support for multiple narrators when matching. We were accidentally missing a split in the same way we handle it for genres and tags.

Screenshot of this fix with the ASIN data linked in #719:

image

### DIFF
--- a/client/components/modals/item/tabs/Match.vue
+++ b/client/components/modals/item/tabs/Match.vue
@@ -428,7 +428,7 @@ export default {
             )
             updatePayload.metadata.authors = authorPayload
           } else if (key === 'narrator') {
-            updatePayload.metadata.narrators = [this.selectedMatch[key]]
+            updatePayload.metadata.narrators = this.selectedMatch[key].split(',').map((v) => v.trim())
           } else if (key === 'genres') {
             updatePayload.metadata.genres = this.selectedMatch[key].split(',').map((v) => v.trim())
           } else if (key === 'tags') {


client/components/modals/item/tabs/Match.vue
```diff
@@ -428,7 +428,7 @@ export default {
             )
             updatePayload.metadata.authors = authorPayload
           } else if (key === 'narrator') {
-            updatePayload.metadata.narrators = [this.selectedMatch[key]]
+            updatePayload.metadata.narrators = this.selectedMatch[key].split(',').map((v) => v.trim())
```
jmt-gh, line 431: @advplyr , I noticed in server/providers/Audible.js line 22, we are taking the response from Audible that is already an array, and creating a singular string out of it:

22: narrator: narrators ? narrators.map(({ name }) => name).join(', ') : null,

We do this with genres and a few other fields that come back as arrays as well:
28: genres: genresFiltered.length > 0 ? genresFiltered.map(({ name }) => name).join(', ') : null,

I was just curious why we were changing it to a string, just for us to need to change it back later?

advplyr: There is no good reason to still be doing that. Originally we didn't support multiple narrators, so it was a string everywhere, but now we should return an array from there.
I also thought I fixed the multiple narrators already in the matching so I'm looking into that now.

advplyr: Nevermind, it was actually genres that I fixed the other week.

jmt-gh: We're changing authors, narrators, genres, and tags all to strings here just to convert them back to arrays on the front end. Should I use this PR to update all of them to have Audible.js just return arrays?

advplyr: Yeah I think that would be good. The only thing is that the GoogleBooks.js would need to be updated to return genres as an array. The providers should all return the same formats if they are returning the same data.
Audible is the only provider that returns narrators and tags. Then it looks like GoogleBooks is the only other one not returning genres as an array.

jmt-gh: I dug a bit deeper in to this and I think we need to leave it for now -- or at the least, I'm not sure I'd like to push this further for the refactor in this ticket.

If we change the providers to return Arrays, things work fine if you don't edit any of the returned metadata in the Update Book Details portion of the modal during the match flow. But if you edit any of that data, the changes what is passed in to selectedMatch from the Array returned by the provider, to a string returned by the input box. Either we'd need to add logic to these specific input boxes to return an array still (can we even do that?), or have the provider return an array, and then do conditional || logic to check if the .map failed, and try a split first (at which point we may as well just keep the logic in the provider to join it all to a string). 

I think we should merge this as is, and look at this either in the future as a tech debt todo, or just decide we'll live with it for now

advplyr: I'm good with that, thanks!

Issue #624, opened 2022-05-23
[Enhancement]: Date formatting


jt196:
### Describe the feature/enhancement

Would it be possible for the user to be able to set their date formatting? Switch between European/US and maybe ISO8601? Can't see anywhere to do this in the settings section.

cassieesposito:
I'd be willing to take this on if you can point out places that dates appear.

jt196:
So, date is displayed when the user listened info (currently mm/dd/yy):

![image](https://user-images.githubusercontent.com/8290163/174255724-fea74598-1537-424e-9fde-0c70981e7a3c.png)

Also, on the sorting by date added (currently mm/dd/yy):

![image](https://user-images.githubusercontent.com/8290163/174256068-0f18931f-d971-406e-8ce2-99dc7ce97476.png)

In the Users section in the settings (currently mmm dd, yyyy):

![image](https://user-images.githubusercontent.com/8290163/174256194-99e3b2d5-4d1e-4d12-83f8-a28d7f3cf088.png)

cassieesposito:
https://github.com/advplyr/audiobookshelf/pull/736 deals with the first two. I'm going to leave the users section alone for now.

PR #736, opened 2022-06-17
Date format

cassieesposito:
Implementing feature request: https://github.com/advplyr/audiobookshelf/issues/624


client/store/globals.js
```diff
@@ -12,7 +11,21 @@ export const state = () => ({
   selectedCollection: null,
   selectedAuthor: null,
   isCasting: false, // Actively casting
-  isChromecastInitialized: false // Script loaded
+  isChromecastInitialized: false, // Script loaded
+  dateFormats: [
```
cassieesposito, line 15: I don't love my choice to put dateFormats in state since it's not really.
But I was struggling to access the list when I put it other places, and also I'm not sure where the right place to put it is.
I'm open to suggestions about this.

advplyr: I'm not sure we have a great place to put this yet. I think it's good here for now, no need to overthink it. We can consolidate data like this later.

advplyr:
The way you opened the PR does not allow me to add contributions to it. I'm not sure how to enable that but in some PRs I can commit directly. Can you see if you can update this so that I can commit?

cassieesposito:
> Can you see if you can update this so that I can commit?

Try now?

advplyr:
What did you do to update it?

cassieesposito:
I checked a box on the side that says "Allow edits and access to secrets by maintainers"

advplyr:
I decided to break out the appearance related settings into their own section.

image

PR #742, opened 2022-06-18
Fetch author photo from external URL

mrdth:
At the moment, there is no easy way to add a photo for an author, if one isn't found during a (quick)match

This PR adds a new text field 'Photo URL' on the author edit modal, shown only if there is no existing image for an author. When submitted the image is saved from the URL provided, using the existing AuthorFinder.saveAuthorImage() method.

### DIFF
--- a/client/components/modals/authors/EditModal.vue
+++ b/client/components/modals/authors/EditModal.vue
@@ -25,8 +25,8 @@
                 <ui-text-input-with-label v-model="authorCopy.asin" :disabled="processing" label="ASIN" />
               </div>
             </div>
-            <div class="p-2" v-show="!author.imagePath">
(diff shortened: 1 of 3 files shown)


advplyr:
I wasn't sure how this was going to be implemented because I also wanted to allow uploading images like you can do with item covers. The UI/UX for this is not great, but for now I figure we can just make it more like items where the image input is used for the image path but also accepts a URL.

Since we also need to add better matching for authors I thought we could mirror the edit modal for items by adding tabs for "Details", "Covers" and "Match".  That can be a future update.

mrdth:
> I wasn't sure how this was going to be implemented because I also wanted to allow uploading images like you can do with item covers. The UI/UX for this is not great

Agreed - I should have put in the PR that I see this as a temp solution, at least as far as UI goes

PR #758, opened 2022-06-20
Update Settings page UI

jmt-gh:
This PR updates the settings page UI. This biggest changes here are a more distinct visual hierarchy of the content, as well as moving the content to 2 columns when above the small break-point to reduce scrolling. With the groupings introduced in 8ab0f164, I also streamlined the text to be less redundant, which helps with the cognitive load of looking at a bunch of text at once

**Before (desktop):**
image

**After (desktop):**
image

**Before (mobile-web):**
image

**After (mobile-web):**
image

### DIFF
--- a/client/pages/config/index.vue
+++ b/client/pages/config/index.vue
@@ -5,9 +5,9 @@
         <h1 class="text-xl">Settings</h1>
       </div>
 
-      <div class="sm:flex">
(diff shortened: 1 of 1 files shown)


advplyr:
Nice, I like it! I made a few tweaks.

PR #770, opened 2022-06-23
Show chapter duration in the "now playing" chapter modal (Issue #767)

### LINKED ISSUE #767: [Feature Request]: Show Chapter Length

At the moment in the chapter menu we can see the title of the chapter and the absolute start time of it inside the audiobook.
While it might be useful to know when each chapter start, I think that knowing the duration (relative difference between each chapter) can be more interesting.
Having a switch in the settings to toggle between the two display mode would be ideal.

jmt-gh:
This PR resolves #767 by adding the chapter duration alongside the chapter start timestamp. I went with showing them both at once rather than a toggle in the settings to change between these, as having it be a toggle could lead to more confusion (forgetting you had it set and not knowing why the times were not linear, or sharing with multiple users who have different preferences). If we end up adding the concept of user-scoped settings in the future, we should look at moving this over to there. 

Before
image

After
image

### DIFF
--- a/client/components/modals/ChaptersModal.vue
+++ b/client/components/modals/ChaptersModal.vue
@@ -1,11 +1,14 @@
 <template>
-  <modals-modal v-model="show" name="chapters" :width="500" :height="'unset'">
+  <modals-modal v-model="show" name="chapters" :width="600" :height="'unset'">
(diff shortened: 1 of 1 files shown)


advplyr:
Updating sizes for mobile

image

PR #776, opened 2022-06-25
fix: use chapter duration when seeking on track bar

alexmaras:
Fixes an issue where seeking by clicking the track bar acts as if the whole book's duration is being seeked.


alexmaras:
Just found a bug with this - will fix

alexmaras:
Sorted - I forgot to add the chapter start time to the seek time

advplyr:
Thanks, missed that one!

PR #791, opened 2022-06-29
fix: disable workbox to prevent failure with service worker

alexmaras:
This fixes a variety of lockups and failures to play on both chrome and firefox. Previously, rapidly clickiing through the application with workbox enabled (the default in nuxt production environments) would cause chrome to practically lock up, and firefox would fail to play files until the entire m4b had downloaded. 

In theory, this may cause issues with offline caching and playback. I'm not sure what level of offline functionality the audiobookshelf app is supposed to support. This is probably worth a fairly thorough test.

The change makes the service significantly more robust in my short time testing with the change.

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -120,6 +120,9 @@ module.exports = {
           sizes: "512x512"
         }
       ]
+    },
+    workbox: {
+      enabled: false,
     }
   },
 


advplyr:
I didn't change any of the defaults of workbox that come with the nuxt PWA module. Since the PWA is not going to play audio offline I don't think it is important to cache.

advplyr:
I realized that the PWA has not been installable since this update. I added back the workbox module but attempting to prevent caching.
This will probably require some more testing.

Issue #800, opened 2022-07-03
[Bug]: Cache directory not recreated after purge


mcdinner:
### Describe the issue

After purging the cache directory in settings the cache folder is not recreated.
After restart / container restart the cache folder is recreated.

As I can see the cachePathsExist property in the CacheManager does not get reset on purge.

### Steps to reproduce the issue

1.  Go to Settings and purge cache folder.

### Audiobookshelf version

v2.0.22

### How are you running audiobookshelf?

Docker

advplyr:
Fixed in [v2.0.24](https://github.com/advplyr/audiobookshelf/releases/tag/v2.0.24)

PR #801, opened 2022-07-03
Remove cachePathExists property (Issue #800)

mcdinner:
Remove cachePathsExist property to ensure missing cache directories are recreated when EnsureCachePaths() called.


advplyr:
I put that in to try to avoid checking every time but you're right this will cause problems if clearing cache.

PR #821, opened 2022-07-09
Add ability to view current version's changelog from within ABS

jmt-gh:
This PR adds in the ability to view the current running version's changelog directly within ABS. We were already grabbing this information to derive the latest version, so we just hook in to that to extract the changelog for the current running version. There is a future enhancement here if we want it, to prompt this modal to the user once upon the first time they log in for a new version.

I noticed a few, what felt like at first glance, slightly different variants in our modal implementations, so I mimiced the RSSFeedModal since it was relatively new.

### New Package Added
**Name** [Marked](https://github.com/markedjs/marked)
**Number of dependencies**: 0
**Reason**: The Github API returns the changelog in the raw [Github Flavor Markdown](https://github.github.com/gfm/) that it was written in. Marked allows us to parse that to HTML, with explicit support for Github Flavor Markdown. 

### **Screenshots** (1920x1080)
**Home screen**
image
image

**Config screen**
image
image


client/components/app/SideRail.vue
```diff
@@ -150,17 +154,21 @@ export default {
     hasUpdate() {
       return !!this.versionData.hasUpdate
     },
-    latestVersion() {
```
jmt-gh, line 153: this wasn't being used on this page, so I did a drive by cleanup

PR #842, opened 2022-07-18
Add restart and fix volumes

revilo951:
Added `restart: unless-stopped`
Adjust volumes - probably shouldn't be scattering the volumes around the root dir? Could also do this to put them in the home dir instead of relative to the .yml file location:
```
      - ~/abs/audiobooks:/audiobooks
      - ~/abs/metadata:/metadata
      - ~/abs/config:/config
```
https://docs.docker.com/compose/compose-file/compose-file-v3/#short-syntax-3

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,7 @@ services:
     ports:
       - 13378:80
     volumes:
-      - /audiobooks:/audiobooks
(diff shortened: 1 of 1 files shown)


advplyr:
I think this is fine. Not everyone is running docker-compose on linux

PR #946, opened 2022-08-28
Clear selectedEpisodes on download.  

### LINKED ISSUE #931: [Bug]: [Podcasts][Find Episodes] Downloaded Episodes Remain Selected Post-Download, Results In Unnecessary Extra Downloads
### Describe the issue

This is a bug in the Find Episodes dialog that comes up when you click the Find Episodes button from within any podcast in the library.  It looks like the logic for when finishing the batch download isn't clearing out everything at the end like it needs to. Also there's no check for an existing file before a download is done?

1. open a podcast where you have episodes left to download
2. open Find Episodes
3. Click the checkbox on 1 or more episodes and then click "Download # Episode(s)" button
4. open Find Episodes again
5. the "Download # Episode(s)" button is now stuck with the count of all currently downloaded episodes. If clicked, all episodes will be re-downloaded again

To escape this condition you must refresh the page.

### Audiobookshelf version

2.1.4

Docker

barrycarey:
Clears the selectedEpisodes once the download is started.  Fixes #931.  

Also, clicking on the checkbox for existing downloads still added that item to the list.  Moved onClick to checkbox instead of container div to prevent this.

### DIFF
--- a/client/components/modals/podcast/EpisodeFeed.vue
+++ b/client/components/modals/podcast/EpisodeFeed.vue
@@ -12,11 +12,11 @@
           :key="index"
           class="relative"
           :class="episode.enclosure && itemEpisodeMap[episode.enclosure.url] ? 'bg-primary bg-opacity-40' : selectedEpisodes[String(index)] ? 'cursor-pointer bg-success bg-opacity-10' : index % 2 == 0 ? 'cursor-pointer bg-primary bg-opacity-25 hover:bg-opacity-40' : 'cursor-pointer bg-primary bg-opacity-5 hover:bg-opacity-25'"
-          @click="toggleSelectEpisode(index)"
+
         >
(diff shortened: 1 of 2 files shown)


barrycarey:
Idiot move on my part making a PR 10 minute after looking at the code for the first time.  Will reopen tomorrow with a proper fix.

advplyr:
I'm looking through this modal and it is a mess so I'm going to make some more updates in your PR. I also think the toggle shouldn't requiring clicking the checkbox.

barrycarey:
Sounds good. 

As for the toggle, do you mean it should active by clicking clicking anywhere on the line? 

Also, the setting of selected episodes in the init seems unnecessary.

advplyr:
Yeah that's what I mean and yeah the init is unnecessary. I'm going to do a refactor now

PR #955, opened 2022-08-30
Ability to decode HTML Entities when all tags are stripped. Fixes #929

### LINKED ISSUE #929: [Bug]: Ampersand symbol not parsed correctly in description
### Describe the issue

I'm not sure if this is a metadata source issue, I'm using the Audible provider, or a ABS issue but the ampersand symbol is being displayed as "& a m p ;" in the description.

Using the Audible provider search the book "I'm Glad My Mom Died" by Jennette McCurdy. The description will contain "& a m p ;" where it should be &

### Audiobookshelf version

v2.1.4

Docker

barrycarey:
Using the list of entities linked in the issue.  Made the decoding optional in case there's a future need to skip the decode.


server/utils/htmlSanitizer.js
```diff
@@ -17,12 +18,22 @@ function sanitize(html) {
...
   }
 
-  return sanitizeHtml(html, sanitizerOptions)
+  let sanitized = sanitizeHtml(html, sanitizerOptions)
+  return shouldDecodeEntities ? decodeHTMLEntities(sanitized) : sanitized
+}
+module.exports.stripAllTags = stripAllTags
+
+function decodeHTMLEntities(strToDecode) {
+  return strToDecode.replace(/\&([^;]+);/g, function (entity) {
```
advplyr, line 33: It looks like some of the entities don't end in a semicolon. At quick glance that is what this regex looks like it is matching for.

barrycarey: Good catch.  Just pushed an update making it optional.

advplyr:
Works great, thanks!

PR #961, opened 2022-09-02
Pass lib ID to toolbar so it can refresh state when clearing issues.

barrycarey:
This seems like a clean way to do this.  Although I've never used Vuex so feel free to slap my wrist if this isn't right.

### DIFF
--- a/client/components/app/BookShelfToolbar.vue
+++ b/client/components/app/BookShelfToolbar.vue
@@ -89,7 +89,8 @@ export default {
     authors: {
       type: Array,
       default: () => []
-    }
+    },
+    libraryId: String
   },
   data() {
     return {
@@ -215,6 +216,7 @@ export default {
             this.$toast.success('Removed library items with issues')
             this.$router.push(`/library/${this.currentLibraryId}/bookshelf`)
             this.processingIssues = false
+            this.$store.dispatch('libraries/fetch', this.libraryId)
           })
           .catch((error) => {
             console.error('Failed to remove library items with issues', error)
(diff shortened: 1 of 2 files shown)


client/pages/library/_library/bookshelf/_id.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="page" :class="streamLibraryItem ? 'streaming' : ''">
-    <app-book-shelf-toolbar :page="id || ''" :view-mode.sync="viewMode" />
+    <app-book-shelf-toolbar :page="id || ''" :view-mode.sync="viewMode" :library-id="libraryId"/>
```
advplyr, line 3:
```suggestion
    <app-book-shelf-toolbar :page="id || ''" :view-mode.sync="viewMode" />
```

client/components/app/BookShelfToolbar.vue
```diff
@@ -89,7 +89,8 @@ export default {
     authors: {
       type: Array,
       default: () => []
-    }
+    },
+    libraryId: String
```
advplyr, line 93:
```suggestion
    }
```

client/components/app/BookShelfToolbar.vue
```diff
@@ -215,6 +216,7 @@ export default {
             this.$toast.success('Removed library items with issues')
             this.$router.push(`/library/${this.currentLibraryId}/bookshelf`)
             this.processingIssues = false
+            this.$store.dispatch('libraries/fetch', this.libraryId)
```
advplyr, line 219:
```suggestion
            this.$store.dispatch('libraries/fetch', this.currentLibraryId)
```

advplyr (changes requested):
I think this is the right solution so we reload the library data but we can use the `currentLibraryId` instead of passing the library id into the component.

barrycarey:
Right on.  Updated. Didn't catch that the ID was already available.

PR #1005, opened 2022-09-23
Multi select quick match

Undergrid:
Adds a feature to allow the user to batch Quick Match the selected audiobooks.

A button is added to the Appbar when one or more audiobooks are selected using the multi-select option (left most button on the below image).

button

Once clicked, a dialog is shown allowing the user to select the metadata provider and to select whether exiting covers or metadata should be overwritten.

dialog

When "Continue" is selected, a quick match is run against each selected audiobook, updating metadata and covers as required.

### DIFF
--- a/client/components/modals/BatchQuickMatchModel.vue
+++ b/client/components/modals/BatchQuickMatchModel.vue
@@ -1,0 +1,70 @@
+<template>
+  <modals-modal v-model="show" name="batchQuickMatch" :processing="processing" :width="500" :height="'unset'">
+    <template #outer>
+      <div class="absolute top-0 left-0 p-5 w-2/3 overflow-hidden">
+        <p class="font-book text-3xl text-white truncate">{{ title }}</p>
+      </div>
+    </template>
+
+    <div ref="container" class="w-full rounded-lg bg-primary box-shadow-md overflow-y-auto overflow-x-hidden" style="max-height: 80vh">
+      <div v-if="show" class="w-full h-full">
+        <div class="py-4 px-4">
+          <h1 class="text-2xl">Quick Match {{ selectedBookIds.length }} Books</h1>
+        </div>
+      </div>
+    </div>
+  </modals-modal>
+</template>
+
+<script>
+export default {
+  data() {
+    return {
+      processing: false
+    }
+  },
+  computed: {
+    show: {
+      get() {
+		  console.log("Getter")
+        return this.$store.state.globals.showBatchQuickMatchModal
+      },
+      set(val) {
+		  console.log("Setter")
+        this.$store.commit('globals/setShowBatchQuickMatchModal', val)
+      }
+    },
+    title() {
+      return `${this.selectedBookIds.length} Items Selected`
+    },
+    showBatchQuickMatchModal() {
+      return this.$store.state.globals.showBatchQuickMatchModal
+    },
+    selectedBookIds() {
+      return this.$store.state.selectedLibraryItems || []
+    },
+    currentLibraryId() {
+      return this.$store.state.libraries.currentLibraryId
+    }
+  },
+  methods: {
+  },
+  mounted() {}
+}
+</script>
+
+<style>
+.list-complete-item {
+  transition: all 0.8s ease;
+}
+
+.list-complete-enter-from,
+.list-complete-leave-to {
+  opacity: 0;
+  transform: translateY(30px);
+}
+
+.list-complete-leave-active {
+  position: absolute;
+}
+</style>
--- a/client/components/modals/BatchQuickMatchModel.vue
+++ b/client/components/modals/BatchQuickMatchModel.vue
@@ -36,7 +36,7 @@
               </p>
             </ui-tooltip>
           </div>
-          <div class="mt-4 py-4 border-b border-white border-opacity-10 text-white text-opacity-80" :class="isScrollable ? 'box-shadow-md-up' : 'border-t border-white border-opacity-5'">
+          <div class="mt-4 py-4 border-b border-white border-opacity-10 text-white text-opacity-80 border-t border-white border-opacity-5">
             <div class="flex items-center px-4">
               <ui-btn type="button" @click="show = false">Cancel</ui-btn>
               <div class="flex-grow" />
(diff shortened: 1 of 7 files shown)


advplyr:
Awesome. I think we should initialize the provider to whatever the default provider is for the library.

There are some getters in `libraries.js` to get a library provider or get the current library.

Also, we can show the user how many items were updated and how many items were unable to find a match.
Currently the `quickMatchLibraryItem` function is returning a warning if it doesn't find a match. https://github.com/advplyr/audiobookshelf/blob/master/server/scanner/Scanner.js#L695

Undergrid:
How would you like to show the updated and unmatched count?  Would in the toast be okay?

advplyr:
Yeah that works, the toast message can have line breaks `\n`.

Eventually we will probably have to change this so that the API request doesn't need to wait for the matches to finish and instead gets the result from a socket event. The reason for that is because some people have very slow systems and the API request will timeout before it finishes.

Undergrid:
If you can point me at an example of doing that, I can probably get it working that way.  Might as well do it properly from the start.

skyzuma:
is possible to select details if the switch goes to on, like (update only) language, genre, ... (for selected) ?

advplyr:
> If you can point me at an example of doing that, I can probably get it working that way. Might as well do it properly from the start.

I don't have a good example to point to yet for this. We would need to set some temporary id locally for the quick match request and setup a socket event listener (something like `batch_quickmatch_complete`). Once finished the backend emits that event including the temporary id so know it is this clients request that was completed.
Most of the socket event listeners are set in the `default.vue` layout.
If you want to go for this that would be great but up to you.

advplyr:
> is possible to select details if the switch goes to on, like (update only) language, genre, ... (for selected) ?
> 
> li > switch > details ul > swicth > language ul > switch > genre

We don't have support for this in the backend and would be a lot of re-working that is out of the scope of this PR

Undergrid:
So I'm going to look at the websocket part, but in the mean time I've set the provider and added the counts to the toast.  The toast looks a little different depending on the counts:

match
nomatch
matchunmatch
nomatchunmatch

Undergrid:
@advplyr PR has now been updated as request, including with the websocket feedback etc.

client/components/modals/BatchQuickMatchModel.vue
```diff
@@ -0,0 +1,145 @@
...
+          this.$store.commit('setProcessingBatch', false)
+          this.show = false
+        })
+        .catch((error) => {
+          this.$toast.error('Batch quick match failed')
+          console.error('Failed to batch quick match', error)
+          this.processing = false
+          this.$store.commit('setProcessingBatch', false)
+          this.show = false
+        })
```
advplyr, line 139:
```suggestion
        .then(() => {
          this.$toast.info('Batch quick match of ' + this.selectedBookIds.length + ' books started!')
        }).catch((error) => {
          this.$toast.error('Batch quick match failed')
          console.error('Failed to batch quick match', error)
        }).finally(() => {
          this.processing = false
          this.$store.commit('setProcessingBatch', false)
          this.show = false
        })
```

server/controllers/LibraryItemController.js
```diff
@@ -305,6 +305,42 @@ class LibraryItemController {
...
+    var itemsUpdated = 0
+    var itemsUnmatched = 0
+
+    var matchData = req.body
+    var options = matchData.options || {}
+    var items = matchData.libraryItemIds
+    if (!items || !items.length) {
+      return res.sendStatus(500)
+    }
+    res.sendStatus(200);
```
advplyr, line 324:
```suggestion
    res.sendStatus(200)
```

server/controllers/LibraryItemController.js
```diff
@@ -305,6 +305,42 @@ class LibraryItemController {
...
+        } else if (matchResult.warning) {
+            itemsUnmatched++
+        }
+    }
+    
+    var result = {
+      success: itemsUpdated > 0,
+      updates: itemsUpdated,
+      unmatched: itemsUnmatched
+    };
```
advplyr, line 340:
```suggestion
    }
```

client/components/modals/BatchQuickMatchModel.vue
```diff
@@ -0,0 +1,145 @@
...
+      lastUsedLibrary: undefined,
+      options: {
+        provider: undefined,
+        overrideDetails: true,
+        overrideCover: true,
+        overrideDefaults: true
+      },
+      tooltips: {
+          updateCovers: 'Allow overwriting of existing covers for the selected books when a match is located.',
+          updateDetails: 'Allow overwriting of existing details for the selected books when a match is located.'
```
advplyr, line 67:
```suggestion
        updateCovers: 'Allow overwriting of existing covers for the selected books when a match is located.',
        updateDetails: 'Allow overwriting of existing details for the selected books when a match is located.'
```

client/components/modals/BatchQuickMatchModel.vue
```diff
@@ -0,0 +1,145 @@
...
+        <div class="py-4 px-4">
+          <h1 class="text-2xl">Quick Match {{ selectedBookIds.length }} Books</h1>
+        </div>
+        
+        <div class="w-full overflow-y-auto overflow-x-hidden max-h-96">
+          <div class="flex px-8 items-center py-2">
+            <p class="pr-4">Provider</p>
+            <ui-dropdown v-model="options.provider" :items="providers" small />
+          </div>
+          <p class="text-base px-8 py-2">Quick Match will attempt to add missing covers and metadata for the selected books.  Enable the options below to allow Quick Match to overwrite existing covers and/or metadata.</p>
```
advplyr, line 20:
```suggestion
          <p class="text-base px-8 py-2">Quick Match will attempt to add missing covers and metadata for the selected books. Enable the options below to allow Quick Match to overwrite existing covers and/or metadata.</p>
```

client/components/modals/BatchQuickMatchModel.vue
```diff
@@ -0,0 +1,145 @@
...
+    currentLibraryId() {
+      return this.$store.state.libraries.currentLibraryId
+    },
+    providers() {
+      if (this.isPodcast) return this.$store.state.scanners.podcastProviders
+      return this.$store.state.scanners.providers
+    },
+    libraryProvider() {
+      return this.$store.getters['libraries/getLibraryProvider'](this.currentLibraryId) || 'google'
+    },
```
advplyr, line 105:
```suggestion
    }
```

advplyr (changes requested):

advplyr:
Works beautifully, thanks!

Issue #1004, opened 2022-09-23
Audiobook Scanner- Files using the xHE-AAC Codec cause the library scan process to crash


SuburbanVampire:
Effected Version: Docker 2.1.5 running Debian x64 host

The codec xHE-AAC is a proprietary version of AAC from Fraunhofer. It is protected by patents and has no compatible licensed version that can be used with FFMPEG.

It is like this format will never be supported by FFMPEG. 

Currently, if the audiobook library scanner encounters a file using this codec, the process halts immediately, looks like an unhandled exception.

Steps to reproduce: Place any xHE-AAC media file in a scannable path and let scan run automatically or manually.

Example files here: https://www2.iis.fraunhofer.de/AAC/xhe-aac-abr.html

advplyr:
Fixed in [v2.2.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.0)

This fix just prevents the crash, the audio file will still be invalid.

PR #1007, opened 2022-09-24
Issue 1004

Undergrid:
Possible fix for Issue #1004

### DIFF
--- a/server/libs/nodeFfprobe/index.js
+++ b/server/libs/nodeFfprobe/index.js
@@ -20,7 +20,14 @@ module.exports = (function () {
 
       proc.on('exit', code => { exitCode = code })
       proc.on('error', err => reject(err))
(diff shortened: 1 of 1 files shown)


server/libs/nodeFfprobe/index.js
```diff
@@ -20,7 +20,14 @@ module.exports = (function () {
 
       proc.on('exit', code => { exitCode = code })
       proc.on('error', err => reject(err))
-      proc.on('close', () => resolve(JSON.parse(probeData.join(''))))
+      proc.on('close', () => {
+        try {
+            let result = JSON.parse(probeData.join(''))
+            resolve(result);
```
advplyr, line 26:
```suggestion
            resolve(JSON.parse(probeData.join('')))
```

advplyr (changes requested):

advplyr:
Yep, this should do it, thanks.

PR #1031, opened 2022-10-03
Fix a crash under certain circumstances when updating tags when quick…

Undergrid:
… matching.


advplyr:
Yep that's definitely it. Thanks!

Issue #960, opened 2022-09-01
[Enhancement]: Eliminate Search for Episode modal, all episodes available on default view


gbakeman:
### Describe the feature/enhancement

From issue #615,  @advplyr:

- [ ] Remove the search for episode modal and put the full episode list derived from RSS feed in the same spot downloads currently are.
- [ ] Filtering (move to second issue?):
- - [ ] Listened state (Complete/in progress/never started)
- - [ ] Downloaded state (Complete/in progress/never started)
- - [ ] Live filtering text search in-page for titles and descriptions
- - [ ] Range of published dates? Idk, maybe unnecessary.
- [ ] Smooth animations of filter changes would be cool

PR #1047, opened 2022-10-05
WIP: Show all podcast episodes in podcast view

rhyst:
Very WIP:

- Removes "search for episode modal"
- Fetches podcast feed when adding a new podcast
- Shows all podcast episodes in the episodes list and shows which are downloaded
- Allows downloading from the episodes list
- Allows refreshing the episode list

Thought I'd stop there and get feedback before I carried on because I don't normally work with Vue so I am sure this is terrible 😆 

Questions based on what I have done:
- I create PodcastEpisodes for the undownloaded episodes, isthat a good idea?
- It uses PodcastEpidose.audioFile being null/not null to determine if the podcast is downloaded. Should the `downloaded` state be its own property?
- What should the behaviour of the "Latest" page be? I think other podcast apps would show you new epsidoes that you have not downloaded yet. Maybe that should be a filter.
- WIll this need a new config option alongside autodownload for "autorefresh" the list.
- This kind of changes the concept of removing/adding episodes. So the episodes are always there, and removing them just deletes the file. This seems like desirable behaviour for podcasts at least?
- No idea how this interacts with the mobile apps

I am happy to work on it or you can say this is completely the wrong direction.

Screenshot of what it looks like with one episode downloaded:
image


advplyr:
I agree we should have the feed there instead of having to press a button to view the feed and download.

I'm not sure about using PodcastEpisode for non-downloaded episodes and putting it in the `libraryItem.media.episodes` array. It makes it easier to build the front-end but the episode is not yet part of the library item in my mind.

My first thought would be making a separate request to get the podcast feed that would return an array of PodcastEpisode objects or some similar object. Not all podcasts will have an RSS feed so we have to keep that in mind as well when designing this we may not need to make this additional request.

For the frontend part we aren't going to be able to use the EpisodesTable.vue component as it is built now because the UI will get very slow when rendering thousands of episodes. I didn't take this into account at first when I made that and have since had reports of the UI getting slow with a lot of downloaded episodes. When I made the latest page recently I built a simplified list view that isn't using a component for each episode row and isn't using tooltips. Basically trying to limit the number of event listeners required to render each podcast episode. This will be especially tricky if we plan to keep the multi-select and is the main reason I didn't switch that over yet.

There is a lot more to discuss here but I'll keep it at that for now. I'm not sure if you are using VSCode or I messed up the `.vscode/settings.json` but your editor is adding semicolons and doing other auto-formatting that doesn't align with the rest of the project. It makes it harder to identify which lines were changed.

advplyr:
> * Should the frontend handle merging the feed episodes list with the library item episodes list?

This could go either way. If it is handled on the server side we will still need to update the client if an episode gets downloaded.

> * Should the server redownload the feed every time this request is made, or store a copy of the feed?

I think the server should store a copy of the feed that could be used by any other task. This would need to also store the timestamp it was fetched and have some set amount of time before it fetches new data.

rhyst:
Hallo, I have updated this to another very WIP state to get feedback before continuing.

Changed:
- On adding a podcast library item it saves a copy of the podcast feed to the `media` object
- The client merges actual downloaded episodes with the podcast feed to construct a big list of podcasts
- Refresh button allows refetching podcast feed

Things that are janky that I would make better if this is the right path:
- Merging/sorting the podcast list 
- Pagination/lazy load of podcast list not considered
- Various bits of UI (download/queued state)
- Have not really tested the case of a podcast without a feed yet (what is that use case? manually adding files to the server?) but I don't think it will be a problem.

Questions:
- Is saving the feed to the `libraryItem.media.podcastFeed` object correct? Or should it be saved as a file (like the cover art)?
- How should the automatic fetch work? Could:
  - Reuse download schedule
  - Have own schedule configuration
  - Have hard coded daily refresh (maybe with toggle on/off)

rhyst:
@advplyr  One more thought - I think it's useful to be able to mark podcasts as finished even if they are not downloaded. Do you think that is a useful feature, and if so do you think there is a way to implement it that you'd prefer?

rhyst:
@advplyr bump for feedback 😬

advplyr:
Right now I don't think it is useful to mark episodes you don't have downloaded as finished. Unless we decide to build a full-on podcast player, which we could do.

I don't think we should be storing the podcast feed on the Podcast object at all. I think this needs to be a separate API endpoint that fetches the RSS feed and returns data that includes the PodcastEpisode objects.

We also need to re-do the podcast episodes table because it will get very slow if there are thousands of episodes. The `latest.vue` page was built with performance in mind by reducing the number of Vue components.

I've had this on my radar for a while but since it is such a major rewrite I haven't prioritized it. I've been primarily focusing on the mobile apps and haven't spent much time on podcasts. There are some other things with podcasts that need to be done like embedding meta tags when the episodes get downloaded and parsing meta tags when scanning in episodes.

Also, not every podcast library item has an RSS feed. For example, I put Hardcore History series that I purchased in Abs as a podcast but it doesn't have an RSS feed.

rhyst:
When I load a library item page for a podcast are you thinking that it makes a request to this new rss feed endpoint which then fetches the rss feed xml from the source every time the endpoint is called? Or does it need to store the rss feed somewhere. Previously you mentioned tracking downloaded time etc. so if the feed should be stored then how do we know when to refetch it?

In the interest of limiting the scope of this as well, would adding simple pagination to the episodes table be enough to satisfy performance concerns for now?

If you think that the podcasts features need a rewrite then I am happy to drop this PR, it seems like you have a specific idea of how it should work. IMO for podcasts it is important to decouple the played state from actually having the downloaded file. As you point out podcasts can have thousands of episodes and storing the files is unnecessary as I suspect most people will listen to a podcast episode just once.

advplyr:
> When I load a library item page for a podcast are you thinking that it makes a request to this new rss feed endpoint which then fetches the rss feed xml from the source every time the endpoint is called? Or does it need to store the rss feed somewhere. Previously you mentioned tracking downloaded time etc. so if the feed should be stored then how do we know when to refetch it?

I'm thinking that it would be a separate request or maybe a flag passed in as a query string. The main thing is I don't think the RSS feed belongs on the library item.
We may not need to cache the response from the RSS feed. I would be curious if other podcast clients do any sort of caching.

> In the interest of limiting the scope of this as well, would adding simple pagination to the episodes table be enough to satisfy performance concerns for now?

One of the main performance issues is batch selecting and pagination can complicate this. It would probably be just as much effort to setup pagination as it would be to re-design that table.

> If you think that the podcasts features need a rewrite then I am happy to drop this PR, it seems like you have a specific idea of how it should work. IMO for podcasts it is important to decouple the played state from actually having the downloaded file. As you point out podcasts can have thousands of episodes and storing the files is unnecessary as I suspect most people will listen to a podcast episode just once.

I don't have a specific idea of how this should work. The addition of podcasts was originally to be for downloading episodes or managing your own episode files independent of an RSS feed. Since it has been added there has been more of an expectation that Abs would be a full podcast client. Streaming an episode that is not on your server will require a significant update to the media progress & playback session data models.

This update would be for just displaying the episodes in the RSS feed in the same list as the downloaded episodes, which I think is a good idea.

advplyr:
This no longer fits with the new data model and I'm not sure about going this route anymore. I wrote a little something about it here https://github.com/advplyr/audiobookshelf/issues/2216#issuecomment-1763423433

I will leave the github issue related to this open for discussion https://github.com/advplyr/audiobookshelf/issues/960

Issue #1044, opened 2022-10-04
[Bug]: Fail to embed meta tags when a description starts with a double quote


andrewgdunn:
### Describe the issue

The ability to match then M4B encode is __incredible__.

I'm running into an issue when using the M4B encoder that looks like this:

```
[2022-10-04 16:05:59] ERROR: [toneHelpers] tagAudioFile: Failed for "/metadata/cache/items/li_9y6kvoa37spanbbweg/A Single Thread (2019).m4b" Error: The path '/(People) "The best-selling novelist has done a masterful job 
of depicting the circumstances of a generation of women we seldom think about: 
the mothers, sisters, wives and fiances of men lost in World War I, whose job it
was to remember those lost but not forgotten." (Associated Press) 1932. After 
the Great War took both her beloved brother and her fiancé, Violet Speedwell has
become a "surplus woman," one of a generation doomed to a life of spinsterhood 
after the war killed so many young men. Yet Violet cannot reconcile herself to a
life spent caring for her grieving, embittered mother. After countless meals of 
boiled eggs and dry toast, she saves enough to move out of her mother's place 
and into the town of Winchester, home to one of England's grandest cathedrals. 
There, Violet is drawn into a society of broderers - women who embroider 
kneelers for the Cathedral, carrying on a centuries-long tradition of bringing 
comfort to worshippers. Violet finds support and community in the group, 
fulfillment in the work they create, and even a growing friendship with the 
vivacious Gilda. But when forces threaten her new independence and another war 
appears on the horizon, Violet must fight to put down roots in a place where 
women aren't expected to grow. Told in Chevalier's glorious prose, A Single 
Thread is a timeless story of friendship, love, and a woman crafting her own 
life.' is too long, or a component of the specified path is too long.
```
```
[2022-10-04 16:21:24] ERROR: [toneHelpers] tagAudioFile: Failed for "/metadata/cache/items/li_zdfe9izqh5hxrx95ym/A Spool of Blue Thread (2015).m4b" Error: The path '/This is how Abby Whitshank always begins the story of how she 
fell in love with Red that day in July 1959. The Whitshanks are one of those 
families that radiate togetherness: an indefinable, enviable kind of 
specialness. But they are also like all families, in that the stories they tell 
themselves reveal only part of the picture. Abby and Red and their four grown 
children have accumulated not only tender moments, laughter, and celebrations, 
but also jealousies, disappointments, and carefully guarded secrets. From Red's 
[...]

PR #1076, opened 2022-10-18
Fix:Unescaped characters in description causing issues with m4b merging

conumey:
When the merge tags the merged file, and the metadata contains an unescaped double quote (or potentially only when it is the first character), it causes the merge to fail, and potentially other issues if the tagging action happens to succeed anyway.

Here's a nice error where the description after the initial quoted text gets added to the filepath:
image

This happened with a few other merges that completed "succesfully":
image

Escaping the characters may have stopped this Missing Parts problem happening as well, though I'm not certain as it doesn't consistently happen when they're unescaped.

My fix is a bit of a lazy one in that I just duplicated the getToneMetadataObject function, as escaping by default meant that the descriptions in the library were full of backslashes and I didn't fancy rewriting the frontend to render escaped characters normally 😇

### DIFF
--- a/server/managers/AbMergeManager.js
+++ b/server/managers/AbMergeManager.js
@@ -131,7 +131,7 @@ class AbMergeManager {
       }
     }
 
-    const toneMetadataObject = toneHelpers.getToneMetadataObject(libraryItem, chaptersFilePath)
+    const toneMetadataObject = toneHelpers.getEscapedToneMetadataObject(libraryItem, chaptersFilePath)
     toneMetadataObject.TrackNumber = 1
     task.data.toneMetadataObject = toneMetadataObject
 
--- a/server/utils/stringEscaper.js
+++ b/server/utils/stringEscaper.js
@@ -1,0 +1,5 @@
+function escapeString(string) {
+  return string.replace(/[.*"+?^${}()|[\]\\]/g, '\\$&');
+}
+
+module.exports.escapeString = escapeString
--- a/server/utils/toneHelpers.js
+++ b/server/utils/toneHelpers.js
@@ -2,6 +2,7 @@ const tone = require('node-tone')
 const fs = require('../libs/fsExtra')
 const Logger = require('../Logger')
 const { secondsToTimestamp } = require('./index')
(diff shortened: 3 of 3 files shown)


advplyr:
This is because the first character is a quote. It was reported #1044 and I looked into it but wasn't able to easily solve it by escaping.
Your solution works because it is adding a backslash as the first character but that backslash stays in the description.

image

The wrapper I wrote for tone using nodejs is handling all the escaping for us but doesn't work with the first quote. I went in a little more detail on that in the issue #1044

https://github.com/advplyr/node-tone

advplyr:
Also, I think the missing parts alert is a separate issue having to do with the scanner not resetting the "missingParts" array on the `Book.js` object. That probably goes a bit deeper with just a poorly designed track scanner.

The quote as the first character issue also breaks embedding metadata so if we don't get this fixed before the next release I'll add some alert and/or put a space in-front.

conumey:
> This is because the first character is a quote. It was reported #1044 and I looked into it but wasn't able to easily solve it by escaping. Your solution works because it is adding a backslash as the first character but that backslash stays in the description.
> 
> image
> 
> The wrapper I wrote for tone using nodejs is handling all the escaping for us but doesn't work with the first quote. I went in a little more detail on that in the issue #1044
> 
> https://github.com/advplyr/node-tone

Ah yes, silly of me to not check the resulting metadata with Tone. I was assuming since it was using Tone to tag the metadata that it would work the same way as running that from the command line, so characters would be escaped properly and also that a simple workaround of adding a space at the beginning wouldn't work as it would be ignored, I didn't realise it was using a wrapper you'd written. I'll have a poke at that and see if I can work out the first character issue there.



> Also, I think the missing parts alert is a separate issue having to do with the scanner not resetting the "missingParts" array on the `Book.js` object. That probably goes a bit deeper with just a poorly designed track scanner.
> 
> The quote as the first character issue also breaks embedding metadata so if we don't get this fixed before the next release I'll add some alert and/or put a space in-front.

Yeah I was thinking it was a little bit too lucky to have sorted two problems, I guess it was just coincidental in my testing.

PR #1105, opened 2022-10-30
Patching handling of titles with multiple series

ruoti:
I noticed that there was an issue with the "Collapse Series" option in the Library view that happens when a book is in two series at the same time. For example, in the Pern series of books, the first three books are part of a trilogy (series A) and part of the overall Pern universe (series B). If I label them as such, one of those series will disappear from the view, even though both should be shown. They still show up in the series list, so this is just a display issue.

While fixing this problem, I also noticed that code for sorting books attached to multiple series wouldn't work correctly as sorting happened before collapsing. I've reworked the sorting to have collapsing occur first, then sorting after that. I also fixed a couple of other small issues with sorting:

1. When series are put at the end of the list of titles (happens when not sorting by title), they are now sorted alphabetically.
2. In the above case, a series will appear after books beginning with 'zzzzz', which they wouldn't have before. Admittedly, this would probably never happen, but I was here so I fixed it anyways.
3. If filtering by series, the code would resort after the initial sort. I moved this into the normal sort workflow.

The one non-obvious thing I did was make it so that when filtering by a series and collapsing series at the same time, I do not collapse the titles that are only in the series being filtered. This didn't make sense as you would select a series just to see a single item with that series (unless you had titles with multiple series). Along with my change #3 above, this results in cool functionality where subseries are displayed in order in this case. For example, If there was a trilogy, then a standalone book, then another trilogy, it would show them in this order. Nice for larger series made up of many sub-series (e.g., Dragonlance).

### DIFF
--- a/client/components/app/BookShelfToolbar.vue
+++ b/client/components/app/BookShelfToolbar.vue
@@ -28,13 +28,19 @@
             <span class="font-mono">{{ numShowing }}</span>
           </div>
           <div class="flex-grow" />
-          <ui-btn color="primary" small :loading="processingSeries" class="flex items-center" @click="markSeriesFinished">
+          <ui-checkbox v-model="settings.collapseBookSeries" label="Collapse Series" checkbox-bg="bg"
+            check-color="white" small class="mr-2" @input="updateCollapseBookSeries" />
+          <ui-btn color="primary" small :loading="processingSeries" class="flex items-center ml-1 sm:ml-4"
+            @click="markSeriesFinished">
             <div class="h-5 w-5">
-              <svg v-if="isSeriesFinished" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="rgb(63, 181, 68)">
-                <path d="M19 1H5c-1.1 0-1.99.9-1.99 2L3 15.93c0 .69.35 1.3.88 1.66L12 23l8.11-5.41c.53-.36.88-.97.88-1.66L21 3c0-1.1-.9-2-2-2zm-9 15l-5-5 1.41-1.41L10 13.17l7.59-7.59L19 7l-9 9z" />
+              <svg v-if="isSeriesFinished" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"
+                fill="rgb(63, 181, 68)">
+                <path
+                  d="M19 1H5c-1.1 0-1.99.9-1.99 2L3 15.93c0 .69.35 1.3.88 1.66L12 23l8.11-5.41c.53-.36.88-.97.88-1.66L21 3c0-1.1-.9-2-2-2zm-9 15l-5-5 1.41-1.41L10 13.17l7.59-7.59L19 7l-9 9z" />
               </svg>
               <svg v-else xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor">
-                <path d="M19 1H5c-1.1 0-1.99.9-1.99 2L3 15.93c0 .69.35 1.3.88 1.66L12 23l8.11-5.41c.53-.36.88-.97.88-1.66L21 3c0-1.1-.9-2-2-2zm-7 19.6l-7-4.66V3h14v12.93l-7 4.67zm-2.01-7.42l-2.58-2.59L6 12l4 4 8-8-1.42-1.42z" />
+                <path
+                  d="M19 1H5c-1.1 0-1.99.9-1.99 2L3 15.93c0 .69.35 1.3.88 1.66L12 23l8.11-5.41c.53-.36.88-.97.88-1.66L21 3c0-1.1-.9-2-2-2zm-7 19.6l-7-4.66V3h14v12.93l-7 4.67zm-2.01-7.42l-2.58-2.59L6 12l4 4 8-8-1.42-1.42z" />
               </svg>
             </div>
             <span class="pl-2"> Mark Series {{ isSeriesFinished ? 'Not Finished' : 'Finished' }}</span>
@@ -278,6 +284,9 @@ export default {
       this.$eventBus.$emit('series-sort-updated')
     },
     updateCollapseSeries() {
+      this.saveSettings()
+    },
+    updateCollapseBookSeries() {
       this.saveSettings()
     },
     saveSettings() {
--- a/client/components/app/LazyBookshelf.vue
+++ b/client/components/app/LazyBookshelf.vue
@@ -122,6 +122,9 @@ export default {
     collapseSeries() {
       return this.$store.getters['user/getUserSetting']('collapseSeries')
     },
+    collapseBookSeries() {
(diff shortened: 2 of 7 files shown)


ruoti:
I've rebased my commit as there were conflicts with the code pushed last night.

I've also made it so that when collapsing in series view, instead of showing the count for collapsed series, I should the sequence numbers they represent in the filtered series. This matches stylistically with how the sequence number is shown for other books. 

Pre-patch look:
Screenshot_20221030_114457

New look:
Screenshot_20221030_114253

Some things to note:
1. With these changes, I think it would make sense to add a "Collapse Series" checkmark to the series view. When dealing with sub-series, I think this works really nicely. Or perhaps, make this just how that view works. It wouldn't have worked in the old scheme where this would have collapsed the main series into itself, but that is no longer the case.
2. I think it is a little weird that the sort order is ignored when filtering by series. For the series tab, this makes sense, but it is a little odd for the others. With the proposed patch, removing this functionality when there are sortBy parameters will work as expected (basing sort by the first box in a collapsed series). However, doing so would make it impossible to sort by sequence numbers in this view, so the drop-down would need to be changed to include that option, though that option wouldn't make sense with other filers. Just something to consider. In this pull request, I've kept the existing functionality the same in this regard.

advplyr:
> I've also made it so that when collapsing in series view, instead of showing the count for collapsed series, I should the sequence numbers they represent in the filtered series. This matches stylistically with how the sequence number is shown for other books.

One problem with this is if you have a sub-series with a lot of books then the sequence numbers will overflow.

image

advplyr:
> 2\. I think it is a little weird that the sort order is ignored when filtering by series. For the series tab, this makes sense, but it is a little odd for the others. With the proposed patch, removing this functionality when there are sortBy parameters will work as expected (basing sort by the first box in a collapsed series). However, doing so would make it impossible to sort by sequence numbers in this view, so the drop-down would need to be changed to include that option, though that option wouldn't make sense with other filers. Just something to consider. In this pull request, I've kept the existing functionality the same in this regard.

This is definitely a bug I wasn't aware of. The sort order should not be ignored.

ruoti:
Ok, I'm most of the way there. I've fixed the overflow issue, allowed it to display sequence ranges, fixed the item count in the toolbar, and fixed sorting within series view (as this was stated to be a bug). With the latter, it is no longer possible to sort books by sequence in the library view (still works fine in 'book-series' view). To address this, I tried to add another sort field when filtering by series. While I got that working, if I switch to another type of filter, the order by column ends up being blank because the (selected) sequence value is no longer in the list. Considering I've only started using Vue tonight, I'm not expert on how to fix this. I was wondering if you (@advplyr) have any suggestions on how I could reset the order by when no longer filtering by a series.

server/controllers/LibraryController.js
```diff
@@ -218,30 +252,55 @@ class LibraryController {
...
+          nameIgnorePrefix: li.collapsedSeries.nameIgnorePrefix,
+          libraryItemIds: li.collapsedSeries.books.map(b => b.id),
+          numBooks: li.collapsedSeries.books.length
+        }
+
+        // If collapsing by series and filtering by a series, generate the list of sequences the collapsed
+        // series represents in the filtered series
+        if (filterSeries) {
+          json.collapsedSeries.seriesSequenceList = li.collapsedSeries.books
+            .map(b => parseFloat(b.filterSeriesSequence))
```
advplyr, line 278: An issue with this solution to overflow sequence values is that we don't enforce a sequence to be a number. For example, a valid sequence could be roman numerals or `1a`. In that case this would crash but if we chose to just ignore those then the possibility of the sequence overflowing the book cover would remain.

client/components/app/BookShelfToolbar.vue
```diff
@@ -28,13 +28,19 @@
             <span class="font-mono">{{ numShowing }}</span>
           </div>
           <div class="flex-grow" />
-          <ui-btn color="primary" small :loading="processingSeries" class="flex items-center" @click="markSeriesFinished">
+          <ui-checkbox v-model="settings.collapseSeries" label="Collapse Series" checkbox-bg="bg" check-color="white"
+            small class="mr-2" @input="updateCollapseSeries" />
```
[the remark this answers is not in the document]
ruoti: Updated to use a different setting.

I also added a new commit that moves all of these options to local storage. It took longer than I thought it would, but the utility class I created should help you add any other items you want to move to local storage. Unfortunately, I couldn't get it to work as a mixin, but it isn't too much to add.

server/controllers/LibraryController.js
ruoti: I fixed the logic to use a natural sort. Even without that change, it still wouldn't overflow. The overflow before was caused by a lack of whitespace. I implemented ranges because I think they are more readable, not because they fixed overflow. Below you can see how it looks. I also updated it to use the same styling as normal sequence numbers (I think it looks better).

Screenshot_20221104_090103

ruoti:
I believe I have addressed all your comments

advplyr:
Thanks for setting up that local storage solution. We will need to set default values for those settings also.
Otherwise we end up with
image

client/components/app/LazyBookshelf.vue
```diff
@@ -437,6 +435,9 @@ export default {
         searchParams.set('filter', this.seriesFilterBy)
       } else if (this.page === 'series-books') {
         searchParams.set('filter', `series.${this.$encode(this.seriesId)}`)
+        if (this.collapseSeries) {
```
advplyr, line 438:
```suggestion
        if (this.collapseBookSeries) {
```

ruoti:
I've rolled back the localStorage changes. I saw that you were doing a fair bit of consistency checking on the server side, and I wasn't ready to go through that logic and make sure it is working in localStorage. I've staged those changes and may go back and add back localStorage in another pull request. This way we don't have to block these changes on localStorage.

ruoti:
Also, I probably won't work on the local storage stuff until this is merged so I don't have to worry about rebasing that stuff.

advplyr:
Great thanks, working well for me

server/controllers/LibraryController.js
```diff
@@ -218,30 +252,55 @@ class LibraryController {
...
-    } else {
-      libraryItems = libraryItems.map(li => payload.minified ? li.toJSONMinified() : li.toJSON())
+    if (sortArray.length) {
+      libraryItems = naturalSort(libraryItems).by(sortArray)
     }
 
-    if (payload.limit) {
-      var startIndex = payload.page * payload.limit
-      libraryItems = libraryItems.slice(startIndex, startIndex + payload.limit)
-    }
```
advplyr, line 243: Is there a reason this was removed here? Just got notified on Discord about an issue and saw the limit is not applied.

ruoti: No, that was an accident. My guess is that this happened the first time I rebased the patch, where we had both made changes to this file, and I just made a mistake. Sorry.

PR #1106, opened 2022-10-30
Makes the dev target support auto reloading of the server

ruoti:
Simple commit that makes it so that the server code supports live reload. I've tested it and it works really nicely. I know the client already had this option, so adding it to the server only made sense.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Self-hosted audiobook and podcast server",
   "main": "index.js",
   "scripts": {
-    "dev": "node index.js",
+    "dev": "nodemon --watch server index.js",
     "start": "node index.js",
     "client": "cd client && npm ci && npm run generate",
     "prod": "npm run client && npm ci && node prod.js",
@@ -37,5 +37,8 @@
     "node-tone": "^1.0.1",
     "socket.io": "^4.4.1",
     "xml2js": "^0.4.23"
+  },
+  "devDependencies": {
+    "nodemon": "^2.0.20"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,9 @@
       },
       "bin": {
         "audiobookshelf": "prod.js"
+      },
+      "devDependencies": {
+        "nodemon": "^2.0.20"
       }
     },
     "node_modules/@types/component-emitter": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Self-hosted audiobook and podcast server",
   "main": "index.js",
   "scripts": {
-    "dev": "node index.js",
+    "dev": "nodemon index.js",
```
advplyr, line 7:
```suggestion
    "dev": "nodemon --watch server index.js",
```
Only need to watch files inside /server since /client is watched by Nuxt.

advplyr:
npm-watch was attempted to be added by someone before and I had to remove it because the github action was failing to build due to the dev dependency. I didn't have time to debug that and didn't seem to make sense because the Dockerfile is specifying --production when building, but just a heads up.

ruoti:
I've updated the command to watch only the server.

I'm not sure how to make sure this won't break your CI tasks. Any tips for that? I can manually build the docker image and run it without any problem. I would hope the same would be true running on github's CI backend. With that said, npm-watch and nodemon take entirely different approaches to reloading the server. From my understanding npm-watch hijack's the module loading in npm, whereas nodemon just looks for changes, and kills then relaunches the server if it detects them. There is no special config needed that could mess up the normal `npm run start` task. I'm happy to try some more testing if you have suggestions.

advplyr:
I don't think it should break anything, the last time it didn't make sense to me I just didn't have time to troubleshoot. We'll see how it goes.

---

I'm curious why you are force pushing into this branch.
image

I wasn't sure why git was rejecting my attempt to checking out your other branch then I just noticed the force push here. When checking out a branch locally I've been using the command `gh pr checkout <# of PR>`

```bash
> gh pr checkout 1106
remote: Enumerating objects: 4, done.
remote: Counting objects: 100% (4/4), done.
remote: Compressing objects: 100% (4/4), done.
Unpacking objects: 100% (4/4), 25.66 KiB | 656.00 KiB/s, done.

From https://github.com/advplyr/audiobookshelf
 ! [rejected]        refs/pull/1106/head -> livereload-server  (non-fast-forward)
exit status 1
```

In order to checkout I have to first delete the existing branch on my local like this
```bash
> git branch -D livereload-server
Deleted branch livereload-server (was 8fe42db).
```

advplyr:
this built successfully

PR #1107, opened 2022-11-01
Development documentation

ruoti:
As requested in #259, I have tried to create documentation about setting up the development environment and running this project. As part of this effort, I also updated the .devcontainer used by visual studio code. For this, I regenerated the base configuration using VSCode, then added the necessary modules based on the Dockerfile in the root directory. I also added post-build commands to ensure everything was setup for new users. Finally, I added VSCode build/test tasks and debug targets.

If people use the dev container, it takes running one command in VSCode and hitting F5 before they are running the code for the first time, which is pretty nice.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,15 @@
-FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:16
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get install ffmpeg gnupg2 -y
+# [Choice] Node.js version (use -bullseye variants on local arm64/Apple Silicon): 18, 16, 14, 18-bullseye, 16-bullseye, 14-bullseye, 18-buster, 16-buster, 14-buster
+ARG VARIANT=16
+FROM mcr.microsoft.com/devcontainers/javascript-node:0-${VARIANT} as base
+
+# Setup the node environment
 ENV NODE_ENV=development
+
+# Install additional OS packages.
+RUN apt-get update && \
+  DEBIAN_FRONTEND=noninteractive apt-get -y install --no-install-recommends \
+    curl tzdata ffmpeg && \
+  rm -rf /var/lib/apt/lists/*
+
+# Move tone executable to appropriate directory
+COPY --from=sandreas/tone:v0.1.2 /usr/local/bin/tone /usr/local/bin/
--- a/.devcontainer/dev.js
+++ b/.devcontainer/dev.js
@@ -1,0 +1,9 @@
+// Using port 3333 is important when running the client web app separately
+const Path = require('path')
+module.exports.config = {
+  Port: 3333,
+  ConfigPath: Path.resolve('config'),
+  MetadataPath: Path.resolve('metadata'),
+  FFmpegPath: '/usr/bin/ffmpeg',
+  FFProbePath: '/usr/bin/ffprobe'
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,12 +1,40 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/javascript-node
 {
-    "build": { "dockerfile": "Dockerfile" },
-    "mounts": [
-        "source=abs-node-node_modules,target=${containerWorkspaceFolder}/node_modules,type=volume"
-    ],
-    "features": {
-        "fish": "latest"
+	"name": "Audiobookshelf",
+	"build": {
+		"dockerfile": "Dockerfile",
+		// Update 'VARIANT' to pick a Node version: 18, 16, 14.
+		// Append -bullseye or -buster to pin to an OS version.
+		// Use -bullseye variants on local arm64/Apple Silicon.
+		"args": {
+			"VARIANT": "16"
+		}
 	},
-	"extensions": [
-		"eamodio.gitlens"
-	]
+	"mounts": [
+		"source=abs-server-node_modules,target=${containerWorkspaceFolder}/node_modules,type=volume",
+		"source=abs-client-node_modules,target=${containerWorkspaceFolder}/client/node_modules,type=volume"
+	],
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
+		3000,
+		3333
+	],
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "sh .devcontainer/post-create.sh",
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"dbaeumer.vscode-eslint",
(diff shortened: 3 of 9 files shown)


ruoti:
Setting up the dev container is a little slow as the project relies on `npm ci`. While I understand the rationale, destroying the node_modules folder is wasteful. @advplyr, would you be opposed to me submitting a pull request that moves the project to yarn? It can better handle this situation and make downloading packages much quicker.

taxilian:
I'm new so maybe I'm missing something -- there is no `.devcontainer/dev.js`, is that expected? everything seems to be indicating that it expects that there will be one

ruoti:
I'm not sure there is a standard for devcontainers in general, but the standard used by vscode just requires a devcontainer.json file: <https://code.visualstudio.com/docs/devcontainers/create-dev-container>.

taxilian:
right, there is information in the project for creating the devcontainer, but all the scripts reference "if there isn't a dev.js, get it from `.devcontainer/dev.js` -- `.devcontainer` exists in the project with files for creating the container, but `dev.js` is not there.

ruoti:
Oh, I see what you mean. Yep, that was missing. It was on the .gitignore list, but I didn't noticed. I've checked it in and rebased.

Weldawadyathink:
As requested, I took a look through this PR. Here are a few modifications I would recommend based on the research I have done into dev containers. I explained my reasoning in the commit messages, and I would be happy to go into more detail. 

[My modified repo](https://github.com/Weldawadyathink/audiobookshelf/tree/dev-documentation)

I can confirm this is working on macOS arm docker and GitHub codespaces.

ruoti:
@Weldawadyathink, lots of great changes there. I've updated my pull request incorporating your changes and rebasing on the latest version. Some changes I didn't make yet, pending more discussion, are as follows:

* I left the forwarded ports in the devcontainer.json. Yes, vscode opens these automatically, but I don't think that is required of devcontainer tools. For example, I haven't observed this behavior when using docker's own devcontainer tools. I've left it here for completeness, but am not tied to this change.
* I've kept the comments and commented out functionality. I think this is helpful for other developers who might contribute to this project. While neither I nor @Weldawadyathink needs this, it could help a new developer understand what is happening inside their devcontianer. It would also help others if in the future the devcontainer needs to be updated. It doesn't impede understanding of what is happening in the file, so I think it can stay.
* In your code, you move from `npm ci` to `npm i`. While this (substantially) speeds up the process of starting the devcontainer, it does not match how development on the project is supposed to go. Most importantly, it does not install packages based on package-lock.json, meaning it can result in different package versions than what will be used when the docker image is built, causing potential breakage. Unfortunately, I can't find any way to install packages based on the package-lock.json without using `npm ci`. A solution would be to move to yarn, which does support the desired functionality, but that seemed outside the scope of this pull request.
* The postCreateCommand doesn't need the change directory command because the devcontainer will automatically set the workspace directory as the working directory when executing the container. The reason not to include it here is that it potentially builds an incorrect mental model of how the containers are setup, but I'm not tied to this and will switch out if you disagree with me.

Otherwise, I've adopted the changes and updated the PR. I look forward to your thoughts.

Weldawadyathink:
Those changes all look good to me. 

- Ports: Yeah I don't feel too strongly either way. It makes sense to follow the container spec. 
- Comments are usually good, so this is fine in my opinion. One weird side note: the comment says that the -bullseye variants should be used on arm64. But testing on an apple silicon Mac, it seems to work just fine without the variant flag. Not sure what that is about. Maybe Microsoft updated their dev containers. 
- That is a bummer about npm. It's annoying to wait for node to reinstall everything after every start. Maybe yarn would be better. I don't have much experience with node or yarn, so I can't comment on that. 
- I couldn't find anything about this on the containers.dev spec website, but it does appear to work fine everywhere I tested. 

I have tested this PR in vs code desktop and GitHub codespaces. It works perfectly in vs code desktop. It almost works perfectly in codespaces. If codespaces is setup to open a local vs code instance, vs code doesn't pick up the launch.json for some reason. I am not familiar enough with vs code to figure out what is going on here. I do all of my development locally, but it would be nice to have it working on codespaces. 

Also it is super nice to be able to run and debug natively in vs code. Thanks for all of this improved documentation.

ruoti:
@Weldawadyathink Just tried it and codespaces pulled launch.json configurations for me. Not sure what is happening there.

@advplyr I think this is ready to merge. I know you aren't too aware of devcontainers, but this is better than what was there, so it is a net win that doesn't impact your current development processes. The files that have an impact on people not using containers are the changes in `.vscode/` and `.gitignore`, so if those look good I think it would be nice to see this committed to ease first-time development.

Also, if you want, I can create a pull request moving to `yarn` to improve the build process (both in the devcontainer and out).

ruoti:
So I don't lose this and as reference for @advplyr, I'm uploading a patch that moves the repo to support yarn. It works locally for me, but I didn't test all possible paths (e.g., autobuilding through git hooks). I believe it should work for those use cases, but it would need testing.

At least for me, yarn is much faster. It is especially faster when you need to update packages as you no longer need to clean the directory before installing.


[yarn.patch](https://github.com/advplyr/audiobookshelf/files/10326845/yarn.patch)

advplyr:
I still haven't read up on devcontainers but I'm poking around and not understanding how the run the client/server using the Debug commands.

image

Weldawadyathink:
The vscode left panel has a run and debug tab. It has three different run options for the server, client, and client+server. Is that what you are looking for? Also once the dev container is setup you can just run the commands in the terminal. It automatically attaches to the docker container.

hypnotoad08:
Any updates on this PR? it seems like it would be very beneficial for people like me who use VSCode for regular development and would like to eventually contribute to audiobookshelf.

ruoti:
@advplyr, if you are willing to merge this, I'll go back and rebase it off the latest commit and make sure everything is still building before pushing. It doesn't change anything in the main build process, so it seems like there is minimal risk in merging. Just let me know.

advplyr:
Thanks, yeah I'm willing to merge this. I wasn't able to get it going but if it helps others that would be great. I did comment about a possible bug if you can check that out also.

ruoti:
Ok, I've rebased off of the latest commit. I also tweaked the devcontainer based on the suggestions from @Weldawadyathink. I've tested this and it build and runs on my system.

@advplyr I cannot replicate your bug. I think it might have happened because the first time your script ran, the node_modules folder didn't exist. The next time you ran it (after removing parenths) worked because it did exist that time. Not entirely sure, but that is my hunch. If this is the case, I've fixed this by doing the checks for directories explicitly.

Hopefully this can get committed now.

P.S., the dev container is built on debian, not alpine. In that one small way it diverges from the main Dockerfile. I tried to get stuff working with alpine, but would require a much longer Dockerfile as I would have to setup nvm myself as Alpine is not supported by the Node.js devcontainer feature. Hope this isn't too big of an issue.

P.P.S. It might be good to upgrade Tone at some point. There have been some important bug fixes since 1.0.2. I didn't test out the new version and kept it the same as the main Dockerfile in this pull request.

advplyr:
Great, thanks for setting this up. I was able to get it running but the only thing that didn't work out of the box for me was the socket connection.

This is probably just something I need to do on my end. Using Windows WSL2

ruoti:
Thanks. I just tested and on my end, the websocket connects without any problem. Maybe if you told me what the error the socket reports is, I could look into it some more.

advplyr:
`ws://localhost:3333/socket.io/?EIO=4&transport=websocket' failed: WebSocket is closed before the connection is established.`

PR #1109, opened 2022-11-02
Add support for forward proxy auth. Tested with authentik.

pgodwin:
This is small change to enable proxy forward authentication. This is useful for providers which can proxy authentication, and can be used a simple way of providing SSO to Autobookshelf.

Tested with Authentik. See details on https://goauthentik.io/docs/providers/proxy/, but supported by other reverse proxies. Eg [traefik](https://doc.traefik.io/traefik/middlewares/http/forwardauth/), [caddy](https://caddyserver.com/docs/caddyfile/directives/forward_auth).

To use it with Authentik set the following variables:

```bash
export PROXY_FORWARD_AUTH_ENABLED=1
export PROXY_FORWARD_AUTH_USERNAME=X-authentik-username
export PROXY_FORWARD_AUTH_CREATE=1
```


pgodwin:
I just saw there's an app for Audiobookshelf (neat!). Enabling forward auth seems to break it though. 

Will take a look whether it's my proxy config blocking a call the app needs, or whether this change has just broken the auth flow entirely.

pgodwin:
I had a quick look at this last night, and it doesn't work properly with the app without changing some settings on the proxy. 
The following URLs need to be whitelisted (at least from what i've found so far)
```
/ping
/login
/api
```
This would allow the app to use regular authentication (username/password), but it may or may not be appropriate for all users in all cases (especially those who want to force all access to the app via an authenticated proxy).

To enable forward auth with the app, it would need to display the `/login` page once a server is selected. The proxy can then force whatever authentication flow it needs at that point. 

I'm not familiar enough with the app to know the best way to implement it though.

Sapd:
Quickly tested @pgodwin  branch successfully

with:
- Authentik
- docker-compose

What needs to be done to get this feature merged? I can help with further testing and/or development

In my branch I also have a draft for login using Jellyfin users (with the idea of also OpenID to work). However I think the approach in this PR is much simpler.

advplyr:
I don't have Authentik setup to test this PR but I looked over the code and have some thoughts.

1. It doesn't seem like the create flag would be useful. Anytime you want to change users you would have to shut down the server and change the env vars.
    - I may be missing something but this would mean that this solution is only viable for a single user to use Abs?
2. The login page is already making a call to the server with `/status` which could return whether or not forward auth is enabled. If that request returns back that forward auth is enabled and the server is initialized then go ahead with the `/forwardauth` request.
3. Whitelisting those paths for the mobile would defeat much of the purpose to having an authentication layer between Abs. Currently the mobile app is making a `/ping` request once you enter a server. I think this could probably be changed to a `/status` request.

Sapd:
> I don't have Authentik setup to test this PR but I looked over the code and have some thoughts.
> 
> 1. It doesn't seem like the create flag would be useful. Anytime you want to change users you would have to shut down the server and change the env vars.


Im not sure how much experience you have with Single Sign On, so I will start from the beginning:
The idea with Authentik (or any similar software, i.e. Authentik, Authelia, Keyclock, Auth0) is that you have your user database separate from your application. That has that all friends and family members login in a central place (auth.example.com) and they are logged in automatically in all applications with a single account (audiobookshelf.example.com, jellyfin.example.com, nextcloud.example.com etc.).

The feature of this PR implements Forward Auth[1], which is the most simple way to implement it.
Basically you *must* have a reverse proxy like NGINX in front of the application. Using auth_ directives[2], one can check against software like authentik if he is authenticated, if not he gets redirected to the auth software.
When he is authenticated, the software will simply set together with nginx a header containing the username and the e-mail. Audiobookshelf can then simply read this header,  **create an account if none exists**  or simply log the user in.

> * I may be missing something but this would mean that this solution is only viable for a single user to use Abs?

This works with as many accounts as you wish, the authentication software will modify the header accordingly.

Ofc the feature must never be enabled without an reverse proxy with auth enabled, as otherwise an attacker can log himself into any account.

I also tested it with multiple accounts, and it works as expected.

[1] https://goauthentik.io/docs/providers/proxy/forward_auth
[2] https://nginx.org/en/docs/http/ngx_http_auth_request_module.html



> 2. The login page is already making a call to the server with `/status` which could return whether or not forward auth is enabled. If that request returns back that forward auth is enabled and the server is initialized then go ahead with the `/forwardauth` request.

Shouldn't it behave same way like the function`checkAuth`?

> 3. Whitelisting those paths for the mobile app would defeat much of the purpose to having an authentication layer between Abs. Currently the mobile app is making a `/ping` request once you enter a server. I think this could probably be changed to a `/status` request.

API is often whitelisted in Forward Auth, as the /api route is in most cases protected by an API key anyway.
The purpose of authentication proxy is mostly not to provide the same security as a VPN or a private network, but to provide an easer to implement alternative to OAuth ( https://en.wikipedia.org/wiki/OAuth ); or to protect applications which do not have any authentification at all.

---

Also something which is missing in that PR, but can also be implemented later: There must be a way to overwrite the link to the logout button. Otherwise it is not possible to do so.

advplyr:
Thanks for the writeup. I now just have some issue with the way the login is setup in this PR so I'll comment on the code

client/pages/login.vue
```diff
@@ -202,6 +230,10 @@ export default {
     }
   },
   async mounted() {
+
+    // Check for proxy auth
+    this.checkForwardAuth();
+
     if (localStorage.getItem('token')) {
       var userfound = await this.checkAuth()
```
advplyr, line 238: This `checkForwardAuth` function is asynchronous and since it isn't being waited for there may be a token that gets validated and/or the `/status` call being made simultaneously.

Even if we added an `await` this would not be good for servers that haven't been initialized yet. The `/status` request is how we find out if the server has been setup.

I'm not sure how this should be structured but we want to make the `/status` call first. We could move the `checkAuth` to happen after the `/status` that way we can first find out if forward auth is configured. I assume if forward auth is configured we want to use that instead of using the persisted user token.

Sapd: > I'm not sure how this should be structured but we want to make the /status call first. We could move the checkAuth to happen after the /status that way we can first find out if forward auth is configured. I assume if forward auth is configured we want to use that instead of using the persisted user token.

Would it then make sense to implement the Info directly into `/status` ? Given that also other options like the ConfigPath are passed.

> I assume if forward auth is configured we want to use that instead of using the persisted user token.

If I understand the code correctly, I think it can be done in https://github.com/advplyr/audiobookshelf/blob/0e292c64c44c3dd0af4cd5f0338bdda23b90c620/server/Auth.js#L55

One could copy the `this.user.find` and feed it directly with the value from the username header. However the problem is that the JWT verifier here also uses the userID to find an user: https://github.com/advplyr/audiobookshelf/blob/0e292c64c44c3dd0af4cd5f0338bdda23b90c620/server/Auth.js#L112
Not sure if this required from ABS. Authentik can also supply a UserID (just a normal unsigned integer) over the headers, but most software implementing Forward Auth only rely on either unique usernames or unique e-mail-adresses.

advplyr: Actually if we still use the auth token in Abs then we don't need to update the `authMiddleware`.
Here is what I'm thinking for the login page.

If a token is stored then `checkAuth` and login using the token.
Otherwise call `/status` which will:
 - If the server is not initialized it responds how it does now
 - If the PROXY_FORWARD env variables are set then attempt to get/create the user. In this case the user will be returned in with the response and the web ui can just `setUser`
 - Else regular login is requested

With this method if you are already using token login before setting this up you will need to logout first.

Your comment on the logout button also made me realize we would need to store a flag on the User if it is created or authenticated at all via the FORWARD_PROXY. This flag is how we could tell whether the logout button needs an additional step. How would logout be implemented for Authentik?
Users created from the forward proxy won't have passwords set so we wouldn't want them using the regular login form.

advplyr: This would be weird if the server hasn't been initialized yet.

Sapd: > Here is what I'm thinking for the login page.
> 

Sounds good, I will wait a bit until the PR creator (or you) started work on it as I cannot do commits on this PR and don't want to directly hijack with a new PR.  


> Your comment on the logout button also made me realize we would need to store a flag on the User if it is created or authenticated at all via the FORWARD_PROXY. This flag is how we could tell whether the logout button needs an additional step. How would logout be implemented for Authentik? Users created from the forward proxy won't have passwords set so we wouldn't want them using the regular login form.

Actually this also should be configurable with a ENV variable, as every SSO software has an potentially different URL for that. On Authentik the default value for that is `/outpost.goauthentik.io/sign_out` (which is a special location directive in nginx proxied to authentik instead to ABS). This could be also delivered via `/status` so the Button can be adjusted. 
When the user clicks on it, he gets forwarded to `audiobookshelf.example.com/outpost.goauthentik.io/sign_out`, authentik clears his session, and wont allow the user on audiobookshelf any more (and gives him then the option to log off from authentik completely or to log back in into ABS).

> store a flag on the User

Temporarily for handling it in the UI? Yes that sounds smart. However I wouldn't store that permanently in the DB, so that one can again deactivate Forward Auth.

> Users created from the forward proxy won't have passwords set so we wouldn't want them using the regular login form

A good way is also to simply set a random password for the user on creation or to not allow a login with an empty password. This way the feature can be completely disabled again while still keeping the users in tact and without creating a security issue.

And the password change function should be probably disabled, too. This could also be then done by just checking wether the user is currently logged in via SSO.

---

If you want to make it even more advanced. One can also implement the header for the SSO Groups (Authentik provides a list via the X-authentik-groups header). 
Example:
```
export PROXY_FORWARD_AUTH_GROUP_HEADER=X-authentik-groups
export PROXY_FORWARD_AUTH_GROUP_ADMIN=AudioBookAdmin
export PROXY_FORWARD_AUTH_GROUP_USER=AudioBookUser
export PROXY_FORWARD_AUTH_GROUP_GUEST=AudioBookGuest
```

This way a user is set automatically to admin if he has in authentik the group AudioBookAdmin. But this is not soo important and probably shouldnt be done in the first PR.

advplyr: > > store a flag on the User
> 
> Temporarily for handling it in the UI? Yes that sounds smart. However I wouldn't store that permanently in the DB, so that one can again deactivate Forward Auth.

I think I'm following everything so far except how we would detect whether a user is signed in via SSO when they are logging in via a persisted token.

Back to my post earlier I thought we would just use the persisted token if it existed.

> If a token is stored then `checkAuth` and login using the token. Otherwise call `/status` which will:

In this case I guess we could add an additional step in the `/api/authorize` endpoint that checks if the auth forward is configured then if the username from the token matches the user retrieved from the token.

Sapd: > In this case I guess we could add an additional step in the /api/authorize endpoint that checks if the auth forward is configured then if the username from the token matches the user retrieved from the token.

Ah yes, yep that seems to be indeed necessary. And should work this way; otherwise one ignores the header. The endpoint can then like you suggest also return if the user is logged in via SSO (when configured and header and token matches).

advplyr:
I did the refactor as discussed but I think the token authorization step will not allow for mobile use. I haven't tested any of it so let me know what you think.

server/controllers/MiscController.js
```diff
@@ -120,6 +120,13 @@ class MiscController {
       Logger.error('Invalid user in authorize')
       return res.sendStatus(401)
     }
+
+    // When authorizing an API token and forward auth is enabled then validate user using forward auth headers
+    if (global.ForwardAuth.Enabled && !this.auth.validateForwardAuthUser(req)) {
+      Logger.error(`[MiscController] Authorize token with forward auth enabled failed for user "${req.user.username}"`)
+      return res.sendStatus(403)
+    }
+
```
advplyr, line 129: This is the code that would break the mobile app but I'm realizing now that we probably don't need to return an error code here.
So if forward auth is enabled and the user does not match then just proceed with the user from the token.

server/Auth.js
```diff
@@ -94,6 +99,10 @@ class Auth {
     })
   }
 
+  generateRandomPasswordHash() {
+    return hashPass(getId())
```
Sapd, line 103: must be `this.hashPass`

```
ReferenceError: hashPass is not defined
    at Auth.generateRandomPasswordHash (/server/Auth.js:103:5)
    at Auth.getUserFromProxyAuth (/server/Auth.js:213:20)
    at /server/Server.js:246:59
    at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
    at next (/node_modules/express/lib/router/route.js:144:13)
    at Route.dispatch (/node_modules/express/lib/router/route.js:114:3)
    at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
    at /node_modules/express/lib/router/index.js:284:15
    at Function.process_params (/node_modules/express/lib/router/index.js:346:12)
    at next (/node_modules/express/lib/router/index.js:280:10)
    ```

Comment (advplyr) [reply server/Auth.js]: Fixed

File: server/controllers/MiscController.js (hunk 2 above)
Comment (advplyr) [reply server/controllers/MiscController.js]: Just removed that return and changed the error log to a warning.

**Comment by Sapd** [comment]
I created a test case, which I went through using the latest commits

**Existing Installation**
- [x] Do: Login existing User A
    - [x] **Result**: Logged in
    - [x] Do: Log out
    - [x] Do: Choose Log back in in Authentik
    - [x] **Result**: Logged back in
    - [x] Do: Log out
    - [x] Do: Choose Completly Log out
    - [x] **Result**: Completely logged out, no access any more, authentik redirects to authentik's login
- [x] Do: Login with user B (not yet in ABS existing) with same browser session
    - [x] **Result**: Logged as user B
    - [x] Do: Log out and Log out completly
    - [x] Do: Log in as User A
    - [x] **Result**: Logged in as User A

Works as expected.

---

** Completely new installation - Auth forwarding already set up **

- [x] Do: Login User A
- [x] **Result**: Initial server set up screen
- [x] Do: Create root user in server set up
- [ ] **Result**: Server set up complete, **however** loggedin as User A instead of root

This is a bit confusing if someone sets up Auth Forward before setting up the server (or if he uses the Auth Forward/external URL instead of the local machine url for that). As it will create an root user, but to give himself Admin he has to use the local url to bypass the reverse proxy to be able to log in as root and give himself admin.

We can decide to either fix this somehow, or to simply document this.

Everything else works also with the fresh installation.

---

**Mobile App (iOS)**

- On default settings, Ping fails
```
"OPTIONS /ping HTTP/1.1" 302 145 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 16_0_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148" "ipomitted"
```
This is because it returns a 302. As Authentik wants to redirect the request to the login page

- Authentik set to allow /ping without authorization
```
10.10.10.101 - - [18/Dec/2022:21:43:23 +0100] "GET /ping HTTP/1.1" 200 16 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 16_0_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148" "ipomitted"
```
- Lets me through and gives me the login screen of authentik
```
10.10.10.101 - - [18/Dec/2022:21:44:36 +0100] "OPTIONS /login HTTP/1.1" 302 145 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 16_0_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148" "ipomitted"
```
- Gives me a Login failed because of the 302

Adding /login as exception Authentik actually would not make much sense. I can login temporarily using i.e. the root account, however each subsequent request fails bc. also auf 302 answers.

To implement the Forward auth for the mobile app, it would have to recognise wether SSO is enabled (by checking where 302 leads to and comparing that with an additional ENV variable I would suggest; or by using the /ping route to provide that info and make it a requirement that it needs to be whitelisted). If it is enabled, an inline browser should open up with the URL of the 302; when he is back on `audiobookshelf.example.com` it can close again, however preserving the cookies.

**Comment by advplyr** [comment]
We can change the `/ping` on the mobile app to a `/status` request

**Comment by Sapd** [comment]
> We can change the `/ping` on the mobile app to a `/status` request

I have some time end of week and over the holidays. If you also have time and link a PR for that I can compile it via Xcode and test the whole process for mobile.

File: server/Auth.js [hunk 4]
```diff
@@ -173,6 +182,89 @@ class Auth {
...
+    }
+
+    let username = req.headers[global.ForwardAuth.UsernameHeader]
+
+    if (!username) {
+      Logger.warn(`[Auth] Forward Auth username header ${global.ForwardAuth.UsernameHeader} has no username`)
+      return false
+    }
+
+    let user = this.users.find(u => u.username.toLowerCase() === username)
```
Sapd, line 204: I noticed just another critical bug. 

If you login with Browser A, everything works. If you then use the same account with Browser B, ANOTHER account with the same name is created. That happens when you have an account with an Uppercase letter

I think the error is here; `username` on the rightside also needs to be lowercased.

Sapd:
So I am using it since December successfully, just with the change in line 204: https://github.com/Sapd/audiobookshelf/commits/feat-forwardauth

pgodwin:
> So I am using it since December successfully, just with the change in line 204: https://github.com/Sapd/audiobookshelf/commits/feat-forwardauth

Thanks @Sapd . I haven't had a chance to come back to this. Is it working properly with the mobie app? I'd have expected it would need some changes to support the redirection?

pgodwin:
> Would this work with something like Cloudflare Tunnels and their Access framework?

I don't think so.  I think proper JWT support would be required.

Sapd:
> > So I am using it since December successfully, just with the change in line 204: https://github.com/Sapd/audiobookshelf/commits/feat-forwardauth
> 
> Thanks @Sapd . I haven't had a chance to come back to this. Is it working properly with the mobie app? I'd have expected it would need some changes to support the redirection?

No, the mobile app will still need to handle redirection and show a popup then, preserving the state.

Sapd:
I did take a deep look into the App functionality and experimented a lot. Implementing it will be really difficult as Capacitor (which is the web runtime for mobile which audiobookshelf is using) does not make it easy.

Capacitor basically only supports `Browser.Open` for external browser-popup links: https://capacitorjs.com/docs/apis/browser

Now the problem is, when doing that, it does not preserve the state of the in-app browser but simply discards it. Displaying it in-app is also not possible as Capacitor prevents going to different urls directly in-app (one could use allowNavigation to prevent that, but that doesnt sound like a good idea; also the variable cannot be dynamically adjusted). 

To solve it, Audiobookshelf probably would need an API endpoint like `/authredirect` which is called similar to /ping for the first request, leading in authentik redirecting back to /authdirect after successful authentication. The API can then redirect again to audiobookshelf://applogin?sessionparameters opening up the app, while the app can store the session state. Another solution would be to write a plugin for capacitor...

So both solutions are really overly complicated and not feasible in my opinion.

The other PR here which suggests oauth #1636 would be easier to implement in the App, simply for the fact that there is actually a plugin for oauth and capacitor: https://github.com/moberwasserlechner/capacitor-oauth2

PR #1110, opened 2022-11-03
Added tone json file support

keaganhilliard:
This is an academic PR, I haven't been able to test it, but the logic checks out in my head haha.


keaganhilliard:
I was able to get everything up and running, and it all seems to work correctly, now.

advplyr:
Note sure what to do about the multiple files and embedding chapters either. I just added a note on that page if embedding in a multi-track audiobook that the chapters won't get embedded.
Did some testing and seems to be working well, thanks!

PR #1111, opened 2022-11-03
Makes it so the when sorting and ignoring prefixes, prefixes are actually ignored

ruoti:
I noticed that when sorting with ignore prefix on, items were not actually sorted correctly. For example, imagine having three items:

"The A"
"The A B"
"The A C"

This sorted to
"A B, The"
"A C, The"
"A, The"

This happens because the ", The" is part of the sort operation. I've modified the code so that is no longer the case. I chose to leave the Vue side of things alone, so the true sort value is only available to the server.


advplyr:
That makes sense, thanks. I didn't consider this one

PR #1120, opened 2022-11-07
Update zh-ch.json

springsunx:


advplyr:
thanks, I'll have a bunch more in a few days!

PR #1124, opened 2022-11-07
Update polish translation

konradorlinski:


advplyr:
thanks, I'll have a bunch more in a few days!

PR #1127, opened 2022-11-08
Update zh-ch.json

springsunx:


advplyr:
I think I accidently mislabeled this JSON file. It should be `zh-cn.json`, does that sound correct?

Would that be 汉语 (简化字)?

springsunx:
> I think I accidently mislabeled this JSON file. It should be `zh-cn.json`, does that sound correct?
> 
> Would that be 汉语 (简化字)?

Yes, it is 'zh-cn.json'. Language name is 中文简体.

Issue #1142, opened 2022-11-11
[Bug]: Incorrect value for total number of listening sessions on User Detail Page


austinphilp:
### Describe the issue

When navigating to `/config/users/<USER_ID>` on the web app, under the "Listening Stats" section there is a counter indicating the number of Listening Sessions over all time. It appears that this counter will always yield a value of `10` or lower, even if more than 10 listening sessions exist.

It appears at first glance like the code is taking the length of `recentSessions` in the `api/users/root/listening-stats` response, but `recentSessions` will only ever have at most 10 entries. We probably want to add a field to `listening-stats` `totalListeningSessions` (or something similar) which will have an int value representing the total number of sessions.

Happy to take this on myself if no one objects, I just started using this project a few weeks ago and would love to contribute!

### Steps to reproduce the issue

1. On the `config/users`  webpage, select a user with more than 10 listening sessions
2. Scroll down to the `Listening Stats` section
3. Observe that the `Listening Sessions` value is 10.

### Audiobookshelf version

v2.2.1

### How are you running audiobookshelf?

Docker

advplyr:
Yep that is a mistake for sure. The solution you mentioned with a `totalListeningSessions` variable sounds good to me. Let me know if you have questions setting up the project locally. I'm also active on our discord/matrix servers.

austinphilp:
@advplyr Sounds good! I'll probably work on getting it setup tonight, I'll find you on discord if something goes amiss

austinphilp:
fixed in #1144

austinphilp:
Re-opened - just realized we should leave this open until release

advplyr:
Fixed in [v2.2.3](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.3)

PR #1144, opened 2022-11-11
Fix listening sessions count bug

austinphilp:
Fixes the bug reported [here](https://github.com/advplyr/audiobookshelf/issues/1142) where listening sessions weren't being reported properly on the user details page.

Rather than adding a new field to `listening-stats` as discussed in the issue, I opted to utilize the `total` field already given by `listening-sessions`, necessitating a very minor refactor of how we get the `latestSession`


advplyr:
Nice, thanks!

PR #1156, opened 2022-11-14
Add French Translation

tomazed:
This is a French Translation for audiobookshelf


tomazed:
I have noticed a problem that is more of a syntaxical issue, there are two labels used in the library statistics page that sum up the total length of the library in hours:
image
but this seems to be two different labels:
`"LabelStatsOverall": "Overall",
"LabelStatsHours": "Hours",`
that are used.
I have translated
`"LabelStatsOverall": "Total",
"LabelStatsHours": "Heures",`
Problem being in French those should be inverted to read "727 Heures au total" or but mine is  727 Total Heures which is grammatically incorrect.
Any idea how to circumvent this? switching the translation Overall<>Hours seems to be a dirty fix to me and I do not know if the label are used elsewhere.

advplyr:
That label can either be "Overall Hours" or "Overall Days". If you have over 10,000 hours of audio then it will show "Overall Days".

If I create 2 separate strings for those would that be a viable solution?

I would add:
```json
"LabelStatsOverallDays": "Overall Days",
"LabelStatsOverallHours": "Overall Hours"
```

advplyr:
This is updated now and those 2 strings were added. https://github.com/advplyr/audiobookshelf/commit/5a60bb82677451f751972f0c9fd3238b79d313b8

If you can update this PR with those that would be great.

tomazed:
I will update, thanks

advplyr:
thanks!

PR #1175, opened 2022-11-18
Update it.json old version

burghy86:
minor fix update


advplyr:
Please update with the new it.json file that I had to update the order of strings.

PR #1182, opened 2022-11-19
Update fr.json from new strings in  advplyr/audiobookshelf@d24ed98bcd8ea6b36dd211b8a0829de7c09f165d

tomazed:
just updated the new strings (and two fix mistranslated)


advplyr:
Hah, just missed the release! Thanks

PR #1187, opened 2022-11-20
Add a few tooltips

lkiesow:
Starting to use audiobookshelf, the function of some buttons weren't very clear to me and while some buttons have tooltips, others have not.

This patch adds some additional tooltips to the user interface, further explaining some of the functionality.

_This doesn't modify any of the translations since that's often a separate process. I can adjust the other files if necessary._


client/strings/en-us.json
```diff
@@ -282,6 +284,7 @@
   "LabelRecentSeries": "Recent Series",
   "LabelRegion": "Region",
   "LabelReleaseDate": "Release Date",
+  "LabelRemoveCover": "Remove book cover",
```
advplyr, line 287:
```suggestion
  "LabelRemoveCover": "Remove cover",
```
This is also used for podcasts

client/strings/en-us.json
```diff
@@ -39,6 +39,7 @@
   "ButtonOk": "Ok",
   "ButtonOpenFeed": "Open Feed",
   "ButtonOpenManager": "Open Manager",
+  "ButtonPause": "Pause",
```
advplyr, line 42:
```suggestion
```
not used

advplyr:
Thanks!

lkiesow:
Thanks for the quick review and sorry for missing `ButtonPause`

PR #1188, opened 2022-11-20
Make Tooltips Accessible

lkiesow:
When using accessibility tools like screen magnifiers, dynamic screen content can be quite problematic. In particular content, which only appears if you interact with elements somewhere else on the screen. That is the case, for example, with the current implementation of tooltips used by audiobookshelf.

This patch provides a slight adjustment, keeping the tooltips open if you hover over them. This allows users to have better access to the content.

For those unfamiliar with e.g. screen magnification tools, here is a short example:

https://user-images.githubusercontent.com/1008395/202920990-3df731c9-29a6-442d-b845-61364d45c20b.mp4

You couldn't have read the right half of the tooltip if it would still disappear immediately.

### DIFF
--- a/client/components/ui/Tooltip.vue
+++ b/client/components/ui/Tooltip.vue
@@ -21,7 +21,8 @@ export default {
     return {
       tooltip: null,
       tooltipId: null,
-      isShowing: false
+      isShowing: false,
+      hideTimeout: null
     }
   },
   watch: {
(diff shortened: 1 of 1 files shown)


advplyr:
Works great, thanks!

itzexor:
@lkiesow Hello sir, rather than create a new issue I figure I could bring up an issue here where the feature was implemented (I am assuming). I think this change introduced this usability regression:

Recording_2023-12-18_at_03 25 20
which I summarized and commented upon on discord as follows:
`because of the way tooltips are designed here, this UI element is unapproachable from like 110°`
`tooltips really shouldn't even receive events, in most cases they should bubble through and hit on the actual interactive UI element underneath or be dismissed (then bubble it) unless they are an interactive popup (like a notification or something) `
However, advplyr mentioned that it was intentionally changed to do this, and so I searched and found this.

I can totally agree with the rationale behind implementing it this way, but it does seem to come with some slight issues in these edge cases. I think the value of accessibility support outweighs minor "navigation inconveniences" like this, and so I would just ignore it most likely...but I wanted to know if you had any comment on this situation.

Perhaps moving forward the best resolution is to consider this limitation of tooltips when designing adjacent elements or something... Alternatively, a delay before showing the tooltip would likely completely solve this issue and is pretty much the standard for tooltip "display timing", though personally I do kind of prefer the immediate info.

PR #1192, opened 2022-11-21
Update Author Modal on Changes

lkiesow:
If you are on the home page and open the edit author modal, you can automatically update all data by clicking “Quick Match” or you can remove a set image by clicking 🗑.

Both options will update the actual data, but not the data in the open modal. This means that, for example, a picture is still shown in the modal after deleting it. That's confusing.

This patch fixes the bug and makes sure the modal is updated if the data is updated.


springsunx:
I also found this problem. I prefer to automatically get the author's photos in the author directory.

advplyr:
Nice, thanks!

PR #1193, opened 2022-11-21
No feed log level

lkiesow:
This patch drops the log level for logging requests to non-existing feeds from error to debug. The reasoning behind this is that this is a client error and a client error is returned and can be handled by the client.

Otherwise anyone can easily spam the logs with error messages by just requesting non-existing feeds.


advplyr:
Yeah that's true since these are unprotected routes. We may also consider using the rate limit library already in this project to prevent spamming these.

PR #1196, opened 2022-11-21
Update fr.json

tomazed:


tomazed:
- Updated the newest strings
- I've changed some labels that are used for both books and series to include both genders to be grammaticaly correct

advplyr:
Thanks! We can separate those out if you think it is necessary.

tomazed:
> Thanks! We can separate those out if you think it is necessary.

I don't want to burden the translation process more than neccessary. The notation appending "(e)" is valid to refer to both feminine and masculine gender of words, although not very nice to read.
Most language do not bear a gender attribute. Maybe put a pin on this for a later time or if other translator thinks it's necessary for their language.

advplyr:
Makes sense!

PR #1212, opened 2022-11-24
Improve title naming for single file audiobooks when opening RSS feed

Weldawadyathink:
When opening an RSS feed, Audiobookshelf uses the track titles for the individual episodes in the rss feed. This leads to problems with single file audiobooks. The entire book will be titled, for example "Opening Credits" or "Chapter One". This changes the rss feed generator to use the book title instead of the audio title if the book only contains one audio file.


advplyr:
This makes sense to me but I think we should implement it slightly differently.
The function `setFromAudiobookTrack` could receive an additional parameter with either the number of tracks or a flag for if it is a single track audiobook.
Using that parameter would only require a few extra lines of code around setting the title.

Also, there are 2 places where we set the FeedEpisode. The other is `updateFromItem` in `Feed.js`

Weldawadyathink:
Thanks for the feedback. I haven't used JavaScript much before, so I didn't really know what the best way to implement this was. I will reimplement and update the PR.

Weldawadyathink:
Updated. It seemed more elegant to move the implementation details to the setFromAudiobookTrack function. This way, the Feed.js only deals with the high level "is this a podcast or audiobook" decision. It also nicely handles creating and updating an rss feed.

server/objects/FeedEpisode.js
```diff
@@ -92,11 +92,15 @@ class FeedEpisode {
...
+    if (libraryItem.media.tracks.length == 1) { // If audiobook is a single file, use book title instead of chapter/file title
+      var title = libraryItem.media.metadata.title
+    } else {
+      var title = audioTrack.title
+      if (libraryItem.media.chapters.length) {
+        // If audio track start and chapter start are within 1 seconds of eachother then use the chapter title
+        var matchingChapter = libraryItem.media.chapters.find(ch => Math.abs(ch.start - audioTrack.startOffset) < 1)
+        if (matchingChapter && matchingChapter.title) title = matchingChapter.title
+      }
     }
```
advplyr, line 104:
```suggestion
    let title = audioTrack.title
    if (libraryItem.media.tracks.length == 1) { // If audiobook is a single file, use book title instead of chapter/file title
      title = libraryItem.media.metadata.title
    } else {
      if (libraryItem.media.chapters.length) {
        // If audio track start and chapter start are within 1 seconds of eachother then use the chapter title
        var matchingChapter = libraryItem.media.chapters.find(ch => Math.abs(ch.start - audioTrack.startOffset) < 1)
        if (matchingChapter && matchingChapter.title) title = matchingChapter.title
      }
    }
```

advplyr:
Great thanks! 
I just moved the `title` variable out of the if/else block. One of the ugly parts of Javascript is a `var` variable is not block scoped which means you can define it inside a an if/else block and it is available outside of it.
That's why JS introduced the `let` variable since it IS block scoped. I've been trying to use that more recently but you'll see I used `var` all over the place out of old habit.

PR #1220, opened 2022-11-25
Restrict Library Paths

lkiesow:
This patch introduces the new environment variable `MEDIA_BASE_PATH` allowing to restrict the media paths that can be added to libraries and therefore limiting which sections of the filesystem audiobookshelf will scan and potentially serve to the web.

This means, you can, for example, specify that all media are in `/audiobooks`, preventing users to access file system paths not starting with this.

This also prevents things like users specifying `/` as a library path which seems to be a good way for a denial of service attack right now.

The patch also adds the new environment variable to the example Docker commands and docker-compose files, matching the media file specified in those examples.


advplyr:
I'm not sure about this. In the example docker command I wouldn't know what that is and if I included it in my docker run command I wouldn't be able to create the `/podcast` folder that I mapped.

I don't understand the use-case fully. Only admin users can create libraries so you are concerned about an admin user creating a library with a bad folder path?

advplyr:
In Docker only the folders you map to the container will be accessible to Abs. So you should only map the folders you want Abs to have access to.

lkiesow:
You are right, the example isn't the best and there should be either a base path in the example covering both audiobooks and podcasts, or you should be able to specify multiple paths.

The concern is that someone exposes something that shouldn't be exposed. As you pointed out, you need high privileges in audiobookshelf to do that, but for one, limiting the risk if a breach happens is never a bad idea and also, this limits the risk of accidents like e.g. adding `/`. Audiobookshelf didn't really like when I did that.

Docker is also only one deployment method and this is usable for other deployment methods as well.

This isn't a high risk issue, but I don't think hardening this a bit more hurts anyone.

lkiesow:
I thought a bit more about this. Trying to cover multiple paths with a mechanism like this makes it easily very complicated. Simply using something like a comma separated list would mean that we would need to escape actual comma characters if the path contains any, … Better to stay with a very simple mechanism.

I removed this from the example, but you can now still easily use this to limit the library access to a part of the file system, e.g. having all library paths in `/data/`.

What do you think?

advplyr:
I think it will work but you also need to handle the create library function at the top of that script.

lkiesow:
> …but you also need to handle the create library function at the top of that script.

Updated the patch

PR #1221, opened 2022-11-26
Implement a Settings Content component

jmt-gh:
In previous commits, we consolidated all of the settings pages to have the same styling. This commit takes that one step further by extracting out all of that logic to a singular component. This component wraps the content of the settings pages themselves, so it is very easy to use. As well, it has the functionality to pass through the header text, description text, and a method that could be needed to "add a new one" of something for the page (users, libraries, etc.).


advplyr:
Makes sense, thanks!

PR #1231, opened 2022-11-29
Server respond with objects

k9withabone:
As discussed on matrix, this updates the server so that it always responds with an object, never an array. This simplifies processing the response for clients in other languages. I didn't actually test the changes so that probably should be done before merging.


advplyr:
I realize now a lot of these are used in the mobile app. Collections, libraries, shelves, etc. so this is going to have to wait until the next mobile app release.

advplyr:
Unfortunately I missed the `/personalized` and the `/search/podcast` route in the mobile app.
I reverted those as well as `/search/books`.

Since I want to merge this without waiting for another mobile app release we will have to lease those 3 out.

PR #1247, opened 2022-12-05
Update GoogleBooks.js

aurokon:
Further updated to check for extraLarge, then fallback to Large then to thumbnail and then to null if not available.


advplyr:
We'll go with another approach as discussed in Discord

PR #1251, opened 2022-12-05
No PermissionsStartOnly=true

lkiesow:
This patch removes `PermissionsStartOnly=true` from the systemd unit file used for packaging. This shouldn't be necessary for any commands run by the unit.

Note that while I'm pretty sure that this is not necessary, I did not test this.
We should do that before merging this.

### DIFF
--- a/build/debian/lib/systemd/system/audiobookshelf.service
+++ b/build/debian/lib/systemd/system/audiobookshelf.service
@@ -11,7 +11,6 @@ ExecReload=/bin/kill -HUP $MAINPID
 Restart=always
 User=audiobookshelf
 Group=audiobookshelf
-PermissionsStartOnly=true
 
 [Install]
 WantedBy=multi-user.target


advplyr:
Was this requiring the systemd to be run with sudo?

lkiesow:
> Was this requiring the systemd to be run with sudo?

I'm not actually sure what you mean?

- Systemd is the init system and should always have the UID 0 and therefor superuser privileges.
- If you are asking if this changes, if you can start or stop the system service with(out) root privileges now (`systemctl start audiobookshelf.service`), this patch should cause no difference in behavior.
- If your question is if you can now run this using the service manager of the calling user (`systemctl --user start audiobookshelf.service`), this still shouldn't work, given the rest of the unit specification.

The difference `PermissionsStartOnly` should make is only if all commands are run using the specified user and group, or just the start command. In practice, that means that without this patch `ExecReload=/bin/kill -HUP $MAINPID` is run as user `root` which are unnecessarily high privileges. If the service is run as user `audiobookshelf` it should suffice to run `kill` as that user as well. That's what this patch should change.

advplyr:
Thanks for the clarification. I saw 2 users post that they needed to use root to run the service. I can't remember now where I saw those reports but I popped over here to mention it incase it was related.

PR #1268, opened 2022-12-14
Improve docker dev container support

Weldawadyathink:
The dev container config was missing a few necessary features. This adds the following features to the dev container:

- Adds tone to the dev container
- moves client/node_modules to a docker volume for performance reasons
- configures git project directory to be trusted, preventing git complaining about ownership
- Automatically install node modules and generate client
- adds dev.default.js. This provides a default config that should work for most linux dev environments. Can be overridden by creating dev.js, so this will not interfere with existing setups.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,10 @@
 FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:16
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-    && apt-get install ffmpeg gnupg2 -y
+
+COPY --from=sandreas/tone:v0.1.2 /usr/local/bin/tone /usr/local/bin/
+
 ENV NODE_ENV=development
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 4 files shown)


advplyr:
I don't know anything about devcontainers. Is this an overlap of #1107 ?

Weldawadyathink:
Oh, yes it definitely is. I really need to learn to search before putting the work in. Oh well, it was a good chance to learn more about docker. 

Dev containers just allow you to develop inside a docker container. This normalizes the dev environment, so everyone can develop with the same tools. (Extra important for me, it includes tone. I develop on an M1 Mac, and tone currently has some issues with the M1. It works just fine on docker.)

The other PR looked like it was missing a few things that I had to solve, in particular with GitHub code spaces. When I get a chance, I can go through that PR and test it out. Also it doesn't package client/node_modules in a docker volume. This isn't necessary, but is good for performance reasons. Docker bind mounts are not very good for large numbers of small files like node creates. This can be closed as duplicate for now.

advplyr:
Awesome thanks. I haven't reviewed the other PR just because I haven't dug into devcontainers yet. Your review/suggestions will be helpful

PR #1271, opened 2022-12-15
Missing Localization in Appbar.vue

tomazed:
A string which is translated was hard coded...
I just changed it.


client/components/app/Appbar.vue
```diff
@@ -68,7 +68,7 @@
           <ui-icon-btn :disabled="processingBatch" icon="collections_bookmark" @click="batchAddToCollectionClick" class="mx-1.5" />
         </ui-tooltip>
         <template v-if="userCanUpdate">
-          <ui-tooltip text="Edit" direction="bottom">
+          <ui-tooltip text="$strings.LabelEdit" direction="bottom">
```
advplyr, line 71:
```suggestion
          <ui-tooltip :text="$strings.LabelEdit" direction="bottom">
```

PR #1282, opened 2022-12-17
Use HTTPS for Google Books Images

lkiesow:
The API for Google Books will return HTTP image URLs when matiching any books using it as a search provider. In a secure environment, this causes browser warnings.

Screenshot from 2022-12-18 00-05-36

Screenshot from 2022-12-18 00-02-08

All Google image links support HTTPS and we can safely switch to HTTOS to avoid these warnings.


advplyr:
Looks good. I wonder why they return the insecure url even when requested by a secure url.

PR #1286, opened 2022-12-18
Update fr.json with new strings from d7cc8a0

tomazed:
Localization strings for creating a Playlist from a Collections


advplyr:
Thanks!

PR #1297, opened 2022-12-20
Add Caddyfile example to readme

Eschguy:
An example Caddyfile based on what I use

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -265,18 +265,8 @@ From [@Dondochaka](https://discord.com/channels/942908292873723984/9429141542541
 
 ```
 subdomain.domain.com {
-        tls internal
-        header {
-                Strict-Transport-Security max-age=31536000
-                X-Content-Type-Options nosniff
-                X-Frame-Options DENY
-                Referrer-Policy no-referrer-when-downgrade
-                X-XSS-Protection 1
-        }
         encode gzip zstd
-        reverse_proxy LOCAL_IP:PORT {
-                trusted_proxies xxx.xx.xxx.x/xx xxx.xx.xxx.x/xx
-        }
+        reverse_proxy <LOCAL_IP>:<PORT>
 }
 ```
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -261,6 +261,26 @@ Middleware relating to CORS will cause the app to report Unknown Error when logg
...
+                X-Content-Type-Options nosniff
+                X-Frame-Options DENY
+                Referrer-Policy no-referrer-when-downgrade
+                X-XSS-Protection 1
+        }
+        encode gzip zstd
+        reverse_proxy LOCAL_IP:PORT {
+                trusted_proxies xxx.xx.xxx.x/xx xxx.xx.xxx.x/xx
+        }
+}
```
k9withabone, line 280: I'm not sure including `tls internal` in the example is a good idea. Also, I'm not sure if those headers are necessary.
The `trusted_proxies` option of `reverse_proxy` is only necessary if you are using Cloudflare tunnels or a CDN, I think.

Personally, this is what I do in my Caddyfile:

```
abs.example.com {
	reverse_proxy * <LOCAL_IP>:<PORT>
	encode zstd gzip
}
```

k9withabone (changes requested):

readme.md
Eschguy: > I'm not sure including `tls internal` in the example is a good idea. 

Probably not, that's a situational thing.

> Also, I'm not sure if those headers are necessary.

They are not, just tested on my setup and it worked without them. Not sure why I had them there.

> The trusted_proxies option of reverse_proxy is only necessary if you are using Cloudflare tunnels or a CDN, I think.

You are correct, I use it because I have Cloudflare as my DNS provider.

advplyr:
Is this going to be updated?

Eschguy:
@advplyr  Sorry I thought I had made the changes, they're up on master now.

Issue #584, opened 2022-05-11
[Bug]: AUDIOBOOKSHELF_UID & AUDIOBOOKSHELF_GID values not being honored


wshyang:
### Describe the issue

I started a container following the sample docker command provided on the main page

`docker run -d \
    -e AUDIOBOOKSHELF_UID=65500 \
    -e AUDIOBOOKSHELF_GID=65500 \
    -p 13378:80 \
    -v $PWD/audiobooks:/audiobooks \
    -v $PWD/config:/config \
    -v $PWD/metadata:/metadata \
    --name audiobookshelf \
    --rm ghcr.io/advplyr/audiobookshelf`

However subsequently when checking the contents of the files created in the volumes, I found that they are all created under root:root. It appears that within the container itself, the process is running as root too.

### Steps to reproduce the issue

1. `docker run -d \
    -e AUDIOBOOKSHELF_UID=65500 \
    -e AUDIOBOOKSHELF_GID=65500 \
    -p 13378:80 \
    -v $PWD/audiobooks:/audiobooks \
    -v $PWD/config:/config \
    -v $PWD/metadata:/metadata \
    --name audiobookshelf \
    --rm ghcr.io/advplyr/audiobookshelf`
2. Check permissions of created files

### Audiobookshelf version

2.0.12

### How are you running audiobookshelf?

Docker

advplyr:
A lot of updates were made to setting ownership of files/folders created by abs. If this is still an issue we can re-open this but please provide specific folders/files that are not set correctly.

advplyr:
@raviolispy Thanks. I can confirm the database files being created in /config and the logs in /metadata/logs are not getting the ownership set correctly.

advplyr:
Fixed in [v2.2.9](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.9)

PR #1301, opened 2022-12-20
Dockerfile update

ceramicwhite:
- Separate deps and build stage
- Using [linuxserver/docker-baseimage-alpine](https://github.com/linuxserver/docker-baseimage-alpine) for base image of final stage
  - 30% reduced image size 
  - using s6 for pid 1
  - No more running app as root
  - Ability to add Docker Mods such as Apprise or Cloudflared via `envs`:
  ```
     environment:
        DOCKER_MODS=linuxserver/mods:universal-apprise
        DOCKER_MODS: linuxserver/mods:universal-cloudflared
        CF_ZONE_ID: zone_id
        CF_ACCOUNT_ID: acct_id
        CF_API_TOKEN: token
        CF_TUNNEL_NAME: example
        CF_TUNNEL_PASSWORD: pleasedontusethisexamplepassword
        CF_TUNNEL_CONFIG: |
          ingress:
            - hostname: test.yourdomain.url
              service: http://localhost:80
            - service: http_status:404
  ```


advplyr:
I don't know Docker well enough yet to understand all that is going on here.

With the `build/docker/root/etc/services.d/audiobookshelf/run` script if this is getting executed before the server is started then all of those config files won't exist yet. Those config files are created in `/server/Db.js`. Is it possible that you already had these config files created before v2.2.9?
The ownership of those config files was fixed in 2.2.9 https://github.com/advplyr/audiobookshelf/issues/584

What is "abc" here `chown -R abc:abc /config/`?

ceramicwhite:
@advplyr Yeah those files wouldn't exist for new users, I wanted to make sure that those upgrading wouldn't have any problems also. I've cleaned it up, I was checking the folders permissions and only changing if the user `abc` was not already owner of the folder, but the base image creates `/config` already, anyways I've fixed it.
I tested the image with server files from your image pre-2.2.9 and I had no problems switching over.

The `abc` user is just the name linuxservers gives as the default unprivileged user across their fleet. I've set it so `abc` will have `UID:GID` based on what's set for `AUDIOBOOKSHELF_UID:AUDIOBOOKSHELF_GID`

Personally I would also remove the `healthcheck`, if people want it they can add it to their docker-compose.yml

PR #1303, opened 2022-12-20
Add AD Integration

BrianCArnold:


BrianCArnold:
Just a simple implementation designed to not interfere with how internal authentication works, if the required environment variables are not set.

advplyr:
I'm not familiar with LDAP but I'm looking at this activedirectory package and thinking we should find an alternative.
That package would introduce many dependencies, some of which are no longer necessary in modern node version (e.g. `async`).

I'll need to spend more time learning about LDAP to provide better suggestions.

BrianCArnold:
I can implement it with a different package pretty easily. I'll take look around.
________________________________

BrianCArnold:
I reimplemented things using ldapjs, and moved the implementation details to `server/authProviders/ldap.js`, so that other auth providers (forward auth/OIDC/etc.) can be implemented and put there, instead of filling up the Auth.js file with lots of complex implementations. 

I'd like to provide TS interfaces and implementations, so that each auth provider implements the same interface. Let me know if you'd be comfortable with me doing that.

Thanks.

BrianCArnold:
Just realized I had modified the Dockerfile for testing, I removed that modification and force pushed. It's also rebased on top of the updated master.

BrianCArnold:
I'm going to work on fixing some (very minor) security issues with the current implementation, then try to generalize authorization and provide alternate ways to do auth, including forward auth, ldap, OAuth, etc.

PR #1310, opened 2022-12-22
Socket fixes

k9withabone:
Some fixes for various server socket events I've noticed as I've been working on documenting the socket events.


server/controllers/UserController.js
```diff
@@ -48,7 +48,7 @@ class UserController {
     var newUser = new User(account)
     var success = await this.db.insertEntity('user', newUser)
     if (success) {
-      SocketAuthority.adminEmitter('user_added', newUser)
+      SocketAuthority.adminEmitter('user_added', newUser.toJSONForBrowser())
```
k9withabone, line 51: Full user with password hash was being sent.

server/objects/Feed.js
```diff
@@ -61,6 +61,15 @@ class Feed {
...
 
+  toJSONMinified() {
+    return {
+      id: this.id,
+      entityType: this.entityType,
+      entityId: this.entityId,
+      feedUrl: this.feedUrl
+    }
+  }
+
```
k9withabone, line 72: Added `Feed.toJSONMinified()` for `RssFeedManager` lines 109 and 123.

server/routers/ApiRouter.js
```diff
@@ -476,7 +476,7 @@ class ApiRouter {
         }
         if (newAuthors.length) {
           await this.db.insertEntities('author', newAuthors)
-          SocketAuthority.emitter('authors_added', newAuthors)
+          SocketAuthority.emitter('authors_added', newAuthors.map(au => au.toJSON()))
```
k9withabone, line 479: Mapping `toJSON()`.

server/scanner/Scanner.js
```diff
@@ -477,7 +477,7 @@ class Scanner {
       })
       if (newSeries.length) {
         await this.db.insertEntities('series', newSeries)
-        SocketAuthority.emitter('series_added', newSeries.map(se => se.toJSON()))
+        SocketAuthority.emitter('multiple_series_added', newSeries.map(se => se.toJSON()))
```
k9withabone, line 480: Added `multiple_series_added` event to distinguish from a single series being added.

server/routers/ApiRouter.js
```diff
@@ -500,7 +500,7 @@ class ApiRouter {
         }
         if (newSeries.length) {
           await this.db.insertEntities('series', newSeries)
-          SocketAuthority.emitter('authors_added', newSeries)
+          SocketAuthority.emitter('multiple_series_added', newSeries.map(se => se.toJSON()))
```
k9withabone, line 503: Pretty sure this was supposed to emit series added.

server/scanner/Scanner.js
```diff
@@ -984,7 +984,7 @@ class Scanner {
         Logger.info(`[Scanner] matchLibraryItems: Library match scan canceled for "${libraryScan.libraryName}"`)
         delete this.cancelLibraryScan[libraryScan.libraryId]
         var scanData = libraryScan.getScanEmitData
-        scanData.results = false
+        scanData.results = null
```
k9withabone, line 987: Makes the emitted output consistent with `scan()`, see line 158.

server/controllers/SeriesController.js
```diff
@@ -41,7 +41,7 @@ class SeriesController {
     const hasUpdated = req.series.update(req.body)
     if (hasUpdated) {
       await this.db.updateEntity('series', req.series)
-      SocketAuthority.emitter('series_updated', req.series)
+      SocketAuthority.emitter('series_updated', req.series.toJSON())
     }
     res.json(req.series)
```
k9withabone, line 46: Should `toJSON()` be added for the API response?

```suggestion
    res.json(req.series.toJSON())
```

k9withabone (commented):
Mostly just added `toJSON()` to things to make sure it's consistent with the API responses and other socket events.

server/controllers/SeriesController.js
advplyr: It's not required since the socket library is going to make that call but in our case I think it is a good idea because we differentiate between toJSONExpanded, etc.

advplyr:
Excellent, thank you. The `multiple_series_added` made me realize we should probably standardize these similar to what we did with the API responses.
I'm thinking we combine the single & multi socket events. Then if it is a single update then it would just be an array of one object.
Also, changing the socket events to always emit an object same as the API. This so that if we implement socket listeners in other languages they have an easier time parsing the JSON.

This would require a lot of updates in web and mobile

k9withabone:
> I'm thinking we combine the single & multi socket events. Then if it is a single update then it would just be an array of one object.

Makes sense to me.

> Also, changing the socket events to always emit an object same as the API.

There are also a handful of events (on the client and server side) that just emit strings or integers. I think wrapping those in objects would be a good idea as well.

PR #1320, opened 2022-12-26
Fix undefined string assignment

lkiesow:
Assigning something to `process.env.profile`, Node stringifies the value. This means that assigning `undefined` to an environment variable in Node will result in it holding the string `undefined`.

This means, for example, that `module.exports.FFPROBE_PATH || 'ffprobe'` in `server/libs/nodeFfprobe/index.js` will actually result in the string `undefined`.

This patch fixes several such assignments in the `index.js`, potentially causing problems in the development mode.


advplyr:
Makes sense, thanks!

PR #1333, opened 2022-12-28
Highlight items when navigating via keyboard

lkiesow:
This patch highlights items in the app bar if a user uses the keyboard to navigate in audiobookshelf. This ensures that users actually know which item they have selected.

Screenshot from 2022-12-28 22-51-24

This also modifies the text for the library selector, so that users which are using a screen reader understand that it is a selector for libraries and not only a button related to the current library.


advplyr:
Thanks!

PR #1335, opened 2022-12-28
Fix keyboard navigation in library selection

lkiesow:
This patch fixes the keyboard navigation in the library selection of the main app bar. Without this patch, no options are selectable via keyboard and selecting an option and hitting return has no effect.


advplyr:
Works great

PR #1336, opened 2022-12-29
Make User Settings Accessible via Screen Reader

lkiesow:
This patch should fix most of the problems for users trying to access the user settings via screen reader. It makes sure user interface elements can be reached via keyboard and provides proper labels, roles and values so you not only can interact with elements but also know what you are actually changing.

While not focused on other views, this should also already fix a number of accessibility issues with other settings pages.


advplyr:
Looks great, thanks!

PR #1338, opened 2022-12-29
Fix incorrect series and seriespart tag codes, they were swapped

naleo:
When reading in metadata from tags, they series and series part were swapped.  I checked, and it appears that MVIN and MVNM were swapped in these files.  I have corrected this, so that it is mapped like the following:
series -> MVNM
series part/number -> MVIN


advplyr:
I wonder how that was backwards so long without anyone noticing..
Thanks!

PR #1341, opened 2022-12-29
Fix keyboard navigation in user settings

lkiesow:
This patch makes sure that the option in the user settings are accessible via keyboard navigation and that the labels, if users use a screen reader, actually make sense.

This patch introduces new strings which need to be translated. Although I did already provide a German translation.

As a minor visual details, this patch also aligns the delete and edit icons:

Screenshot from 2022-12-29 21-03-12


advplyr:
Great, thanks!
Also mentioned in Discord about given permission to commit to your branches.

PR #1343, opened 2022-12-30
Accessibility Improvements for Main Settings

lkiesow:
This patch fixes some accessibility problems on the main settings page. Most notably, it makes sure that the different options have labels which are picked up by screen readers.

As a more generic addition, this also makes sure that the dropdown component will always have a proper label constructed, explaining what the dropdown is for and what its current value is.


advplyr:
Great!

PR #1350, opened 2023-01-01
Fix Hidden Settings Menu

### LINKED ISSUE #1334: [Bug/Usability]: Hidden Settings Menu
### Describe the issue

If you limit the screen _height_ for audiobookshelf the main settings menu is hidden and there is no obvious way to open the menu. For me, the only way to access this was unrecognizable.

- Open Audiobookshelf
- Go to settings
- Limit browser height

https://user-images.githubusercontent.com/1008395/209881189-5b1f912f-5155-47fa-bb37-93183cdf7161.mp4

### Audiobookshelf version

v2.2.11

lkiesow:
This patch fixes several problems of the settings menu related to display on mobile devices or small(ish) windows:

- The `isMobileLandscape` is now calculated correctly. Previously, this was set to `true` if a device was in portrait mode.

- Showing the button to collapse the settings menu and making the menu collapsible now use the same mechanism. Previously, it could happen that the menu was opened and not fixed, but no button to close it again was shown.

- The icons fore opening and closing the settings menu are now both arrows, indicating that their functionality is reversed.

- The button to open the menu now always has the string “Settings”, instead of using the name of the current page. The current page hader is listed below that anyway and this is the action component to open the settings menu after all.

This fixes #1334


advplyr:
Thanks!

PR #1364, opened 2023-01-04
Update Server Libraries

lkiesow:
This patch updates the server libraries to their latest state.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
       "version": "2.2.11",
       "license": "GPL-3.0",
       "dependencies": {
-        "axios": "^0.26.1",
+        "axios": "^1.2.2",
         "express": "^4.17.1",
         "graceful-fs": "^4.2.10",
         "htmlparser2": "^8.0.1",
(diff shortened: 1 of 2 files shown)


lkiesow:
I didn't use this in production, but I did click through several things in the ui including functions like fetching metadata and cover images (where I guess axios is being used) and didn't see any issues. Do you have anything in particular you would want me to test?

advplyr:
I haven't gone through it yet but axios is used for podcast RSS feeds and fetching data from metadata providers.

lkiesow:
I now got around testing that podcasts also work and not only fetching metadata for books. There seems to be no problem and everything works well.

For Podcasts, I've tested adding new Podcasts via RSS URL as well as searching for Podcasts, searching for episodes, updating and downloading episodes. I didn't detect any problems.

advplyr:
Thanks!

lkiesow:
I'm going to test the RPM auto-updates later today.
If things work there should indicate if this is a general problem.
I'll let you know.

But Docker seems towork at least.

PR #1369, opened 2023-01-05
Slightly increase width of left navigation bar

lkiesow:
This patch slightly increases the width of the left navigation bar to allow for a little bit of padding on the “Collections” link which looked odd pressed right to the edge of the navigation.

### DIFF
--- a/client/components/app/SideRail.vue
+++ b/client/components/app/SideRail.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="w-20 bg-bg h-full fixed left-0 box-shadow-side z-50" style="min-width: 80px" :style="{ top: offsetTop + 'px' }">
+  <div class="w-20 bg-bg h-full fixed left-0 box-shadow-side z-50" style="min-width: 90px" :style="{ top: offsetTop + 'px' }">
     <!-- ugly little workaround to cover up the shadow overlapping the bookshelf toolbar -->
     <div v-if="isShowingBookshelfToolbar" class="absolute top-0 -right-4 w-4 bg-bg h-10 pointer-events-none" />
 


advplyr:
What browser is that from? Mine isn't like that in Brave

image

k9withabone:
Here's Firefox on Fedora KDE Plasma for comparison:
image

lkiesow:
I dug a bit deeper and the difference seems to be the requested font-family. Requested is `Gentium Book Basic, serif`. Seems like none of my systems have `Gentium Book Basic`, so they fall back to some `serif` font.

In Google Chrome (Fedora Linux, Cinnamon), that's “Times New Roman” which looks okay:

Screenshot from 2023-01-06 11-00-50

In Firefox (Fedora Linux, Cinnamon), that's “DejaVu Serif” which is less condensed and that's why “Collections” looks odd:

Screenshot from 2023-01-06 11-00-57

On Android (both Firefox and Chrome) it's somewhat in between:

index

Maybe it would make sense to make the width dynamic (e.g. text width with ~10px margin) or include the selected font as a web font?

advplyr:
That font is included in the client already so I'll have to investigate why it is not loading.

https://github.com/advplyr/audiobookshelf/tree/master/client/static/fonts

advplyr:
Can you check if the font is loading?
image

I'm not able to reproduce this

springsunx:
image
It looks like this. I'm not sure it is?

GetOutOfMyBakery:
I'm seeing something similar on Firefox and Safari with Mac 12.6.2 (Montery):

- Font has loaded:
Screenshot 2023-01-07 at 11 44 55 am

- Falling back to Times New Roman (locally installed), because of the 'serif' fallback option:
Screenshot 2023-01-07 at 11 44 10 am

For what it's worth, I'm a fan of Source Sans Pro being used rather than a serif font here:
Screenshot 2023-01-07 at 12 16 02 pm

It's also in-keeping with the Settings sidebar:
Screenshot 2023-01-07 at 12 13 27 pm

Unrelated, Audiobookshelf is fantastic, @advplyr. Thanks for this project!

lkiesow:
> Can you check if the font is loading?

The file `/_nuxt/fonts/GentiumBookBasic.d46ef40.woff2` is being loaded, but not being used. Out of curiosity, I just downloaded a WOFF2 file for “Gentium Book Basic” and replaced `./client/static/fonts/GentiumBookBasic.woff2`. As soon as I did that, the font was picked up.

Are you sure the font is not picked up from your system? The inspector in Firefox has a pretty good overview of available fonts and where they are being loaded from:

Screenshot from 2023-01-07 19-45-50

advplyr:
I found the issue. It was the way `@font-face` was being declared. Fixed in https://github.com/advplyr/audiobookshelf/commit/b00510855e4ce859a9d463f321667c134b2b5f3f

I don't think we need to expand the side nav now

lkiesow:
Looking at the German translation of ABS, it may still make sense. This is using Gentium Book Basic:

lkiesow:
> I found the issue. It was the way `@font-face` was being declared. Fixed in [b005108](https://github.com/advplyr/audiobookshelf/commit/b00510855e4ce859a9d463f321667c134b2b5f3f)
> 
> I don't think we need to expand the side nav now

That doesn't seem to fix the problem for me. The font is still not being used.
I simply re-downloaded the font. That seems to fix the problem. I'll create a pull request for that patch.

advplyr:
No matter what there is going to be some languages that are going to have to wrap the text. It probably makes more sense to add some x padding so it doesn't touch the edges.

PR #1374, opened 2023-01-07
Update Gentium Book Basic Font

lkiesow:
This patch updates the Gentium Book Basic font file [1]. While I couldn't get any client to use the previous file, it doesn't seem to be a problem with this file and now the text is being rendered correctly.

The previous attempt at fixing this (b005108) doesn't seem to work.
For more details about the issue, take a look at #1369.

[1] https://gwfh.mranftl.com/fonts/gentium-book-basic?subsets=latin


advplyr:
Thanks, I must have gotten a copy with limited Unicode range or something.

PR #1388, opened 2023-01-10
Show next book only if previous book is finished

### LINKED ISSUE #1382: [Enhancement]: Don't show the next book in series if you are still listening to the previous book

If I start listening to a book in a series, let's say I start listening to “A Game of Thrones”, the next book in the series, “A Clash of Kings”, would immediately show up on the home page below the one I listened to in the category “Continue Series”. Wouldn't it make more sense to only show the book there once I finish the first one?

Reasoning:

- Assuming, I'm still listening to the first book, I'm not jumping straight into the next one before finishing. In this case, I don't need that book there.
- Assuming I abandon a book, it's unlikely that I will start the next one in the series at all. That means, I don't need a prominent listing.
- While still listening to a book, I might decide to take a break. In that case I might decide to pick up another book from a series I have read before. That means, the next one in the series I haven't yet finished the last one of is kind of in the way.

I've been wondering for a while why they are listed there, and I don't see a good reason. Any thoughts on this?

lkiesow:
This patch changes the books displayed in “Continue Series”, avoiding books if another book from the series is played back right now. This prevents Audiobookshelf suggesting books to which users will not listen to because they are still listening to the last one.

Once a book is finished, the next book in the series will pop still be suggested to the user.

This fixes  #1382

### DIFF
--- a/server/utils/libraryHelpers.js
+++ b/server/utils/libraryHelpers.js
@@ -499,6 +499,7 @@ module.exports = {
           for (const librarySeries of libraryItem.media.metadata.series) {
             const mediaProgress = allItemProgress.length ? allItemProgress[0] : null
             const bookInProgress = mediaProgress && (mediaProgress.inProgress || mediaProgress.isFinished)
+            const bookActive = mediaProgress && mediaProgress.inProgress && !mediaProgress.isFinished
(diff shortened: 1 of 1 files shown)


advplyr:
Works great, thanks

PR #1390, opened 2023-01-10
Toggle switch shouldn't submit form

lkiesow:
This patch fixes the problem that toggling one of the options in the user account dialog will automatically submit the form.

The problem got introduced as a combination of the recent accessibility fixes where some elements got turned into HTML button elements to make them keyboard accessible. Doing that, I did not realize that [the default type of a button is `submit`](https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html#attr-button-type). This causes no problems at most places, but will cause problem within a form (e.g. the user account settings) where toggling an option is now identical to clicking submit.

This patch fixes the issue by setting the `type` attribute to `button`. Not only for the toggle switch, but also for a few other elements which have been recently converted to buttons.


advplyr:
Thanks, I overlooked that also.

PR #1397, opened 2023-01-12
Add total duration info on the podcasts page.

mfcar:
For now, the total duration info is not shown for podcasts, just for audio files or audio tracks:

image

This pull request adds this information on the podcast page:

image

My goal is to have a piece of information about how long is the duration of the podcast, similar to what iTunes does:

image

### DIFF
--- a/client/pages/item/_id/index.vue
+++ b/client/pages/item/_id/index.vue
@@ -113,6 +113,14 @@
                   {{ durationPretty }}
                 </div>
               </div>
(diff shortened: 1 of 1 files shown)


advplyr:
For this you should use the existing `durationPretty` computed property and use the same HTML

advplyr:
I implemented this here https://github.com/advplyr/audiobookshelf/commit/024d052a7b614a5066a18af4637561fbc48cb17d

PR #1398, opened 2023-01-12
Don't list book twice in continue series

lkiesow:
Sometimes, a book belongs to more than one series. If you listen to and finish such a book, Audiobookshelf will list the next book in “Continue Series” twice, right next to each other. That is not helpful.

This patch fixes the problem by not adding books to the list if they are already in the list.


advplyr:
Thanks!

PR #1404, opened 2023-01-14
update it

burghy86:
and add a month and day a abbreviated for annual table


advplyr:
Thanks! These are kept in alphabetical order even when it doesn't make as much since with the weekdays/months.

PR #1420, opened 2023-01-21
Fix HTTP links not working in podcast show notes.

### LINKED ISSUE #1418: [Bug]: HTTP links broken in show notes
### Describe the issue

Links in show notes don't work if the linked site uses HTTP, but they work fine for HTTPS. I've compared the original RSS feed of a podcast to the one generated by audiobookshelf, and it seems that the `href="..."` part in the `<a>` tag is removed if the address starts with http. The links aren't clickable in the audiobookshelf web UI, so the problem must be occurring earlier than RSS feed generation.

1. Subscribe to a podcast (e.g. No Such Thing as a Fish, https://audioboom.com/channels/2399216.rss)
2. Click link in the show notes. If it's an HTTPS link, it will work, if it's HTTP, it won't. As an example, episode 292 of No Such Thing as a Fish contains both (the nosuchthingasafish.com link is HTTP, the amazon and apple links are HTTPS).
3. Generate an audiobookshelf RSS feed and check the resulting file to see that the <a> tags are altered for HTTP links, compared to the original RSS.

### Audiobookshelf version

v2.2.12

Docker

Bostrolicious:
Fixes #1418. HTTP links in show notes didn't work because `sanitize` in `server/utils/htmlSanitizer.js` removes them, as only `https` was set in `allowedSchemes`. This commit adds `http`.

`htmlSanitizer/sanitize` is only used for podcast and episode descriptions (in `server/utils/podcastUtils.js` and `server/providers/iTunes.js`) , so I don't think there's any reason to disallow HTTP.

### DIFF
--- a/server/utils/htmlSanitizer.js
+++ b/server/utils/htmlSanitizer.js
@@ -10,7 +10,7 @@ function sanitize(html) {
     allowedAttributes: {
       a: ['href', 'name', 'target']
     },
-    allowedSchemes: ['http', 'https'],
+    allowedSchemes: ['http', 'https', 'mailto'],
     allowProtocolRelative: false
   }
 
--- a/server/utils/htmlSanitizer.js
+++ b/server/utils/htmlSanitizer.js
@@ -10,7 +10,7 @@ function sanitize(html) {
     allowedAttributes: {
       a: ['href', 'name', 'target']
     },
-    allowedSchemes: ['https'],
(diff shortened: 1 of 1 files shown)


Bostrolicious:
I realised `mailto` links weren't allowed either, so I added them to `allowedSchemes`  as well, since they're often used in podcast show notes.

advplyr:
I think this should be fine. Thanks!

PR #1421, opened 2023-01-21
Reduce Fingerprinting

lkiesow:
As DieselTech#6997 pointed out in Matrix, it is a good idea to reduce fingerprinting by removing the `X-Powered-By` response header as pointed out by the Express security best practices:

See http://expressjs.com/en/advanced/best-practice-security.html#reduce-fingerprinting


advplyr:
Thanks!

PR #1423, opened 2023-01-21
Ensure default are similar in different environments

lkiesow:
This patch tries to make sure that defaults and options work in the same way regardless of users using Docker or a package deployment.

- Commit 10295b0 updated `Host` to be empty by default to better support IPv6. This fixes a missed occurrence of the old default.
- This makes sure the `SOURCE` environment variable is honored when installing the packages. The `--source` option will still take precedence though.


advplyr:
Thanks!

PR #1425, opened 2023-01-22
Skip AUDIOBOOKSHELF_UID/GID if undefined

lkiesow:
This patch slightly changes the behavior of the `AUDIOBOOKSHELF_UID` and `AUDIOBOOKSHELF_GID` options. Instead of defining a default user and group, trying to modify files and silently failing if the filesystem mode cannot be changed, this patch will just skip the entire process in the first place.

If these options are defined, Audiobookshelf should behave exactly as before. If they are not defined, Audiobookshelf will now cause fewer file modifications (or less failures when trying to modify files).

If this patch gets applied, it should probably be highlighted in the release notes. This usually shouldn't cause problems for migrations since the Docker guides explicitly configure the options and the package installations do not seem to use this at all, but there is still a change that it will and users should be aware of that.

If a problem arises, users can easily fix the problem by either setting the permissions once manually to the audiobookshelf user or by simply defining the `AUDIOBOOKSHELF_UID/GID` options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,8 +18,8 @@ const PORT = process.env.PORT || 80
 const HOST = process.env.HOST
 const CONFIG_PATH = process.env.CONFIG_PATH || '/config'
 const METADATA_PATH = process.env.METADATA_PATH || '/metadata'
-const UID = process.env.AUDIOBOOKSHELF_UID || 99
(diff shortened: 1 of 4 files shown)


advplyr:
Great, thanks. Will highlight in the release notes.

PR #1438, opened 2023-01-24
Localized stats page

tomazed:
I modified the page a bit to accept the translated strings for abbreviated Month and Days

I have not changed the non translated elapsed string in the Recent Session Pane because it is an output of lib date-fns formatDistance function and I need to dig a bit into it to see how to do local with date-fns


tomazed:
I've seen my vim keeps adding new line at the end of file. Sorry.
I've disabled this.

advplyr:
Great, thanks! I took a look at localization for date-fns also and didn't see a quick way to implement it so haven't prioritized that.

PR #1453, opened 2023-01-28
One Default Bookshelf View

lkiesow:
This patch updates the default value of the home page bookshelf view so that it is identical to the default library view. Having different styles by default seems odd. I picked the non-wooden view as default based on the recent discussion on Matrx/Discord that that one looks more modern.


advplyr:
This works for me. I also updated the mobile app to use the standard view by default.

PR #1456, opened 2023-01-29
Fixes m4b chapters only taken from first file.

jramer:
Noticed that I was only getting chapters from the first `m4b` file and this fixes that.
Only tested with `m4b`. Don't know if other file formats support chapters.

Code uses `let` and optional chaining. Hope that's ok.


advplyr:
There are some cases where the chapters are embedded in every audio file the same. This is assuming that each audio file has its own separate chapters, which is what makes the most sense but is not guaranteed.

jramer:
> There are some cases where the chapters are embedded in every audio file the same. This is assuming that each audio file has its own separate chapters, which is what makes the most sense but is not guaranteed.

Interesting. 
Ok but from my point of view that case is giving wrong input and should not really be considered. I mean my files have correct chapters but only get them from the first file.
 If you provide multiple files with the same chapters. It's the chapters that are wrong. Not the parsing of the chapters.

Or do you mean that those chapters actually span multiple files already? If so I guess I could check that the last chapter is contained within the file duration.

And if not. I guess the check would be to compare the last chapters start/end to the previous file.

jramer:
@advplyr Added a check that keeps original implementation when the first and second audio files chapters titles are the same.
Hope that's enough for the special case.

server/objects/mediaTypes/Book.js
```diff
@@ -420,10 +420,38 @@ class Book {
...
+        let currStartTime = 0
+
+        includedAudioFiles.forEach((file) => {
+          if (file.duration) {
+            let chapters = file.chapters?.map((c) => ({
+              ...c,
+              id: c.id + currChapterId,
+              start: c.start + currStartTime,
+              end: c.end + currStartTime,
+            }))
```
advplyr, line 447:
```suggestion
            })) ?? []
```

server/objects/mediaTypes/Book.js
```diff
@@ -420,10 +420,38 @@ class Book {
...
+        this.chapters = includedAudioFiles[0].chapters.map((c) => ({ ...c }))
+      } else {
+        Logger.debug(`[Book] setChapters: Using embedded chapters from all audio files ${includedAudioFiles[0].metadata?.path}`)
+        this.chapters = []
+        let currChapterId = 0
+        let currStartTime = 0
+
+        includedAudioFiles.forEach((file) => {
+          if (file.duration) {
+            let chapters = file.chapters?.map((c) => ({
```
advplyr, line 442:
```suggestion
            const chapters = file.chapters?.map((c) => ({
```

advplyr:
Thanks, this should do it. The chapters are only parsed again if any of the audio files are updated. This can be improved but we want to make sure to not overwrite changes someone made in the chapter editor that haven't yet been embedded.

advplyr:
This is in `MediaFileScanner.js` line 278

PR #1463, opened 2023-02-01
Update fr.json

Machou:
fr_FR update
fr_FR reworked
fr_FR cleaned


tomazed:
thanks for the cleanup

advplyr:
Thanks!

PR #1468, opened 2023-02-01
Integration Test

lkiesow:
This patch adds a minimal integration test building Audiobookshelf as a binary, running it and checking if the server is available on each push and pull request.

We can easily extend this with a Selenium or Playwright test later, but it should already alert us about problems in the build pipeline without the need for any developer to take a look at the new patches.


advplyr:
Thanks!

PR #1469, opened 2023-02-02
Adding tags to metadata.abs

yuuzhan:
Modified the reading / saving of metadata.abs to allow for the reading and writing of tags.  Used JSON functions to stringify the objects to handle escape characters automatically, then JSON.parse to get it back in.

Tried to make the read / write as error proof as possible so it would function with files that do not have the tags entity in the metadata file.

### DIFF
--- a/server/objects/mediaTypes/Book.js
+++ b/server/objects/mediaTypes/Book.js
@@ -255,9 +255,16 @@ class Book {
       const abmetadataUpdates = abmetadataGenerator.parseAndCheckForUpdates(metadataText, this, 'book')
       if (abmetadataUpdates && Object.keys(abmetadataUpdates).length) {
         Logger.debug(`[Book] "${this.metadata.title}" changes found in metadata.abs file`, abmetadataUpdates)
-        metadataUpdatePayload = {
-          ...metadataUpdatePayload,
-          ...abmetadataUpdates
+
+        if (abmetadataUpdates.tags) { // Set media tags if updated
+          this.tags = abmetadataUpdates.tags
+          tagsUpdated = true
+        }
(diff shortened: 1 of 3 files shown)


Hallo951:
Oh great verry thanks, I've been waiting for this for a while. How can I update older abs files so that the tags are also saved in this one?

yuuzhan:
> Oh great verry thanks, I've been waiting for this for a while. How can I update older abs files so that the tags are also saved in this one?

At this time you have to manually trigger an update on each metadata file.  This should happen automatically when you modify a field in the library entry.  Depending on your library size, I might suggest doing a bulk edit and appending a temporary tag to all entries then saving them all.  Ive done this kind of operation with 300-400 items and it takes quite a few minutes to complete (you might have to click "Wait" when your browser prompts you to kill the tab as it thinks its stuck)

advplyr:
I hesitated on adding tags to this file because the whole abmetadata file concept needs to be re-written. 

As it stands now `tags` will be the only key in there that is using JSON, which is weird. JSON is preferred because we ran into issues using genres with commas because the current abmetadata file expects comma separated genres.

That being said I just cleaned this up for release since it is being asked for. I made updates to your original logic

yuuzhan:
I was sensing that was what you were planning however I wanted to do something and I needed tags in the file for now.

I debated refactoring some of the other code to make it JSON and not CSV but thought better of doing too much at once :p

I had also written some code to dump out LibraryItem.Book.ToJSON() into metadata.json.  that would be simple to pull those both in and out in total without having as much logic around reading and writing, but not sure about the efficiency of the application overall doing that

PR #1484, opened 2023-02-06
Reduce memory usage when scanning large folders

magnww:
When audiobookshelf scans a folder with a lot of files, it will take up a lot of memory, because audiobookshelf will start a ffprobe process for each media file at the same time. In order to reduce memory usage, it is necessary to limit the number of ffprobe processes started at the same time.
In addition, after the ffprobe process is executed, it does not exit immediately and becomes a zombie process. It is necessary to call kill() in the exit event to end the process.

Before:
图片

After:
图片

### DIFF
--- a/server/libs/nodeFfprobe/index.js
+++ b/server/libs/nodeFfprobe/index.js
@@ -18,10 +18,7 @@ module.exports = (function () {
       proc.stdout.on('data', function (data) { probeData.push(data) })
       proc.stderr.on('data', function (data) { errData.push(data) })
 
-      proc.on('exit', code => {
-        exitCode = code
-        proc.kill()
-      })
+      proc.on('exit', code => { exitCode = code })
       proc.on('error', err => reject(err))
       proc.on('close', () => {
         try {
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks! I haven't done any testing on the zombie processes to confirm but I'm wondering if calling `kill()` in the exit event could cause some issues if the process exited.

In node docs https://nodejs.org/api/child_process.html#subprocesskillsignal

> The [ChildProcess](https://nodejs.org/api/child_process.html#class-childprocess) object may emit an ['error'](https://nodejs.org/api/child_process.html#event-error) event if the signal cannot be delivered. Sending a signal to a child process that has already exited is not an error but may have unforeseen consequences.

magnww:
After some investigation I found a docker configuration issue. The zombie process problem can be solved by adding `--init` parameter to docker run command. So I will revert the change of ffprobe.
[https://docs.docker.com/engine/reference/run/#specify-an-init-process](url)

advplyr:
Thanks!

whmountains:
Hmmm, is it really a good idea to create zombie processes?  Having the init system reap them seems like it should be a last resort.  I am concerned that few will realize they should be starting their containers with `--init`.  I vaguely knew that was an option, but I had not seen any examples use it and so was not in the habit.

PR #1505, opened 2023-02-14
Add library tags variable to podcast notifications

p-rintz:
This adds the `libraryTags` variable to the `onPodcastEpisodeDownloaded` notification handler.

This helps in the usecase where multiple users have access only to limited tags.
In this case they can filter the sent out notifications only for the tags that they are interested in and are no longer pinged for every podcast downloaded.

Example:

image

image

### DIFF
--- a/server/managers/NotificationManager.js
+++ b/server/managers/NotificationManager.js
@@ -24,7 +24,7 @@ class NotificationManager {
       libraryItemId: libraryItem.id,
       libraryId: libraryItem.libraryId,
       libraryName: library ? library.name : 'Unknown',
-      libraryTags: libraryItem.tags,
+      libraryTags: libraryItem.media.tags,
       podcastTitle: libraryItem.media.metadata.title,
       episodeId: episode.id,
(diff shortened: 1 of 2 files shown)


advplyr:
Were you able to put through the test email for this?

p-rintz:
I am guessing you mean a notification in general?

I was under the impression that it was only using the Apprise notification system in this case.

The test notification was the second screenshot. 

I did find out that there was an error still anyway, that is now fixed.

I tested it with an actual podcast download now as well:

grafik


Though I am honestly wondering wether it doesnt make sense to add a whole lot of variables that people might use in notifications. (ie. Genre, subtitle, description, length, chapters)

advplyr:
Yeah I was just meaning the Test button that triggers the notification with the test data defined in `notifications.js`

image


I went ahead and added a bunch more variables there. I also updated it so that we are only storing strings in `eventData` objects for now and changed the name to `mediaTags` because that is more accurate.

PR #1517, opened 2023-02-19
Add episodes search

mfcar:
- Add an episode search, filtering the episodes list by title or description based on the search input
 - Added to the Episode List and Episode Feed
 - Add translation strings

**Episode List**
Screenshot 2023-02-19 at 20 45 39

**Episode Feed**
Screenshot 2023-02-19 at 20 45 57

**Demo**

https://user-images.githubusercontent.com/814828/219974584-548559ca-2a12-420e-b72b-4c5ae84d02d4.mov


mfcar:
I have added translation strings and changed the input type to search (some browsers added a clean button when using this input type).

Screenshot 2023-02-21 at 19 40 08

advplyr:
We may be refactoring this so that the episodes show in the main list instead of having to click the search button to lookup.

advplyr:
Since the DB migration is happening it will be a while before podcast episode lists are refactored. In the meantime I think this will be useful. Thanks!

PR #1518, opened 2023-02-19
Add largest item stats

mfcar:
- Add a new stats item listing the top 10 largest podcasts/books
 - Affects the podcast and book stats page
 - Like the longest items stats that use the total duration for an item (summing all the duration of all media in a podcast/book), on the largest items, I am summing all the sizes of all media within a podcast.

Screenshot 2023-02-19 at 21 40 31

### DIFF
--- a/client/pages/config/library-stats.vue
+++ b/client/pages/config/library-stats.vue
@@ -60,6 +60,25 @@
             </div>
           </template>
         </div>
+        <div class="w-80 my-6 mx-auto">
+          <h1 class="text-2xl mb-4">{{ $strings.HeaderStatsLargestItems }}</h1>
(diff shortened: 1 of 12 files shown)


hobesman:
Is this specific files, or is it entire podcasts? Does it also affect the books stats page?

mfcar:
> Is this specific files, or is it entire podcasts? Does it also affect the books stats page?

Like the longest items stats that use the total duration for an item (summing all the duration of all media in a podcast/book), on the largest items, I am summing all the sizes of all media within a podcast.
If an item (podcast or book) has 10 files. I sum up all the sizes of the 10 files, and I will have a total for each podcast/book.

This change also adds stats for the books because both use the same component.

advplyr:
Excellent, thanks!

I also realized when looking at this that the Podcast library stats is still showing the Authors sections.

Issue #1522, opened 2023-02-20
[Enhancement]: Include iTunes and google play block tags to keep RSS feeds private


Weldawadyathink:
### Describe the feature/enhancement

http://support.megaphone.fm/en/articles/2700280-itunes-and-google-block-tags

iTunes and google play both support block tags to prevent podcast feeds from being publicly indexed and available. Since audio bookshelf servers can be expected to host copyrighted content, these feeds should not be made available in podcast directories. If we add these tags, RSS feeds can still be manually added to podcast apps, but will not be listed publicly. 

This seems like a pretty easy fix to me, so I am going to have a go at this when I get some free time.

apiweb:
I think this needs to have an option on the Settings page.
Or a setting per podcast (to be chosen when we open/close a podcast feed).
This block can be a default choice, but we need to have the option to disable the blocking because some people can use ABS to host their legal podcasts.

Weldawadyathink:
Do you know anyone who hosts podcasts with ABS? It just seems like a really poor tool for the job. There are a ton of really good podcast hosting software designed for public feeds. ABS has some weird hangups for a public podcast. For example itunes:email is hardcoded to an empty string. And ABS doesn't host a public per-episode webpage or an overall webpage. I don't know if Apple would even accept an ABS generated RSS feed because of these issues. 

It is ultimately up to @advplyr, but my impression is that public rss feeds are out of the scope of this project.

advplyr:
There is a feature request open for hosting your podcast through Abs. https://github.com/advplyr/audiobookshelf/issues/1446

I'm considering this because we already built the foundation so adding additional options to control the XML might be reasonable. I'm not sure how the UI looks for that.

advplyr:
Added in [v2.2.16](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.16)

PR #1532, opened 2023-02-22
Added block tags based on Patreon private rss feed

Weldawadyathink:
Xml formatting is based on how Patreon sets up their private rss feeds. Patreon is pretty protective of their stuff, so this should cover all bases and keep personal books off of iTunes and google podcasts. 

#1522


advplyr:
Implemented in https://github.com/advplyr/audiobookshelf/pull/1554 with configurability

PR #1554, opened 2023-02-25
Add rss feed configuration

mfcar:
This pull request adds an option to prevent indexing of your feed (turned on by default) and some new fields used to set up RSS  feed metadata details like Owner Name and Owner Email (used by Spotify to validate your ownership).
Works for podcasts and audiobooks.
Fix: #1522

-----

Screenshot 2023-02-25 at 13 19 15

### Advanced details tab
Screenshot 2023-02-25 at 13 19 25

### Viewing details
Screenshot 2023-02-25 at 13 19 01

### Details if you do not fill owner name and email input fields
Screenshot 2023-02-25 at 13 25 42

### XML
Screenshot 2023-02-25 at 13 20 15

PR #1556, opened 2023-02-25
Fix incorrect tags when blocking public feeds

Weldawadyathink:
#1554 was way better than what I could do, had 1 minor issue. iTunes:block should be "Yes". Any other result is ignored. [Documentation here](https://help.apple.com/itc/podcasts_connect/#/itcb54353390). Although I cannot find the documentation for google:block, pattern podcast feeds suggest it should be lowercase "yes". 

image


advplyr:
Thanks!

Issue #1331, opened 2022-12-28
[Enhancement]: View Podcast download Queue


aurokon:
### Describe the feature/enhancement

Would be pretty nifty to be able to see what's queued up to be downloaded for our podcasts

advplyr:
What do you mean? Can you share more details and how you think it should designed?

aurokon:
Just a window with a list of what's waiting to be downloaded with some buttons to stop/cancel/pause/start the downloads. Nothing special.

Issue #764, opened 2022-06-21
[Enhancement]: Improved user feedback for podcast download functionality


RandomWare:
### Describe the feature/enhancement

The only feedback offered to users on podcast downloads appears to be a couple of info balloons on the individual podcast's page. There is no indication of progress other than a spinner and no clear indication of a failure and the cause of that failure. This extends to the logs even in DEBUG mode. 

I recommend implementing a separate section for managing podcast downloads with the following functionality:

- list which individual podcasts have downloads queued
- list what files are queued for each podcast
- display the status of each queued item (queued, downloading, failed, success) and controls for deleting items, clearing completed downloads, viewing logs relevant to failures, potentially changing priority/order
- have a separate area highlighting the current status of the current download with information on status and speed, and controls for pausing, stopping, putting back in queue to allow the next download, potentially changing priority.

PR #1559, opened 2023-02-27
Add download queue page

mfcar:
This pull request supports a universal download queue (per podcast library).
You can see the current downloading episode and the next on the queue on the download queue page.
The queue table has four columns:
 - Podcast name
 - [If available] Season, Episode number and type indicator (just for trailer or bonus)
 - Episode title
 - Publication date

In the future, I want to add a "clear queue" button and "remove episode" from the queue button.
I would like to add a badge with the queue length on the queue menu button. However, this needs a massive investigation with sockets to update the number in real-time for each podcast library (I will investigate because I need to learn more about sockets).

image

----------------

And I also add an indicator for the podcast types Bonus and Trailer. This is to make it easier to identify special episodes (bonus or trailer. The default type is full).

### Trailer
image

### Bonus
image

------------
Fix: #1331
Partial fix: #764

----
### Download queue
Screenshot 2023-02-27 at 02 30 21

### Bonus-type episode
Screenshot 2023-02-27 at 02 32 48

### Trailer-type episode
Screenshot 2023-02-27 at 02 32 58

### Type indicator in the latest episodes page
Screenshot 2023-02-27 at 02 36 10


advplyr:
There is a `Task` model that I started but haven't fully built out yet. I think it can be useful here.

The plan was to show a notification in the appbar when "background" tasks are running. This would include scans instead of showing the big toast when a scan is happening.

image

Currently the `NotificationWidget.vue` is just this spinner in the Appbar, but I thought it could be more robust with a dropdown menu showing the tasks.

The `TaskManager.js` would be in charge of storing the active tasks and sending socket events with `Task.js` objects to the clients.
The `Task.js` is very broad right now to try to encompass any event.

Right now it is only being used for the m4b merge tool in `AbMergeManager.js`, you can see the Task created on line 49.

I don't exactly know how this would work with episode downloads since there can be many. I just wanted to dump this here since I'm out of time today.

mfcar:
Good, Kavita has a dropdown menu on the tasks menu button. It is handy because I can see all the background tasks simultaneously.

Screenshot 2023-02-28 at 08 17 39

On the Synology DSM, for more complex task notifications, they have a link to get more details.

Screenshot 2023-02-28 at 08 36 30

I think a similar way can be used for episode downloads, for example: 
Showing the current download as a notification (using the Kavita notification layout)

> Line 1 (Notification title/category): Current downloading
> Line 2: {{Episode Number}} {{Episode Title}}
> Line 3: {{Queue lenght}} downloads remaining

And this notification can have a link to the download queue for more details (to reorder the queue, cancel some items, etc.).

Screenshot 2023-02-28 at 08 17 39 copy


----

Thanks for your tip. I found the `Task` object on the `AbMergeManager.js` file. It's a very good starting point. I can start my investigation about the background tasks and notifications.

mfcar:
Hi @advplyr, I have updated my pull request to include a dropdown showing the currently running tasks. The new update also includes the following changes:

 - The dropdown now shows tasks running or failing (failed at the last minute).
 - Each task now has a different icon to represent its action type.
 - A link to the task is now provided. However, only tasks that involve downloading episode links or encoding M4B links will have this link available. Generic items will not have a link.

In the future, I plan to create a task page that displays all the tasks with their details. Additionally, I would like to add more task actions to other fields, such as Backup Running, Folder Scanning, Rss Checking, etc. However, I would like this pull request to focus solely on the download queue.

Screenshot 2023-03-05 at 11 16 40

advplyr:
I made some updates to the UI so that the current download is left aligned with the table. The table is shorter though so titles will wrap.

image

I updated the library item page and download queue page to show the current download on initial page load. Previously you would only see it if the event came in while on that page.

For future reference, Nuxt handles importing components for us so as long as you create the component in the components folder then you can use it. That is also why when using the components you have to put the folder names in there `tables-podcast-download-queue-table`. [ref](https://github.com/nuxt/components)

Also, early on I wasn't taking advantage of [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) because of old habits before these were available in node.
e.g.
`this.podcastEpisode ? this.podcastEpisode.episode : null` becomes `this.podcastEpisode?.episode ?? null`

Thanks for implementing this. The tasks dropdown UI is great.

advplyr:
Oh also, I set it to hide the download queue table if there is no queue. I thought it looked cleaner.

Issue #1491, opened 2023-02-08
[Enhancement]: Show 'Next backup in [duration]' on the backup dialog


raviwarrier:
### Describe the feature/enhancement

For those who use cron to schedule their backup, it might take a little mental maths to figure out when the next backup is scheduled to take place. It would be nice if there was a label somewhere on the dialog that calculated the cron schedule day/date and displayed it as 'Next backup in [duration in days]'.

Example (note: mine is every two days and is easy, but someone might have a really complicated cron schedule set up).
![image](https://user-images.githubusercontent.com/42061703/217508313-79ab7571-f74f-4f48-a096-2615f73ec962.png)

advplyr:
Added in [v2.2.16](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.16)

PR #1562, opened 2023-02-27
Improve dates, times and schedule backup info

mfcar:
This pull request improves the dates throughout the entire system:
 - Add more Date Formats options
 - Add a Time Format
 - Add an example for the date and time formats selected
 - Almost all the dates on the ABS try to respect the date and time formats selected
 - The backup page shows the next scheduled run based on the cronjob (the date is shown respecting the date format selected) - Fix: #1491
 - The schedule podcast modal received an improvement to show the next scheduled run based on the cron expression inputted (if the expression is validated)
 - Using [Cron Parse](https://github.com/harrisiirak/cron-parser) to parse the cron expressions. Cron parse is a tiny and well-documented library and receives updates with frequency.

------
### New date formats with the example
Screenshot 2023-02-27 at 17 44 27

### New time format with the example
Screenshot 2023-02-27 at 17 44 33

### Dates on the ABS respecting the date and time format selected
Screenshot 2023-02-27 at 17 52 21

### Improvement on the backup screen showing the next scheduled run and dates respecting the date format
Screenshot 2023-02-27 at 17 50 47

### Improvements on the schedule modal showing the next scheduled run
Screenshot 2023-02-27 at 17 51 49

Screenshot 2023-02-27 at 17 51 59


client/pages/config/index.vue
```diff
@@ -293,7 +299,18 @@ export default {
     },
     dateFormats() {
       return this.$store.state.globals.dateFormats
-    }
+    },
+    timeFormats() {
+      return this.$store.state.globals.timeFormats
+    },
+    dateExample() {
+      const date = new Date(2014, 2, 25);
```
advplyr, line 307: Any reason to not use the current date for these? Just to not have confusion?

client/plugins/utils.js
```diff
@@ -144,4 +185,26 @@ export function supplant(str, subs) {
       return typeof r === 'string' || typeof r === 'number' ? r : a
     }
   )
-}
\ No newline at end of file
+}
+
+function getNextValue(cronField, currentValue, maxValue) {
+  if (cronField === '*') {
+    return currentValue + 1 <= maxValue ? currentValue + 1 : 0;
```
advplyr, line 192: There is an incorrect next date because of this increment.

image

The correct next date is March 6th:

image

I'm not sure a quick fix because for hour we only want to increment if not incrementing would cause the date to be in the past.

client/plugins/utils.js
```diff
@@ -136,6 +136,47 @@ Vue.prototype.$parseCronExpression = (expression) => {
...
+  const currentMonth = currentDate.getMonth() + 1;
+  const nextMonth = getNextValue(month, currentMonth, 12);
+  nextDate.setMonth(nextMonth - 1);
+
+  // Calculate next day of week
+  const dayOfWeek = cronFields[4];
+  const currentDayOfWeek = currentDate.getDay();
+  const nextDayOfWeek = getNextValue(dayOfWeek, currentDayOfWeek, 6);
+  const daysUntilNextDayOfWeek = getNextDaysUntilDayOfWeek(currentDate, nextDayOfWeek);
+  nextDate.setDate(currentDate.getDate() + daysUntilNextDayOfWeek);
```
advplyr, line 174: This causes issues with `0 12 1 * *`, which is running on the 1st of every month.

The correct result is Match 1st but we are getting March 28th.

client/pages/config/index.vue
mfcar: Yes, I choose this specific date and time (to be honest, I'm using the same date example from Sonarr) because it can avoid confusion. For example, day two and month two can be represented 02-02. However, this needs to be clarified.
Choosing a late month day helps to avoid this.

client/plugins/utils.js
mfcar: Thanks for your guidance. At first glance, I thought to install an external library like [cron-parser](https://github.com/harrisiirak/cron-parser), but I wanted to avoid adding new libraries. 
I will check the code and validate the results against more cron job expression validators. Thanks.

mfcar:
I investigate other systems that offer to change the date and time format, like iOS, WordPress, XenForo, vBulletin, Radarr, Sonarr, and others the result is variable. 

 - Some, like WordPress, show the current data as an example label. But this can be confusing. For example, day two and month two can be represented 02-02. 
 - Others like vBulletin and XenForo don't show an example with the format chosen.
 - iOS, Radarr, and Sonarr show an example date with a default value to avoid confusion. iOS has a minor change because they use the current year.

I added an example date (with a default value) and the current date to make it easier for the user.

Prototype 1
Screenshot 2023-03-04 at 11 03 11

Prototype 2
Screenshot 2023-03-04 at 11 04 01

Final result
Screenshot 2023-03-04 at 12 01 56

advplyr:
I thought the UI was looking a bit cluttered with both lines and the text color of example. I think we can keep it simple
image

advplyr:
cron-parser is working great. The only downside with this is now the client has 2 date libraries, date-fns and luxon.

I think reducing the number of client libraries can be a problem for another time.

Thanks!

PR #1581, opened 2023-03-04
Improve podcast editing

mfcar:
This pull request adds a new action to save podcast editing without closing the modal.
 - Like the Library Item editing, we have two buttons, Save and Save & Close actions.
   - `Save` button updates the episode and keeps the modal opened
   - `Save & Close` updates the episode details and closes the modal
 - I also added chevron buttons to navigate to the next or previous episode to make editing many episodes in sequence easier.

----
Screenshot 2023-03-04 at 19 05 32


advplyr:
Great, thanks!

PR #1586, opened 2023-03-07
fr_FR update

Machou:


client/strings/fr.json
```diff
@@ -39,8 +39,8 @@
   "ButtonMatchBooks": "Rechercher les Livres",
   "ButtonNevermind": "Oubliez cela",
```
Machou, line 40: can i get context please ?

advplyr: Restoring backups

image

Machou:
Is good. You can valide:)

advplyr:
Thanks~!

Issue #793, opened 2022-06-29
[Enhancement]: Support multiple author/narrator separated by semicolon from audio metadata


DarkShortie:
### Describe the feature/enhancement

Please add support for multiple author and narrator from file metadata like it's already supported for genre.

advplyr:
I'm not sure what you mean. This should already be supported. Can you be more specific?

hobesman:
> I tried to add multiple author and narrator by setting artist to Author 1;Author 2;Author 3 and narrator to Narrator 1;Narrator 2;Narrator 3.
> 
> I wanted that those where added as separate entries like it's done for multiple genres.
> 
> This is how it looks like if it's filled by the metadata. ![image](https://user-images.githubusercontent.com/54549395/177597873-a094abfd-3e38-463a-a3c9-b5f9dbe5557f.png)
> 
> This is after editing it manually in Audiobookshelf and like it would be great if it will be imported like this. ![image](https://user-images.githubusercontent.com/54549395/177598160-1ab13f41-8ae8-4c0b-a69e-07e232009bbf.png)
> 
> I've tested this with m4b and the attributes aART for author and ©wrt for narrator.

I believe if the authors are comma separated it interprets them as separate authors. 

This works for me but may conflict for people who list authors as "Last, First" so it might be worth revisiting how this should work.

advplyr:
Multiple authors works when separating by comma. This also works when you are using Last, First, Last, First.

I'm not sure about using semicolon to separate authors, I haven't seen anyone use that before. I'll leave this open to see if anyone else does this

szemlicka:
Semicolon is a typical delimiter for mp3 files for multibles fiels like Author, genre or narrator. Squeezebox use this as default. Please add this or make it as a option which you can choose. As an example, in squeezebox is this an option (default ;). You can use different delimiter seperated with a space like ", ; /" if you want this 3 charactes as delimiter

calebj:
Importing m4b files from AAX Audio Converter had semicolons in the Artist field seperating the author, foreword author (where present) and narrator when I did that today.

scop22:
Semi-colon delimiters are pretty much standard. The other common delimiter is a slash. 

Commas are tricky, because of the use of commas in names like "Joe Smith, M.D." I already had that problem, where a name like this was interpreted as two different authors. There are other types of names that have commas in them too.

I think allowing ";" as a delimiter would be a very welcome enhancement.

[...]

PR #1609, opened 2023-03-16
Support for scanning semicolon seperated author and narator lists.

Linden-Ryuujin:
#793

Duplicate "&" style multi author/narrator support for semicolon. I did consider a loop for handling separators, but it seems like overkill, however happy to be persuaded.


advplyr:
Nice and easy, thanks!

Issue #838, opened 2022-07-14
[Bug]: Encoding for stream fails when playing m4b on iOS


0xcd0e:
### Describe the issue

When continuing an audiobook on safari on iOS, the stream starts playing all the way at the beginning and the elapsed time in the player is frozen.
The Logs show some warnings:

```
2022-07-14 18:58:33 INFO [Server] Socket Connected IUK4SgpYEaQLBWuiAAAF

2022-07-14 18:58:49 INFO [PlaybackSessionManager] startSession: Closing open session "Audiobook 123" for user "username"

2022-07-14 18:58:49 INFO [STREAM] START STREAM - Num Segments: 6479

2022-07-14 18:58:49 INFO [STREAM] Starting Stream at startTime 5:27:15.0 (User startTime 5:27:45.0) and Segment #3272

2022-07-14 18:58:49 INFO [INFO] FFMPEG transcoding started with command: ffmpeg -seek_timestamp 1 -f concat -safe 0 -ss 19635.4s -noaccurate_seek -i /metadata/streams/play_tei3ov6i7l60nhv83i/files.txt -y -loglevel error -map 0:a -c:a copy -f hls -copyts -avoid_negative_ts make_non_negative -max_delay 5000000 -max_muxing_queue_size 2048 -hls_time 6 -hls_segment_type mpegts -start_number 3272 -hls_playlist_type vod -hls_list_size 0 -hls_allow_cache 0 -hls_segment_filename /metadata/streams/play_tei3ov6i7l60nhv83i/output-%d.ts /metadata/streams/play_tei3ov6i7l60nhv83i/final-output.m3u8

2022-07-14 18:58:50 WARN File path does not exist /metadata/streams/play_tei3ov6i7l60nhv83i/output-0.ts

2022-07-14 18:58:50 WARN [STREAM] Segment #0 Request @0:00.0 is before start time (5:27:45.0) - Reset Transcode

2022-07-14 18:58:50 INFO [FFMPEG] Transcode Killed

2022-07-14 18:58:51 INFO Stream Reset New Start Time 0:00.0

2022-07-14 18:58:51 INFO [STREAM] START STREAM - Num Segments: 6479

[...]

PR #1625, opened 2023-03-21
Fix: Resseting time is not taking account codec and mime type resulting in isResetting loop

fidoriel:
The long resetting time is not triggered if transcode is triggered because mimetype or codec. This resulted in a resetting loop, always starting a new transcode.

```
audiobookshelf  | [2023-03-21 06:39:12] INFO: 
audiobookshelf  | [2023-03-21 06:39:13] INFO: [STREAM] Clearing isResetting
audiobookshelf  | [2023-03-21 06:39:13] WARN: File path does not exist /metadata/streams/play_ege9befho8m2xsrhrt/output-289.ts (HlsRouter.js:37)
audiobookshelf  | [2023-03-21 06:39:13] INFO: Segment #289 requested is 287 segments from latest (28:54.0) - Reset Transcode
audiobookshelf  | [2023-03-21 06:39:13] INFO: [FFMPEG] Transcode Killed
audiobookshelf  | [2023-03-21 06:39:14] INFO: Stream Reset New Start Time 28:42.0
audiobookshelf  | [2023-03-21 06:39:14] INFO: [STREAM] START STREAM - Num Segments: 793
audiobookshelf  | [2023-03-21 06:39:14] INFO: [HlsRouter] Resetting Stream - notify client @1734s
audiobookshelf  | [2023-03-21 06:39:14] INFO: [STREAM] Starting Stream at startTime 28:12.0 (User startTime 28:42.0) and Segment #282
audiobookshelf  | [2023-03-21 06:39:14] DEBUG: [Stream] Forcing AAC for tracks with mime type audio/mp4 and codec alac (Stream.js:279)
audiobookshelf  | [2023-03-21 06:39:14] INFO: [INFO] FFMPEG transcoding started with command: ffmpeg -seek_timestamp 1 -f concat -safe 0 -ss 203s -noaccurate_seek -i /metadata/streams/play_ege9befho8m2xsrhrt/files.txt -y -loglevel error -map 0:a -c:a aac -f hls -copyts -avoid_negative_ts make_non_negative -max_delay 5000000 -max_muxing_queue_size 2048 -hls_time 6 -hls_segment_type mpegts -start_number 282 -hls_playlist_type vod -hls_list_size 0 -hls_allow_cache 0 -hls_segment_filename /metadata/streams/play_ege9befho8m2xsrhrt/output-%d.ts /metadata/streams/play_ege9befho8m2xsrhrt/final-output.m3u8
audiobookshelf  | [2023-03-21 06:39:14] INFO: 
audiobookshelf  | [2023-03-21 06:39:14] INFO: [STREAM] Clearing isResetting
audiobookshelf  | [2023-03-21 06:39:15] WARN: File path does not exist /metadata/streams/play_ege9befho8m2xsrhrt/output-289.ts (HlsRouter.js:37)
audiobookshelf  | [2023-03-21 06:39:15] INFO: Segment #289 requested is 287 segments from latest (28:54.0) - Reset Transcode
audiobookshelf  | [2023-03-21 06:39:15] INFO: [FFMPEG] Transcode Killed
audiobookshelf  | [2023-03-21 06:39:16] INFO: Stream Reset New Start Time 28:42.0
audiobookshelf  | [2023-03-21 06:39:16] INFO: [STREAM] START STREAM - Num Segments: 793
audiobookshelf  | [2023-03-21 06:39:16] INFO: [HlsRouter] Resetting Stream - notify client @1734s
audiobookshelf  | [2023-03-21 06:39:16] INFO: [STREAM] Starting Stream at startTime 28:12.0 (User startTime 28:42.0) and Segment #282
audiobookshelf  | [2023-03-21 06:39:16] DEBUG: [Stream] Forcing AAC for tracks with mime type audio/mp4 and codec alac (Stream.js:279)
audiobookshelf  | [2023-03-21 06:39:16] INFO: [INFO] FFMPEG transcoding started with command: ffmpeg -seek_timestamp 1 -f concat -safe 0 -ss 203s -noaccurate_seek -i /metadata/streams/play_ege9befho8m2xsrhrt/files.txt -y -loglevel error -map 0:a -c:a aac -f hls -copyts -avoid_negative_ts make_non_negative -max_delay 5000000 -max_muxing_queue_size 2048 -hls_time 6 -hls_segment_type mpegts -start_number 282 -hls_playlist_type vod -hls_list_size 0 -hls_allow_cache 0 -hls_segment_filename /metadata/streams/play_ege9befho8m2xsrhrt/output-%d.ts /metadata/streams/play_ege9befho8m2xsrhrt/final-output.m3u8
audiobookshelf  | [2023-03-21 06:39:16] INFO: 
audiobookshelf  | [2023-03-21 06:39:16] INFO: [STREAM] Clearing isResetting
audiobookshelf  | [2023-03-21 06:39:17] WARN: File path does not exist /metadata/streams/play_ege9befho8m2xsrhrt/output-289.ts (HlsRouter.js:37)
audiobookshelf  | [2023-03-21 06:39:17] INFO: Segment #289 requested is 287 segments from latest (28:54.0) - Reset Transcode
audiobookshelf  | [2023-03-21 06:39:17] INFO: [FFMPEG] Transcode Killed
audiobookshelf  | [2023-03-21 06:39:18] INFO: Stream Reset New Start Time 28:42.0
audiobookshelf  | [2023-03-21 06:39:18] INFO: [STREAM] START STREAM - Num Segments: 793
audiobookshelf  | [2023-03-21 06:39:18] INFO: [HlsRouter] Resetting Stream - notify client @1734s
audiobookshelf  | [2023-03-21 06:39:18] INFO: [STREAM] Starting Stream at startTime 28:12.0 (User startTime 28:42.0) and Segment #282
audiobookshelf  | [2023-03-21 06:39:18] DEBUG: [Stream] Forcing AAC for tracks with mime type audio/mp4 and codec alac (Stream.js:279)
audiobookshelf  | [2023-03-21 06:39:18] INFO: [INFO] FFMPEG transcoding started with command: ffmpeg -seek_timestamp 1 -f concat -safe 0 -ss 203s -noaccurate_seek -i /metadata/streams/play_ege9befho8m2xsrhrt/files.txt -y -loglevel error -map 0:a -c:a aac -f hls -copyts -avoid_negative_ts make_non_negative -max_delay 5000000 -max_muxing_queue_size 2048 -hls_time 6 -hls_segment_type mpegts -start_number 282 -hls_playlist_type vod -hls_list_size 0 -hls_allow_cache 0 -hls_segment_filename /metadata/streams/play_ege9befho8m2xsrhrt/output-%d.ts /metadata/streams/play_ege9befho8m2xsrhrt/final-output.m3u8

```

`const clearIsResettingTime = this.transcodeForceAAC ? 3000 : 500`might change the `clearIsResettingTime` to 10 seconds. On low powered hardware 3 seconds is to short, and it might falls into the isResetting loop. Maybe the waiting time should be an env var or ui-setting to customize for the device?

### DIFF
--- a/server/objects/Stream.js
+++ b/server/objects/Stream.js
@@ -314,7 +314,7 @@ class Stream extends EventEmitter {
       Logger.info('')
       if (this.isResetting) {
         // AAC encode is much slower
-        const clearIsResettingTime = this.transcodeForceAAC ? 3000 : 500
+        const clearIsResettingTime = this.transcodeForceAAC ? 10000 : 500
         setTimeout(() => {
           Logger.info('[STREAM] Clearing isResetting')
           this.isResetting = false
--- a/server/objects/Stream.js
+++ b/server/objects/Stream.js
@@ -121,7 +121,7 @@ class Stream extends EventEmitter {
     return ['mp4', 'm4a', 'm4b'].includes(this.tracksAudioFileType)
   }
   get transcodeForceAAC() {
-    return !!this.transcodeOptions.forceAAC
+    return !!this.transcodeOptions.forceAAC || this.mimeTypesToForceAAC.includes(this.tracksMimeType) || this.codecsToForceAAC.includes(this.tracksCodec)
   }
 
   toJSON() {
@@ -275,7 +275,7 @@ class Stream extends EventEmitter {
     const logLevel = process.env.NODE_ENV === 'production' ? 'error' : 'warning'
 
     let audioCodec = 'copy'
-    if (this.transcodeForceAAC || this.mimeTypesToForceAAC.includes(this.tracksMimeType) || this.codecsToForceAAC.includes(this.tracksCodec)) {
+    if (this.transcodeForceAAC) {
       Logger.debug(`[Stream] Forcing AAC for tracks with mime type ${this.tracksMimeType} and codec ${this.tracksCodec}`)
       audioCodec = 'aac'
     }


server/objects/Stream.js
```diff
@@ -314,7 +314,7 @@ class Stream extends EventEmitter {
       Logger.info('')
       if (this.isResetting) {
         // AAC encode is much slower
-        const clearIsResettingTime = this.transcodeForceAAC ? 3000 : 500
+        const clearIsResettingTime = this.transcodeForceAAC ? 10000 : 500
```
advplyr, line 317: Do you think 10 seconds is necessary here? This means after starting a stream you have to wait 10 seconds before you can seek.

fidoriel: the resetting time depends on the power of my machine and my disk speed. My RasPi does need 10s with hdd while on my dev machine needs < 100 ms with ssd. 10s helped on my raspberry pi to stop resetting over and over because he did not finish transcode the first .ts file. File not finished, Path not found => reset loop. My Desktop is fast enough not have the first .ts ready before the frontend request comes in, so he does not fall into the reset loop.

fidoriel: Today I did a little research on the "WHY this time is needed" part of this topic. [https://stsaz.github.io/fmedia/audio-formats/](url)
Turns out alac in my case is very slow to decode in comparison to wav or flac and aac slow to encode. On slow devices like the raspi this scales.

If ABS detects an transcode format it defaults to aac. If there is a jump or a bookmark position abs provides a starting point with `-ss` to ffmpeg. Therefor ffmpeg starts decoding 30 seconds before this time stamp. It generates handy `.ts` files to be stramt via hls and generates a `.m3u8` for them. If the transcode starts from the beginning, abs sends out the m3u8  via startloop, startloop handles the connection messages and its good to go. If the client resets the position, the old stream is cleaned up, and a new transcode at a new position is started. There is a case that the transcode has not reached the playing position because of the 30 seconds (slow device), now the client requests this time and because it is not transcoded, the stream rests. This is the resetting loop. In any non resetting case startloop delivered a `.m3u8` file to the client which is used #1625 . THis is contains the chunk info for the audiobook. If a chunk is requested, the info if it is avaliable is not needed. If it is retrieved, other isReset. The client starts to stream if or of not startloop sends ready, because it already happned. With this time we just give the device more time to process the transcode without reset. The client knows where to look for the audio to jump to the position.

advplyr: The socket events for "stream_open" and "stream_ready" aren't actually used for anything anymore so this original implementation you see in `Stream.js` can be misleading. You are mostly correct though but what is most important is the `isResetting` flag.

When the `isResetting` flag is enabled then the stream will not be restarted if a request is made for a segment that doesn't exist yet. This happens on line 48 of `HlsRouter.js`.

The one issue I have with using 10 seconds is that if you start an audiobook and seek to a new position within that 10 second window then the stream is not going to restart. I don't have a good solution for this right now. 

There is more issues with this general approach to HLS streaming with the iOS app. We have no control over how the iOS audio player requests segments and what happens is it will continue requesting segments with no buffer resulting in the stream resetting over and over.

fidoriel: This does not sound good. For users in a hurry 3 sec might be to long as well. Did you mean #838?
One possible solution would be to create a stream lock file like output-42.lock file if the first reset comes in. If the second reset try arrives. the lock file exist and the stream is not reset. If the real ts file is finished, and requested, the lockfile is deleted. Maybe a db save with the last resetting chunk number is better. Somehow we need to save this for the next request.
---
Not relevant but during testing I was wondering which format does transcode for which device. alac does transcode for Firefox/Chrome, wma and vorbis does as for some devices others? Right now I am converting alac to flac because since I am not using iTunes anymore there is no reason to keep them. But this is not an valid solution ;)

fidoriel: So I quickly build a lockfile solution. It is working on my desktop and my raspi. It does not use the timeout at all, so we do not have to decide on a exact amount of seconds. Seeking now is not blocked at all, it is possible immediately after the stream started. The downside: client tries to get chunk until it exists. This may not the best solution, file system reads, but it seems to work. my working branch: https://github.com/fidoriel/audiobookshelf/tree/lockfilestream

fidoriel:
I did convert all my alac to flac for various reasons, but maybe my findings are interesting for future development. I do not need this fix anymore. @advplyr if you like please close this pr.

advplyr:
Thanks for following up

PR #1627, opened 2023-03-21
Save Progress for EPUBs

vincentscode:
As discussed here the EPUB Reader in its current form is hardly usable.
I think moving forward EPUB.js is still our best choice for reading EPUBs as it is also used by calibre web for example.

This pull implements persistent progress for EPUBs so reading can be resumed which is one of the biggest problems at the moment (https://github.com/advplyr/audiobookshelf/discussions/75#discussioncomment-1966441).

It also lays the groundwork for other features such as text size control or a ToC.


advplyr:
Thanks, this is a great start. What is the benefit of storing the locations in duration?

We'll need to update the data model for media progress since a library item can have an ebook and audiobook.

vincentscode:
I've modified the code to use ebook specific fields for progress now. I've also added a simple sidebar with a table of contents.

advplyr:
Thanks, it is working great in my testing.

There are a few things we may want to address before the release but can be done separate from this. I'll have to think on it and message in Discord.

Also, if you are using VSCode there is a settings.json there and if you have vetur extension installed it should follow the same formatting used in the rest of the project. It is harder to see the changes being made when we are using different formatting.

Issue #998, opened 2022-09-21
[Enhancement]: Add SSO Support (OIDC)


DDriggs00:
### Describe the feature/enhancement

The previous attempt at implementing SSO is not valid for the current version of the app.

OIDC, SAML and LDAP appear to be the most popular methods of doing this.

pto199:
I would like to see this.  ODIC is the best method I think.

Im currently setting up Authentik and have some users now with Authentik usernames and passwords.  It would be great if they can just visit audiobookshelf and have an account auto-create for them so I dont need to make them an account.  Also they could sign on via the android/ios app with their authentik username and password.  I cant bear to give my users yet another username an password for yet another app (Which they will again independently forget eventually).

 2FA should be put on the backburner I say until SSO is done, cause most SSO solutions these days also do 2fa

kurokay:
OIDC would be awesome !

BCNelson:
I think that I'm going to take a crack at SSO. @advplyr you said in [this](https://github.com/advplyr/audiobookshelf/pull/351#issuecomment-1154169722) commented that you are going to want some more design work done on this. What do you feel is the best way to get this design work done. I see a few options.

1) I write up a draft on my plan. 
2) We chat about it on discord
3) We chat about it here
4) Discord call

or any combination of the above. Let me know.

Also it seems like a bit problem with the last PR was you not being able to easily setup a SSO server. I think that a vscode devcontainer could help out with that. I will set one up one the branch that I end up using.

advplyr:
@BCNelson Your help would be great! With something like this that I'm not familiar with it's useful to highlight some self-hosted software that already has this feature. I know a few have been mentioned but I haven't looked into them yet.

It seems like the mobile app especially complicates things here. There is a related discussion here https://github.com/advplyr/audiobookshelf-app/issues/254

It's probably best to discuss on Discord to start off then if we get a PR going we could discuss there as well.

Eschguy:
I'd love LDAP support

michaelkrieger:
Adding my comment from another issue:

[...]

PR #1636, opened 2023-03-24
Integrate passportjs for muti-strategy authentication and SSO

lukeIam:
As I wrote in the SSO issue (https://github.com/advplyr/audiobookshelf/issues/998#issuecomment-1454735891) I would suggest to integrate a well supported auth middleware instead of implementing single protocols by hand.

This PR shows a possible integration of [passortjs](https://passportjs.org) into the server code of audiobookshelf.
Login with username + password and google oauth2 integrated and working so far. But other auth providers can be integrated really easily with a few lines of code ([supported strategies](https://www.passportjs.org/packages/)).

My current implementation is not final nor ready to be merged - it is more meant as a PoC.
But if you (@advplyr and also others) decide that you want to go in this direction I'm happy to develop/contribute to a stable implementation. 

I don't have a client implementation atm - I'm testing with postman...
Login via:
`POST /login` for username + password
`GET /auth/google` redirects to google login

The response returns the known json with a token but also a cookie is set containing the session.
For the further API communication either the token (Bearer header) or the cookie can be used if authentication.
Maybe the cookie simplifies the handling in the webclients of the apps.

Most of the changes where done in the `Auth.js` file - I tried to not touch the other code where possible.

So what do you think? Is `passportjs` a good idea or do you prefer to go in another direction?


advplyr:
I think passportjs will be a good solution from the looks of it. It was attempted here https://github.com/advplyr/audiobookshelf/pull/351 but never got finalized and then the v2 rewrite happened.

@mfcar also discussed implementing this via passportjs in Discord but I don't think anything happened there.

I haven't got around to checking the PR yet but wanted to follow up.

mfcar:
> I think passportjs will be a good solution from the looks of it. It was attempted here #351 but never got finalized and then the v2 rewrite happened.
> 
> @mfcar also discussed implementing this via passportjs in Discord but I don't think anything happened there.
> 
> I haven't got around to checking the PR yet but wanted to follow up.

I still need to start the work with SSO. Sorry about this.
The plan in mind would allow registering multiple SSOs using OAuth2 standard on a Configuration Page. Allowing add provider details like Client Secret, Identifier, etc. With this, the admins could add support to login using Plex, Google, Twitter and other providers.
But this will take much more time.

I think the @lukeIam's PR it's a very good start.

lukeIam:
passport-openidconnect implementation tested with authentik

advplyr:
When getting the local strategy working I ran into an issue where it didn't allow an empty password to be passed in. Since that was already an allowable option and some users may have be doing that I included the package in `libs` folder with the changes.

I haven't tested oauth2 yet but I just got authentik setup.

lukeIam:
> When getting the local strategy working I ran into an issue where it didn't allow an empty password to be passed in. Since that was already an allowable option and some users may be doing that

oh, wasn't aware of that...

> I included the package in libs folder with the changes.

ok - that maintainer seems not want to have such a feature (as it's not longer an authorization).
But maintaining an own version makes further updates (and potential security fixes) harder (to a manual process).
But I see the problem here... (only other workaround I currently see: we tweak the client to send `EMPTY_STRING` if no password is given - but it's also not a clean solution).

> I haven't tested oauth2 yet but I just got authentik setup.

Let me know when I can assist.
My current config is:
```json
"authActiveAuthMethods": [
    "local",
    "google-oauth20",
    "openid"
  ],
  "authGoogleOauth20ClientID": "[ID]apps.googleusercontent.com",
  "authGoogleOauth20ClientSecret": "[SECRET]",
  "authGoogleOauth20CallbackURL": "/auth/google/callback",
  "authOpenIDIssuerURL": "https://[URL]/application/o/audiobookshelf/",
  "authOpenIDAuthorizationURL": "https://[URL]/application/o/authorize/",
  "authOpenIDTokenURL": "https://[URL]/application/o/token/",
  "authOpenIDUserInfoURL": "https://[URL]/application/o/userinfo/",
  "authOpenIDClientID": "[ID]",
  "authOpenIDClientSecret": "[SECRET]",
  "authOpenIDCallbackURL": "/auth/openid/callback"
}
```

advplyr:
I think it's fine. The last change in that library was many years ago. All the commits I see over the last several years are documentation updates. It is also a small library so I think we are fine including this internally.

DDriggs00:
What needs to be done before this can be implemented?

lukeIam:
I have not fully understand how the current socket connection is working and is authenticated.
But looks like `socket.io` is used - so maybe something like this could be a solution:

```typescript
// convert a connect middleware to a Socket.IO middleware
const wrap = middleware => (socket, next) => middleware(socket.request, {}, next);

io.use(wrap(sessionMiddleware));
io.use(wrap(passport.initialize()));
io.use(wrap(passport.session()));
```
from https://github.com/socketio/socket.io/blob/master/examples/passport-example/index.js#L75-L80

and on client side:

```typescript
const socket = io.connect('', {
    extraHeaders: {
        Authorization: "[...]",
    },
});
```

But again I'm just guessing here...

lukeIam:
took a bit longer than promised here (https://github.com/advplyr/audiobookshelf/issues/998#issuecomment-1681093489) - but I found some time to program again 😃 

In the dev container I'm now able to login with google auth (openid should also work but untested).
We are matching to the existing users by email.
But the current implementation it quite rough and may things needs to be solved...

Main blocker at the moment:

- [ ] How to "find" back to a "local" app page that can receive the token.

Currently I'm redirecting away from the app completely and have problems finding back.
I've hardcoded `localhost:3000` or `localhost:3333` for the moment - but this will most likely only work in the dev env.

But there is also other stuff to do:

- [ ] clean up
- [ ] writing some comments
- [ ] add a few asyncs
- [ ] currently I killed the passport-login for the rest-api in favor of the ui -> find a solution for both
- [ ] figure out why the ci is failing
- [ ] only show buttons for available providers 
- [ ] user creation via passport login (optional I guess)
- [ ] more testing 😉 
- [ ] add more providers

advplyr:
Thanks, I'd like to get back into this one again. We need to merge this with master which is going to have a lot of conflicts at this point

lukeIam:
@advplyr I already merged in master from time to time and resolved to conflicts (but we should double check)

lukeIam:
@advplyr not 100% finished (see above) - but we might can start having a look (especially in the apps - I'm just testing with the webapp so far).
Login itself is now functional...

lukeIam:
FYI: currently on "user creation via passport login"

advplyr:
Are you running authentik locally for testing?

lukeIam:
> Are you running authentik locally for testing?

docker (compose) on my mini-server

advplyr:
I got my authentik test environment up and running. My first test was with an authentik user that did not have an email set. This threw an error because you set it up to `getUserByEmail`. 

I'm not familiar with other projects that use this method so I don't know if it is normal to require an email. Is there something more secure about this then using the username to lookup the user?

advplyr:
I removed the `/auth_methods` function in favor of returning the available auth methods in the `/status` request that already goes out when hitting the login page.

I changed openid to lookup the user by username instead of by email.

> * replace `http://localhost:3333` with serverUrl (where to get it from)

I updated this

advplyr:
I setup an authentication page for password and openid. I haven't tested with google oauth yet.

When relying on the cookie to keep the session the user has to re-login if the server restarts. This is a change from before where the token was saved in local storage. This is for the best I think and we should probably remove the local storage token if it's not needed.

I updated openid to redirect to /login on failure but haven't tested this properly yet. I didn't see a failure message getting passed through but may be because of the test environment being on a different URL.

I'm also not understanding why we need an authorize URL, token URL, and Userinfo URL when this info is returned in the Issuer URL. I may be missing something there but looking at other self-hosted services they are only asking for the Issuer URL.

image

lukeIam:
> I'm not familiar with other projects that use this method so I don't know if it is normal to require an email. Is there something more secure about this then using the username to lookup the user?

> I changed openid to lookup the user by username instead of by email.

komga for example also uses the mail for matching - and it makes sense if you have more than one provider (Local + X).
Imagine you have a local user with the username `asdqwe` and you have github oauth enabled: 
I register the github username `asdqwe` and can impersonate you (or I register with this username at your Authentik while you are a "local" only user).
With emails it (normally) does not work this way: most of the auth providers actually check if you have control over the email account you are registering with.
Nice side effect: if you have the same email in multiple auth provider accounts it does not matter which one you use at login time.

lukeIam:
> I'm also not understanding why we need an authorize URL, token URL, and Userinfo URL when this info is returned in the Issuer URL. I may be missing something there but looking at other self-hosted services they are only asking for the Issuer URL.

Both should be possible - issuer-uri (if supported) just serves a document with all the other info (but I had problems at the beginning with `issuer-uri` but maybe I should check again and we might should allow both).

> When relying on the cookie to keep the session the user has to re-login if the server restarts. This is a change from before where the token was saved in local storage. This is for the best I think and we should probably remove the local storage token if it's not needed.

Session-Cookies are only used for REST Endpoints atm. For the normal websocket connection we still use `jwt`.

lukeIam:
But getting closer - Thanks for your help!

advplyr:
I haven't looked into mobile yet but depending on the complexity of that we may need to just start with openid. With the [browser capacitor plugin](https://capacitorjs.com/docs/apis/browser) we can open an in-app browser and get back events so my first thought is using that.

kaivol:
> > > I'm not familiar with other projects that use this method so I don't know if it is normal to require an email. Is there something more secure about this then using the username to lookup the user?
> > 
> > 
> > > I changed openid to lookup the user by username instead of by email.
> > 
> > 
> > komga for example also uses the mail for matching - and it makes sense if you have more than one provider (Local + X). Imagine you have a local user with the username `asdqwe` and you have github oauth enabled: I register the github username `asdqwe` and can impersonate you (or I register with this username at your Authentik while you are a "local" only user). With emails it (normally) does not work this way: most of the auth providers actually check if you have control over the email account you are registering with. Nice side effect: if you have the same email in multiple auth provider accounts it does not matter which one you use at login time.
> 
> That makes sense. We can switch it back to email and add an email input in the user account modal. I guess we would specify that the user must have a matching email, or maybe that is implied. When adding email I also wonder if we should allow username or email to login with password.

I would kindly ask you to keep in mind that in the [OpenID Connect specification](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) both `preferred_username` and `email` are not guaranteed to be unique, and only the `sub` (subject) claim (together with the `iss` (issuer)) can be used as a stable identifier for the end user.

server/Auth.js
```diff
@@ -123,15 +135,25 @@ class Auth {
         httpOnly: true
       })
 
+      // persist state if passed in
+      if (req.query.state) {
```
lukeIam, line 139: good idea 👍

server/Auth.js
```diff
@@ -16,6 +16,18 @@ class Auth {
   constructor() {
   }
 
+  static cors(req, res, next) {
```
lukeIam, line 19: For what do we require cors to be * ?
They are a security function we are disabling completely here...

advplyr:
> > I haven't looked into mobile yet but depending on the complexity of that we may need to just start with openid. With the [browser capacitor plugin](https://capacitorjs.com/docs/apis/browser) we can open an in-app browser and get back events so my first thought is using that.
> 
> I dont know how capacitor works - but it's basically a website on localhost what is displayed? Can't we just redirect the main view away from "localhost" do the login (Authentik, Google, ...) and redirect back to the same localhost page we where before? The calling website can currently set the callback url (we store the callback url in a small temp 2min cookie to find the way back).

I got openid working on android using [capacitor oauth](https://github.com/moberwasserlechner/capacitor-oauth2). It redirects to login and redirects back using an android intent. I still have to test on iOS.

server/Auth.js
advplyr: I spent some more time on this and wasn't able to get the ebooks to work with capacitor/http. There is an open issue and PR being worked on to fix an issue with returning data of the `blob` type for that plugin.

I think this will have to be updated to allow cors for only that `/ebooks` API endpoint and only allow `http://localhost` and `capacitor://localhost` ([described here](https://ionicframework.com/docs/troubleshooting/cors#what-is-cors))

server/Auth.js
```diff
@@ -21,12 +28,267 @@ class Auth {
...
+    }
+
+    // Check if we should load the google-oauth20 strategy
+    if (global.ServerSettings.authActiveAuthMethods.includes("google-oauth20")) {
+      passport.use(new GoogleStrategy({
+        clientID: global.ServerSettings.authGoogleOauth20ClientID,
+        clientSecret: global.ServerSettings.authGoogleOauth20ClientSecret,
+        callbackURL: global.ServerSettings.authGoogleOauth20CallbackURL
+      }, (async function (accessToken, refreshToken, profile, done) {
+        // TODO: do we want to create the users which does not exist?
```
Sapd, line 47: This should be definitely supported as I think almost all want to have that.
Also we can read the profile variable for the groups the idp provider sends and also automatically set admin status on register (and on login, potentially adding or revoking it). For full flexibility the groupname to be expected from the IdP should be configurable.

server/Auth.js
```diff
@@ -21,12 +28,267 @@ class Auth {
...
+      // params (isRest, callback) to a cookie that will be send to the client
+      this.paramsToCookies(req, res)
+      auth_func(req, res, next)
+    })
+
+    // openid strategy callback route (this receives the token from the configured openid login provider)
+    router.get('/auth/openid/callback',
+      passport.authenticate('openidconnect'),
+      // on a successfull login: read the cookies and react like the client requested (callback or json)
+      this.handleLoginSuccessBasedOnCookie.bind(this))
```
Sapd, line 229: It needs better error handling.
When there is an error by passport, it will just return a 403 without any information or logs and makes developing quite hard.

Can be easily tested with https://youraudiobookurl/auth/openid/callback?code=asdasd&state=3333 which should return a state error instead of forbidden (at least in logs for the admin).

During development I replaced the code like this for that:
```
app.get('/auth/openid/callback', (req, res, next) => {
  passport.authenticate('openidconnect', (err, user, info) => {
    if (err) {
      console.error('Passport authenticate error:', err);
      return res.status(500).send(err.message);
    }
    if (!user) {
      console.error('Authentication failed:', info);
      return res.status(403).send(info.message);
    }
    req.logIn(user, (err) => {
      if (err) {
        return res.status(500).send(err.message);
      }
      // here should be the success code
    });
  })(req, res, next);
});
```
but did not test it how to fully integrate it

server/Auth.js
```diff
@@ -21,12 +28,267 @@ class Auth {
...
+      // TODO: invalidate possible JWTs
+      req.logout((err) => {
+        if (err) {
+          res.sendStatus(500)
+        } else {
+          res.sendStatus(200)
+        }
+      })
+    })
+  }
```
Sapd, line 242: When using SSO it needs to invalidate the JWT and redirect to something like `application/o/audiobookshelf-dev/end-session/`.
Currently not sure how to do it best on mobile. Either the mobile app should catch that redirect, or the backend should directly check if it's mobile and redirect to an in-app link like `audiobookshelf://logout`

server/objects/settings/ServerSettings.js
```diff
@@ -97,6 +114,50 @@ class ServerSettings {
...
+    this.authGoogleOauth20ClientSecret = settings.authGoogleOauth20ClientSecret || ''
+    this.authGoogleOauth20CallbackURL = settings.authGoogleOauth20CallbackURL || ''
+
+    this.authOpenIDIssuerURL = settings.authOpenIDIssuerURL || ''
+    this.authOpenIDAuthorizationURL = settings.authOpenIDAuthorizationURL || ''
+    this.authOpenIDTokenURL = settings.authOpenIDTokenURL || ''
+    this.authOpenIDUserInfoURL = settings.authOpenIDUserInfoURL || ''
+    this.authOpenIDClientID = settings.authOpenIDClientID || ''
+    this.authOpenIDClientSecret = settings.authOpenIDClientSecret || ''
+
```
Sapd, line 128: I would recommend adding an option to change the name of the OpenID button in the frontend. Most apps provide this functionality. Reason is, that most friends or family don't know what SSO/OpenID is, so one could name it just "Family Auth", "Click here" or something like that depending on preference

server/Auth.js
```diff
@@ -21,12 +28,267 @@ class Auth {
...
+   * @param {import('express').Request} req
+   * @param {import('express').Response} res
+   */
+  async handleLoginSuccessBasedOnCookie(req, res) {
+    // get userLogin json (information about the user, server and the session)
+    const data_json = await this.getUserLoginResponsePayload(req.user)
+
+    if (req.cookies.is_rest === 'true') {
+      // REST request - send data
+      res.json(data_json)
```
Sapd, line 176: I think there should be a single comment what is_rest is exactly for as it confused me a bit during the app implementation. I used is_rest=true for the mobile app bc a redirect is not needed.

Sapd:
> > > I haven't looked into mobile yet but depending on the complexity of that we may need to just start with openid. With the [browser capacitor plugin](https://capacitorjs.com/docs/apis/browser) we can open an in-app browser and get back events so my first thought is using that.
> > 
> > 
> > I dont know how capacitor works - but it's basically a website on localhost what is displayed? Can't we just redirect the main view away from "localhost" do the login (Authentik, Google, ...) and redirect back to the same localhost page we where before? The calling website can currently set the callback url (we store the callback url in a small temp 2min cookie to find the way back).
> 
> I got openid working on android using [capacitor oauth](https://github.com/moberwasserlechner/capacitor-oauth2). It redirects to login and redirects back using an android intent. I still have to test on iOS.

After a tip from @advplyr in Discord I also noticed that Capacitor oauth2 plugin does seem to have unresolved problems with iOS: https://github.com/moberwasserlechner/capacitor-oauth2/issues/13

So I just implemented it by myself just using API requests, opening up a browser window, and passing information back into the app. Tested on iOS for now: https://github.com/advplyr/audiobookshelf-app/pull/906

lukeIam:
I'm currently very busy with other stuff - sorry

Here is my current working stage for optional automatic user creation:
https://github.com/lukeIam/audiobookshelf/compare/auth_passportjs...lukeIam:audiobookshelf:auth_passportjs_autocreate?expand=1

If something is blocking you where I can help please let me know and I will make some time available...

Sapd:
@lukeIam 
Hey, there is a very active discussion in audiobookshelfs discord (in dev-chat) about that. You probably should join in 

Main points we are discussing and testing:
- passport-openidconnect has to be replaced. It does not support PKCE and does not give back for example groups claim. Also it's rather unmaintained. The current idea is to try https://github.com/panva/node-openid-client . It comes with a passport plugin and on top is certified 
- The idea is to match users by sub instead of usernames or email.  It is defined in the OIDC standard and the recommended approach. "Old" users might be matched by name or email if the admin enables that 
- a lot more discussions 

So the PR is progressing fast

advplyr:
I just made the change to [openid-client](https://www.passportjs.org/packages/openid-client/)

That required the JWKS url so I added that to the server settings and UI. We will also need the `/end-session` URL when logout is implemented so I put that in there.

image

This is getting a bit crowded and seems unnecessary since I think we can get all of that from the config url `/application/o/audiobookshelf/.well-known/openid-configuration`. If that is the case then the UI could be updated to only ask for that URL and then do a GET request to fill in the other URLs.

The other change I made was using the `email` and `email_verified` returned from the userinfo to match with an existing user. This is temporary since we discussed making this a configurable setting.

advplyr:
I added a button next to the Issuer URL that can be used to automatically fill in the other URLs using the openid config url at `/.well-known/openid-configuration`

This will also work if users put in the config URL

openid_autopopulate

advplyr:
I added the server settings for auto-registering and matching existing users.

`sub` is now persisted on the user in Abs database when it first authenticates. On successful login it first checks if a user is already matched with the `sub`.
Next it will check for existing user matches by email or username depending on the server setting.
Lastly, it will auto register the user if none was found.

image

image

There are still a few details to workout.
1. Sending back appropriate error messages. For example, a match is found but it already has a `sub` we may want to return an error.
2. Permissions for new user. For now this is just the default permissions. This may be fine to start with.
3. Handling logouts

advplyr:
Another part I keep forgetting about is applying/updating the settings to passport. Currently the settings are only applied when first starting the server, so you have to restart after updating settings.

advplyr:
This is mostly wrapped up now. I removed the google oauth stuff so we can handle that in a separate PR.

Before this merges we may want to add a logout button. I'm thinking the client just stores the auth method locally and shows an additional logout allowing you to use the provided logout url.

We may also want to support setting the account type of auto-registered users using a user provided claim

Sapd:
> I removed the google oauth stuff so we can handle that in a separate PR

To add to this. This PR as-is should work with Google out of the box, because Google simply uses OIDC. However I did not test it. So a specific Google implementation is not required.

However for the mobile app a change is probably required in the app and in the backend (a small additional route), because Google is has some additional rules regarding the callback URL. I can implement that if someone needs that and the PR here was released.

advplyr:
I've been doing some more testing and I think things are solid the way they are now. 

In separate PRs we can implement setting the account type when auto-registering and a logout button using the openid logout url. As it stands auto-registered users will be the "User" account type with default user permissions (download only).

Another future update can clean up `Auth.js` and put the auth routes in their own file.

Further testing can be done on `edge` docker image before it gets released.

Thanks to everyone contributing on this!

Issue #1595, opened 2023-03-09
[Enhancement]: Add "Updated" to the Series Sort Options


raviwarrier:
### Describe the feature/enhancement

This is to make sure that the series that had the latest book added to it, comes on top (or bottom) when sorted by "Updated"...

![image](https://user-images.githubusercontent.com/42061703/224004908-21c1b9ee-cbc3-4260-b4f0-8d0871187ef0.png)

The series may have been added a while back, but a newly released book could have been added recently.

advplyr:
Added in [v2.2.19](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.19)

PR #1670, opened 2023-04-12
Add sortBy Last Book Added and Updated to series

divyangbw:
This allows us to sort a series by when a book was last added and when last updated. The current `addedAt` applies to the series itself, and wont change when a new book gets added. Didn't use `updatedAt` since that could apply to making changes to the series object itself in the future.

Am I correct to assume the user settings object is purely FE for now?

Handles #1595


advplyr:
Sorting happens on the server https://github.com/advplyr/audiobookshelf/blob/master/server/controllers/LibraryController.js#L410

Do you have a dev environment up and running for testing?

advplyr:
I just realized you updated that on the server so disregard the previous comment. I was thrown off by your mention of the front end.
What did you mean by 
```Am I correct to assume the user settings object is purely FE for now?```

advplyr:
I originally interpreted that feature request to be using a timestamp of when the book was added to the series. That would have been more difficult since we don't store that timestamp.
This probably makes sense though.

divyangbw:
I can add the value being used when sorting soon 👍🏽 

> I originally interpreted that feature request to be using a timestamp of when the book was added to the series.

I did consider that, but as you said, but addition to what you said, I don't think that would have mattered as much. When using sort by book last added for example, I want to see ongoing series which recently got the "new book". 

And I'm assuming a books `updatedAt` will change when adding it to a series anyways, so the sort by book last updated should be able to handle that use case anyways...I think.

advplyr:
Thanks! Works great

PR #1680, opened 2023-04-16
Use region for author queries

lukeIam:
as this should be supported now by audnex.us
https://github.com/laxamentumtech/audnexus/issues/609#event-9015163203


advplyr:
There was a bunch left to do here. The `region` had to be passed up to the API and there is no setting in the client currently to select the region to use.
I used the region from the library provider so for this to work for a different region the selected provider for the library would need to be set

PR #1684, opened 2023-04-17
blur the url input when clicking submit to add info currently in input

ThinkSalat:
Current behavior: if user has entered information in url input box but it hasn't been defocused, an error will be thrown. This blurs the input when the user taps the submit button to add what's in the input as a url


ThinkSalat:
> This wasn't working. Are you able to run the project locally for testing?

I haven't tried yet tbh. I threw this up bc I figured I could more or less pinpoint the issue. If I do any more in depth PRs I will set up a dev environment.

Issue #402, opened 2022-03-09
[Enhancement]: Labels/Tags


Kussie:
### Describe the feature/enhancement

Essentially this would be have similar to the labels features in Plex or even as a simple tagging system that is fairly common amongst these sort of applications. 

What i would like to see is the ability to place a label/tag on collections/series/individual books, these labels could then be used for searching and another level of pseudo organisation especially in the web app.  Ideally you would then be able to use these labels to control what other users can see as well.  At the moment we are limited to sharing on a per library basis, this feature would allow some greater flexibility in what is shared without having to break all of the media up into a bunch of libraries. As per Plex each user account could then have two additional fields, "Labels to include" and "Labels to exclude".  

- Labels to include: Would limit the user to only see content with the selected tags.
- Labels to exclude: Any content with any of these tags will not be shown to that user.

Jdiesel87:
Just noticed this after creating a new discussion. I'm going to cross post it for now

https://github.com/advplyr/audiobookshelf/discussions/404

M-archand:
`Labels to exclude: Any content with any of these tags will not be shown to that user.`

I would also love this feature. I use it a lot in Plex. It's a lot easier to just exclude 1 tag (e.g. _private)_ rather than allowing every single tag except _private_.

Nuuki9:
> `Labels to exclude: Any content with any of these tags will not be shown to that user.`
> 
> I would also love this feature. I use it a lot in Plex. It's a lot easier to just exclude 1 tag (e.g. _private)_ rather than allowing every single tag except _private_.

Agreed. In Plex I have a mixture of adults and children using it and I use a couple of tags to ensure both groups only see things of interest/suitability - children only see things tagged "Kids" or "Kids Only" and adults exclude "Kids Only". That way I can have things appear for one, other or both groups of users.

I'd love to be able to do something similar with ABS, as about half my library is children's audiobooks. I know I could split into two libraries, but then everything of interest to both (e.g. Harry Potter) has to sit in one or the other library and be inaccessible to some interested users.

advplyr:
There is another feature request specifically for denying user access based on tags #1545

The way the data model is setup now I don't think tags would be added to series/collections/etc. When restricting access based on tags these should be on library items (books or podcasts) otherwise we have to add logic that loops through books in a series and restrict each book. Maybe we do that but you could just bulk tag the books in the series.

advplyr:
[...]

PR #1686, opened 2023-04-17
Invert Tag Selection

### LINKED ISSUE #1545: [Enhancement]: Ability to deny user access to chosen tag(s)

Currently admins can limit a user's access to a specific tag or tags. I would like to be able to allow general access _except for_ a chosen tag or tags.

divyangbw:
Fixes #1545
Feel free to suggested UI changes as well

## Changes
Updates:
- Removed Libraries to keep this simple for now

Support for:
* Tags NOT Accessible to User
* Renamed `itemTagsAccessible` to `itemTagsSelected`.

UI & UX Changes:
* Allow the modal to scroll
* New Invert options to chose if the list is for access or prevent access.

Screenshot 2023-04-27 at 4 39 59 PM

Issues:
- #1545
- Partially #402


advplyr:
An issue with this method is that if you do the invert then add more tags to the library it would need to continually keep updating the `itemTagsAccessible` array or else those users won't have access to anything with new tags.

Also, if someone has thousands of tags then we would be increasing the payload of a user by a notable amount.

divyangbw:
Hmm, true. Guessing this holds true even now though, right? Anytime a new tag is added, the admin would need to go in add access to each user? Also I guess no one will go ahead and add a thousand tags to begin with, but if they ever did, the payload problem would happen even now?

Could go and create 2 separate arrays (boolean clears one or the other on save) or maybe better, use the invert boolean to determine if user has access. So rather than `itemTagsAccessible`, we make it `itemTagsSelected`, and the boolean determines if its for access or not access. Problem with both of these approaches is that we are essentially changing the way the core permissions would work...not sure how that effects mobile, for example.

Do you have a strategy in place for something like this? I'm not sure I'd have the time to touch the mobile app right now.

advplyr:
Ugh, I had typed a whole nice response to you then my browser randomly closed.

I think the best approach is what you said about using `itemTagsSelected`. This will not impact the mobile clients at all because these permissions are only used by the API to determine whether it should return an item to the user or not.

Every API that is returning library items will call `checkCanAccessLibraryItemWithTags` function in `User.js` to check if the user has access. That should be the only place as far as permissions go that needs to be updated. The rest is just UI and the `update` function in `User.js`

apiweb:
I have a suggestion regarding the option to invert the selection on GitHub. Instead of just having the option to invert the selection, it would be great if we could have both an allow and deny list. This would be especially useful in cases where we only want to allow certain libraries or deny certain access.

Having both lists would make it much easier to manage compared to having to invert the selection constantly. This would also solve the issue of updating the list every time a new library is added, as users can simply add the new library/tag to the allow/deny list themselves.

divyangbw:
@apiweb With the changes I'm making, there's always a single list to maintain -- has access to X tags OR doesn't have access to X tags. That means, we no longer need to update the list every time a new tag. So if I block access to the tag `Adult`, the user will just get access to everything (including any new tags that are added) except `Adult.

I'll be removing the Library related changes for now to keep this simple. I'm confused by `having to invert the selection constantly`...how so? When creating/updating your user, you're simply deciding if you wish to give per tag access or block per tag access. Same toggle as give access to everything.

advplyr:
Thanks! This is working well

PR #1692, opened 2023-04-18
fix: Make sure all existing genres also show up on the match tab

### LINKED ISSUE #1690: [Bug]: when previewing a "Match", Genres do not autocomplete
### Describe the issue

When editing Genre in the Match editor, autocomplete is partially present, but doesn't connect to the normal list of Genres available in the regular book editor -- it only allows autocompletion of already-present Genres (which of course isn't allowed). It would make more sense if either it would include full autocompletion (allowing me to check whether what I've typed matches a genre I used in the past).

1. I click on a book's "..." button, choose Match from the dropdown
2. Wait for match to load, click the correct matching data
3. Edit Genre. Type in something, notice that autocomplete only works to complete genres already present in the list.

### Audiobookshelf version

2.2.18

Docker

divyangbw:
Fixes #1690


advplyr:
Thanks!

PR #1695, opened 2023-04-18
Improve German (de.json) translation

jkuehnemundt:


client/strings/de.json
```diff
@@ -20,7 +20,7 @@
   "ButtonCreate": "Erstellen",
   "ButtonCreateBackup": "Sicherung erstellen",
   "ButtonDelete": "Löschen",
-  "ButtonDownloadQueue": "Queue",
+  "ButtonDownloadQueue": "Download Warteschlange",
```
advplyr, line 23: Does `Warteschlange` translate okay on its own? This is displayed in the side rail so shorter text is preferred.

image

PR #1710, opened 2023-04-20
Add AudiobookCovers.com metadata provider

Weldawadyathink:
I recently launched [AudiobookCovers.com](audiobookcovers.com). [Source code and documentation](https://github.com/Weldawadyathink/Audiobook-Covers). The goal of the project is to archive all the audiobook covers posted to [/r/audiobookcovers](reddit.com/r/audiobookcovers), and make them searchable using google vision optical character recognition. 

This pull request adds support for AudiobookCovers.com as a metadata provider. Unlike other metadata providers, it only provides cover images. Therefore, this PR also allows providers to be listed only in the covers modal, so the new provider will not be listed under match or the library config screens.

### DIFF
--- a/client/components/modals/item/tabs/Cover.vue
+++ b/client/components/modals/item/tabs/Cover.vue
@@ -49,7 +49,7 @@
     </div>
     <form @submit.prevent="submitSearchForm">
       <div class="flex items-center justify-start -mx-1 h-20">
-        <div class="w-45 px-1">
+        <div class="w-48 px-1">
           <ui-dropdown v-model="provider" :items="providers" :label="$strings.LabelProvider" small />
         </div>
         <div class="w-72 px-1">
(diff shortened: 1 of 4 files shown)


advplyr:
Awesome, great addition. Thanks!

damajor:
Anyone though about foreign languages and how to manage titles in different lang, and also how to match foreign titles to english covers.....

Weldawadyathink:
Foreign titled covers should be handled just fine already. My website has no concept of languages. I use OCR to recognize whatever text is in the image, so the search should still pull it up. That being said, the only source I currently have is /r/audiobookcovers, and I have never seen a non English cover there. 

As far as applying English titles to non English books, that should still be doable. You would just have to search for the English title of the book. In order to have multi language search, I think something like translating the recognized text with an api could work. I'll check out the apis that are available and see if pricing would be prohibitive.

kevingatera:
Thank you. This is awesome. One thing it only shows up on the cover page:
image

On the Match tab Audiobookcovers, shows up as empty:
image

advplyr:
Thanks, I noticed that also. Just created an issue #1764

Issue #385, opened 2022-03-01
[Enhancement]: Add custom path settings


maartenbrakkee:
### Describe the feature/enhancement

For reverse proxy usage with a subfolder instead of a subdomain. E.g., `/audiobookshelf/` instead of `/`.

unclebacon-live:
I second this.

whiskywaffel:
Would be a nice. Is there any chance this gets implemented?

sandreas:
+1

sandreas:
@advplyr 

How is the state of this issue? You said in #570: 

> we need to add a base url from env variables for the client side to work. Then there is the server and mobile apps. I've attempted to add this once before and it was more complicated then it seems. 

So there is work to do in:
- Client
- Server
- Mobile Apps

I did not find a branch with your failed attempt, but I would really like to help you with this issue and submit a PR, if you don't have the time. I would just need your help to point me into the right direction... the collected parts of code above might be a first step, but I do not know, what code parts may be also critical / necessary... 

Should I fork and create a branch / PR for discussion? Or would you just list the things you tried out and failed with?

Thank you.

advplyr:
Thanks, I could use some help.
The first step is the base url will need to be passed in as an environment variable for docker or specified in the config file for debian.

The part where I got hung up was passing that environment variable to the Nuxt.js app so that we can set the base url in the nuxt.config.js file. I'm not sure the best way to handle that. I think setting that would handle the web app client portion of this.

I'm not sure exactly what would be needed on the server yet.
The mobile apps would need to store the base url separately. I haven't gotten too deep into this since I got hung up on the web app client.

advplyr:
Oh great, how are you passing the env variable to the nuxt app?

sandreas:
> Oh great, how are you passing the env variable to the nuxt app?

That is something I am working at. I tried some static options first and it worked with the options `router` and `baseUrl`:

```
  router: {
    base: '/audiobooks'
  },
  axios: {
    // baseURL: process.env.serverUrl || ''
    baseURL: "/audiobooks/"
  },
```

The `nuxt`-config is very fiddly here. A missing `/` or one too much at the wrong place and it does not work at all. Seems that this is influenced by the server config in `dev.js`. 

[...]

PR #1724, opened 2023-04-25
Fix: apply base path to client and websocket

1nVitr0:
### Adresses #385

Although in my tests this seems to work reliably, this is at best a temporary solution. As discussed in #385, this should be replaced by a long-term solution, such as running nuxt programatically or as it's own thread and proxying requests to it.

Feel free to absolutely obliterate it, or keep this open until the long-term solution is finished. I think we should decide on a direction before I begin tackling that one. 

Also, this was done outside the development container (Node 16.20, Ubuntu 22.04), so you might want to look over the lockfile.

### DIFF
--- a/client/layouts/default.vue
+++ b/client/layouts/default.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="text-white max-h-screen h-screen overflow-hidden bg-bg">
+  <div class="text-white max-h-screen h-screen overflow-hidden bg-bg">
     <app-appbar />
 
     <app-side-rail v-if="isShowingSideRail" class="hidden md:block" />
--- a/client/layouts/default.vue
+++ b/client/layouts/default.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="text-white max-h-screen h-screen overflow-hidden bg-bg">
+    <div class="text-white max-h-screen h-screen overflow-hidden bg-bg">
     <app-appbar />
 
     <app-side-rail v-if="isShowingSideRail" class="hidden md:block" />
@@ -368,6 +368,7 @@ export default {
     initializeSocket() {
       this.socket = this.$nuxtSocket({
         name: process.env.NODE_ENV === 'development' ? 'dev' : 'prod',
+        channel: this.$config.routerBasePath,
         persist: 'main',
         teardown: false,
         transports: ['websocket'],
--- a/server/Server.js
+++ b/server/Server.js
@@ -344,8 +344,11 @@ class Server {
       return
     }
 
+    const urlKeys = ["href", "src", "url", "scope", "workerUrl"].join("|");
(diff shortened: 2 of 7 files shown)


client/layouts/default.vue
```diff
@@ -368,14 +368,15 @@ export default {
     initializeSocket() {
       this.socket = this.$nuxtSocket({
         name: process.env.NODE_ENV === 'development' ? 'dev' : 'prod',
+        channel: this.$config.routerBasePath,
```
1nVitr0, line 371: This selects the "basePath" for sockets

server/Server.js
```diff
@@ -323,6 +327,62 @@ class Server {
...
+    const currentBasePath = indexHtml.match(/routerBasePath:"([^"]*)"/)[1]
+    if (currentBasePath === basePath) {
+      Logger.info(`[Server] Client base path already set to ${basePath}`)
+      return;
+    }
+
+    // Matches e.g `href="/_nuxt/..."` or `src='/...'` or `"url":"http://localhost:3333/...`
+    const replaceUrls = new RegExp(`((?:href|src|url|scope)"?(?:=|:)\\s*["'](?:http://localhost:3333)?)${currentBasePath || "/"}([^"']*["'])`, "g")
+    const replaceNuxtConfig = new RegExp(`<script>window.__NUXT__=.*?</script>`, "g")
+    const replaceAssetPaths = new RegExp(`${currentBasePath}/(_nuxt/[^"]*|sw.js)`, "g")
```
1nVitr0, line 350: I looked through every single change this made, and they all looked fine. No guarantee though that it will stay this way, nor that it will always catch every path to be updated.

server/SocketAuthority.js
```diff
@@ -68,12 +68,12 @@ class SocketAuthority {
...
     this.Server = Server
 
-    this.io = new SocketIO.Server(this.Server.server, {
+    this.io = (new SocketIO.Server(this.Server.server, {
       cors: {
         origin: '*',
         methods: ["GET", "POST"]
       }
-    })
+    })).of(global.RouterBasePath || "/")
```
1nVitr0, line 76: This will always use the base path as namespace

advplyr:
Awesome, this is working in my quick test. I'll have to do more testing when I have time. I spotted one path for the comic reader that still needs to get updated https://github.com/advplyr/audiobookshelf/blob/master/client/components/readers/ComicReader.vue#L59

Thanks for getting this moving along!

1nVitr0:
The `workerUrl` should now update as well, it should now also be easier to update other key-value pairs. I'll be on discord tomorrow if you have time to chat about more permanent solutions.

1nVitr0:
@advplyr I guess we close this? Feel free to do so :smile:

Issue #1486, opened 2023-02-06
[Enhancement]: Disable Version Check


Artain:
### Describe the feature/enhancement

I really love what you guys did with Audiobookshelf but is possible to disable the version check in the UI. It makes a request to Github. I would like my self hosted server to not make calls to resources outside of my server.
Thanks.

PR #1741, opened 2023-04-28
Add New Options: Hide Version & Disable Checking for Updates

divyangbw:
Handles #1486 

Add 2 new options:
- Disabling checking for version updates
- Hide the version from the User (still shows up in the settings view)

@advplyr I noticed that in `client/store/index` -> `checkForUpdate`, we return `res && res.hasUpdate`. Basically return is a boolean. This means we will never tigger this:

```
.then((res) => {
  if (res && res.hasUpdate) this.showUpdateToast(res)
})
```
I personally like that a toast doesn't show, but thoughts?

## Options

Screenshot 2023-04-28 at 7 39 59 PM


advplyr:
I'm not sure about hiding the version entirely. That seems like only a disadvantage for developers and maybe a minor convenience for users not having to see it down there.
The toast was useful early on when critical updates were being made weekly but not necessary anymore.

We already get a good amount of bug reports and feature requests that were fixed/implemented in later versions. Hopefully we don't start getting a bunch more of these with this feature

advplyr:
https://github.com/advplyr/audiobookshelf/issues/1744

divyangbw:
@advplyr I'll get back to this soon and go over some reasoning as well. Busy atm 😪

divyangbw:
👋🏽 ok, so this is something I was basically doing on my local instance anyways, but here's some reasons why:

Prevent GH checks: My server has been locked down pretty tight (in parts due to qnap security issues they have had). Meaning the "prod" version has zero access to GH and the request gets blocked and creates noise in the logs. 

As for hiding the version: the devs/admins will  still see it on the settings page and that will never be hidden. This removes it for "end users" essentially. There's no value that _I_ see. Feel free to add insights I might have missed. I think  that vertical space is pretty valuable (esp for  those with smaller displays, or like me, use snap window). See  the screenshot below. Could just fix the css behind it, but overall, the question is for a user who is reading or listening (in most cases my friends and family who aren't devs), what value does seeing the current version provide?

If someone sees an issue, they'll reach out to me, I can debug, and either create a fix or a ticket or explain why to them. In all instances, I can see the version on the settings page anyways. And to those who don't care, they can simply not toggle the setting 🤷🏽‍♂️ 

Screenshot 2023-05-08 at 3 01 52 PM

divyangbw:
> The toast was useful early on when critical updates were being made weekly but not necessary anymore.

In that case, the next time this is touched, are we ok with just cleaning up that code/removing it? Less code to try and understand and confuse us 🤷🏽‍♂️

jrpetersjr:
When will this be merged and deployed? This is a killer on a small screen. 

@advplyr, @divyangbw

Issue #1721, opened 2023-04-25
[Enhancement]: seekForward/Backward instead of next/previousTrack option


Alistair1231:
### Describe the feature/enhancement

Hi!
Love the app. Thanks for your work.
I would love it, if there were an option to allow me to use my media keys, for skipping tracks forward/backwards, to instead of skipping an entire chapter, only skipping 10 seconds. In audiobooks I often simply miss what was said and want to hear it again. On android, the media keys simply skip back a bit, on desktop it skips entire chapters however. I rarely need to do that, and if I do I can just use the UI. Since other people may have different needs I feel an option is better, than just changing the current way it works.  
I was playing around with the code, and found that in `StreamContainer.vue` on Lines 369 - 371 can be changed to easily accomplish the change in behaviour. I would've made a pull request. But I am hopelessly lost in how to make a settings option with my change instead of hardcoding it. My change is simply.
From this:
```js
navigator.mediaSession.setActionHandler('previoustrack', this.mediaSessionPreviousTrack)
const hasNextChapter = this.$refs.audioPlayer && this.$refs.audioPlayer.hasNextChapter
navigator.mediaSession.setActionHandler('nexttrack', hasNextChapter ? this.mediaSessionNextTrack : null)
```
To this:
```js
navigator.mediaSession.setActionHandler('previoustrack', this.mediaSessionSeekBackward)
navigator.mediaSession.setActionHandler('nexttrack', this.mediaSessionSeekForward)
```

advplyr:
Hey are you using a keyboard for the media keys?

advplyr:
In the browser controls there is both seek and next track buttons.

![image](https://user-images.githubusercontent.com/67830747/234431603-57a727cd-8084-48b1-b7d5-529ceee4ff81.png)

Are you using something that doesn't have the seek buttons?

advplyr:
Ah ok so it doesn't support the seek forward/backward media keys?

Alistair1231:
No, neither my Keyboard, nor Autohotkey support those. I also checked the Powertoys Keyboard Manager and couldn't see such a key. The typically well supported keys are Play/Pause, Stop, Prev Track and Next Track.
Outside of UI elements I have never encountered seekforward / seekbackward.

advplyr:
Are there other websites you use the media session keys for?

I just checked youtube.com and it looks like next/prev track isn't set for regular video. Maybe for a playlist. They are setting the seek forward/backward.

[...]

PR #1760, opened 2023-05-07
seek instead of skip with action handlers

Alistair1231:
Related to https://github.com/advplyr/audiobookshelf/issues/1721


advplyr:
Thanks!

Issue #1766, opened 2023-05-10
[Enhancement]: Let me use bookmarks without losing reading stats


zblesk:
### Describe the feature/enhancement

I often make bookmarks when listening to a book, with the intent of revisiting them sometime after finishing the book. (I use the web version for this.)

Currently, I can only view bookmarks by going to the books' page, hitting Play, going to the Player and displaying the bookmarks there. 

The biggest disadvantage of this is that it deletes my previous progress. I like seeing which book I finished when; but when I go to revisit the bookmarks, ABS forgets when the book was finished and just keeps tracking it as a currently-reading book. I can only fix that by 'finishing' it again, which puts today's date there, messing up my history. 

I'd like to be able to play bookmarked locations without changing the 'finished' state and date of the book.

Dr-Blank:
#1653 would fix this as a side effect since multiple plays would not affect the finished status

advplyr:
I can see #1653 partially handling this by still showing that you have finished it before, but I don't see how this could be built so that it doesn't save your progress when resuming a bookmark.

This would require some type of preview audio player or an extra button saying to play without tracking progress. I don't see this being implemented unless you have some other idea.

zblesk:
Frankly, if I understood #1653 correctly - that is to say: that it would track progress, but count it as a _new_ reading, keeping the old one as-is, and letting me delete the new progress as easily as I can do it now - I'd be perfectly happy with that. 
I understand not wanting to add convoluted niche features to the app, and there's really no need

mostlyvirtual:
Don't want to hijack this thread, but I can't find the answer anywhere, and this is as close as I've gotten to my question:

Is there any way to export the bookmarks?

advplyr:
@zblesk Yeah we'll have to workout the details on how that functions but you are understanding correctly.

@mostlyvirtual I responded in Discord.

advplyr:
Duplicate of #1653

Issue #1653, opened 2023-04-07
[Enhancement]: Play count


burnshroom:
### Describe the feature/enhancement

I mentioned this last year under [issue #527](https://github.com/advplyr/audiobookshelf/issues/527#issuecomment-1152207707).

Still holding out hope for it as I have several books I listen to over and over again and would like to see how many times I may have accomplished this.  

Similar to what Plex has and that it also allows you to sort by this play count.  

![image](https://user-images.githubusercontent.com/123109/230566966-f988bfa5-4004-4bc5-bb14-51af08a82465.png)

Not something that makes or breaks the app at all, but would be nice to have.

advplyr:
The difficulty I have with this is how is a play counted? What if you toggle the "Mark as Finished" button a few times, would those be plays?
Maybe testing what Plex considers a play would be helpful

Dr-Blank:
Here is how plex counts the plays:

```mermaid
flowchart TD
    A["Play count = 0 (Unwatched State)"] -->|Watch more than 90%| B["Play count = 1"]
    A["Play count = 0 (Unwatched State)"] -->|Mark it as watched| B["Play count = 1"]
    B -->|Watch more than 90%| C["Play count +=1"]
    C -->|Watch more than 90%| C["Play count +=1"]
    C -->|Mark as unwatched| A
```

Jmanko16:
Honestly a simple counter (with a plus/minus) would work.  So if you mark watched etc but did on accident jus hit the negative button.  Obviously have the play count update on finished as above as well but this would give more control (and allow for reading outside ABS If used as a front end for audiobook/ebook)

nichwall:
To track the plays, I like the idea of the plus/minus buttons to simplify tracking and accidentally "Making as Finished" messing up the counter.

Incorporating the idea from #1905, could this be added as an array of completion times (separate item entry in the database)? My idea for the UI would look something like the History view in the Android app (for checking sync status), where each line is the "completion date", and then have an Edit and Remove button on the right to fix errors easily. In a web browser, it would just be a popup because that information doesn't need a ton of horizontal space.

To get to this view, the user can either use the three dot menu OR click on the listening progress box.

![Screenshot_20230715-073638.png](https://github.com/advplyr/audiobookshelf/assets/5686638/3a2757b5-f9c9-428a-a61e-6018e10a182e)

[...]

PR #1778, opened 2023-05-17
Don't sync progress for finished content

watsonbox:
This is just a drafted idea for the moment, but the goal is essentially to allow finished content to be re-visited without (necessarily) updating progress or listening stats.

This relates to a few topics currently being discussed, but most pertinently [Let me use bookmarks without losing reading stats](https://github.com/advplyr/audiobookshelf/issues/1766#top). It explores [a comment](https://github.com/advplyr/audiobookshelf/issues/1766#issuecomment-1544529902) on that issue which proposed the idea of a sort of preview player (one that won't track progress), but instead of being activated by a button, here it is enabled or disabled depending on finished status. That issue has now been marked as a duplicate of https://github.com/advplyr/audiobookshelf/issues/1653, so I'll add any further comments there (even if it doesn't quite feel like the same thing).

In terms of implementation, I think ultimately it might be better to have this as a setting rather than default behavior. When set, it would still of course be possible to mark the item as unfinished again to re-enable progress tracking. I created a distinction between `isFinished` at the session/progress level, and `syncProgress` at the player level, mostly because `syncProgress` need not necessarily only be deactivated for finished content going forward, even if that's the current proposition.

Some things which remain to be addressed:

* Since the behavior of the player depends on finished status, changing the finished status should probably close or better yet update the player state accordingly.
* Should a listening session be opened at all for finished content? I believe right now a listening session is required in order to get the media details, so I've left it like this for now to keep things simple.

Again, this is just an idea to see whether it's something potentially workable. I'm fine with throwing it away of course, but it turned out to be surprisingly easy to put together so I figured I'd share it all the same.

### DIFF
--- a/client/components/app/StreamContainer.vue
+++ b/client/components/app/StreamContainer.vue
@@ -41,6 +41,7 @@
       :sleep-timer-set="sleepTimerSet"
       :sleep-timer-remaining="sleepTimerRemaining"
       :is-podcast="isPodcast"
+      :is-finished="isFinished"
       @playPause="playPause"
       @jumpForward="jumpForward"
       @jumpBackward="jumpBackward"
@@ -82,7 +83,8 @@ export default {
       sleepTimer: null,
       displayTitle: null,
       currentPlaybackRate: 1,
-      syncFailedToast: null
+      syncFailedToast: null,
+      isFinished: false
     }
   },
   computed: {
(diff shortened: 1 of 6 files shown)


advplyr:
I can't seem to think of a good UI/UX for this other then having a separate button like "Preview" to open a player that doesn't sync.
If we had a server setting what would it say? Would they never be able to track progress again on a finished item unless they disable the server setting?
Is there examples of this being implemented in other servers that we can reference?

advplyr:
Having a separate "Preview" button I don't think is a good strategy either though because it isn't intuitive. It would also need to be in a context menu so at that point you wouldn't be saving any clicks by just marking the item as complete again.

advplyr:
Closing this as abandoned

PR #1786, opened 2023-05-21
Update Dutch translation

92Kev:
Some minor changes to fix errors or replace unintuitive terms.

Below is a number of strings I ran into that are not in the string files yet, can they be added to the string files?
ABS strings 1
ABS strings 2
ABS strings 3
ABS strings 4


advplyr:
Thanks! Yes I can add those. I'm not sure about `16 hr 14 min`. I don't know if all translations have a shorthand for hour, minute and second.

PR #1814, opened 2023-06-03
Update readme.md - Expand required Apache modules for reverse proxy and detail how to handle acme challenges

depwl9992:
After setting up the reverse proxy for Apache per the current instructions, I received the error:

> AH01144: No protocol handler was valid for the URL / (scheme 'http')

Installing `proxy_http` and `proxy_balancer` modules in addition to those listed (as recommended [here](https://stackoverflow.com/a/26045183)) fixed this issue.

`a2enmod` command does not include the redundant `_module` suffix.

Let's Encrypt was not able to validate certificates either automatically or manually as ABS does not respond to ACME challenges directly. Editing the virtual environment configuration to not reverse-proxy `/.well-known`, and serving that directly from Apache allowed LE to validate the challenge.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -117,9 +117,11 @@ Add this to the site config file on your Apache server after you have changed th
 
 For this to work you must enable at least the following mods using `a2enmod`:
   - `ssl`
-  - `proxy_module`
-  - `proxy_wstunnel_module`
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #1819, opened 2023-06-04
Fix: useTempFiles=true, upload use tmp instead of ram

### LINKED ISSUE #1817: [Enhancement]: allow or default to use express-fileupload/useTempFiles=true

https://github.com/richardgirges/express-fileupload/blob/master/README.md

image

ABS actually default to use memory for upload, which isn't quite an option for large audio book collections.

mayli:
Fix #1817

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -152,6 +152,7 @@ class Server {
       defCharset: 'utf8',
       defParamCharset: 'utf8',
       useTempFiles: true,
+      tempFileDir: Path.join(global.MetadataPath, 'tmp')
     }))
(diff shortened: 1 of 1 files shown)


mayli:
Seems to be working for me, uploaded a 5G collection while keep rss at 103M.

advplyr:
This is the right solution but we just need to make sure the temp files dir is appropriate for the debian package

advplyr:
I think the best solution is to make the temp files dir inside the `/metadata` folder. I tested and is working well for me.

PR #1838, opened 2023-06-12
Updates to Email settings/manager to include test email

daVinci2793:
Made minor updates to Email manager and settings to include a "Test Address."

Justification: Some smtp relays and/or APIs do not inherently support receiving emails. Ideally, the Email Test button would include a customizable recipient. Allowing users to arbitrarily input an email address to send the test email to instead of assuming the "From Address" is capable of receiving emails. 

Here are the changes made to the client:
image

Additionally, I added a simple fixed string Subject to the Ebook Email transporter. There was something jarring about seeing (NO SUBJECT) and that might get filtered out.


advplyr:
I updated it so it falls back to using the from address if the test address is not set.

I also noticed when using this that the Test button is confusing since it is using the saved data, so you have to save first then press test. As a quick workaround to this I put in a reset button that is shown instead of the test button when there are changes to the form.

The translation keys are in alphabetical order. I haven't automated this yet but for now I just add new keys to `en-us` and I have a script locally that adds the new string to the rest of the files.

daVinci2793:
Smart suggestion with the fall-back! 

I was wondering how the translation key's worked. For future reference, is the best-practice to simply update the en-us and allow you to run your script on the other keys?

PR #1946, opened 2023-07-19
Fix byte conversion

rasmuslos:
Right now ABS converts `206291193b` into `196.73MB`. This is only partially correct, because the bytes are converted using a factor of 1024. The correct amount would be `206.29MB` The are converted to [Mebibyte](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) not Megabytes. I got confused by it, because if was wondering why my file size did not match with the one I uploaded. Changing the factor to 1000 fixes this. The unit could also be renamed to `MiB` but I think this would only confuse most people.

Bildschirmfoto 2023-07-20 um 00 00 38
Bildschirmfoto 2023-07-20 um 00 00 56
Bildschirmfoto 2023-07-20 um 00 02 49

Bildschirmfoto 2023-07-20 um 00 02 11

And yes linux used Mebibyte here, but they do it throughout the entire system. My supposed 5TB Hard drive shows up as a 4.5TB one, because it is only technically 5TB large. Actually it has only 4.5TiB of usable space. But I think this is trivial, as most people probably don't use the linux command line to find out what the size of a file is.

Bildschirmfoto 2023-07-20 um 00 11 06

### DIFF
--- a/server/utils/fileUtils.js
+++ b/server/utils/fileUtils.js
@@ -83,7 +83,7 @@ function bytesPretty(bytes, decimals = 0) {
   if (bytes === 0) {
     return '0 Bytes'
   }
-  const k = 1024
+  const k = 1000
   var dm = decimals < 0 ? 0 : decimals
   const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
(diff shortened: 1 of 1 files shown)


advplyr:
I agree. If anything we could setup a tooltip that shows the Mebibyte value but it doesn't make a difference for me.

PR #1965, opened 2023-07-25
Added Lithuanian translation

petras-sukys:
Added Lithuanian language L10N string file and added entry for available languages.

Brunt of the work was done by the one that shall be not named, but each string hand-verified as much as possible by native speaker :)


advplyr:
Thanks!

Nighthad:
Dekui

PR #1994, opened 2023-08-06
Fix: server crash when uploading invalid backup file

NiclasHaderer:
If the backup file is not a valid .zip file, or the .zip file is corrupted the server crashes. 
This catches the error and returns an error description to the user

### DIFF
--- a/server/managers/BackupManager.js
+++ b/server/managers/BackupManager.js
@@ -83,7 +83,7 @@ class BackupManager {
       return res.status(500).send('Invalid backup file')
     }
 
-    const tempPath = fileUtils.sanitizeFilename(Path.join(this.BackupPath, backupFile.name))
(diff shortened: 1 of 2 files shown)


server/managers/BackupManager.js
```diff
@@ -82,7 +83,7 @@ class BackupManager {
       return res.status(500).send('Invalid backup file')
     }
 
-    const tempPath = Path.join(this.BackupPath, backupFile.name)
+    const tempPath = fileUtils.sanitizeFilename(Path.join(this.BackupPath, backupFile.name))
```
advplyr, line 86:
```suggestion
    const tempPath = Path.join(this.BackupPath, fileUtils.sanitizeFilename(backupFile.name))
```
This function was just written for filenames

advplyr:
Thanks!

PR #2026, opened 2023-08-22
Add config page for all RSS feeds

shawnphoffman:
# Overview

I tend to lose track of what RSS feeds I have open so I decided to add a config screen for them. Basically, it just lists all of the open RSS feeds with some basic metadata and a button to close them. A modal lists the episodes associated with the feed and copy-to-clipboard option. 

Discussion: #2014 

## Open Questions

- [ ] Localization. Is there a process for this? Do we leave the keys out of the JSON files or include them with English strings? 

## Screenshots

Full-width

Narrow view

View feeds modal

Close feed confirmation


advplyr:
Thanks, this is a good idea. For localization just updating the `en-us.json` file is perfect. At some point a more scalable localization solution will be put in place but for now I just run a script that adds the English strings to all the other json files.

PR #2027, opened 2023-08-23
Added podcasts volume to docker compose

realmain:
The example docker compose file in github does not have the podcasts volume in it, unlike on [the official website](https://www.audiobookshelf.org/docs#docker-compose-install).


advplyr:
Thanks!

PR #2030, opened 2023-08-25
Fix system file ownership for Debian package

bluecmd:
Extract ffmpeg and tone but ignore the ownership information from the tar archive. This will ensure that those files are owned by root, not the UIDs/GIDs the tar files happen to contain.


advplyr:
Thanks!

PR #2054, opened 2023-09-05
Follow-up update for German strings

JBlond:
Follow-up update for German strings to 469167df661a6537f5aba34cdba03111dfdda95e. 
Plus addtional string translation


advplyr:
Thanks!

PR #2089, opened 2023-09-12
Update de.json

Lionfox2:
The more suitable translation for "discover" is "Entdecken", same wording as in other media apps like Spotify


advplyr:
Thanks!

PR #2120, opened 2023-09-18
[server] x-accel: encode all paths to URIs

itzexor:
updates util function  encodeUriPath to use node:url with a file:// path prefix, and updates all instances x-accel redirection to use this helper util instead of sending unencoded paths into the header.

Using a local dev container with an external nginx reverse proxy for x-accel: I tested covers, audio files, and ebooks which all played correctly, even with unicode chars. I was unable to get download working, but I think that was just because of my test setup. I don't see any reason it should function differently in a proper setup...

fixes errors such as:
```
[2023-09-16 22:46:05] DEBUG: Use X-Accel to serve static file /audiobooks/Craig Alanson/Ascendant/1. Ascendantː Book 1/Ascendantː Book 1.webm (LibraryItemController.js:578)
node:internal/errors:478
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["X-Accel-Redirect"]
    at ServerResponse.setHeader (node:_http_outgoing:647:3)
    at ServerResponse.header (/node_modules/express/lib/response.js:794:10)
    at ServerResponse.header (/node_modules/express/lib/response.js:797:12)
    at ApiRouter.getLibraryFile (/server/controllers/LibraryItemController.js:579:30)
    at Layer.handle [as handle_request] (/node_modules/express/lib/router/layer.js:95:5)
    at next (/node_modules/express/lib/router/route.js:144:13)
    at ApiRouter.middleware (/server/controllers/LibraryItemController.js:742:5) {
  code: 'ERR_INVALID_CHAR'
```
[the error in this specific file is the colon which is `modifier letter triangular colon (U+02D0)`]

this also changes the log to output the full encoded x-accel path (which means it now includes the prefix):
```
[2023-09-17 22:36:24] DEBUG: Use X-Accel to serve static file /protected/audiobooks/Craig%20Alanson/Ascendant/1.%20Ascendant%CB%90%20Book%201/Ascendant%CB%90%20Book%201.webm (LibraryItemController.js:579)
```

### DIFF
--- a/server/controllers/BackupController.js
+++ b/server/controllers/BackupController.js
@@ -1,4 +1,5 @@
 const Logger = require('../Logger')
+const { encodeUriPath } = require('../utils/fileUtils')
 
 class BackupController {
   constructor() { }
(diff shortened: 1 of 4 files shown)


advplyr:
Thanks!

PR #2133, opened 2023-09-20
Add more information to the backup page

mfcar:
This pull request adds more information to the backup page:
 - The backup location path and 
 - Name when downloading a backup file

Screenshot 2023-09-20 at 22 37 51

Screenshot 2023-09-20 at 22 38 05


advplyr:
I thought that adding another api endpoint for getting the location felt like overkill so I added the `backupLocation` to the get all backups api endpoint instead.
I also moved the backup location path to be above the "Enable automatic backups" setting. It doesn't look as clean but it wasn't specific to automatic backups so I didn't want to group it in with that setting.
Thanks!

PR #2161, opened 2023-09-30
Support markdown in book descriptions

### LINKED ISSUE #617: [Enhancement]: Support Markdown in comments/description

beets-audible converts audible descriptions with html into markdown. Please support displaying markdown as html in web and mobile apps. 

image

https://api.audnex.us/books/B01IW9TQPK

Thank you for this wonderful program.

### LINKED ISSUE #1820: [Enhancement] Support HTML/rich text descriptions for audiobooks
### Describe the issue

Metadata fetching from audible is working and a great feature. But there is an issue with the comment field, that contains the description of the book.

The comments fetched from audible are loosing all their line breaks, which makes the text harder to read. Hopefully an easy that can improve the ABS even more :)

1. Fetch metadata (including comment) for an audiobook from audible.
2. Compare the text from the audible website with the text in ABS.
3. You will notice, that the line breaks are missing.

### Audiobookshelf version

2.2.22

Docker

MxMarx:
It would be pretty nice to have a little rich text in book descriptions, mostly so line breaks are preserved when fetching metadata.
This pull request uses [turndown](https://github.com/mixmark-io/turndown) to convert book descriptions grabbed from Audible or iTunes into markdown, and since [marked](https://github.com/markedjs/marked) is already included to render the changelog, rendering markdown in the descriptions is pretty easy. 

Alternatively, we could use the trix editor currently used for podcast episodes instead of using markdown. If this is something planned for the future, it might make more sense to store rich text descriptions as sanitized html instead of markdown. I could also make a pull request with the trix editor instead if wanted.

resolves #617 
resolves #1820

### DIFF
--- a/client/pages/item/_id/index.vue
+++ b/client/pages/item/_id/index.vue
@@ -125,7 +125,7 @@
           </div>
 
           <div class="my-4 max-w-2xl">
-            <p class="text-base text-gray-100 whitespace-pre-line">{{ description }}</p>
+            <div class="text-base text-gray-100 default-style" v-html="description" />
           </div>
 
           <div v-if="invalidAudioFiles.length" class="bg-error border-red-800 shadow-md p-4">
@@ -153,6 +153,8 @@
 </template>
 
 <script>
+import { marked } from '@/static/libs/marked/index.js'
+
 export default {
   async asyncData({ store, params, app, redirect, route }) {
     if (!store.state.user.user) {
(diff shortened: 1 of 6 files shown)


advplyr:
One issue I had with implementing this is descriptions are also embedded in audio file meta tags with the embed tool. I'm not sure if we want to be embedding markdown or HTML in meta tags.

If we keep this as a single `description` field then it makes more sense to me to use HTML since both Audible and iTunes use HTML.

PR #2175, opened 2023-10-02
Completed Spanish translation and corrected typo in an English string

MarshDeer:
The Spanish translation has some odd capitalization choices I tried to match for stylistic consistency, but it should probably get another pass to improve that, along with some other bad grammar and awkward phrasing I might have missed while looking for untranslated lines. 

I could do that second pass in a couple of days if it's okay for me to overwrite much of the previous translator's work. I'm a professional EN<>ES translator, but I don't want to be rude in case you would rather keep the original translator's choices.


advplyr:
Thanks! It's okay to make corrections to existing translations. If there is a discrepancy in the future I'm sure someone will bring it up.

PR #2182, opened 2023-10-05
Add Amazon Kindle provider

ahobsonsayers:
Adds a initial implementation of a provider for Amazon Kindle. This is able to provide limited metadata for the moment, but is particularty useful for grabbing covers as they tend to be pretty high quality.

Implementation is shamelessly copied from Calibre (see [here](https://github.com/kovidgoyal/calibre/blob/master/src/calibre/gui2/store/amazon_base.py) and [here](https://github.com/kovidgoyal/calibre/blob/master/src/calibre/gui2/store/amazon_live.py)) but reimplemented in js.

I plan to upgrade this further soon by pulling the page at the url for a particular book entry in the search list, allowing us to pull much more infomration such as description, publisher, year etc.

Please let me know if you have any questions or concerns!


advplyr:
Thanks for contributing! I have been declining requests for new providers that require web scraping, but this may be an exception since Calibre has been using this method successfully.

The reason to not include web scraping is it requires more maintenance due to DOM changes and is more likely to get throttled/ip banned.

If we can implement this without adding any dependencies and making it an cover-only provider like [AudiobookCovers.js](https://github.com/advplyr/audiobookshelf/blob/b447cf5c1ccb273fd45af8a7e8b11f3844c28fe6/server/providers/AudiobookCovers.js) then I think it'll be a nice addition.

ahobsonsayers:
@advplyr I completely understand why you closed this given my inactivity (apologies, i have had very little free time. Can I ask recently). Can I ask if the recent addition of custom metadata providers makes adding things like this easier?

advplyr:
For sure, this use-case is what the feature was added for. You have to host your own server to host the metadata provider but maybe once we write up some guides users will host public ones that others can use

PR #2186, opened 2023-10-05
Fuzzy matching continued

mikiher:
This is a continuation of [Fuzzy Matching V1](https://github.com/advplyr/audiobookshelf/pull/2099).
This includes some cleanups and refactoring, a few improvements, and one major enhancement.

Cleanups, refactoring, and small fixes:
- Refactor title candidates logic (addition, variants, sorting) into a separate class (TitleCandidates) (https://github.com/advplyr/audiobookshelf/commit/46b0b3a6efb7f31ac7d67ee5fff6dcbd2ff28542) 
- (minor) Rewrite the logic that makes sure we don't run the original Title/Author search twice, to make it more readable (https://github.com/advplyr/audiobookshelf/commit/10f5bc8cbeeacd3c47f7115f387dd7d5817982e7)
- (minor) Refactor OpenLib-specific sorting into getOpenLibResult (https://github.com/advplyr/audiobookshelf/commit/1d3ad38187708ca0c6efefce2d04b82820f19522)
- (minor) Add back lower-casing in the cleanTitle/Author methods, since they are called from other places in the code (https://github.com/advplyr/audiobookshelf/commit/5d7c197c893d10277f59c753e2d324837185a78f)

Enhancements & Improvements:
- (major) Move Author candidates logic into its own class (AuthorCandidates), and introduce author extraction and validation (from both author and title parts) using parallel requests to Audnexus. This helps in cases where: the author field includes additional data, or when the author hides in one of the title parts. Fuzzy search logic now has an external loop that goes over author candidates (including empty author in the end), and an internal loop that goes over title candidates (https://github.com/advplyr/audiobookshelf/commit/9eff471afaa87572bfcb312af64d756511fde2a3)
- Handle initials in author normalization (separate initials, and remove middle initials, as they sometimes mismatch with providers) (https://github.com/advplyr/audiobookshelf/commit/f3555a12ceff25d328b7dd1637668874e181946e)
- Added/fixed a couple of title transformer regular expressions. (https://github.com/advplyr/audiobookshelf/commit/b2acdadcea6fa52636d816166beac24cb370e127)
- Improved title candidate sorting (preferring transformed title parts over original ones, and title parts in their order of appearance) (https://github.com/advplyr/audiobookshelf/commit/8979586404a1ca4a46b0eff3d1cc23582ffbfbb5)
- Add just one title variant after all all transformers have been applied, and not after each transformer (https://github.com/advplyr/audiobookshelf/commit/752bfffb1109e8fadf87775ecacf588365608b03)
-  Treat underscores as title part separators (improves some corner cases) (https://github.com/advplyr/audiobookshelf/pull/2186/commits/bf9f3895db17f2172cda4e32caab559eda9c05a1)
- Reduce spurious audnexus author matching (by reducing the max Levenshtein distance, and looking only at the top 10 results) (https://github.com/advplyr/audiobookshelf/pull/2186/commits/b0b7a0a61817671b15e2687a32399aea6f0bdb51)
- If no authors have been validated, use an aggressively cleaned version of the author field (in many cases, it is better than nothing) (https://github.com/advplyr/audiobookshelf/pull/2186/commits/f44b7ed1d0f8ba538e194632f98660893d9206a6)

The code is now more robust, and handles various hard corner cases it didn't handle before.

It fixes one case in the previous eval set, and keeps a 98% found rate and 96% found@1 in a new 50 title/author pairs eval set.
In the new eval set, I also measured the average number of fuzzy searches - 1.18 (note that the eval sets are picked from an unstructured torrents folder, where the initial search with the original title and field almost always fails. 1.18 means that in most cases, only one fuzzy search request is needed).

The additional parallel author validation requests (usually between 2-4) to audnexus seem to run very quickly, and most of the network time seems to be spent in the search provider requests.


server/finders/BookFinder.js
```diff
@@ -179,35 +188,166 @@ class BookFinder {
...
+      this.candidates.add(cleanAuthor)
+
+      let candidate = cleanAuthor
+
+      for (const transformer of authorTransformers) {
+        candidate = candidate.replace(transformer[0], transformer[1]).trim()
+        if (candidate) {
+          this.candidates.add(candidate)
+        }
+      }
```
advplyr, line 311: This is unused, were you planning on adding to this?

mikiher: This was meant for future additions, but I wasn't planning on having any just right now. I removed it (will add it back if I need it).

mikiher:
By the way, I can share my eval sets and hard cases privately, if you'd like to see them.

advplyr:
@mikiher I tested this PR with my sample libraries that have random meta tags and titles and it worked really well. It would be nice to have a better test set to test with in the future.

advplyr:
I think the AuthorCandidates class can be useful for the AuthorFinder as well

mikiher:
@advplyr I just sent you a link to the evals spreadsheet to your personal email I found [here](https://www.audiobookshelf.org/support). Feel free to send me any additional hard cases you find.

server/finders/BookFinder.js
```diff
@@ -179,35 +188,152 @@ class BookFinder {
...
+      this.candidates = new Set()
+      this.cleanAuthor = cleanAuthor
+      this.priorities = {}
+      this.positions = {}
+    }
+
+    add(title, position = 0) {
+      // if title contains the author, remove it
+      if (this.cleanAuthor) {
+        const authorRe = new RegExp(`(^| | by |)${this.cleanAuthor}(?= |$)`, "g")
```
advplyr, line 204: For future reference if you are working on this, an edge case came up here with invalid regex. https://github.com/advplyr/audiobookshelf/issues/2265

Fixed by adding a util function to escape the string.

mikiher:
Nice lesson in defensive coding, thanks!

I got sidetracked by this external docker-on-windows abs watcher project
(which I hope to release sometime this week), but I'm definitely planning
to go back to the matching code. This code requires some serious unit
testing, but I didn't get a clear response to my question on discord
regarding which unit testing framework was your favorite (jest, mocha,
cypress, something else?). I think you just need to make a decision and
we'll all stick with it, but I believe you should drive this.

advplyr:
I don't have a preferred framework. I've used mocha and jest a little bit but not enough to have a preference. Do you have a preference?

mikiher:
I'm quite new to Node.js and Javascript (my background is C++ and Python). 
Let me do some digging and testing of my own and I'll come back with a recommendation.

PR #2188, opened 2023-10-06
fix: HTTP/429 when requesting authors information, resolves #1570

### LINKED ISSUE #1570: [Bug]: "Match All Authors" with a big library will hit rate limits on Audible API and start getting 429s
### Describe the issue

**Expected behaviour**
Ideally the requests would be queued server-side to throttle them to avoid exceeding the rate limit threshold

**Actual behaviour**
All lookups start to fail with 429s from Audible API. If your library is big enough, you may even get an IP ban

1. Load hundreds of books from different authors into a library
2. Visit the "authors" page
3. Click "Match All Authors"
4. Observe backend 429 errors from Audible API after several dozen lookups

### Audiobookshelf version

v2.2.4.  I can update and re-test in a little while but for the next week or so I need it to be perfectly stable as I'm traveling

Docker

jfrazx:
- Utilize a rate limiter when requesting matching author information 
- Ensure Audnexus class is a singleton


advplyr:
I put in a limiter for the login using [express-rate-limit](https://github.com/express-rate-limit/express-rate-limit) but added in the `server/libs/` folder here https://github.com/advplyr/audiobookshelf/blob/master/server/libs/expressRateLimit/index.js

Is this a library we could use instead? If we can use express-rate-limit then we can pull it out of the libs folder and add it to the package.json instead.

The `limiter` library adds an additional dependency `just-performance` that seems unnecessary at a quick glance.

jfrazx:
I don't think `express-rate-limit` will work. Being express middleware it expects to receive request and response objects. The use case of each is different. `express-rate-limit` throttles incoming requests while `limiter` intends to assist in managing outgoing requests.

nichwall:
Could the `throttleIt` package work instead as a wrapper around existing functions?

https://github.com/sindresorhus/throttleit/blob/main/index.js

jfrazx:
I don't think so. It appears `throttleIt` is for synchronous functions. But, the same author does have an async throttling package: [`p-throttle`](https://github.com/sindresorhus/p-throttle).  I'll swap out `limiter` and see how well it works.

jfrazx:
@advplyr Any other concerns with this PR?

advplyr:
This is working well in my tests. Thanks!

PR #2189, opened 2023-10-06
Try #2 at Forward Proxy Auth

igor47:
This is a somewhat-different take on #1109 

Instead of using environment variables, I've modified the UI to allow you to dynamically configure the forward auth. Screenshot:

image

image

A lot of the concerns in the previous PR had to do with the question, "how do you do forward proxy auth in the mobile app". my answer to that question: you don't. we keep both the current authorization-token based auth, as well as the forward proxy auth. if forward proxy auth is disabled, there is no change to current behavior. to log into the mobile app when using forward proxy auth, do the following:

1. log into the web app using forward proxy auth
2. set your password. the UI makes it clear that this password is only used for the mobile app login
3. use the password to log into the mobile app

in this case, the in-app passwords are more like access tokens. i had an idea for a future improvement where we display a QR code with a per-device auth token, and you can log into the app by scanning the QR code, this is not in scope right now but i think would be a pretty nice UX.

to make this work, we need support in the reverse proxy layer. in particular, we need conditional logic:

1. if there is an `Authorization` header in the request, allow it through to audiobookshelf
2. if there is not, forward the request to the authentication mechanism first

i am using traefik and authentik. i still have to test this setup, but my idea:

1. in authentik, pass through `/login` and `/ping` requests with no auth needed (this enables the app to log in and set an `Authorization` header)
2. in traefik, i need two routes and two services. one route has a rule for just the `audiobookshelf` domain, and includes the authentik middleware. the second route has a rule for the `audiobookshelf` domain, and also checks that the `authorization` header is not null. this service does *not* include the authentik middleware.

the mobile app would use the second service. i'll include config samples once i test it.


igor47:
This is very close to working. I am running it in prod behind `authentik` and `traefik`. here is my `traefik` labels:

```yaml
  audiobookshelf:
    build:
      context: ./custom/audiobookshelf
    # see: https://github.com/advplyr/audiobookshelf/pull/2189
    #image: ghcr.io/advplyr/audiobookshelf:latest
    restart: unless-stopped
    container_name: audiobookshelf
    labels:
      traefik.enable: true
      cluster.name: mittens.services
      traefik.http.routers.pods.rule: Host(`audio.example.com`)
      traefik.http.routers.pods.tls: true
      traefik.http.routers.pods.tls.certresolver: le
      traefik.http.routers.pods.entrypoints: https
      traefik.http.routers.pods.middlewares: authentik@file
      traefik.http.routers.podsNoAuth.rule: Host(`audio.example.com`) && HeadersRegexp(`Authorization`, `^Bearer .+$`)
      traefik.http.routers.podsNoAuth.tls: true
      traefik.http.routers.podsNoAuth.tls.certresolver: le
      traefik.http.routers.podsNoAuth.entrypoints: https
```

i can log into the web UI using proxy auth, and i can set my password. then, i open the mobile app and try to connect to the server using the password i just set. it... **almost** works? i get the following in the logs:

```
2023-10-07 02:42:59

INFO

Received ping

2023-10-07 02:43:19

WARN

[Auth] Failed login attempt 1 of 10 from 107.222.25.48

2023-10-07 02:43:28

INFO

[Auth] igor47 logged in from 107.222.25.48
```

the app connects, and i get a list of books, but then it gets stuck in a `Attempting server connection...` loop. something is not working with the web socket connection, maybe?

in my `authentik` config, i excluded the following paths:
```
/ping
/status
/login
```

igor47:
okay, so, the problem in the webapp at least is that the request to `/socket.io/` is made without an `Authorization` header, at least in the webapp and probably also in the mobile app. instead, we expect to connect to the web socket, and then send an `auth` event containing the token.

what does this mean? if i add `/socket.io/.*` to the non-authenticated paths, then i can successfully log in both via a browser (with username/password) and also in the mobile app. but lots of stuff is still broken. why?

ah -- it's because we often put the `token` into the query params and not into the `Authorization` header.  this is unfortunate and makes this much more challenging. is there a reason we use a query param and not a header for so many requests?

i tried changing my `traefik` matcher for the non-authenticated proxy to:

```
  traefik.http.routers.podsNoAuth.rule: Host(`audio.example.com`) && (HeadersRegexp(`Authorization`, `^Bearer .+$`) || Path(`/.+?\?.*?token=.+$`))
```

but that didn't fix it... maybe the `Path` matcher in traefik doesn't include query params? i can't tell.

i also tried adding an exclude path to `authentik` like:

```
.*?\?.*?token=.*
```

but again it's not doing anything. i saw some hint that [query params are not actually passed down to authentik from traefik using a forward auth middleware](https://community.traefik.io/t/cannot-pass-query-params-to-forword-auth-server-not-headers/15127/2).

so, i guess i'm stuck now. 

(HeadersRegexp(`Authorization`, `^Bearer .+$`) || Path(`/.+?\?.*?token=.+$`))

igor47:
ok, looking around, we are basically almost always generating URLs with `?token=` query params. i think this makes any auth system besides the built-in one difficult to implement. also -- the existing `token`s have no expiry, so if someone gets a hold of a `token` (like in a shared URL) then they can become that user.

i think a more sensible approach would be to migrate to session cookies, and remove all the `token=` . if we use a cookie, then there's no point in having an `Authorization` header. we would get to stop constantly getting `user/getToken` in every component, and we wouldn't need to mess with `localStorage` and the logout flow would be easier... in short, i think cookies are the way to go.

this would be a pretty big change, i would open a separate PR for that, and then re-base this change on top. before i begin that work, just wanted to get a sense from @advplyr if that would be welcome, i don't think i understand the current rationale behind the `Authorization` header/query-param auth system.

advplyr:
I'm currently focusing on getting this PR https://github.com/advplyr/audiobookshelf/pull/1636 ready to merge which is rewriting the auth system. I haven't looked into how this would work with forward auth yet. We will be using [passport.js](https://www.passportjs.org/) if you want to look into that

Sapd:
> you don't. we keep both the current authorization-token based auth, as well as the forward proxy auth. if forward proxy auth is disabled, there is no change to current behavior. to log into the mobile app when using forward proxy auth, do the following:
> 
> 1. log into the web app using forward proxy auth
> 2. set your password. the UI makes it clear that this password is only used for the mobile app login
> 3. use the password to log into the mobile app
> 
> in this case, the in-app passwords are more like access tokens. i had an idea for a future improvement where we display a QR code with a per-device auth token, and you can log into the app by scanning the QR code, this is not in scope right now but i think would be a pretty nice UX.

This mechanism does not really provide an advantage over oauth2 then. Actually your approach is similar to what oauth2 does (generate temporary access codes, which are however hidden to the user). The main benefit of forward auth (in my opinion) is that the whole application is shielded behind the reverse proxy's authentication mechanism (so one could not even now that you are actually hosting audiobookshelf when you would open the site - as you would only ever be redirected to your auth provider on any given path/request). 

The issue with mobile authentication can be actually solved. Im currently investigating the workaround Ive written at the bottom of the other PR. Basically open a browser window from the app, and then redirect back to `audiobookshelf://applogin?code=asdsdfs` or something like that. This way the browser can pass the temporary code (oauth) or a cookie (forward auth) back to the app. This will be probably also required for oauth2 on iOS and actually solves the problem for both oauth and forward auth.

I would suggest (however Im currently a non-contributer) to wait until the other PR is through and the mobile auth-problem on iOS is solved and then to rediscuss forward auth

Sapd:
The current SSO implementation with oauth2 was merged. I think that this PR with forward auth probably has to be completely redone or considered if it should be done at all.

In theory mobile app authentication with forward auth can be done by opening up a browser with the audiobookshelf URL and a special route. The user logs in and a cookie is set for the audiobookshelf domain, audiobookshelf can redirect back to the app and send the cookie as GET parameter with.

<details>
  <summary>Very detailed explanation how it works in a secure way</summary>

0. User types in server URL. The request is being redirected to a different hostname -> start following forward auth flow with the redirected URL
1. In the mobile app, generate a challenge
```
      # See also current mobile app implementation in components/connection/ServerConnectForm.vue
      const verifier = generateRandomString()
      const challenge = base64URLEncode(await sha256(verifier))
```
2. Open up a popup Browser, with the audiobookshelf URL and a special path
```
# Include the challenge from above
Browser.Open('https://audio.example.com/forwardauth?challenge=challenge')
```
3. The user will be redirected to `auth.example.com` where he logs in
4. The auth provider should now redirect back to `https://audio.example.com/forwardauth?challenge=challenge` if it is properly configured.
5. In the forward auth route (the user is now logged in inside the popup browser):
    1. Get the Cookie of - for example - Authentik. Authentik names it `authentik_proxy_RANDOMID=RANDOMSTRING` . Note that this cookie is inside the same domain as the audiobookshelf server (not authentik domain)
    2. Generate a temporary random token
    3. Add it to a global list, like `{ challenge: challengeFromGetParmaters, token: randomtoken, cookie: authentikcookie, expires: +2 minutes in future }`
    5. Clean up expired list entries
    6. Redirect to `audiobookshelf://forwardauth?token=temporarytoken`
7. Because it was redirected to an app link. The browser will now close
8. Do an API call to `https://audio.example.com/forwardauth/exchangetoken?token=randomtoken&verifier=verifier` <-- This is the only call which MUST be in the whitelist
9. If **both** `challenge==base64URLEncode(await sha256(verifierFromGetParameters))` **AND** token are correct, send back the cookie `authentik_proxy_RANDOMID`
10. Set the received cookie `authentik_proxy_RANDOMID` to be sent with subsequent requests

The user is now authenticated against the reverse proxy inside the app. A request to /login can be done to receive userinfo, which the backend can lookup by taking a look at the `username` header the user provides.

Using `verifier` is required, because otherwise a malicious app could get a hold of the cookie by also registering `audiobookshelf://`.
</details>

But it has multiple disadvantages:
* The code is complicated if this needs to be done in a secure way
* It is highly dependent on the forward auth provider like Authentik. Other providers name the cookie differently and could have requirements like a matching user-agent
* One API URL still needs to be in the whitelist

@igor47 approach has the advantage if that works, that it is simpler code but:
* He possibly need a big whitelist rule

As such both approaches negates the advantages of forward auth. I think **IF** someone wants to implement forward auth, it should be done simply without mobile app support.

However the question would be if it should be supported at all, given oauth2 is here. Because forward auth is meant as a compatibility layer for application which do not have support for oauth2. Additionally, forward auth is not standardised.

advplyr:
Thanks for the detailed breakdown @Sapd. I'm not sure forward auth is worth the effort since there is a more robust and secure SSO solution in place now. We'll see if it is the case that most users just wanted an SSO option and thought forward auth would be the easiest option in the short term.

PR #2176, opened 2023-10-03
Support rich text descriptions

MxMarx:
Alternative to  #2161 

This uses the trix editor that's currently used to include rich text for podcast episodes to enable rich text in book descriptions, and preserves descriptions grabbed from iTunes, Audible, and epubs in html. A `descriptionPlain` field is also returned by compatible providers so that the description preview when searching in match tab doesn't show html tags.
Additionally, the current description line in the match tab is truncated with css and shows the full description when hovering.

To preserve compatibility, html is stripped from the descriptions when embedding metadata in audio files.

The hover text for buttons doesn't seem to work if `getDefaultHTML` is provided as it is currently to remove unused toolbar buttons, so I also hardcoded the button labels.

Also, this is literally the first thing I've done with javascript so I'm probably not the right person to do this but I tried anyway, hopefully this doesn't cause too much annoyance!

resolves #617 
resolves #1820

PR #2198, opened 2023-10-09
Highlight updated metadata during matching

MxMarx:
This pull request makes it so that in the match tab, the values returned by the metadata provider are highlighted if they are different from the previous values. I wasn't sure what the best way to highlight things is, so I just went with green.

image

Also removed some [dynamic class names]( https://tailwindcss.com/docs/content-configuration#dynamic-class-names) to keep tailwind from missing things.

If #2176 goes through, `ui-textarea-with-label` can be replaced with `ui-rich-text-editor`.


jrpetersjr:
I think would be great to include in the next version of the app.

Issue #1789, opened 2023-05-24
[Enhancement]: Match New Authors feature


raviwarrier:
### Describe the feature/enhancement

In the authors' page, there is only one possible action: "Match All Authors". This action starts working on all authors, irrespective of whether they have info or not. See image below:
![image](https://github.com/advplyr/audiobookshelf/assets/42061703/5974c518-4647-4a05-bf89-6b8dd0574330)

In addition to "match all authors", also have "match new authors" and/or "match unmatched authors" (phrasing can be better) to work only on those authors that are newly added or those who don't have info matched. 

This should save processing time, audible limits and overall time for library updates.

catduckgnaf:
I agree, can it simply do a "match on new author" like when a new author is found, try a quickmatch?

advplyr:
What would the criteria be for a new or unmatched author? It doesn't have an ASIN? Image? Description?

raviwarrier:
> What would the criteria be for a new or unmatched author? It doesn't have an ASIN? Image? Description?

It would, I presume, be the same as what you do for "Match All Authors", but "added date > last matching done date", or "has description = false"

This depends on whether you do have a way of tracking 'last matching done date', or maintain an 'added date' field for authors.

the easiest way would be to see if the book associated with that author has been added recently, and if yes, check if the author record has details, and if not, then do a quick match. since you do maintain "added date" for the book.

If book new > if author details missing > quick match.

if the book is old, one can presume that the author has been updated: searched, matched and updated OR searched, not found, not updated - are the only two possible outcomes.

catduckgnaf:
> > What would the criteria be for a new or unmatched author? It doesn't have an ASIN? Image? Description?
> 
> It would, I presume, be the same as what you do for "Match All Authors", but "added date > last matching done date", or "has description = false"
> 
> This depends on whether you do have a way of tracking 'last matching done date', or maintain an 'added date' field for authors.
> 
> the easiest way would be to see if the book associated with that author has been added recently, and if yes, check if the author record has details, and if not, then do a quick match. since you do maintain "added date" for the book.
> 
> If book new > if author details missing > quick match.
> 
> if the book is old, one can presume that the author has been updated: searched, matched and updated OR searched, not found, not updated - are the only two possible outcomes.

I agree with this approach, and as far as details, I think description and/or image. I would avoid looking at ASIN, or at least exclusively. I have added many authors manually, and I may not add or have their ASIN.

Issue #1490, opened 2023-02-08
[Enhancement]: A setting to allow auto lookup for books and chapters [edit] and authors


raviwarrier:
### Describe the feature/enhancement

It would be nice to have a toggle-able setting that if set to 'on' does a scan of the books (and its chapters) in the background when the book is discovered by the library scanner.

In case of conflict or non-identification (couldn't find), it could show up on a new list (just like Issues show up), so that one can manually resolve those.

Also, the same can be done for every new author found in the library.

Use case:
Sometimes I add many books and do not get time to manually update its metadata and then I have to go through anywhere between 10-30 books. It is painful to do this because of the multiple clicks required per book (first on the book edit icon, then quick match - to get ASIN, then to change tab to 'Chapters', then to click 'Edit', then to click 'Lookup, then to click 'Save').

PR #2205, opened 2023-10-11
Automatically quickmatch new authors

MxMarx:
Automatically do a quick match when new authors are found by the scanner. 
It might be better if this was a toggleable option instead, but I'm still super new at this and wasn't sure if this was the best approach.

Somewhat related to #1789, #1490


advplyr:
I think this should be handled as a background task after the scan is complete. Mainly because if you are scanning in a new library it may be creating many new authors and come up against the rate limit https://github.com/advplyr/audiobookshelf/issues/1570

If we implement the rate limiter in #2188 it would slow down the scanner. The `TaskManager` was started to allow for notifying the client of background tasks. I was thinking of implementing an auto quick match of books setting that would run after the scan for new books.

PR #2240, opened 2023-10-23
add cypress tests

MxMarx:
I got bored enough to try learning how to write tests. I don't know how useful this is but I did it anyway!

`npm run test` will start the server with a fresh database, create some synthetic audiobooks to test with, and run the tests with coverage. There are currently five main tests:

- Use either the UI or API to initialize the server, add a library, and start a scan
- Log in with the UI
- Create and delete a user with the UI
- Check that the scanner added the test books and that their metadata is correct with the API
- Navigate to an audiobook, press play, and verify that audio is playing

To seed an audiobook library for testing, I made a script that takes json files from a few known library items, and uses ffmpeg and tone to create some small audiobook files with silent audio but real metadata and chapters.

Cypress seems to need a desktop to run, but adding ghcr.io/devcontainers/features/desktop-lite:1 to devcontainers seems to work.

It could probably still use some cleaning, but I guess I'd like to know if this approach is worth continuing first (for audiobookshelf in addition to personal growth) before I get too far into it.


advplyr:
We ended up going with Mocha since the implementation is more straightforward https://github.com/advplyr/audiobookshelf/pull/2300
I'm not sure what the effort would be to port some of these unit tests over to Mocha

PR #2245, opened 2023-10-23
Fix incorrect subpath checks in server/watcher.js

### LINKED ISSUE #2244: [Bug]: Sub-directory checks in watcher.js are incorrect
### Describe the issue

In server/watcher.js, there are two places in which the code checks if a path B is the same or a sub-path of another path A:
```
  // Get file folder
  const folder = libwatcher.folders.find(fold =>  path.startsWith(filePathToPOSIX(fold.fullPath)))
  if (!folder) {
    Logger.error(`[Watcher] New file folder not found in library "${libwatcher.name}" with path "${path}"`)
    return
  }

```
and:
```
checkShouldIgnorePath(path) {
  return !!this.ignoreDirs.find(dirpath => {
    return filePathToPOSIX(path).startsWith(dirpath)
  })
}
```

B.startsWith(A) is an incorrect check for sub-paths. For example, we might have A='/audiobooks' and B='/audiobooks2/book.m4b', so B.startsWith(A), but B is not a sub-path of A. This can of course lead to issues.

This [Stackoverflow question](https://stackoverflow.com/questions/37521893/determine-if-a-path-is-subdirectory-of-another-in-node-js) has a solution, which is still not watertight, but I think it's acceptable for this case.

1. In the same library, put two folders: '/audiobooks' and '/audiobooks2'.
2. Add a file to /audiobooks2.
3. The wrong folder will be selected in the first code snippet above, and consequently a wrong relPath will be calculated, likely leading to errors.

### Audiobookshelf version

2.4.4

Docker

mikiher:
Fixes #2244.
This fix applies [this solution](https://stackoverflow.com/questions/37521893/determine-if-a-path-is-subdirectory-of-another-in-node-js) to perform proper subpath checking.

### DIFF
--- a/server/Watcher.js
+++ b/server/Watcher.js
@@ -6,7 +6,7 @@ const LibraryScanner = require('./scanner/LibraryScanner')
 const Task = require('./objects/Task')
 const TaskManager = require('./managers/TaskManager')
 
-const { filePathToPOSIX } = require('./utils/fileUtils')
+const { filePathToPOSIX, isSameOrSubPath } = require('./utils/fileUtils')
 
 /**
  * @typedef PendingFileUpdate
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks! Works great

PR #2246, opened 2023-10-24
show a modal with cover images when clicked

### LINKED ISSUE #1853: [Enhancement]: Enlarged view of covers

Currently, when clicking on some point of a cover, other than the "Play" or "Edit" buttons, nothing happens. It would be nice to get an enlarged view of the cover instead. I'm thinking a pop-up, like you get with the "Edit" button, but just showing the cover.
Mockup:
image

MxMarx:
This makes it so that clicking on the cover in the item view or the player bar brings up a modal with an enlarged view of the cover so we can see all the juicy details of high-res covers.

Closes #1853


advplyr:
I moved the modal out to a separate component because it adds overhead to every BookCover component on the page if we put it inside. The cover overlay on the item page was also always showing so updated that, I don't think that was intentional.

Thanks!

PR #2253, opened 2023-10-26
Option to change the font family in epub viewer

MxMarx:
Adds an option to switch between serif and sans-serif fonts in the epub viewer.
Translations are machine translated.


advplyr:
Thanks!

PR #2255, opened 2023-10-27
Search epub text

### LINKED ISSUE #2045: [Enhancement]: search in ebooks

Most ebook readers I've seen have the ability to search for a particular phrase in the book and jump to it.  It's a very handy feature if you're trying to pull up a quote from a book, or if you want to synchronize your progress between devices that don't otherwise support it.

MxMarx:
Adds a search box in the table of contents sidebar to search for text within epubs.

Search results are displayed under the chapter they are found in, but at the moment, only the first two chapter levels are displayed in the ToC, so search results in chapters more than two levels deep won't appear. I could see about using a recursive component to display ToC entries instead, although I haven't found a book with more than three chapter levels yet.

This also makes adds a flattenChapters method, which converts ebook chapters into something that looks like audiobook chapters so that the seek bar will work with ebooks too (in a future pull request)

Closes #2045 

image

### DIFF
--- a/client/components/readers/EpubReader.vue
+++ b/client/components/readers/EpubReader.vue
@@ -401,6 +401,26 @@ export default {
         this.chapters = tocTree
       })
     },
+    flattenChapters(chapters) {
+      // Convert the nested epub chapters into something that looks like audiobook chapters for player-ui
+      const unwrap = (chapters) => {
+        return chapters.reduce((acc, chapter) => {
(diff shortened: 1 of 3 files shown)


advplyr:
This is great. Works well on all of my books. 

The only thing I noticed is that if the locations aren't loaded yet when first opening the search will only show the last chapter at the top. This was too minor for me to look into but worth noting.

Thanks for this feature!

Issue #2263, opened 2023-10-30
[Enhancement]: Old Apple Podcasts iOS app support


clement-dufour:
### Describe the feature/enhancement

I really like my old iPod to listen to podcasts. However, most of the podcast RSS feeds are not working on the old Apple Podcasts app (version 1.2.3) due to expired SSL certificates (breaking HTTPS). Audiobookshelf is a great app and appears to the perfect solution to regenerate a working RSS from downloaded podcasts.

Nevertheless once the RSS feed added to the app, nothing appears and the episodes list is empty. I suspected the blank spaces in the filename and URL to be the issue :
- [Apple](https://podcasters.apple.com/support/823-podcast-requirements) advises to use only ASCII filenames and URLs that include a-z, A-Z, or 0-9.
- I tried to manually add a file without any blank space in its filename and it appeared on the episode list.

Would it be possible to serve the files through an URL without any blank space in it? Maybe it could be opt-in during the RSS feed creation process.

I understand that this feature would only benefit to a small user base but would allow podcast and audiobooks playback on old devices.

Thanks for this wonderful project :+1:

DieselTech:
> ASCII filenames and URLs that include a-z, A-Z, or 0-9.

This seems like really outdated guidance from the 2008 internet era. Unicode has replaced ASCII everywhere that matters since it has wide support for different characters and languages. 

Problem is if you follows Apple's advice, it is highly likely it will cause problems with services that don't require ASCII and can use unicode characters. So that means in order to ABS to support something like this that it would have to be a toggle when opening the feed so that you can pick what kind of feed it will be. Because taking a filename like "you-👏-can-👏-put-👏-emojis-👏-in-👏-urls-👏" would stop working or you would have to handle a ton of different cases.

PR #2272, opened 2023-11-01
Add support for the old Apple Podcasts iOS app

clement-dufour:
Following my [enhancement request](https://github.com/advplyr/audiobookshelf/issues/2263) to add support for an old version (1.2.3) of the Apple Podcasts iOS app, I found that the RSS feed needed:
1. the artwork URL to include the file extension;
2. the audio file URL not to contain any non-ASCII character and to include the file extension.

With this patch:
1. The artwork URL on the RSS feed now includes the artwork file extension.
2. The content URL does not contain anymore the full filename (`/feed/${slug}/item/${episode.id}/${episode.audioFile.metadata.filename}`) but is now generic keeping a correct file extension (`/feed/${slug}/item/${episode.id}/media${contentFileExtension}`).

Feel free to comment or suggest any prettier code, I digged recently into this project.

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -155,7 +155,7 @@ class Server {
       Logger.info(`[Server] Requesting rss feed ${req.params.slug}`)
       this.rssFeedManager.getFeed(req, res)
     })
-    router.get('/feed/:slug/cover', (req, res) => {
+    router.get('/feed/:slug/cover*', (req, res) => {
       this.rssFeedManager.getFeedCover(req, res)
     })
     router.get('/feed/:slug/item/:episodeId/*', (req, res) => {
--- a/server/objects/Feed.js
+++ b/server/objects/Feed.js
@@ -1,3 +1,4 @@
+const Path = require('path')
 const uuidv4 = require("uuid").v4
 const FeedMeta = require('./FeedMeta')
 const FeedEpisode = require('./FeedEpisode')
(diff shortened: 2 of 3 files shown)


advplyr:
Tested on all feed entity types and it's working well. Thanks for handling your own use-case!

Altares:
I tried using the RSS feeds from ABS in my Apple podcasts app, although it imports fine, I cannot play the feeds nor download them. Any idea? Thanks

clement-dufour:
> I tried using the RSS feeds from ABS in my Apple podcasts app, although it imports fine, I cannot play the feeds nor download them. Any idea? Thanks

I've noticed that the base URL used when creating the RSS feed is used later as the base URL for the feed's media even if the RSS feed is later retrieved via another address.

For example, if I connect to my ABS server via the URL https://audiobookshelf.local/ and create an RSS feed for a podcast. The audio files will be indicated in the XML as accessible at https://audiobookshelf.local/item/<guid\>/media.mp3. If the feed is later retrieved via another URL or simply via HTTP rather than HTTPS, the path to the files remains as https://audiobookshelf.local/item/<guid\>/media.mp3 in the feed XML.

For this reason, when creating RSS feeds on my system, I always open the ABS web page via the target URL through which I will then consume the media. For my iPod Touch 4, which is no longer able to use recent SSL certificates, I create RSS feeds by opening the page in HTTP only so that I can read the media through HTTP.

Do you think this could be the reason why you cannot play or download feeds?

Altares:
> > I tried using the RSS feeds from ABS in my Apple podcasts app, although it imports fine, I cannot play the feeds nor download them. Any idea? Thanks
> 
> I've noticed that the base URL used when creating the RSS feed is used later as the base URL for the feed's media even if the RSS feed is later retrieved via another address.
> 
> For example, if I connect to my ABS server via the URL https://audiobookshelf.local/ and create an RSS feed for a podcast. The audio files will be indicated in the XML as accessible at https://audiobookshelf.local/item/<guid>/media.mp3. If the feed is later retrieved via another URL or simply via HTTP rather than HTTPS, the path to the files remains as https://audiobookshelf.local/item/<guid>/media.mp3 in the feed XML.
> 
> For this reason, when creating RSS feeds on my system, I always open the ABS web page via the target URL through which I will then consume the media. For my iPod Touch 4, which is no longer able to use recent SSL certificates, I create RSS feeds by opening the page in HTTP only so that I can read the media through HTTP.
> 
> Do you think this could be the reason why you cannot play or download feeds?

That was it thank you! I accessed the HTTP version with raw IP then shared the feed with the app and it worked.

PR #2274, opened 2023-11-01
Create cs-CZ.json

radekmuhlfeit2:
Czech strings.


advplyr:
Thanks! Do we need to differentiate between `cs` and `cs-CZ`?

radekmuhlfeit2:
> Thanks! Do we need to differentiate between `cs` and `cs-CZ`?

Well, probably not. It is strings for the Czech language and just didn't know your naming convention for that, hence I used the code from here: https://learn.microsoft.com/cs-cz/cpp/c-runtime-library/country-region-strings?view=msvc-170

advplyr:
There is no naming convention right now, still figuring it out

PR #2288, opened 2023-11-05
Added swedish translation of strings

ScuttleSE:
Well, strings translated to swedish :)


advplyr:
ISO 639-1 has the language code as `sv` for Swedish so I will update it to that

advplyr:
Thanks!

PR #2291, opened 2023-11-06
Hide collection duration if 0

### LINKED ISSUE #2278: [Bug]: Collection view shows ebook duration in seconds
### Describe the issue

The collection view shows a duration in seconds for ebooks which doesn't really make sense as it's always just 0, since ebooks don't have a duration.
It would be better to show the read percentage (or no value at all).

image

1. Add ebook to collection
2. View collection

### Audiobookshelf version

v2.5.0

Docker

brianjaustin:
This fixes #2278 by hiding the duration of a book in a collection if it is 0 (or less). I suppose some audiobooks could also have a duration of 0, but this seems rare (and perhaps hiding the duration in those cases also makes sense?). I opted not to go with progress as that would result in the value having different meanings depending on the media type, which sounds like a potentially confusing user experience.

While this fixes the reported bug, I wonder if it makes sense to have page count as the "duration". Although that would require more effort to implement I think.

### DIFF
--- a/client/components/tables/collection/BookTableRow.vue
+++ b/client/components/tables/collection/BookTableRow.vue
@@ -30,7 +30,7 @@
               ><span :key="author.id + '-comma'" v-if="index < bookAuthors.length - 1">,&nbsp;</span>
             </template>
           </div>
-          <p class="text-xs md:text-sm text-gray-400" v-if="media.duration > 0">{{ bookDuration }}</p>
+          <p v-if="media.duration" class="text-xs md:text-sm text-gray-400">{{ bookDuration }}</p>
         </div>
       </div>
     </div>
--- a/client/components/tables/collection/BookTableRow.vue
+++ b/client/components/tables/collection/BookTableRow.vue
@@ -30,7 +30,7 @@
               ><span :key="author.id + '-comma'" v-if="index < bookAuthors.length - 1">,&nbsp;</span>
             </template>
           </div>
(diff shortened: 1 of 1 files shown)


nichwall:
The only problem I see with using "pages" as duration is that only works for PDF or comics (epub page count is relative based on zoom and screen size).

I can't think of any reason to not hide duration when it's zero. That indicates there are either no audio items, or all of the audio items are zero duration. When people make placeholder books, they tend to use empty files, so there still wouldn't be a duration.

advplyr:
Yeah as @nichwall said, we can't get number of pages for epub. I was thinking of showing the ebook format in-place of the duration

client/components/tables/collection/BookTableRow.vue
```diff
@@ -30,7 +30,7 @@
               ><span :key="author.id + '-comma'" v-if="index < bookAuthors.length - 1">,&nbsp;</span>
             </template>
           </div>
-          <p class="text-xs md:text-sm text-gray-400">{{ bookDuration }}</p>
+          <p class="text-xs md:text-sm text-gray-400" v-if="media.duration > 0">{{ bookDuration }}</p>
```
advplyr, line 33:
```suggestion
          <p v-if="media.duration" class="text-xs md:text-sm text-gray-400">{{ bookDuration }}</p>
```

advplyr:
Thanks!

PR #2297, opened 2023-11-09
Wrong port EXPOSEd in Dockerfile / Wrong port specified in Documentation

Cloufish:
The Exposed Port should be the one that is specified in the documentation.

Port 80 causes routing issues.
I got "Bad Gateway, Connection Refused" when I try to connect to the webpage via Traefik
image
image

**Traefik logs**:

image

**audiobookshelf Docker logs**:
image

Let's see it in another example - anki sync server

image
It's a correct one, because the docker-compose also points to this port:
image

**If it's impossible to change this port in Dockerfile, the docker-compose.yml example should contain 80 port**

image
```
docker run -d \
  -p **80**:80 \
  -v </path/to/podcasts>:/podcasts \
  --name audiobookshelf \
  ghcr.io/advplyr/audiobookshelf
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,6 @@ RUN npm ci --only=production
 
 RUN apk del make python3 g++
 
-EXPOSE 80
+EXPOSE 13378
 
 CMD ["node", "index.js"]


advplyr:
I'm not sure that I understand. The `EXPOSE` in the Dockerfile is documenting that the server running in the docker container is using port 80. It shouldn't have any impact on what port you decide to publish when running the container.

https://blog.ipspace.net/kb/DockerSvc/10-Exposed.html

I may be missing something, let me know what you think

jrpetersjr:
@Cloufish - If you want to run ABS on port 80 then you would need your run command to look like this and you don't need to change the docker file.

docker run -d \
  -p 80:80 \
  -v </path/to/podcasts>:/podcasts \
  --name audiobookshelf \
  ghcr.io/advplyr/audiobookshelf
  
  Then you can point Traefik to port 80 and get it working. Otherwise you can just run the given example with the ports and point Traefik to 13378.

Cloufish:
> @Cloufish - If you want to run ABS on port 80 then you would need your run command to look like this and you don't need to change the docker file.
> 
> docker run -d -p 80:80 -v </path/to/config>:/config -v </path/to/metadata>:/metadata -v </path/to/audiobooks>:/audiobooks -v </path/to/podcasts>:/podcasts --name audiobookshelf ghcr.io/advplyr/audiobookshelf
> 
> Then you can point Traefik to port 80 and get it working. Otherwise you can just run the given example with the ports and point Traefik to 13378.

I you run 
```
docker run -d
-p 80 <= Without external port specified
-v </path/to/config>:/config
-v </path/to/metadata>:/metadata
-v </path/to/audiobooks>:/audiobooks
-v </path/to/podcasts>:/podcasts
--name audiobookshelf
ghcr.io/advplyr/audiobookshelf
```
It will not work.

But now I also see that it's not a Bug in Documentation,
Some people doesn't specify external port in their `docker-compose.yml` file, like me:

```yml
version: "3.5"

services:
  audiobookshelf:
    image: ghcr.io/advplyr/audiobookshelf:latest
    container_name: audiobookshelf
    ports:
      - "80" <====
    networks:
      - web
      - audiobookshelf
    volumes:
      - audiobookshelf-volume:/config
      - /media/Immersion/German/Audiobooks/metadata:/metadata
      - /media/Immersion/German/Podcasts:/podcasts
      - /media/Immersion/German/Audiobooks:/audiobooks
    environment:
      - PUID=1000
      - PGID=1000
      - TZ=Europe/Warsaw
    deploy:
      labels:
        - "traefik.enable=true"
        - "traefik.http.routers.audiobookshelf.rule=Host(`audiobookshelf.home`)"
        - "traefik.http.services.audiobookshelf.loadbalancer.server.port=80" <==
        - "traefik.docker.network=web"
        - "io.portainer.accesscontrol.users=admin"
        - "traefik.http.routers.audiobookshelf.tls=true"

networks:
  audiobookshelf:
    driver: overlay
    attachable: true
    name: audiobookshelf
  web:
    external: true
    name: web

```

If in Dockerfile right port was specified, then people wouldn't need to redirect local port to container's port

jrpetersjr:
@Cloufish - This is how tons of docker containers work and you are fighting an uphill battle that you will not win. You get to expose a port number of your choosing for the docker container. It allows the application in the container to run on any port the developer wants and then you say what port actually accesses it. You could have 60 containers all with port 80 in the container, but then you get to specify and don't have to worry about the containers clashing on ports. 


What other containers are you running where your compose works like that?

Cloufish:
> @Cloufish - This is how tons of docker containers work and you are fighting an uphill battle that you will not win. You get to expose a port number of your choosing for the docker container. It allows the application in the container to run on any port the developer wants and then you say what port actually accesses it. You could have 60 containers all with port 80 in the container, but then you get to specify and don't have to worry about the containers clashing on ports.
> 
> What other containers are you running where your compose works like that?

If I understand it correctly
**With Traefik reverse proxy**
Traefik accepts anything on port 80, and then It forwards it based on the routing rule that you specify.
e.g.:
`  - "traefik.http.routers.audiobookshelf.rule=Host(`audiobookshelf.home`)" `
to the port that the container service uses, e.g.:
`        - "traefik.http.services.audiobookshelf.loadbalancer.server.port=80"`

The main problem is that the documentation can be misunderstood, because It tells that you should bind port 13378:80
Why can't internal port of the container be the same? so 13378?

jrpetersjr:
So Traefik is set up on port 80 then you can't have another container setup listening to port 80 as well. That's the whole point of mapping the ports. You expose port 80 in the container and you can make them to whatever you want. Your docker compose could be 
ports:
      - 81:80

and then
 - "traefik.http.services.audiobookshelf.loadbalancer.server.port=81"

but no matter what if you have 2 applications on the same port it's not going to work.

jrpetersjr:
@Cloufish - In your Anki container you could have done the ports like this:

ports:
 - 12345:27701
 and then point Traefik to port 12345 for Anki Sync instead of 27701 and it would work. 
 
What the above means is "I want my computers port 12345 to point to the containers exposed port of 27701. It just so happens that the maintainer of the Anki container wanted to expose port 27701 instead of 80 or 85 or whatever port. I think you are running into the problem that you have Traefik exposed on port 80 outside of the container and you are also trying to expose port 80 for ABS. That's what docker port mappings are good. Let the maintainer do what they want and also let you do what you want/need so that you don't get a bunch of conflicts. Imagine that you are running a bunch of containers that all of their dockerfiles have EXPOSE 8080 in them...you wouldn't go to every repo and put in a pull request to change the port instead you would do port mappings for each container....for example:

//For Mealie.io
ports:
 - 91:8080
 
//For Paperless-ngx
ports:
 - 92: 8080
 
//For Immich
ports:
 - 93: 8080
 
 
 This are not true to life examples, but they are examples of what it would look like all of those applications docker files had EXPOSE 8080 in them and you needed to get them all working and then with Traefik you would point them individually to the port on the left side.

advplyr:
Abs also exposes a way for you to change the internal port if you really needed to for some reason.

```yaml
version: "3.5"

services:
  audiobookshelf:
    image: ghcr.io/advplyr/audiobookshelf:latest
    container_name: audiobookshelf
    ports:
      - 13378:13378
    volumes:
      - audiobookshelf-volume:/config
      - /media/Immersion/German/Audiobooks/metadata:/metadata
      - /media/Immersion/German/Podcasts:/podcasts
      - /media/Immersion/German/Audiobooks:/audiobooks
    environment:
      - PORT=13378
      - TZ=Europe/Warsaw
```

This should work because the environment variable `PORT` is what will be used when Abs starts the server (I haven't tested this).

I can see an argument to set the internal port to match the recommended external port but since we have already been using port 80 internally for a long time changing it now would force everyone to update their docker compose.

PR #2299, opened 2023-11-09
Update it.json

burghy86:
arrange the additional lines.
how the hell did we get to over 700 lines in less than two months?


advplyr:
Lots of development! Thanks!

PR #2305, opened 2023-11-12
Add NFO metadata source

mikiher:
In many cases, audiobooks come with .nfo files. While there's no standard to this text format, most audiobook .nfo files follow some de-facto standards specifying metadata (e.g. "Title", "Author", "Read by"). In some cases the .nfo files have metadata attributes that are missing/wrong in the folder structure or the audio meta tags.

This adds NFO as an additional metadata source, parsing the .nfo files according to the de-facto standards.

I put the NFO default priority just above audio meta tags, which I find reasonable, but this could be of course changed if needed.

I also have unit test for the parser. If you approve PR #2300, I'll commit that as well.


mikiher:
Unit test now committed.

advplyr:
Can you share one or more sample NFO files so I can test and that we can add to the guide? I don't remember seeing an NFO with an audiobook before

mikiher:
Here are a few examples: [nfo examples.zip](https://github.com/advplyr/audiobookshelf/files/13465362/nfo.examples.zip)
Let's just make sure we sanitize them before adding to the guide.

advplyr:
Since the description at the top of the scanner settings was incorrect now that there are 6 metadata sources I refactored that settings tab.

Previously it was not intuitive because 1 was lowest priority. For the UI only I updated it so that 1 is the highest priority.

image

advplyr:
Thanks!

mikiher:
> I thought it made more sense to trim the metadata description after it gets parsed. For example, one of the sample NFO files had 2 empty lines in the start and end of the description. This didn't seem necessary to include but let me know if I'm missing some edge case there.

Sounds good.

> Another thing to note with the description is that it has to be placed last or the rest of the NFO file will be included in the description. This looked intentional on your part and it works for me but I just wanted to highlight that. I don't see a better way of handling it.

Indeed, it was intentional.

Thanks for the thorough review!

PR #2311, opened 2023-11-15
Update Patch Dependencies

CaseyInHaengsin:
Trying to get familiar with the code and wanted to start with dependency updates.


advplyr:
Thanks. I prefer to do these in increments when we can test everything and see what we are updating. We will probably be updating to Node 20 soon so that'll be a good time for that. If you have a specific package you think needs updating now let me know!

PR #2325, opened 2023-11-19
Add milliseconds to logging

lkiesow:
This patch adds milliseconds to the time string used for logging. This helps when debugging some timing issues and should have no real negative side effect.


advplyr:
Works for me. Thanks!

PR #2326, opened 2023-11-19
Allow enabling dev logs

lkiesow:
This patch allows users to enable dev logs on production systems by setting the `HIDE_DEV_LOGS` environment variable.

Before, you could only use this on a non-production environment. On production, the logs would be disabled. This patch changes the behavior and uses the `NODE_ENV` only as default. On production they are disabled if `HIDE_DEV_LOGS` is undefined but can be enabled by setting `HIDE_DEV_LOGS=0` on dev, they are enabled if undefined, but can be disabled by setting `HIDE_DEV_LOGS=1`.


advplyr:
Works for me, thanks!

PR #2333, opened 2023-11-20
Add ability to fetch book data on upload

kieraneglin:
**IMPORTANT NOTE**

This PR is work-in-progress, mainly when it comes to the finishing touches in the UI. I will likely eventually need some UI direction, but for now the focus is on the general concept's viability.

**Issue:**

When importing media via the upload page, it can be tedious to import lots of media if you also want to ensure the author/series name is correct. This has the most impact to the structure of files stored on-disk - keeping this data correct will namespace media under the desired author/series directories. Keeps things clean!

A secondary issue comes from the details of media not _actually_ being what you expect or may casually call it. For example, people (me) may call [this series](https://en.wikipedia.org/wiki/The_Kingkiller_Chronicle) `The Kingkiller Chronicles`, according to Audible it's just `Kingkiller Chronicle`. Again, this mostly impacts keeping the directory structure of things well-formed and consistent.

**Fix Description**

I've added some UI to allow for automatic fetching of media data based on what data you have entered in so far (title is required). This fires on-selection of media as long as the title gets auto-populated! If it returns the wrong information, you can always edit it to your preference OR give it a little hint by updating, say, the author and fire a re-fetch manually.

As mentioned, this also provides a way to manually fetch details for a single piece of media. This is helpful for re-fetching media if-needed, but it's also good for when the uploaded filename is VERY different than the actual title of the media.

NOTE: this only works on audiobooks.

**Additional Fixes:**

- Fixed `directory` input being smaller and misaligned compared to the `series` input on upload page
- Added `.trim` modifier to upload form inputs to chop leading/trailing whitespace
- Updated error string in the `ItemUploadCard`'s `getData` method to use an I18n string rather than hardcoding

**Regressions:**

None known.

**GitHub Issues Resolved:**

None known.

**Screen Captures**

https://github.com/advplyr/audiobookshelf/assets/569917/68e9f6c4-8741-4cbd-88b4-930fc13d117a


lukeIam:
@kieraneglin I was looking for some feature like that - would love such a feature.
Thanks for you work so far!

client/mixins/apiRequestHelpers.js
```diff
@@ -0,0 +1,12 @@
+export default {
```
kieraneglin, line 1: I'm not married to this mixin existing, but I thought it could provide some utility beyond my changes. If you don't like it, just say the word and I'll nuke it + refactor 🤙

client/pages/upload/index.vue
```diff
@@ -61,9 +75,16 @@
       </widgets-alert>
 
       <!-- Item Upload cards -->
-      <template v-for="item in items">
-        <cards-item-upload-card :ref="`itemCard-${item.index}`" :key="item.index" :media-type="selectedLibraryMediaType" :item="item" :processing="processing" @remove="removeItem(item)" />
-      </template>
+      <cards-item-upload-card
```
kieraneglin, line 78: For viz: minor refactor here to put the `:key` on the iterator. But you can't use a `template` element once you do that, so I removed it. Should be effectively no change

client/pages/upload/index.vue
```diff
@@ -213,27 +250,49 @@ export default {
...
+      }
+    },
+    onItemsSelected(itemResults) {
+      if (this.itemSelectionSuccessful(itemResults)) {
+        // setTimeout ensures the new item ref is attached before this method is called
+        setTimeout(this.attemptMetadataFetch, 0)
       }
     },
-    setResults(itemResults) {
+    itemSelectionSuccessful(itemResults) {
```
kieraneglin, line 269: For viz: renamed this to better match intention

client/components/cards/ItemUploadCard.vue
```diff
@@ -105,9 +128,42 @@ export default {
...
+          title: this.itemData.title,
+          author: this.itemData.author,
+          provider: this.provider
+        })
+        const [bestCandidate, ..._rest] = await this.$axios.$get(`/api/search/books?${searchQueryString}`)
+
+        if (bestCandidate) {
+          this.itemData = {
+            ...this.itemData,
+            title: bestCandidate?.title,
```
kieraneglin, line 150: Just realized I probably don't need the safe navigation operators now that this has been moved into that `if` block

client/components/cards/ItemUploadCard.vue
```diff
@@ -105,9 +128,42 @@ export default {
...
+
+      this.isFetchingMetadata = true
+      this.error = ''
+
+      try {
+        const searchQueryString = this.buildQuerystring({
+          title: this.itemData.title,
+          author: this.itemData.author,
+          provider: this.provider
+        })
```
advplyr, line 144:
```suggestion
        const searchQueryString = new URLSearchParams({
          title: this.itemData.title,
          author: this.itemData.author,
          provider: this.provider
        })
```

I don't think we need the mixin since the same thing can be achieved with URLSearchParams. It encodes the strings properly. I may be missing something though, let me know what you think.

kieraneglin:
@advplyr great to hear! Regarding that bug, I don't necessarily agree that sanitizing file names should be a client-side concern for a few reasons:

1. In general, any important validation should be done server side since you can always circumvent client-side validation
2. The client isn't necessarily aware of what operating system the server is running on (nor should it be IMO), which ties into:
3. Reserved characters, control characters, and reserved file names vary per-os. I believe`:` is totally fair game on UNIX, and you can't name a folder `CON` on Windows, for instance. Given this, it might be best to delegate sanitization to an existing package. Maybe the list is small enough to replicate the functionality in-app, but in any case this really feels like a concern of the server to me. 

Please let me know what you think!

client/components/cards/ItemUploadCard.vue
kieraneglin: totally forgot about that guy! thanks for the tip 🤙

kieraneglin:
Ah, I see the issue - I forgot that the tempfile would also break things so filenames would need to be sanitized before they're written as a tempfile. I'll see what options are available server-side since I don't think it's ideal to be able to crash the server with a well-crafted filename, but as a last resort I'll update it to do client-side sanitizing

advplyr:
I definitely agree this needs to be handled server side. I thought it may be useful to also check for this client side to better inform the user, but as you said it would be difficult for the client to know the proper way to sanitize.
We are sanitizing file paths for podcasts using the Windows restrictions. When I set that up I dug into it a bit and it seemed better to enforce the strictest since podcast episodes are auto-downloaded and the filepath is not a user input. The function for sanitizing the filename is https://github.com/advplyr/audiobookshelf/blob/f59516cc6eaaab792c95b9c991820e3cc057886f/server/utils/fileUtils.js#L284

We could force sanitize it that way server side or we could just catch the exception and return the error. The error will not happen until the upload completes which can be slow for some users.

advplyr:
Oh yeah the sanitize filename function also exists client side https://github.com/advplyr/audiobookshelf/blob/master/client/plugins/init.client.js#L53

client/components/cards/ItemUploadCard.vue
```diff
@@ -87,12 +98,19 @@ export default {
...
       if (this.isPodcast) return this.itemData.title
 
-      if (this.itemData.series && this.itemData.author) {
-        return Path.join(this.itemData.author, this.itemData.series, this.itemData.title)
-      } else if (this.itemData.author) {
-        return Path.join(this.itemData.author, this.itemData.title)
-      } else {
-        return this.itemData.title
+      const outputPathParts = [this.itemData.author, this.itemData.series, this.itemData.title]
+      const cleanedOutputPathParts = outputPathParts.filter(Boolean).map(part => this.$sanitizeFilename(part))
```
kieraneglin, line 102: Sanitizes the output filepath so the preview will show the actual final path - it does NOT try to modify the contents of the inputs because that can be awful to manage if the user is currently working in that input

client/plugins/init.client.js
```diff
@@ -77,6 +77,7 @@ Vue.prototype.$sanitizeFilename = (filename, colonReplacement = ' - ') => {
     .replace(lineBreaks, replacement)
     .replace(windowsReservedRe, replacement)
     .replace(windowsTrailingRe, replacement)
+    .replace(/\s+/g, ' ') // Replace consecutive spaces with a single space
```
kieraneglin, line 80: I noticed that titles like `foo: bar` were being turned into `foo -  bar` (with two spaces between `-` and `bar`). I figured that this is undesirable in all cases so this turns multiple spaces into one

server/controllers/MiscController.js
```diff
@@ -52,43 +50,29 @@ class MiscController {
...
-      Logger.error(`[Server] Upload directory "${outputDirectory}" already exists`)
-      return res.status(500).send(`Directory "${outputDirectory}" already exists`)
-    }
+    // Podcasts should only be one folder deep
+    const outputDirectoryParts = library.isPodcast ? [title] : [author, series, title]
+    // `.filter(Boolean)` to strip out all the potentially missing details (eg: `author`)
+    // before sanitizing all the directory parts to remove illegal chars and finally prepending
+    // the base folder path
+    const cleanedOutputDirectoryParts = outputDirectoryParts.filter(Boolean).map(part => sanitizeFilename(part))
+    const outputDirectory = Path.join(...[folder.fullPath, ...cleanedOutputDirectoryParts])
```
kieraneglin, line 59: Did a bit of a refactor here to slim things down while keeping things readable (imo). This should be functionally identical to the old code, save for my sanitization changes

server/controllers/MiscController.js
```diff
@@ -52,43 +50,29 @@ class MiscController {
...
-        return false
-      })
+      await file.mv(path)
+        .then(() => {
+          return true
+        })
+        .catch((error) => {
+          Logger.error('Failed to move file', path, error)
+          return false
+        })
```
kieraneglin, line 75: okay, I cannot figure this out and I spent way too much time trying to reason around it. `.mv` returns a promise, so why can't I `await` it alone? Why does it need to be chained to another promise? Why can't I use `try`/`catch` here? I dug into the code of `express-fileupload` and the way this promise is called is so contrived that I truly couldn't understand it.  I could do something like this:

```js
await new Promise((resolve, reject) => {
  file.mv(path, (err) => {
    if (err) reject(err)
    resolve()
  })
})
```

But that seems _very_ overkill just to be able to interact with it like a normal promise? I am _very_ tired so maybe I'm missing something obvious, but for now this stays as it was

server/controllers/MiscController.js
advplyr: Ah I didn't know about `filter(Boolean)`, that is a nice shortcut!

advplyr:
Awesome, thanks!

PR #2336, opened 2023-11-21
de language translation follow up

JBlond:
de language translation follow up for 27497451d9847fc57b7e67b8676f35532e299b2d


advplyr:
Thanks!

PR #2365, opened 2023-11-28
SSO/OpenID: Provide error messages to logs

Sapd:


advplyr:
Tested and working well, thanks!

PR #2302, opened 2023-11-10
Preparing for GitLocalize to manage translations 

nichwall:
In order to prepare for using GitLocalize for managing server translations, the following steps need to take place:

- [x] Rename the strings from `%lang%.json` to `%lang%/strings.json` because GitLocalize does not like a flat file structure (without adding the entire repository
- [x] Delete strings which have not been translated yet (matches English version)
- [x] Change `supplant` to fallback to English if the translation does not exist and use the new naming scheme
- [x] Write script to copy translations between repositories
- [ ] Set up GitLocalize

If GitLocalize is decided on, the following steps should be taken:
- Link repository to GitLocalize
- BEFORE adding languages, set the project path in the Settings (it's a pain to get GitLocalize to scan the repository properly if you add languages before fixing the path)
image
- Add languages (`Overview` tab)
- ~~Add "Translation FAQ" to the website and "How to Contribute" on this GH readme.~~ Maybe make an announcement on Discord too. I was thinking of making a thread for people to coordinate since there will probably be multiple people who want to work on per language

Once users who want to help link their GH account to GitLocalize, they should be added as a Moderator (https://docs.gitlocalize.com/understanding_roles.html) so they can open PRs to GH. If a user does not want to use GitLocalize, they can still open PRs like normal, but they will not be able to just look at the `%lang%/strings.json` file to find English lines because they will not exist until they have been translated (since some languages will use the same words as English)

Once this first pass of updating the missing translations for the server is finished, the translations will be copied to the app repository using the included script to help keep the two repositories in sync. ~~will diverge since there is not a lot of overlap beyond the basic book, podcast, and ABS strings, and the UI is updated separately.~~

PR #2372, opened 2023-11-30
Localization Try2

nichwall:
This is a second attempt at getting ready for localization (original was #2302) In order to simplify the merge, a few different steps were taken.

The resource files were moved using `git mv` and were not modified beyond that. This does not remove the English duplicates like I want, but will allow for merging without breaking other PRs by having merge conflicts due to moved files.

I also updated `i18n.js` to use the new resource paths, but did not change the English fallback behavior.

I added the script explained in the original PR to help keep the server and app repository resource files in sync.

EDIT: removed GitLocalize instructions since it was decided against 

<details>
  <summary>Translation workflow</summary>

  The workflow for translations is as follows for both repositories since the sync happens both ways:
1. When adding new strings to the project, only add them to the English resource file.
2. Run the included script to sync existing translations from the other repository (will pull from other repository if translation already exists)
3. Commit and push both repositories as needed
4. Untranslated/updated strings show up in GitLocalize to be translated.
5. Translation PRs from contributors are merged as usual

The script can also be run after translation PRs are merged to ensure that both repositories get new translations of shared strings.

</details>


nichwall:
Note: the included Python script should not be run until the app strings repository is changed to be the same structure for GitLocalize, I am just providing it so all of the initial parts for the transition are included.

PR #2391, opened 2023-12-05
Add a binary manager that finds ffmpeg and ffprobe and installs them if not found

mikiher:
In continuation of our discussion #2352, I created a binary manager that makes sure the ffmpeg/ffprobe binary dependencies exist, and if they don't it installs them. 

A few implementation details:
- The binaries are not bundled with audiobookshelf (and they're also not bundled as pkg assets). Instead I've added a dependency on the ffbinaries package, which can programmatically download and install them if needed. This is good since:
  - it doesn't increase the pkg executable size (which is already quite large, ~54MB for Win64)
  - it is probably better from a legal perspective because of GPL implications.
- The binary manager tries to find the binaries in a number of ways, and only installs if it cannot find it. It tries to:
  - Check the FFMPEG_PATH/FFPROBE_PATH environment variables, if defined.
  - Use which to check if it's somewhere in PATH
  - Checks the main install path (see below)
  - Checks the alt install path (see below)
- If the binary manager does not find the binaries, it tries to install them in one of two paths:
  - the main install path, which is the same directory in which the audiobookshelf binary is installed, if it's pkged, or global.appRoot, if it's not. 
  - in case the main install path is not writable (due to permissions), an alternative install path is used, which is the audiobookshelf config directory (which should always exist and be writable).
- The binary manager is instantiated and initialized with the other managers, in server.js

This should work on all platforms, however I could only test on Win64 (I tested both pkged and non-pkged).
If you don't want to use this on non-windows platforms, we can restrict it to windows only, but I think we should really just make sure it it works on the other platforms.

So with this, we can drop the requirement to install ffmpeg in advance or bundle it with installers. For windows, people just need to download the binary and put it in whatever directory they want.


advplyr:
I didn't think about the server handling this but it makes sense. Most of the install methods are managed by other people but if they are setting FFMPEG_PATH and FFPROBE_PATH correctly then it should be fine. 

The debian package also sets those env variables so that would work without updates also. If this works well the debian package could be updated to handle it this way instead of downloading ffmpeg in the preinstall step.

package.json
```diff
@@ -40,6 +40,7 @@
     "cookie-parser": "^1.4.6",
     "express": "^4.17.1",
     "express-session": "^1.17.3",
+    "ffbinaries": "^1.1.5",
```
advplyr, line 43: I don't think we should include this dependency because it comes with something like 100 more dependencies. Most of them are coming from [request](https://www.npmjs.com/package/request) which is deprecated anyway.

I haven't gone through it yet but looking at the library we should be able to set this up ourselves using our current libraries https://github.com/ffbinaries/ffbinaries-node/blob/master/ffbinaries-lib.js

mikiher: Yeah, my intention was to only include ffbinaries as an intermediate step. 

I was intending to do a rewrite of ffbinaries after this. I also noticed the dependency on request . The library is also written using callbacks rather than promises, and is full of synchronous function calls (although it's probably OK since it's a startup code). I think that I saw some fork that replaces request with axios, but I haven't looked closely yet, since it was convenient to use ffbinaries as-is as a starting point.

Are you ok with including it as an interim step? I think it would take another week or two to rewrite it for our purposes with no/very few additional dependencies. That's definitely next on my todo list.

advplyr: I made a quick run at it but ran out of time today to polish it up more. It should be working though.

I did not implement the progress ticker for the download but shouldn't be difficult if we want to implement that.

I removed `locateBinariesSync` function since it wasn't being used.

`downloadBinaries`, `getVersionData` and `listVersions` are all promises now.

mikiher: Very nice (and quick!).

advplyr:
I happened to test this on a ubuntu server that did not have a home dir so the LOCAL_CACHE_DIR mkdir call crashes.

This got me thinking that we shouldn't even need a cache dir. Even if we do need it we shouldn't be creating it unless it is necessary.

Is there a reason you can think of to have the cache dir instead of downloading to the directory and removing the zip file after extraction?

mikiher:
Hi, I just submitted a refactoring of downloadUrls(), which:
- Made it an async function
- Made the function passed to async.each to be an async function (which makes all the callbacks redundant)
- Refactored spaghetti code and tries within tries, making the code a lot more readable and linear (using await where it makes sense).
- Fixed tickerFn, which now works.

From my perspective, it is now ready for merge (it's probably worth doing another pass on the whole file, but I'm happy with the important stuff).

mikiher:
> I happened to test this on a ubuntu server that did not have a home dir so the LOCAL_CACHE_DIR mkdir call crashes.
> 

I see. I have not encountered this issue on my Windows testing.

> This got me thinking that we shouldn't even need a cache dir. Even if we do need it we shouldn't be creating it unless it is necessary.
> 
> Is there a reason you can think of to have the cache dir instead of downloading to the directory and removing the zip file after extraction?

Since we only expect to download once for every installation of ABS, I don't think that cache matters at all. I'll remove it.

mikiher:
OK, I removed the local cache. The zip file is now downloaded to the destination dir, and removed after binary is extracted.

advplyr:
Awesome, thanks. This is working well so far but I want to do some more testing. Is this a blocker right now on developing the windows install?

advplyr:
I don't think it's worth releasing the exe publicly like that. 

Most of the users who are looking for the Windows install are not as familiar with tech and will be confused by the exe. We will get github issues, emails and posts in Discord about the exe. Some of them will understand and we could provide the exe for them but if we push this as-is it will create a lot of confusion.

The server currently requires Tone for embedding metadata and making m4b files so that functionality would not work as-is. I have to remove Tone before we release the Windows install.

I thought that this change was made so that you don't have to handle the ffmpeg dependency with the installer.

mikiher:
Ok, you're the boss... :)

As I said, I will continue to work on the installer.

mikiher:
It looks like fs.access(dir, W_OK) (which I was using to check if the main install path was writable) does not work on Windows. The last commit works around this by writing and removing a file.

advplyr:
This is good to go. I limited this to Windows or development since we don't need to run this in Debian or Docker.

Thanks!

PR #2400, opened 2023-12-08
A few BookFinder improvements (including a fix for #2238)

### LINKED ISSUE #2238: Sort Book Match Results by closest to length
The new book length on the match page is excellent. A potention suggestion for improvement: Once a list of results has been found for a book, rather than the default ordering from the provider, organise them from closest to book length to biggest difference.

mikiher:
The following improvements were implemented:
- Handle "abridged/unabridged" in titleCandidate generation
- Clean up "et al[.]" in author normalization.
- Sort Audible match results by ascending duration difference
  - this required an API change, since up until now /search/books had no context (it only accepted title, author, and provider). We now pass the libraryItem id from the client, retrieve it from the database, and pass it to BookFinder.search(). 
  - This change is fully backwards-compatible.
  - This change is also generally a good preparation for future context-based matching improvements (e.g. prefer results with same narrator)
  - I believe this resolves #2238


advplyr:
Works great, thanks!

advplyr:
#2517 there is a case I didn't think of where audible returns a bunch of results that are bad matches but if one of those matches has a closer duration it will show up at the top.
I'm not sure the best resolution for this. Maybe we have a max levenshtein distance or something.

mikiher:
So, there's a filterSearchResults function that tries to do this (but only) for OpenLib results. 
I tried to enable it for other providers at some point, but it had some issues and I ditched the idea at the time (I don't remember the exact details).

I'll look into once more, and perhaps change the current implementation so it works well for all providers.

PR #2420, opened 2023-12-17
docs: update synology reverse proxy to use the latest DSM settings

treyg:
Updated the docs for synology reverse proxy to match the current version of synology DSM (7.x). Current docs show settings for outdated DSM 6.x


advplyr:
Thanks!

PR #2431, opened 2023-12-19
Add zoom controls to comic reader

pablojimenezmateo:
I am enjoying audiobookshelf very much, I recently added some comics to my collection and they are hard to read in my laptop due to the lack of zoom.

I managed to enable zoom (by modifying the width of the  tag) to the comicreader without external dependencies.

I am open to discuss other possibilities if that is what you prefer, screenshot:

image

### DIFF
--- a/client/components/readers/ComicReader.vue
+++ b/client/components/readers/ComicReader.vue
@@ -14,19 +14,20 @@
       </div>
     </div>
 
-    <a v-if="pages && numPages" :href="mainImg" :download="pages[page - 1]" class="absolute top-0 bg-bg text-gray-100 border-b border-l border-r border-gray-400 hover:bg-black-200 cursor-pointer rounded-b-md w-10 h-9 flex items-center justify-center text-center z-20" :class="comicMetadata ? 'left-32' : 'left-20'">
-      <span class="material-icons text-xl">download</span>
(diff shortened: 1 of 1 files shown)


advplyr:
I think the scroll wheel will have to work when overtop of the left/right navigation divs

pablojimenezmateo:
I have added an initial implementation, however I am not very happy with it. It comes with a few issues:

- The scroll deltaY from the wheel event is not the same when scrolling on top of the image and scrolling on top of the buttons
- There is no scroll smoothing (this can be enabled, but subsequent wheel events are overriden)

It may be better to just make the Previous and Next button areas smaller, and still deal with not being able to scroll when the mouse is on top of them. This would also allow to interact with the X scrollbar under the areas.

Let me know what you think

advplyr:
This actually works well for me. The scroll delta is the same whether scrolling on the image or on the divs. The only difference is the smooth scroll which isn't bad. I'm sure eventually this will get refactored to show a progress bar or a better way to navigate.

I have a trackpad and when zooming in enough to get a horizontal scroll bar I had to add the `event.deltaX` to allow the trackpad to horizontal scroll. The trackpad scrolls smoothly btw.

Thanks!

Issue #2437, opened 2023-12-22
[Enhancement]: Read ASIN and CDEK tags from m4b (maybe more?) files


johnpyp:
### Describe the feature/enhancement

As far as I can tell, currently the audiobook server only uses ffprobe for metadata fetching from audio files. This is somewhat limiting, at least in the case of m4b, because it only grabs metadata fields that are either well-defined standard fields, or use the standard namespace-prefix scheme.

However, many audiobook files have extra fields to store some of this metadata that don't follow the standard scheme, and of particular interest are the non-standard `ASIN` and `CDEK` fields. CDEK is the field that is most faithful to the AAX file format, and so is included in many 1:1 copies.

Because we can't get to these extra fields currently, another library would likely have to be introduced.
- `tone` is already used and can pickup these extra fields, which would work well, but I see there's some interest in removing it as a dependency in the first place. Reportedly it also has some trouble with some m4b files
- [mediainfo.js](https://github.com/buzz/mediainfo.js) is a pure-js emscripten compilation of the widely used and popular `mediainfo` library, which would fit nicely in this codebase. It also supports extracting these custom fields, and seems to work great for all the files I've tested it on so far.

An example of using mediainfo js on one of my audiobook files, `mediainfo.js The\ Will\ of\ the\ Many/The\ Will\ of\ the\ Many.m4b -f JSON`.

```json
{
  "creatingLibrary": {
    "name": "MediaInfoLib",
    "version": "23.07",
    "url": "https://mediaarea.net/MediaInfo"
  },
  "media": {
    "@ref": "",
    "track": [
      {
        "@type": "General",
        "AudioCount": "1",
        "MenuCount": "1",
        "Format": "MPEG-4",
        "Format_Profile": "Base Media",
        "CodecID": "isom",
        "CodecID_Compatible": "iso2/mp41/M4A /M4B ",
        "FileSize": "809187505",
        "Duration": "101681.072",
        "OverallBitRate_Mode": "VBR",
        "OverallBitRate": "63665",
        "StreamSize": "11065587",
        "HeaderSize": "32",
        "DataSize": "798125517",
        "FooterSize": "11061956",
        "IsStreamable": "No",
        "Title": "The Will of the Many",
[...]

PR #2438, opened 2023-12-22
add mediainfo.js as an additional "backup" tag parser for extra tags …

johnpyp:
Motivation: https://github.com/advplyr/audiobookshelf/issues/2437

This PR adds mediainfo.js as a backup for ffprobe to get metadata information from media files. Currently, it's quite conservative and will only search medainfo if ffprobe doesn't find anything for any of its possible tag values, and will only consider the `extra` fields anyways (i.e non-well-defined/known fields).

A key usecase here is to extract `asin` or `cdek` tags as definitions for the ASIN, which ffprobe can't currently do.

Before:
image

After:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "graceful-fs": "^4.2.10",
         "htmlparser2": "^8.0.1",
         "lru-cache": "^10.0.3",
+        "mediainfo.js": "^0.2.1",
         "node-tone": "^1.0.1",
         "nodemailer": "^6.9.2",
         "openid-client": "^5.6.1",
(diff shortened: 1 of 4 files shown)


nichwall:
I think that looks pretty handy. For the sake of argument, could everything that mediainfo does with parsing be done by ffmpeg? I know that ffprobe doesn't get every tag, but wondering if ffmpeg could get the "extra" tags for the different formats and prevent an additional dependency. This would still use the same way you set up the logic, where ffprobe runs and then only run the backup parser (ffmpeg or mediainfo) if nothing is found. I assume that ffmpeg is slower than ffprobe or mediainfo over many files.

Even if ffmpeg would be the "better" option, mediainfo could still be used in the meantime (like tone is).

johnpyp:
Not quite sure if ffmpeg could get it all, but yeah I imagine it would be significantly slower potentially.

nichwall:
Thanks. I'll probably play with it a bit in a few days once I'm back home and not traveling. From my googling, it *looks* like ffmpeg will still read/write custom tags, but reading documentation and forums online isn't the same as actually testing it.

PR #2442, opened 2023-12-23
Follow up Translations for 76119445a302f0c1109bc4fdb44100f60be7107e

JBlond:
* Update:Listening sessions table for multi-select, sorting and rows per page
- Updated get all sessions API endpoint to include sorting
- Added sessions API endpoint for batch deleting


advplyr:
Thanks!

PR #2445, opened 2023-12-25
Add a SIGINT handler for proper server shutdown

mikiher:
I've noticed that once in a while (can't reproduce consistently), when hitting Ctrl+C to close ABS, the port is not properly released, and the next time I start the server I get EADDRINUSE error. I'm not sure what's the reason for that, but I thought adding a SIGINT handler that calls the server's stop method would be good practice anyway, and It would also provide a way for proper stopping in the upcoming Windows tray app.

I implemented this like many standard SIGINT handlers - the first SIGINT waits for server stop, the second exits immediately (the current behavior).

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -276,6 +276,19 @@ class Server {
     })
     app.get('/healthcheck', (req, res) => res.sendStatus(200))
 
+    let sigintAlreadyReceived = false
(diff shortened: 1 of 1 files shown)


advplyr:
I've not actually had that happen but it makes sense to handle SIGINT. Thanks!

PR #2454, opened 2023-12-27
Add SocketAuthority.close()

mikiher:
This is a small fix for my previous PR #2445.

If there are connected SocketIO sockets, then calling server.close() on SIGINT will hang and not resolve.

Instead, we need to call SocketAuthority.io.close(), which will disconnect all connected sockets, and also close the underlying http server (see documentation [here](https://socket.io/docs/v4/server-api/#serverclosecallback)). Sorry for missing this initially.


advplyr:
Thanks!

Issue #2332, opened 2023-11-20
[Bug]: Server crash when closing playback window before transcoding starts


dmasice:
### Describe the issue

I'm using Firefox from Linux to play some audiobooks.  The server is a raspberry pi 4 with docker using the official image.

If the audiobook I'm trying to play requires transcoding, it can take a bit to start streaming.  If I push the play button to start playback and close the playback window via the close player button before streaming starts, the server crashes.

I think it could be related to #2157.

Here's the log:

```
[2023-11-20 13:18:02] DEBUG: [PlaybackSessionManager] startSessionRequest for device Linux x86_64 / Firefox (PlaybackSessionManager.js:64)
[2023-11-20 13:18:02] DEBUG: [PlaybackSessionManager] "admin" starting direct play session for item "8861a05e-f987-41f6-b8ed-fd70a114d13b" with id f4e3df70-8b10-4f65-a2e1-313c227abae0 (Device: Linux x86_64 / Firefox) (PlaybackSessionManager.js:240)
[2023-11-20 13:18:03] DEBUG: [PlaybackSessionManager] startSessionRequest for device Linux x86_64 / Firefox (PlaybackSessionManager.js:64)
[2023-11-20 13:18:03] INFO: [PlaybackSessionManager] startSession: Closing open session "BOOK_TITLE" for user "admin" (Device: Linux x86_64 / Firefox)
[2023-11-20 13:18:03] DEBUG: [PlaybackSessionManager] closeSession "f4e3df70-8b10-4f65-a2e1-313c227abae0" (PlaybackSessionManager.js:306)
[2023-11-20 13:18:03] DEBUG: [PlaybackSessionManager] Removed session "f4e3df70-8b10-4f65-a2e1-313c227abae0" (PlaybackSessionManager.js:330)
[2023-11-20 13:18:03] DEBUG: [PlaybackSessionManager] "admin" starting stream session for item "8861a05e-f987-41f6-b8ed-fd70a114d13b" (Device: Linux x86_64 / Firefox) (PlaybackSessionManager.js:244)
[2023-11-20 13:18:03] INFO: [STREAM] START STREAM - Num Segments: 12017
[...]

PR #2465, opened 2023-12-29
Add try/catch to fileUtils.getFileMtimeMs

thevoltagesource:
Added a try/catch to the fileUtils.js getFileMtimeMs function.  This addresses the exception thrown when an ephemeral file is added to a library and flagged by Watcher to be scanned.   I tested this against 2.6.0 and 2.7.0 and it resolves my situation mentioned in https://github.com/advplyr/audiobookshelf/issues/2332 and it should resolve the reported issue completely.

### DIFF
--- a/server/utils/fileUtils.js
+++ b/server/utils/fileUtils.js
@@ -81,7 +81,12 @@ module.exports.getFileSize = async (path) => {
  * @returns {Promise<number>} epoch timestamp
  */
 module.exports.getFileMTimeMs = async (path) => {
-  return (await getFileStat(path))?.mtimeMs || 0
+  try {
+    return (await getFileStat(path))?.mtimeMs || 0
+  } catch (err) {
+    Logger.error(`[fileUtils] Failed to getFileMtimeMs`, err)
+    return 0
+  }
 }
 
 /**


advplyr:
I'm not sure how that exception isn't already being caught by the `getFileStat` function

https://github.com/advplyr/audiobookshelf/blob/707451309c16a82989b93d9f84164c613a2fe9dc/server/utils/fileUtils.js#L41-L48

You tested this with your Calibre setup that is triggering the exception?

thevoltagesource:
Yes I tested it by adding a book in Calibre with this code change in place.  Instead of the original error and a container crash, these two error messages are logged and things continue on.
```
{"timestamp":"2023-12-29 15:36:49.021","source":"Logger.js:115","message":"[Watcher] \"Shire Collection\" Ready","levelName":"INFO","level":2}
{"timestamp":"2023-12-29 15:37:03.862","source":"Logger.js:127","message":"[fileUtils] Failed to getFileMtimeMs Error: ENOENT: no such file or directory, stat '/audiobooks/metadata.db-journal'","levelName":"ERROR","level":4}
{"timestamp":"2023-12-29 15:37:04.863","source":"Logger.js:127","message":"[fileUtils] Failed to getFileMtimeMs Error: ENOENT: no such file or directory, stat '/audiobooks/metadata.db-journal'","levelName":"ERROR","level":4}
{"timestamp":"2023-12-29 16:00:00.375","source":"Logger.js:115","message":"[PodcastManager] runEpisodeCheck: \"Self-Hosted\" | Last check: Fri Dec 29 2023 15:00:01 GMT+0000 (Coordinated Universal Time) | Latest 
```
I am still not entirely clear on this so I'm not sure if I handled it the appropriate way.  The issue seems to be a difference between synchronous and asynchronous error handling.  The try/catch in `getFileStat` would have handled  an error in a synchronous call, but because the error happened inside an async call to that function, the promise captured the error and it was what needed to handle it.  Which this lines up with what the original error was showing.
```
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, open '/metadata/streams/987485cd-a3fd-44e8-902a-b57c0374c175/files.txt'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/metadata/streams/987485cd-a3fd-44e8-902a-b57c0374c175/files.txt'
}
```
There are several blogs and stack overflow questions that discuss this but a lot of it went over my head.

advplyr:
Great, thanks for testing!

PR #2468, opened 2023-12-31
Migrate to new pdf viewer

### LINKED ISSUE #2279: [Bug]: PDF Reader is flickering
### Describe the issue

The PDF reader flickers on each page change, which makes using it rather uncomfortable.
The page number also flickers and the green loading bar is rather distracting.

chrome_snkqSByONM

This does not seem necessary since the page is rendered, then white, then rendered again.
Is the first render lower resolution? Would it be possible to not render a white "frame" in between (and remove the loading bar) to make it less jaring?

1. Open a PDF ebook
2. Flip through pages

### Audiobookshelf version

v2.5.0

Docker

vincentscode:
This is an attempt to resolve #2279.
It should retain all the functionality of the current viewer but also resolve the issues outlined there.
The styling isn't perfect yet, but i've been unable to produce a "nice" solution to that, yet.


gh0sti:
Wonder if you should look into implementing this instead for your PR?

https://github.com/johnfactotum/foliate-js

vincentscode:
@gh0sti Is there any reason for switching? It does not seem to have any integration with vue.

gh0sti:
> @gh0sti Is there any reason for switching? It does not seem to have any integration with vue.

I saw in discord chat dev channel some of the devs mentioning it.

vincentscode:
@advplyr would you mind taking a look at this so it may be merged? :)

advplyr:
I checked this out and merged master. Scaling doesn't work properly. It doesn't load the current page because it is listening to the wrong event, should be `@loading`.

This library is abandoned like the one currently used. It is probably worth looking for a library that is up-to-date.
I think we also need a pdf worker.js file to use with pdf.js. I saw a warning in the console about this but haven't spent a lot of time on it due to the library being abandoned and the issues I mentioned above.

PR #2472, opened 2024-01-02
Audiobook converter library specific settings

nichwall:
This PR is designed to help make the Audiobook File Management Tools more useful by adding more information to the Audio Tracks table and adding a new library specific setting for default encoding values for the "advanced settings". It is a WIP while I learn more about how JS and Vue work.

I'm okay if somebody takes this and runs with it since I won't be very fast.

- [x] Add bitrate column
- [x] Set default bitrate to be the maximum bitrate of the existing files (so everything isn't hard coded to 128k)
- [ ] Add database table of default encoder settings as library setting
  - Use [max/min/fixed] bitrate
  - [Keep channels/force stereo/force mono]
  - Codec: [aac, opus], maybe more?
  - File format: [m4b, mp3]
- [ ] Use library specific settings in file management tools

bitrate_column

The default encoding values will include information that can be reused for both audiobooks and podcasts. The default encoding values can then be applied to an entire library, individual library items (book or podcast episodes), or a grouping of library items (podcast). This PR will not include automatic encoding, but hopefully helps provide better support once queueing or automatic conversion is added.

(pardon the terrible mockup, but something kind of like this. Image will be replaced once I get the actual tab built)
automatic_encoding_mockup


nichwall:
A better description is included here. This PR should be good to merge as is to help with setting the bitrate automatically based on the input files and otherwise is purely a minor interface update.

https://github.com/advplyr/audiobookshelf/discussions/2479

slyfoxgh:
I think it looks straightforward. A simple setting of bitrate and preferred codec as the default option is good. I would use a global setting of 64K AAC for all books. I usually manually set 128 for radio drama based books that have music or sound effects.

FreedomBen:
Merging the automatic birate settings soon would be awesome.  I've forgotten to check/adjust the bitrate a number of times so I have dozens of books that are 128k encoded from 64k sources  :sweat_smile:

nichwall:
Superceded by https://github.com/advplyr/audiobookshelf/pull/4253

PR #2485, opened 2024-01-03
Update fr.json

Machou:


advplyr:
We can't have comments in the JSON file.
The context for those:

`LabelDatetime`: This is the column header for a table that is a column of dates with times
image

`LabelDiscFromFilename` and `LabelDiscFromMetadata`: These are column headers when editing tracks to show where the disc number was pulled from, the filename or from the audio file meta tags.
image

`LabelFinished`: This is used in the progress box
image

Machou:
I have other question, about the word "Item", you use this word for many string.

Item is only for book audio ? or other ? because i don't like use "Article" in french.

advplyr:
Item is short for "library item". At the moment a library item can only be a book or a podcast but in the future if we add more media types it could be a music track or something.
I'm not sure if that helps translate.

Machou:
> Item is short for "library item". At the moment a library item can only be a book or a podcast but in the future if we add more media types it could be a music track or something. I'm not sure if that helps translate.

I'll think about it and ask people who are better in French to help me.

You can merge if it's Ok for you, I've double-checked and found nothing to correct =)

advplyr:
Thanks!

Issue #2467, opened 2023-12-30
[Enhancement]: Better Ebook handling and conversion


FlyinPancake:
### Describe the feature/enhancement

Hi!

I really like this project, and would like to ditch calibre-web from my stack.

abs works splendidly for audiobooks, and it could be wonderful for ebooks as well!

For my use-case, custom metadata providers are a must. My friend and I are developing one, that would be compatible with abs, we would like to implement it as a REST API which exposes the metadata required for abs. We would like to collaborate on an OpenApi spec for this REST API, which then could be used as a custom provider. As we are coming up with this spec, I will link it here. This should be really simple, as this would be a generalized provider, that could be specified by the user.

I love abs's send to ereader ability, but as Kindle dropped support for `mobi` it is tedious to manually convert and upload each ebook as epub. For this problem [ebook-converter](https://github.com/gryf/ebook-converter) could be integrated simirarly to tone. The second issue is filesize for email attachments, as right now abs tries to send even 40mb PDFs as attachments, that inevitably fail. Kindle supports `.zip` files for this reason, and this functionality could be incorporated as well.

I know that this is a community-run project, and I am ready to be a part of development efforts (despite my laughable JS skills).  advplyr made a great tool for us data hoarders to admire and manage our piles of data and I am greatful for its existance <3

This feature request is a big one and serves more like a wishlist than actual requests.
Thanks for coming to my TED talk

advplyr:
Hey, I would be interested in hearing more about the custom metadata provider you are working on. Are you aggregating metadata from different sources or putting together your own metadata? Are you hosting this publicly so it could be integrated as a metadata provider for everyone in Abs?

Ebook converter is outside the scope of Abs right now but maybe sometime in the future or as a plugin. Just as an aside, Tone is going to be removed in an upcoming version so the only external dependencies we will have to deal with is ffmpeg/ffprobe.

We can definitely support zipping large ebooks before sending to kindle. We would also need to check if the zipped file still exceeds 50MB

FlyinPancake:
Hi! 

We are currently working on our API, and then will we start tackling the integration with ABS. The provider is currently hosted in a private repo, as the source vert much forbids scraping. We developed a scraper + cache setup to not get IP banned. I plan on open souring the cache and API part, but keeping the scraper private to not get taken down.
That way anyone who wants a custom metadata source, can implement a single rust trait and have the rest taken care of. 

[...]

PR #2183, opened 2023-10-05
Improve formatting configuration

ahobsonsayers:
When developing my first contrbution to this repo in #2182 , I noticed that the configuration of the formatting could be improved slightly to help people developing in VSCode and other IDEs, as well as aid with keeping consistent formatting across the codebase - assuming this is what you want. This was particularly relevant to me as I noticed when editing some existing files in my other PR, that my IDE was trying to format it differently to how it currently was, which i had to be careful to avoid.

I have moved your prettier config from the vscode settings to a `.prettierrc` file so that it is IDE agnostic - vetur will still use this config. Additionally i have added prettier to the dev dependancies and added a task to run prettier across all files should you wish to do this at some point (I did not run this for this PR).

I hope these improvements simply help contributers know the formatting this project would like, allows them to easily ensure they are meeting the style and helps avoid changing the formatting of existing files. 

If this is not what is wanted, please let me know, it just seemed like an easy win that I think is helpful (well for me at least)

PR #2486, opened 2024-01-03
[Feature] Add support for custom metadata providers through a REST API

Dewyer:
# About
This MR is a WiP followup for this issue: https://github.com/advplyr/audiobookshelf/issues/2467

I am trying to implement the provider in ABS. This is my first time contributing to the project so, sorry I am sure I didn't do things exactly as I should, LMK what should I change to better align with the coding style of the project.

## Notes
- I am aware that I am missing a bucnh of transaltion items, thats WiP
- Custom providers should probably get their own page on the guides site, I will at some point write an MR for that


Dewyer:
I am trying to fully test it myself, but probably will miss a lot of edge cases.

server/Database.js
```diff
@@ -694,6 +700,45 @@ class Database {
...
+   */
+  async removeCustomMetadataProviderById(id) {
+    // destroy metadta provider
+    await this.customMetadataProviderModel.destroy({
+      where: {
+        id,
+      }
+    })
+
+    const slug = `custom-${id}`;
```
advplyr, line 730: Is there a reason to use a slug here instead of just the UUID?

client/strings/en-us.json
```diff
@@ -96,6 +96,7 @@
   "HeaderAudiobookTools": "Audiobook File Management Tools",
   "HeaderAudioTracks": "Audio Tracks",
   "HeaderAuthentication": "Authentication",
+  "HeaderCustomMetadataProviders": "Custom metadata providers",
```
advplyr, line 99: These should be alphabetical (ignoring casing). Also, don't worry about adding the strings to the other languages, I have a local script I run for that.

server/controllers/MiscController.js
```diff
@@ -717,5 +717,95 @@ class MiscController {
...
+   * @param {import('express').Request} req
+   * @param {import('express').Response} res
+   */
+  async addCustomMetadataProviders(req, res) {
+    if (!req.user.isAdminOrUp) {
+      Logger.error(`[MiscController] Non-admin user "${req.user.username}" attempted to add admin custom metadata providers`)
+      return res.sendStatus(403)
+    }
+
+    const { name, url, apiKey } = req.body;
```
advplyr, line 766:
```suggestion
    const { name, url, apiKey } = req.body
```
The repo is not using semicolons and not using trailing commas. I have this rule setup for vscode but still working on improvements for this #2183

server/controllers/MiscController.js
```diff
@@ -717,5 +717,95 @@ class MiscController {
...
+    if (!id) {
+      return res.status(500).send(`Invalid delete data`)
+    }
+
+    const provider = await Database.customMetadataProviderModel.findByPk(id);
+    await Database.removeCustomMetadataProviderById(id);
+
+    SocketAuthority.adminEmitter('custom_metadata_provider_removed', provider)
+
+    res.json({})
```
advplyr, line 808:
```suggestion
    res.sendStatus(200)
```

server/models/CustomMetadataProvider.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  static findByPk(id) {
+    return this.findOne({
+      where: {
+        id,
+      }
+    })
+  }
```
advplyr, line 35:
```suggestion
```

This function is built-in to Sequelize. https://sequelize.org/docs/v6/core-concepts/model-querying-finders/#findbypk

server/models/CustomMetadataProvider.js
```diff
@@ -0,0 +1,58 @@
...
+   * @param {import('../Database').sequelize} sequelize 
+   */
+  static init(sequelize) {
+    super.init({
+      id: {
+        type: DataTypes.UUID,
+        defaultValue: DataTypes.UUIDV4,
+        primaryKey: true
+      },
+      name: DataTypes.STRING,
```
advplyr, line 48:
```suggestion
      mediaType: DataTypes.STRING,
      name: DataTypes.STRING,
```
We'll want to support podcast custom metadata providers as well in the future or any other media types that could get added.

server/providers/CustomProviderAdapter.js
```diff
@@ -0,0 +1,70 @@
...
+        const providerId = providerSlug.split("custom-")[1]
+        const provider = await Database.customMetadataProviderModel.findByPk(providerId);
+
+        if (!provider) {
+            throw new Error("Custom provider not found for the given id");
+        }
+
+        const matches = await axios.get(`${provider.url}/search?query=${encodeURIComponent(title)}${!!author ? `&author=${encodeURIComponent(author)}` : ""}`, {
+            headers: {
+                "Authorization": provider.apiKey,
```
advplyr, line 19: If a user is using bearer token authorization or basic authorization would they be instructed to enter that in the `apiKey` input? Let me know if that doesn't make sense

custom-metadata-provider-specification.yaml
```diff
@@ -0,0 +1,124 @@
...
+          type: string
+        subtitle:
+          type: string
+        author:
+          type: string
+        narrator:
+          type: string
+        publisher:
+          type: string
+        published_year:
```
advplyr, line 89: Should we use camelCase here to stay consistent with our data models? publishedYear

client/pages/config/custom-metadata-providers.vue
```diff
@@ -0,0 +1,45 @@
...
+    <app-settings-content :header-text="$strings.HeaderCustomMetadataProviders">
+      <template #header-items>
+        <ui-tooltip :text="$strings.LabelClickForMoreInfo" class="inline-flex ml-2">
+          <a href="https://www.audiobookshelf.org/guides/#" target="_blank" class="inline-flex">
+            <span class="material-icons text-xl w-5 text-gray-200">help_outline</span>
+          </a>
+        </ui-tooltip>
+        <div class="flex-grow" />
+
+        <ui-btn color="primary" small @click="setShowAddModal()">{{ $strings.ButtonAdd }}</ui-btn>
```
advplyr, line 12:
```suggestion
        <ui-btn color="primary" small @click="setShowAddModal">{{ $strings.ButtonAdd }}</ui-btn>
```

server/providers/CustomProviderAdapter.js
Dewyer: Basic auth isn't supported in the current model, as it would require 2 pieces of credentials, but bearer, is essentially just Authorization header + a Bearer prefix and an api key so they could make that work

server/Database.js
Dewyer: it makes it easier to know a provider is a custom one, as they all have this prefix, if it was just an id you could technically have some future provider be prefixed with valid hex characters or something

FlyinPancake:
I have addressed most of the concerns, however I held out on the `MediaType` since I don't know what values should be stored there. 

As of now and according to my knowledge, all the providers are assumed to be book / audiobook metadata providers.

I think one CMP should support multiple `MediaType`s and those could be stored in a comma separated `string`, making it extensible in the future. 

Supporting podcasts would require a separate CMP EP in `custom-metadata-provider-specification.yaml`. We could come up with something, but I would like to keep this PR to the point and addressing it later.

We could rename the `/search` EP to `/searchBooks` to accommodate future expandability.

server/providers/CustomProviderAdapter.js
advplyr: Basic auth would be supported in this model as far as I understand it because the user is required to enter the full authorization header value.
`Basic <credentials>` or `Bearer <api key>`.

My question was about if we are requiring the user to write in the `Bearer` part of the authorization header. If we are then `apiKey` might not be an accurate name for the variable.

advplyr:
I made a lot of updates to get this ready. You will need to delete the table from your db before using the updates. 

In the model I added new properties and changed `apiKey` to `authHeaderValue`. It's technically not just an api key going there but the full authorization header value. I also made this not required which seems reasonable if someone makes a public API. The openapi spec may need to be updated to account for that.

I moved the config page to go under "Item Metadata Utils" instead of being in the main sidebar.

I updated the library api request to return the custom metadata providers so there is no longer a need for a client-only api endpoint to get custom providers. The socket requests have been updated to send all clients the client-safe JSON.

I created a new controller and set the stage to eventually allow custom podcast providers. The `url` is validated to make sure it is a real url.

I want to add more validation to the data being returned from the custom metadata provider. Please test your provider on the edge container when you get a chance.

FlyinPancake:
Thanks for doing this ❤️

Ur a legend


I'll get around to it as soon as I can

FlyinPancake:
Hi! 

I checked the edge container out, it works as expected.
This PR does not call searchBooks it calls search it that is a problem, this is the time to act. 

As this is a merged pr, i'll create the example custom metadata provider and probably a guide for the website if that's ok

advplyr:
I realized a better solution was to pass the `mediaType` to the `/search` endpoint as a query param.

I just made that update here: https://github.com/advplyr/audiobookshelf/commit/4e3e7b10ce70042d2ffa9e391cb18af6de9f8377

So if in the future someone implements a provider for both books and podcasts they can check the `mediaType` passed in.

If you can start on the guide that would be great. I would like to have something up before this goes in the next release.

PR #2491, opened 2024-01-04
Add Podcast Search Region

liaochuan:
Added podcast search regions to the Settings page to control which countries are used when searching for podcasts through iTunes, currently only the US and China have been added.
image


advplyr:
I think this belongs in library settings and not server settings.

liaochuan:
> I think this belongs in library settings and not server settings.

Yes, it should be that way, I'm sorry I'm new to audiobookshelf and not very familiar with it yet, I've adjusted this configuration item.

advplyr:
Thanks!

PR #2506, opened 2024-01-09
Change `Logger.dev` calls to `Logger.debug`

FreedomBen:
Change `Logger.dev` calls to `Logger.debug`

`Logger.dev` is kind of in a weird spot where it doesn't fit into the
standard log level.  It is called directly by some code and it only
checks whether a property is set (which comes from an env var) before
deciding to print out.

This standardizes on `debug` by changing the dev calls to debug. Also
removes the now unused code.


FreedomBen:
Reference:  https://github.com/advplyr/audiobookshelf/pull/2489

The setting of the log level discussed there will be in a separate PR

advplyr:
Looks good, thanks!

Issue #1641, opened 2023-04-01
[Enhancement]: Ignore Synology @eaDir folder and @SynoEAStream files in Library Files


hypnotoad08:
### Describe the feature/enhancement

This is a minor feature request as it doesn't affect functionality. I noticed that the Library files section for some of the audiobook files I have moved to my shared Audiobook folder on my Synology NAS are showing some hidden files/folders that are created by the NAS. Can the Library files section filter out these hidden files/folders?
![@eaDir](https://user-images.githubusercontent.com/842086/229292817-d4e041a1-f7dd-4419-9813-f40e12d96c2a.jpg)

CLHatch:
This is handled by Plex with `.plexignore` files that can be created to specify patterns of files/folders to be ignored.  Something similar could be created for ABS. https://support.plex.tv/articles/201381883-special-keyword-file-folder-exclusion/

CLHatch:
SyncThing has a similar file you can use to set files/folders to ignore.

advplyr:
You can use a `.ignore` file

raphpa:
It is a little bit more complicated with the `@eaDir` folder. The NAS creates it automatically in every folder with media in it. So if you have your library sorted in subfolders you would have to manually put the `.ignore` file manually into every single one of them.
A global ignore option would be great for this case.

hypnotoad08:
Yes I was hoping for a global fix

theolike:
Agreed.  It has to be a configurable global filter.  While this one is specifically a Synology quirk, there are many other systems that use a similar process for indexing, thumbnails, etc.  For example, Windows and it's thumbs.db files.  I haven't been through the indexer code yet (and I guess I should do that), but is there already filter logic in it, even if for only a hard coded list?

(Also, no matter what kind of spoon you are, "Journey before destination", Radiant.)

CLHatch:
> Agreed. It has to be a configurable global filter. While this one is specifically a Synology quirk, there are many other systems that use a similar process for indexing, thumbnails, etc. For example, Windows and it's thumbs.db files. I haven't been through the indexer code yet (and I guess I should do that), but is there already filter logic in it, even if for only a hard coded list?
> 
> (Also, no matter what kind of spoon you are, "Journey before destination", Radiant.)

[...]

PR #2521, opened 2024-01-17
Exclude @eaDir and #snapshot from recursive library searching

arcao:
This partially solve issue #1641. The `@eaDir` is created by Synology DSM to store additional files, e.g. metadata. Synology DSM mount BTRFS snapshots into `#snapshot` folder which is subdirectory of shared folder.

### DIFF
--- a/server/utils/fileUtils.js
+++ b/server/utils/fileUtils.js
@@ -167,7 +167,8 @@ async function recurseFiles(path, relPathToReplace = null) {
     extensions: true,
     deep: true,
     realPath: true,
-    normalizePath: true
+    normalizePath: true,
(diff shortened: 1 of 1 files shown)


Sapd:
I think the proper way to fix it would be to make it configurable.
Or maybe even better, to have a `.absignore` file in the library folder where one can specify things like that.

advplyr:
This doesn't work because that `exclude` array is looking for those strings being including anywhere in the path. I'm not sure who would have the string `#snapshot` in their folder path somewhere but if and when that happens it would be unexpected behavior.

The proper solution for this is a bit more involved and also should be setup in `Watcher.js`. Ideally we implement #1641 in full

PR #2524, opened 2024-01-18
Bump nuxt version

apiweb:
I create this pull request to update the Nuxt version in the AudioBookShelf repo from [v2.15.8](https://github.com/nuxt/nuxt/releases/tag/v2.15.8) to [v2.17.3](https://github.com/nuxt/nuxt/releases/tag/v2.17.3). 
When I tried setting up my environment locally and ran npm run in the client folder, I hit an error:

Screenshot 2024-01-18 at 08 57 04

To fix this I need to put some code on package.json

Screenshot 2024-01-18 at 08 55 27

It looks like there's a bug related to the crypto node library. Nuxt fixed this in version [v2.16.2](https://github.com/nuxt/nuxt/releases/tag/v2.16.2), but AudioBookShelf is currently on [v2.15.8](https://github.com/nuxt/nuxt/releases/tag/v2.15.8). So, I decided to jump to the latest, [v2.17.3](https://github.com/nuxt/nuxt/releases).

I dug into the Breaking Changes between [v2.15.8](https://github.com/nuxt/nuxt/releases/tag/v2.15.8) and [v2.17.3](https://github.com/nuxt/nuxt/releases). The only one I found was in the upgrade to 2.16.0, which was about a new postcss options format. You can get all the details [here](https://github.com/nuxt/nuxt/pull/9671).

After updating the nuxt.config.js file, I tested the app locally. Everything seemed fine - no errors while navigating pages, and both npm run build and npm run generate worked without a hitch.

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -127,7 +127,7 @@ module.exports = {
 
   // Build Configuration: https://go.nuxtjs.dev/config-build
   build: {
-    postcss: {
+    postcssOptions: {
       plugins: {
         tailwindcss: {},
         autoprefixer: {},
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -18,7 +18,7 @@
         "epubjs": "^0.3.88",
         "hls.js": "^1.0.7",
         "libarchive.js": "^1.3.0",
-        "nuxt": "^2.15.8",
+        "nuxt": "^2.17.3",
         "nuxt-socket-io": "^1.1.18",
         "trix": "^1.3.1",
         "v-click-outside": "^3.1.2",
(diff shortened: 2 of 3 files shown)


advplyr:
What version of node are you using? I'm still on v16.14.0 and think that may need to be updated first. I planned to jump to node 20

apiweb:
Oh, my bad. I'm in Node 18, I was on Node 16, but a few weeks ago, I updated it to 18 because 16 reached its end of life.

Also, I ran some tests with Node 20, and it's still working well (but I don't try to use it deeply)

advplyr:
There was a bunch more stuff I had to do to bump to node 20. I did that all here https://github.com/advplyr/audiobookshelf/commit/7f350279fae1ea51fa736c32478f5c650e58fc4f so you should be good now if you pull and switch to node 20.
Thanks for the nudge on getting this done!

PR #2553, opened 2024-01-25
OpenID: Implement Logout + Fix state + Fix URL Regex

Sapd:
**Implement OpenID Logout**

- The **POST** Request to /logout will now optionally response with a `redirect_url` parameter. The client is supposed to follow it if its there (or open it up in a Browser if app).
- The ABS frontend will now redirect to it when supplied.
- For the `redirect_url` (end-session OIDC url) - https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
    - Implemented `id_token_hint` - actually an optional but recommended parameter. After googling a bit, it seems even some SSO providers also require it for security. On provider like Keycloak when its supplied, the User-Experience is a bit better because it will directly skip the keycloak logout screen and redirect back.
    - post_logout_redirect_uri also implemented, which will simply point to `absurl://login`. On mobiles it wont be provided bc of high code complexity for that, but the mobile app (or other mobile apps) can simply append `&post_logout_redirect_uri=audiobookshelf://login` or so to the url if they want optionally. But this URL - if provided - needs to be whitelisted in the SSO provider by the user (often labeled as logout redirect url whitelist). 

Btw. if you test it, Authentik does neither implement `id_token_hint` nor `post_logout_redirect_uri` currently, the redirect back will always be to the application url configured in Authentik (when you click on "log back in again").

**Other changes**

- Introduced a new cookie `auth_method` to store the login method used, such as `openid`, `openid-mobile`, `local`, or `api`. Actually required for logout but I think we can also use it to make the UX in the app a bit easier as we have now a state of how the current session was logged in.
- Removed `is_rest` cookie as no longer required.
- If using mobile, the backend now accepts a `state` parameter if provided. Making it fully compliant with the oauth2 specs. Makes also tests like http://oidcdebugger.com pass against `absurl://auth/openid`
- Reworked the `auth/openid` route. This includes removing unused id_token code, implementing parameter sanity checks, and clearly separating the handling of mobile and non-mobile scenarios. Also only support 'S256' challenge (plain is in the oauth2 specs because there was a time when not every device had the ability for SHA256..., but if a device can do SHA256 it needs to use it anyway according to the spec)
- Fixed a bug in the Regex used for validating redirect URLs, subpaths didn't work.

Old app versions work with the changes. Also I tested implementing the logout process in the app, which also works. However the app also needs some revision regarding the UI when changing/editing servers.


advplyr:
Great, thanks!

PR #2554, opened 2024-01-25
Modify BinaryManager to download version 6.1

mikiher:
This fixes [this issue](https://github.com/mikiher/audiobookshelf-windows/issues/5) on audiobookself-windows

TLDR, the Windows native server requires at at least version 5.1 of ffmpeg/ffprobe, as older versions do not support Windows long paths. This seems like a necessary upgrade, since it might cause all sorts of unexpected problems.

Some more details:
- BinaryManager now downloads the latest ffmpeg and ffprobe versions (6.1) from the ffbinaries repository, instead of version 4.4.1 (which is still set as the 'latest' version on ffbinaries)
- Added a version check to make sure existing versions of the binaries are at least 5.1
- Added removal of old installed binaries (only ones that were installed by audiobookshelf)
- Updated tests
- Fixed a bug in fileUtils.isWritable()


advplyr:
Working great, thanks!

PR #2559, opened 2024-01-26
German localization optimization

bloodscript:
After contributing to the app by working on the german localization I was asked to contribute here as well.
https://github.com/advplyr/audiobookshelf-app/pull/1052
The main focus was to uniformly use one of the two most common ways to speak to a german ("Sie" or "du"). It was decided to use the informal and more widespread form "du". I now introduce this change here as well.
I also adapted the decision regarding the spelling of "ebook".
Please point out any findings and provide me with context of the used text. In most cases the original english version was enough to make a decision, but maybe I missed something there.
Thank you :)


advplyr:
Looks good, thanks!

PR #2570, opened 2024-01-30
Return png from AudiobookCovers.com

Weldawadyathink:
Changes AudiobookCovers.com provider to return the full size png file from the server. The original file url has the incorrect content-type header set, which caused issues downloading new cover images.

### DIFF
--- a/server/providers/AudiobookCovers.js
+++ b/server/providers/AudiobookCovers.js
@@ -14,7 +14,7 @@ class AudiobookCovers {
       Logger.error('[AudiobookCovers] Cover search error', error)
       return []
     })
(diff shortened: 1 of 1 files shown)


Weldawadyathink:
I plan on fixing the content-type of the original files, but this may end up being a quicker fix.

Weldawadyathink:
Well the fix was a lot easier than I expected, so the issue is gone now. Either way, this update will provide more consistent results. All of the images at results.versions.png.original are validated and written by pillow (PIL), so they should be valid images. I don't do any validation on the files at results.original, so there could be issues there.

advplyr:
Thanks!

PR #2573, opened 2024-02-01
Merge cover and media update in Match.vue into a single /media API call

### LINKED ISSUE #2556: [Bug]: After cover update, cover appears broken until page is reloaded
### Describe the issue

Sometimes (seems like this happens more often recently) after a cover update (e.g. when submitting a match result), the book cover appears broken, like this:

Screenshot 2024-01-26 114338

After I reload the page, the problem disappears:

Screenshot 2024-01-26 114458

I suspect there's some race condition between downloading/saving the image and notifying the web-socket that the update is done.

I caught this on a local version built from head on Windows.

1. Update a book cover (e.g. via match result)
2. Cover image sometimes appears broken on the page.

### Audiobookshelf version

v2.7.2

Built from source

mikiher:
Fixes #2556 (see details in the bug).

I tried to do the minimal amount of changes to existing functions on the server, while merging the cover and other media updates under a single API call.


mikiher:
Also fixes #2593.

advplyr:
Works great, thanks!

Issue #2547, opened 2024-01-23
[Enhancement]: Add sort to authors page for number of books


methosinhiding:
### Describe the feature/enhancement

The authors view is currently sorted by Author name by default, but there are no other sorting/filtering options.

Since the addition of authors is automatic but removals are a manual action, a common interaction with the authors page is to use the browser search functionality to locate and then manually delete authors who have 0 books attributed to them (either because their name was wrong/corrected, the 'author' was multiple merged names that had since been split out, or because the file was removed/moved to a different library, etc)

It would be very useful for this repeated author removal task to either be able to sort the author list by book count so that 0-book authors could be deleted en-masse or to provide a shortcut to perform that specific action such as a 'Purge Authors who have no books' button

nichwall:
Duplicate of https://github.com/advplyr/audiobookshelf/issues/2466

advplyr:
I thought there was another feature request open for sorting options on the authors page but I couldn't find it

nichwall:
Author filters (not a comprehensive list, but getting the filter functionality there):
https://github.com/advplyr/audiobookshelf/issues/1678

Author sorting was a discussion: https://github.com/advplyr/audiobookshelf/discussions/2018

advplyr:
Great thanks, this FR can be specifically for sorting by number of books which would be nice to have.

advplyr:
Added in [v2.8.1](https://github.com/advplyr/audiobookshelf/releases/tag/v2.8.1)

PR #2580, opened 2024-02-04
Add sorting to author page

KeyboardHammer:
Adds client side sorting to the author page.  Added lastFirst to authors api as it already is calculated server-side.

image

Implements: https://github.com/advplyr/audiobookshelf/issues/2547


izyspania:
Would be nice if we could add sorting for authors in the API too.

advplyr:
Thanks! This is a good start, eventually it will be moved to the API with lazy loading

PR #2582, opened 2024-02-05
Update AudiobookCovers.com provider to use improved search endpoint

Weldawadyathink:
I was able to get my OpenAI CLIP powered search api to be fast and reliable enough to use as the default for this project. In my testing, the AI search endpoint provides strictly better results in all scenarios.

### DIFF
--- a/server/providers/AudiobookCovers.js
+++ b/server/providers/AudiobookCovers.js
@@ -5,7 +5,7 @@ class AudiobookCovers {
   constructor() { }
 
   async search(search) {
-    const url = `https://api.audiobookcovers.com/cover/bytext/`
+    const url = `https://api.audiobookcovers.com/cover/ai-search`
     const params = new URLSearchParams([['q', search]])
     const items = await axios.get(url, { params }).then((res) => {
       if (!res || !res.data) return []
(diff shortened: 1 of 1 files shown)


advplyr:
Is this able to also return a thumbnail image? I'm seeing that when a lot of results are returned it gets very slow because some of these images are large (5MB+) when we are just displaying a small thumbnail. I will need to do some refactoring to support this but it seems necessary with how slow the page gets on unspecific searches.

Weldawadyathink:
> Is this able to also return a thumbnail image? I'm seeing that when a lot of results are returned it gets very slow because some of these images are large (5MB+) when we are just displaying a small thumbnail. I will need to do some refactoring to support this but it seems necessary with how slow the page gets on unspecific searches.

Hey, I missed the notifications for this issue and happened to see it in my email today. Sorry about not replying. Yeah, it does return a thumbnail, but as you mentioned there wasn't a way to use it in AudiobookShelf without refactoring. The [openAPI spec is hosted here](https://audiobookcovers.com/openapi.yaml). I include 200, 500, and 1000px width images, so the 200 is probably suitable for the thumbnails. It would just need a change on the return value to something like this:
```
return items.map((item) => (
  return {
    cover: item.versions.png.original,
    thumbnail: item.versions.jpeg.200,
  }
))
```
It also returns webp formats, but I don't know if it makes sense to try and optimize that when a 200px jpeg should load fast enough.

PR #2584, opened 2024-02-05
Added location slider and bookmarks to Epub reader

### LINKED ISSUE #1709: [Enhancement]:

I'd love a way to add bookmarks and/or notes to my books, both audio and ebook.
Love using Audiobookshelf, really amazing!

MxMarx:
This pull request adds a location trackbar and bookmarks for EPUBS.

To minimize code changes, the audiobook trackbar was slightly modified to remove audio-only features when used for books, 
Instead of using time to represent position in audiobooks, book position is represented by the percentage.
To convert this to pages, the 'playback rate' is set to `1 / total_pages`.
A bookmark's "time" attribute in the api reference would also need to change from an integer to a float to accommodate percentages.

I don't know if it would be cleaner to make a new component instead of expanding the existing ones, but I'd be happy to refactor stuff if preferred.

closes #1709


MxMarx:
> Does this cover up the audio player if the audio player is open?

It looks like it trackbars are on top of each other, I can try messing around with formatting for a bit

MxMarx:
> Does this cover up the audio player if the audio player is open?

Fixed!

MxMarx:
Bookmarks might actually be a bit premature, I'll take a look at making it less of a bodge first

advplyr:
I'm looking at this now and I think it is actually better to not use the existing PlayerUI and duplicate it into its own component. That PlayerUI is already pretty cluttered and a lot of that logic isn't necessary for the reader.

PR #2610, opened 2024-02-13
make server url configurable in dev mode

freddy36:
Currently when the server is running in dev mode it always assumes it's running on localhost.
With this changes it becomes configurable via the DEV_SERVER_URL environment variable.

Also some - probably no longer neeed - hardcoded localhost URLs in some javascript files have been removed.

### DIFF
--- a/client/components/covers/AuthorImage.vue
+++ b/client/components/covers/AuthorImage.vue
@@ -56,10 +56,6 @@ export default {
     },
     imgSrc() {
       if (!this.imagePath) return null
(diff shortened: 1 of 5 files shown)


advplyr:
What is the use-case?

freddy36:
Running the server and frontend in dev mode on a remote server.

advplyr:
You don't need to change that at all then. That is only used if you are running the nuxt separately with `npm run dev`

freddy36:
That's what we're doing (to have frontend code changes picked up faster).

PR #2617, opened 2024-02-16
Update pwa icon to use iOS icon

### LINKED ISSUE #2520: [Enhancement]: Higher resolution PWA app icon

The PWA icon for Audiobookshelf is low resolution compared to there PWAs. I would like to see that improved. Here is an example with other PWA apps:

IMG_86E9D7CACB4E-1

ipcintron:
resolves #2520

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -30,7 +30,7 @@ module.exports = {
     script: [],
     link: [
       { rel: 'icon', type: 'image/x-icon', href: (process.env.ROUTER_BASE_PATH || '') + '/favicon.ico' },
-      { rel: 'apple-touch-icon', href: (process.env.ROUTER_BASE_PATH || '') + '/icon256.png' }
+      { rel: 'apple-touch-icon', href: (process.env.ROUTER_BASE_PATH || '') + '/ios_icon.png' }
     ]
   },
 


advplyr:
The issue with this is the icon currently looks great on Android. From what I read it looks like you can define an `apple-touch-icon` in the header and it will take precedence over the icon defined in the manifest. I haven't tested this but if it is the case I would prefer that so the android icon stays the same.

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/CycleTracker/Manifest_file#app_iconography

> In Safari, and therefor for iOS and iPadOS, if you include the [non-standard apple-touch-icon](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML#adding_custom_icons_to_your_site) in the [<head>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head) of the HTML document via [<link>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link), they will take precedence over manifest-declared icons.

advplyr:
That would be put in the `link` array underneath the favicon in `nuxt.config.js` if you want to test that out.

nichwall:
For reference, here's a comparison between native app and PWA on Android.

advplyr:
More importantly is if in iOS the icon defined there overrode the icon defined in the manifest. That link should only impact iOS.

nichwall:
Sorry if I caused confusion. The screenshot I shared was how the PWA currently shows up on 2.7.2, I will try again with this PR.

nichwall:
I'm not sure if I did it right (since I'm usually developing in a dev container and had to switch to running from source on my desktop to make the server available over the network), but this is how it looks right now.

The left one is the 2.7.2 PWA, the right one is the PWA from this. Not sure if the little Chrome logo matters or is because I had two PWA?

advplyr:
Thanks for testing that. With brave browser I always get the little icon so I think it's normal. I've never seen it say "2" in front of the app title though.
@ipcintron Can you rename the icon to `ios_icon.png`? Then this is ready to go

nichwall:
> Ive never seen it say "2" in front of the app title though.

Chrome let me name it, so I put the 2 in front so I could tell them apart. It defaulted to "Audiobookshelf"

advplyr:
Thanks!

PR #2620, opened 2024-02-16
PT-BR Strings

pmangro:


pmangro:
Added PT-BR strings and option on i18n

advplyr:
Thanks!

PR #2638, opened 2024-02-21
Updated and fixed Spanish translation

DownloadableFox:
Several strings in Spanish were not updated. This commit translates all strings to Spanish as well as fixing maybe one or two already existing strings.


DownloadableFox:
Feel free to tell me if there are any changes you would like me to do!

advplyr:
Looks good, thanks!

PR #2643, opened 2024-02-22
feat(i18n): add missing localizations

megamegax:
I've noticed, that there are a lot of hardcoded strings in the project. I've started to extract them into localized strings as that is something anyone can do and benefits everyone at the end :)

This is the first batch, I plan to do it in small chunks a few files a time.

May I ask wether filling the existing localization jsons with the new strings is necessary or I can just skip that step in the future?

Also, do you approve the initiation or should I just leave it?


advplyr:
The keys need to be in alphabetical order. Can you also only add the keys to `en-us.json` and another language if you are translating it because it makes it easier to merge. It looks like some more keys have been added since this was opened if you want to do a merge with master

advplyr:
I included these here https://github.com/advplyr/audiobookshelf/commit/6462a50713340bbb38be457c863f257f465fecdc and https://github.com/advplyr/audiobookshelf/commit/8c6c43657cc8ef04d17eea5dd290104e33729f69
Thanks!

PR #2648, opened 2024-02-22
feat(i18n): added missing translation strings to item detail page and added german translation strings

Teekeks:


advplyr:
I included these here https://github.com/advplyr/audiobookshelf/commit/6462a50713340bbb38be457c863f257f465fecdc
Thanks!

PR #2658, opened 2024-02-25
feat(i18n): made the empty library page translatable on home page

Teekeks:
This PR makes the empty library message and buttons translatable on the home and library page


advplyr:
Some of these were added already so I added the last one https://github.com/advplyr/audiobookshelf/commit/4b7b10a901c870af1b2d751052a51dadb0551501

Issue #1889, opened 2023-07-03
[Enhancement]: New filter for ebooks


Demian98:
### Describe the feature/enhancement

The follow filters are available to find audiobooks with ebooks or supplementary ebooks:
![grafik](https://github.com/advplyr/audiobookshelf/assets/77210091/619823eb-2cac-42a1-a3be-057fcc057138)

It would be great to have the opposites too:

- has no ebook
- has no supplementary ebooks

advplyr:
Added in [v2.8.1](https://github.com/advplyr/audiobookshelf/releases/tag/v2.8.1)

PR #2677, opened 2024-02-27
feat: Expanded filter to include "has no ebook" and "has no supplementary ebooks" options

Teekeks:
This PR adds two more filter options: "has no ebook" and "has no supplementary ebooks" as negations of the two existing ebook related filter options.

This PR implements #1889


nichwall:
Tested and working for me with both ebook and supplementary ebooks.

advplyr:
Great, thanks!

PR #2694, opened 2024-03-01
Client side cover image caching

mikiher:
This adds client-side caching for book and podacst cover images.

Caching is performed only for timestamped requests, with cachce-control header set to 'private, max-age=86400'. This means:
- Caching is only done on the client, and not in any proxies. I chose this because currently the URL contains the authorization token, which somewhat of a security flaw, and also makes caching useful only for a specific client and user.
- Caching TTL is one day. Later we can add 'immutable', since the timestamp is updated when the image changes on the server, so cache-busting is possible.

Most important cases in which cover images are displayed on the client already have server-originated timestamp (updatedAt). I [updated a couple of places](https://github.com/advplyr/audiobookshelf/commit/4fe672f09d794c05d31a9375b2a686ed1c0dfe10) where updatedAt was available but not used.

Depending on various parameters (http vs. https, local vs. remote), this can reduce image loading latency by a factor of 2-4.

Example:
Home page before cover caching (home network, HTTP):
Screenshot 2024-02-26 131331
Home page after cover caching (home network HTTP):
Screenshot 2024-02-26 131847

You can see covers load latency goes down from ~500ms to ~200 ms. 

Another example:
Home page before cover caching (home network though reverse proxy, HTTPS):
Screenshot 2024-02-26 134125
Home page after cover caching (home network though reverse proxy, HTTPS):
Screenshot 2024-02-26 134518

Here you can see covers load latency go down from ~270ms to about 70ms

The gains would probably be even more impressive if measured over remote connections.

### DIFF
--- a/client/components/modals/item/tabs/Match.vue
+++ b/client/components/modals/item/tabs/Match.vue
@@ -49,8 +49,8 @@
             </div>
             <div v-if="media.coverPath">
               <p class="text-center text-gray-200">Current</p>
-              <a :href="$store.getters['globals/getLibraryItemCoverSrcById'](libraryItemId, null, true)" target="_blank" class="bg-primary">
-                <covers-preview-cover :src="$store.getters['globals/getLibraryItemCoverSrcById'](libraryItemId, null, true)" :width="100" :book-cover-aspect-ratio="bookCoverAspectRatio" />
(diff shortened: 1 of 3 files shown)


advplyr:
Great, thanks!

PR #2701, opened 2024-03-03
Fix input width in MultiSelect UI components

### LINKED ISSUE #2700: [Bug]: Author text input gets clipped when editing book details
### Describe the issue

When you manually type a new author name, after typing a few letters, the ending gets clipped and you don't see what you type until you press Enter (and then the Authors field is populated correctly with the full author name you typed).

While typing, all the letters you typed briefly flicker on and then get truncated.

1. Open a book's details tab
2. Go the Authors field
3. Add a new author by typing it manually (e.g. "J. K. Rowling")
4. After typing the name, but before hitting Enter, the field looks like this:
Screenshot 2024-03-03 224218

5. After hitting Enter, The field is correctly populated, like this:
Screenshot 2024-03-03 224600

### Audiobookshelf version

v2.8.0

Built from source

mikiher:
Fixes #2700 (see details in the bug).

Note that setInputWidth was removed, and with it the call to this.RecalcMenuPos(). 
I'm not sure why this call is required. If it is, let me know and I'll try to re-incorporate it where it makes sense.

### DIFF
--- a/client/components/ui/MultiSelect.vue
+++ b/client/components/ui/MultiSelect.vue
@@ -11,7 +11,7 @@
             </div>
             {{ item }}
           </div>
-          <input v-show="!readonly" ref="input" v-model="textInput" :disabled="disabled" style="min-width: 40px; width: 40px" class="h-full bg-primary focus:outline-none px-1" @keydown="keydownInput" @focus="inputFocus" @blur="inputBlur" @paste="inputPaste" />
+          <input v-show="!readonly" ref="input" v-model="textInput" :disabled="disabled" style="min-width: 40px; width: fit-content" class="h-full bg-primary focus:outline-none px-1" @keydown="keydownInput" @focus="inputFocus" @blur="inputBlur" @paste="inputPaste" />
         </div>
       </form>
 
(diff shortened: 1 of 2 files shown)


advplyr:
I'm not sure why I put the recalc there but I wrote that a long time ago so it may have gone through a few refactors. This is working great in my testing, thanks!

mikiher:
Actually, I played with it a bit more today, and I'm not happy with my fix.

1. I examined in more detail what is the actual width after each letter is entered. It turns out that setting "width: fit-content" just sets a larger initial value for the width (163px, not sure how it's calculated), and it doesn't change at all when the input changes. It seems like it works because of the larger initial width, but after you type enough, you only see the end what you typed and the beginning is cut-off.
2. I think I now understand why recalcMenuPos is required - assuming width modification really works as expected, you could potentially have the inputWrapper div span more than one line, and the menu position needs to change when this happens.

I think I found another solution that requires less changes and really modifies the width.
Can you please revert the merge?

mikiher:
OK, don't bother to revert, I'll just send a new PR

PR #2706, opened 2024-03-04
📝 Docker Compose and Nginx Proxy Manager setup guide.

fogolin:
Helping with **app documentation**, so, in this PR we're creating a `README` folder that is supposed to hold setups, guides, images and whatever else might be needed for this repository to work properly. As a first contribution, I'm adding a guide on how to properly host it using **Docker** and **Nginx Proxy Manager** (as I couldn't find how to make it work at first, for some reason), but more people are welcome to join and add their own contributions to it.


fogolin:
Also added it to the website guides:
https://github.com/audiobookshelf/audiobookshelf-web/pull/82

nichwall:
Looks good, but I think this specific guide fits more on the website (which you have also written) instead of being part of the main server repo, because the instructions here are more of "how you can set the options for a reverse proxy" instead of "how to set up a reverse proxy".

advplyr:
Thanks! I agree with @nichwall this belongs on the website and don't want to have to two copies of docs to keep up-to-date

PR #2721, opened 2024-03-08
Add keyboard navigation to multi-select components

### LINKED ISSUE #2555: [Enhancement]: Allow keyboard navigation and selection in drop-down lists

There are many places in the product in which you can select an item from a drop-down list (authors, genres, tags, etc...).
Currently the only way AFAIK to select an item from the list is to click on it. 

I'd like to ask for support for keyboard navigation and selection - it is quite annoying to switch between keyboard and mouse, especially when adding tags and other metadata fields.

mikiher:
This (partially) fixes #2555

Adds support for keyboard navigation and selection in the dropdown menu of multi-select components.

Details:
- Support is added for ArrowDown and ArrowUp for navigation
  - ArrowDown/Up highlights the next/previous menu item
- By default, no menu item is highlighted
  - If no menu item is highlighted, Enter simply adds the current text input to the selected list
  - If a menu item is highlighted, Enter adds/removes the highlighted item to/from the selected list
- Highlighted menu element is marked with `text-sky-400` (blue text color)
- Menu scrolling is properly supported
  - After ArrowDown/Up, scroll position is modified to keep the highlighed item visible.
- Mouse navigation and selection continues to work normally as expected

Notes:
1. Using blue text color for highlighting seemed like a reasonable choice, but I'm not a designer - I'll change it to whatever you think is best.
2. There's some refacroring work required to make this code more reusable. I'll do it in a future PR.


advplyr:
I updated the highlight to `yellow-300` in keeping with the theme color. It is working great for me, thanks!

PR #2724, opened 2024-03-09
Fix library filter data direct access

### LINKED ISSUE #2693: [Bug]: Server crashes
### Describe the issue

FATAL: [Server] Unhandled rejection: TypeError: Cannot read properties of undefined (reading 'authors'), promise: Promise {
audiobookshelf  |   <rejected> TypeError: Cannot read properties of undefined (reading 'authors')
audiobookshelf  |       at BookScanner.scanNewBookLibraryItem (/server/scanner/BookScanner.js:446:86)
audiobookshelf  |       at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
audiobookshelf  |       at async LibraryItemScanner.scanNewLibraryItem (/server/scanner/LibraryItemScanner.js:178:24)
audiobookshelf  |       at async LibraryScanner.scanFolderUpdates (/server/scanner/LibraryScanner.js:584:30)
audiobookshelf  |       at async LibraryScanner.scanFilesChanged (/server/scanner/LibraryScanner.js:385:33)
audiobookshelf  | } (Server.js:166)

1. Matching a book using audible.com

### Audiobookshelf version

v2.8.0

Docker

mikiher:
Fixes #2693 (see details in the bug).

Replaces all direct access to Database.libraryFilterData (which caused the crash in the bug) with Database method calls that check if the cached libraryFilterData exists for the given libraryID, and perform a database query if it doesn't exist.


advplyr:
Thanks for digging into that. I updated those functions to only return the `id` because the filter data only stores the `id` and `name`.

PR #2728, opened 2024-03-11
Automating localization

nichwall:
This PR includes 3 new workflows:
- `alphabetization_check` which performs an integration test to verify that all localization keys are alphabetical for all commits and PRs
- `i18n-automatic-branch` which automatically runs the key-copying script on pushes to the main/master branch and opens a branch with the updated files if there were any changes
- `i18n-manual-branch` which can be manually triggered

The `branch` workflows are identical other than the trigger condition to allow for disabling the `automatic` workflow without removing the ability to run the `manual` workflow. These two workflows will need write access to be able to create the branch and PR. These also require defining the `USER_EMAIL` and `USER_NAME` secrets for the git initialization. These don't necessarily  to be secrets, but felt better to do that instead of just including a default value.

These workflows make use of the https://github.com/audiobookshelf/audiobookshelf-i18n-updater GitHub action.

Edit: The action linked above ensures a newline at the end of each i18n file so the last line stops showing up as a change when viewing the diff.

### DIFF
--- a/.github/workflows/alphabetization_check.yml
+++ b/.github/workflows/alphabetization_check.yml
@@ -1,0 +1,29 @@
+name: Verify all i18n files are alphabetized
+
+on:
+  push:
+    paths:
+      - client/strings/** # Should only check if any strings changed
+
+jobs:
+  update_translations:
+    runs-on: ubuntu-latest
+    steps:
+    # Check out the repository
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    # Set up node to run the javascript
+    - name: Set up node
+      uses: actions/setup-node@v4
+      with:
+        node-version: '20'
+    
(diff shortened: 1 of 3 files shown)


nichwall:
I was just reviewing the branch workflow, and realized it seems to have been copied wrong. I'll update that tonight so it's actually triggered manually instead of on every push.

nichwall:
Should I change this PR to just be the check (not automatically running the script that opens a branch) so it's easier to review, and then worry about the automatic branch workflow later?

advplyr:
That would work because I could merge that without digging in

nichwall:
This PR has been broken up into two separate PRs for easier review.

Integration workflow: https://github.com/advplyr/audiobookshelf/pull/2780
Running the script on a new branch and opening a PR: https://github.com/advplyr/audiobookshelf/pull/2781

PR #2731, opened 2024-03-11
Extra logging

hydrian:
Allowing environment variable LOG_LEVEL to be set even in production environment.


advplyr:
Log level can be set on the log page in the web client. 

I know in your case you can't access that because the database isn't initializing but there are no additional logs from Abs that you aren't seeing. At the moment setting that log level when the server isn't starting wouldn't provide a benefit.

The log level is a number so the env variable would need to be cast for this to work but again I don't see where the benefit is in adding that

PR #2733, opened 2024-03-12
Added isbn to CustomProviderAdapter

kaldigo:
This adds the isbn to the CustomProviderAdapter so that it can be used as a identifier like the asin is used by Audible.


advplyr:
Thanks!

Issue #2380, opened 2023-12-02
[Bug]: metadata.json loader misinterprets `series` field when series doesn't exist


wtanksleyjr:
### Describe the issue

When scanner finds a new book with a metadata.json file, it doesn't parse the `series` field correctly if the series doesn't already exist. For example, suppose the file contains `"series": [ "The Pact #1"]`.

If there's already a series named `The Pact` defined, this will add a new book with its ordering=1 (as expected). But if there's no series defined with that name, the loader will create a new series named `The Pact #1` (so it parses the ordering as part of the series name).

### Steps to reproduce the issue

1. Put an audio file and a metadata.json file together as indicated above, using a series name not already present on your setup.
2. Move the folder containing both into the path of the scanner.
3. Edit the resulting book, and edit its Series field. Observe that the whole string is placed in the series name.
4. Delete the book entry and clean up.
5. Add the same series name to some other book (by hand).
6. Repeat steps 1-2 again. Observe that this time, the book's series loads as expected.

### Audiobookshelf version

2.6.0

### How are you running audiobookshelf?

Docker

wtanksleyjr:
> When scanner finds a new book with a metadata.json file, it doesn't parse the series field correctly if the series doesn't already exist. For example, suppose the file contains "series": [ "The Pact #1"].

My mistake. In order to trigger this bug, the series number HAS to include extra text. For an actual example, "Corum #4, Dramatized Adaptation" will trigger this (that's by Moorcock BTW). The intent of this extra text is to allow a series to include multiple editions of each book, so you might have Corum #1 as narrator-only, while #4 might be available both narrated and Graphical Audio. Audible actually includes this text inside the series number, and ABS does support that when doing a Match.

So it turns out your parser's perfectly fine with "The Pact #1" and so on; it's only when you include more stuff in the series-order field that it gets hairy.

Sorry about that, I hope this helps.

advplyr:
Just to clarify, in some series you are storing the sequence as "4, Dramatized Adaptation"?

I'm not sure yet if we should support this now that we have moved to sqlite because in order to sort the sequences properly we have to cast it to a number. Currently there is no restriction as to what you can put in there but I think this should be changed to require it to be a number I just haven't thought of a good way to roll that out without breaking things for users already using alphanumeric.

advplyr:
The parser that is pulling out the sequence number doesn't work for `Corum #4, Dramatized Adaptation` specifically because of the spaces.
[...]

PR #2734, opened 2024-03-12
Make series sequence cleanup slighlty less aggressive

### LINKED ISSUE #2710: [Bug]: Incorrect handling of some series metadata from audible.co.jp
### Describe the issue

[Link to the series](https://www.audible.co.jp/series/%EF%BC%9C%E7%89%A9%E8%AA%9E%EF%BC%9E%E3%82%B7%E3%83%AA%E3%83%BC%E3%82%BA%E3%82%B7%E3%83%AA%E3%83%BC%E3%82%BA/B08WH8LJF5
)
when fetching the metadata of this series the cardinal numbers are discarded.

> シリーズ： ＜物語＞シリーズ , Book ファーストシーズン : 1 

is rendered as 

> ＜物語＞シリーズ #ファーストシーズン 

expected rendering:

> ＜物語＞シリーズ #ファーストシーズン : 1 

or 

> ＜物語＞シリーズ # 1

1. match a book to anything in the above series

### Audiobookshelf version

2.8.0

Docker

mikiher:
Fixes #2710.

The fix makes the sequence cleanup a bit less aggressive, so that the sequence number, which translates to: "first-season : 1", does not get trimmed to "first-season". 

The fix to #2380 still works.

### DIFF
--- a/server/providers/Audible.js
+++ b/server/providers/Audible.js
@@ -29,8 +29,9 @@ class Audible {
      */
     cleanSeriesSequence(seriesName, sequence) {
         if (!sequence) return ''
-        let updatedSequence = sequence.replace(/Book /, '').trim()
-        updatedSequence = updatedSequence.replace(/(\d+)(, .*)/, '$1').trim()
+        // match any number with optional decimal (e.g, 1 or 1.5 or .5)
+        let numberFound = sequence.match(/\.\d+|\d+(?:\.\d+)?/)
+        let updatedSequence = numberFound ? numberFound[0] : sequence
         if (sequence !== updatedSequence) {
             Logger.debug(`[Audible] Series "${seriesName}" sequence was cleaned from "${sequence}" to "${updatedSequence}"`)
         }
--- a/server/providers/Audible.js
+++ b/server/providers/Audible.js
@@ -30,9 +30,7 @@ class Audible {
     cleanSeriesSequence(seriesName, sequence) {
         if (!sequence) return ''
         let updatedSequence = sequence.replace(/Book /, '').trim()
-        if (updatedSequence.includes(' ')) {
-            updatedSequence = updatedSequence.split(' ').shift().replace(/,$/, '')
-        }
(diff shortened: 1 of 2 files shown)


advplyr:
I think we should adjust that `cleanSeriesSequence` function to take the first number. Sqlite cannot do natural sorting I've been thinking of forcing sequence to be a number.
 
Before moving to sqlite all of the series sequence sorting was done in memory using natural sort but now with sqlite we are casting the sequence to int before sorting.

There is an option I saw of extending sqlite to support natural sorting but I'm not sure it's necessary yet. The only reason I haven't forced series to be a number already is because many users may still be using alphanumeric sequence so it would be a breaking change.

mikiher:
I'm not sure I agree here. We obviously have many cases (I quite easily found a few of those in Japanese series) where the sequence is not just a number. We should not throw that information away. In this case specifically, it is really meaningful.

Do we have to use sqlite sorting when sorting series order? we're talking here about maybe tens of books per-series - it's not a performance issue to sort them in memory (but maybe I'm missing the use case that's bothering you).

I say let's try not have constraints dictate the design. My 2c.

advplyr:
When switching to sqlite I brought this up in Discord and several users brought to my attention book series that have thousands of books in them.

If a series has thousands of books the series page needs to have paginated queries. For paginated queries of series books we need to sort the sequence in the query.

One other big issue is the collapse series option. I have wanted to remove that option when moving to sqlite because it is difficult to handle with sqlite queries but it is too popular of a feature to remove now.

We could add a separate column that is the numerical sequence just used for sqlite querying, but that would be ugly and a pain to maintain.

advplyr:
I missed the Continue Series home page shelf. That also needs to sort by sequence in the query to get the next book not finished.

mikiher:
Ok, I'll change it to extract the first number.

What if no number exists in the sequence? Leave it as-is?

mikiher:
OK, ptal again.

advplyr:
Great, thanks for adding the test also

Issue #2687, opened 2024-02-28
[Enhancement]: Continue Series to show only books with sequence higher than a finished book in the same series


justcallmelarry:
## Suggestion

When listening to a series like Wheel of Time for example, there is a prequel book (book#0) which always lists in the "Continue Series" shelf unless it has been marked as finished.

If I do not listen to the book, but instead start at book#1 and want to continue in the series, book#0 will always be shown first on the "Continue Series" shelf.

I would wish for the "Continue Series" shelf (either by default or with a setting on library level for example) to only show me the lowest sequence of the series that is still higher than my highest finished, ie in this example it would show book#2.

In my case this makes it so that I do not get new books of the series showing up in the shelf when they are released.
If I have listened to book#1-3 in series X somewhere else and listen to book#4-6 on ABS, then book#7 is released book#1 will still show up in the shelf.

I know I can mark them as played, but this also messes with my stats, so I would rather leave them un-listened to since this is also the truth for the service in particular.

### Things to consider
* What should be shown (if anything) when you reach the last book of the series? My preference would be to not show it since you are caught up with the last book, but if it circles around I would also be OK with that.
* This would not really help re-listeners, but I think the Continue Series shelf is not great for that right now anyway (unless I'm doing something incorrectly)
* Should this be default-/library-/user-behaviour? (breaking change if default)

theboo:
For my personal usecase, an option(not default) to make it always show the next unread book in the series rather than the first unread book in the series would be perfect... But man, on discord there's sure a lot of people with a lot of different scenarios for this stuff lol, would be hard to please them all I think.

justcallmelarry:
I was tinkering around a bit with it and came to the following conclusions:

## Default behaviour
I tried adding functionality which first tries to get the "next" unread book in the series and if there are none, it loops around to just take whatever the first unread book in the series is.

### Positives

This was pretty easy to implement!
No translations needed.

### Negatives

Breaking change which not everyone might like.

## User setting

### Negatives

Can't really find many user specific settings, so did not spend a lot of time on figuring out how to store it and gave up on this option more or less.

## Library setting

### Positives

This was also pretty easy to implement (I just made a switch just like for hide single book series)
Allows for changes behaviour (not looping back to first unread when no next unread is available)
Does not change default behaviour.

### Negatives

Would need translations for the setting in all languages. Not sure how this is done generally for new copy?

## Additional thoughts

[...]

PR #2737, opened 2024-03-12
Add library toggle setting for skipping earlier instalments in Continue Series

justcallmelarry:
## Issues
Closes https://github.com/advplyr/audiobookshelf/issues/2687 on merge.

## Open questions
* Are there any linting rules I should make sure I follow?
* Did I miss running some script for the other translations?
* Does anyone know of a better name than `onlyShowLaterBooksInContinueSeries` for the setting variable name?
* Does anyone have better copy suggestions for the toggle + info text?

## Pictures
<details>
  <summary>Picture in library, second book read, third up next (first is unread but not next in continue series)</summary>
  image
</details>
<details>
  <summary>Book 2 and 3 read, book 1 still not showing up in Continue Series</summary>
  image
</details>
<details>
  <summary>Library settings page</summary>
  image
</details>

### DIFF
--- a/server/utils/queries/libraryItemsBookFilters.js
+++ b/server/utils/queries/libraryItemsBookFilters.js
@@ -659,7 +659,7 @@ module.exports = {
       [Sequelize.literal('(SELECT max(mp.updatedAt) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.userId = :userId AND bs.seriesId = series.id)'), 'recent_progress'],
     ]
     if (library.settings.onlyShowLaterBooksInContinueSeries) {
-      includeAttributes.push([Sequelize.literal('(SELECT CAST(max(bs.sequence) as FLOAT) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.userId = :userId AND bs.seriesId = series.id)'), 'maxSequence'])
+      includeAttributes.push([Sequelize.literal('(SELECT CAST(max(bs.sequence) as FLOAT) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.isFinished = 1 AND mp.userId = :userId AND bs.seriesId = series.id)'), 'maxSequence'])
     }
 
     const { rows: series, count } = await Database.seriesModel.findAndCountAll({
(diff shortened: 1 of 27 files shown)


server/utils/queries/libraryItemsBookFilters.js
```diff
@@ -676,11 +676,14 @@ module.exports = {
       ],
       attributes: {
         include: [
-          [Sequelize.literal('(SELECT max(mp.updatedAt) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.userId = :userId AND bs.seriesId = series.id)'), 'recent_progress']
+          [Sequelize.literal('(SELECT max(mp.updatedAt) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.userId = :userId AND bs.seriesId = series.id)'), 'recent_progress'],
+          [Sequelize.literal('(SELECT CAST(max(bs.sequence) as FLOAT) FROM bookSeries bs, mediaProgresses mp WHERE mp.mediaItemId = bs.bookId AND mp.userId = :userId AND bs.seriesId = series.id)'), 'maxSequence'],
+          [Sequelize.literal('(SELECT json_extract(libraries.settings, "$.onlyShowLaterBooksInContinueSeries") FROM libraries WHERE id = :libraryId)'), 'onlyShowLaterBooksInContinueSeries']
```
advplyr, line 681: This is unnecessary because we can get the library settings from the function that calls this one. The library in `/objects/Library` is passed into the function that calls this one and has all the settings.

This way we can also check if that setting is set before adding the additional query.

justcallmelarry: great catch! i 👍 

changed so that the entire library object is sent into the function instead, and that it will only make the maxSequence SQL query if the setting is turned on

also found a `var` that i used unconsciously (since I code way more golang than js…) that I changed to a let

advplyr:
I had to update the query attribute to make sure the max sequence is for finished books. This is because it is possible to have media progress for books that are not finished and not in progress.

This is working well, thanks!

justcallmelarry:
> I had to update the query attribute to make sure the max sequence is for finished books. This is because it is possible to have media progress for books that are not finished and not in progress.
> 
> This is working well, thanks!

Ah yeah, I thought about that, but I chose to just base it on the "`recent_progress`" query, and thought that since the series are disqualified if they have anything in progress (by another part of the query) this would work out to the same result anyway. But I like the change, feels like it should be more future proof this way.

Thanks for the responses and pointers for a first time contributer!

advplyr:
Yeah I thought that was your reasoning since it's logical but it doesn't catch the case I mentioned where a book has media progress but it is not in progress.
The first series I tested with had a book like that because it was also an ebook and so continue series was showing the wrong next book.

PR #2738, opened 2024-03-12
Auth: Fix crash on missing logout URL

Sapd:
When using OpenID

Also added debug information on openid errors (so it does not just show OBJECT object)


advplyr:
Thanks!

PR #2740, opened 2024-03-13
Add name labels to login form

Schiriki123:
To increase the likelihood for password managers to detect the username field, I added the `name` attribute to the username input field (and for the password filed too because it cannot hurt).

P.S.: I don't know if this is the right way of implementing this (I never worked with vue.js before).


advplyr:
That works, thanks!

PR #2742, opened 2024-03-14
Fix broken BinaryManager.isBinaryGood test

mikiher:
For some reason the test was broken. 

Not sure how I didn't notice it when I first submitted PR #2554.

I should really add the unit test runs to the PR check flow - will do this in a separate PR.


advplyr:
Thanks and yeah that would be best

PR #2745, opened 2024-03-15
Unit tests flow

mikiher:
This is pretty self-explanatory:

- Trigger on pull_request or push (can also run manually on any ref)
- Checkout
- Run npm ci
- Run npm test


mikiher:
OK, so I believe the failing test can be safely ignored, because it triggers on a pull_request on your reposity, but is running on my repository which does not have this pull_request.

advplyr:
Thanks!

PR #2747, opened 2024-03-15
fix scroll to the current chapter fail

fengqi:
While using `audiobookshelf` in my environment, I encountered an issue where automatic scrolling to the current chapter wasn't functioning. Upon investigation, I discovered that the value of `document.getElementById(chapter-row-${this.currentChapterId})` within the `scrollToChapter()` method was returning null. After adding a `setTimeout`, I observed that the issue was resolved.

As I'm not well-versed in Vue, this solution may not be optimal.

my environment details:
audiobookshelf: 2.8.0
chrome version: 122.0.6261.112

### DIFF
--- a/client/components/modals/ChaptersModal.vue
+++ b/client/components/modals/ChaptersModal.vue
@@ -36,7 +36,9 @@ export default {
   },
   watch: {
     value(newVal) {
-      this.$nextTick(this.scrollToChapter)
+      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks for finding that. I fixed this in commit https://github.com/advplyr/audiobookshelf/commit/630ece82add1ed101e598ed67d3ee0938a39d7bd by putting the scroll call in the component updated hook instead.

PR #2756, opened 2024-03-16
Add Hebrew translation json and Hebrew to i18n.js

arcmagedr:
Add Hebrew translation json and Hebrew to i18n.js


advplyr:
@mikiher I guess that checkout action isn't going to work with PRs

mikiher:
Darn, sorry, I'll fix this ASAP

advplyr:
No rush, thanks!

arcmagedr:
I've gone over all the translation and verified everything makes sense, at least in the context of the file itself.
Once merged, if anything will need changing I'll update as necessary.

One thing though, is there handling for right-to-left languages? Is there anything else that needs to be done in that regard?

nichwall:
I don't think anything has been done for right to left languages. Looks like at a minimum the `dir` tag would need to be added everywhere a string shows up. Not sure if there's anything else that needs to be done for that.

https://www.w3.org/International/questions/qa-html-dir

advplyr:
Thanks! Yeah we will have to setup for rtl languages still

PR #2757, opened 2024-03-17
Change unit tests workflow to include conditional checkout steps

mikiher:
Separates workflow_dispatch and non-workflow_dispatch checkouts into two separate conditional steps.

The initial workflow in PR #2745 was broken when triggered by pull requests.


advplyr:
Thanks!

PR #2780, opened 2024-03-22
Add i18n integration workflow

nichwall:
This workflow runs an integration check to make sure that i18n JSON files have keys in an alphabetical order.


advplyr:
Thanks!

PR #2781, opened 2024-03-22
I18n script runner

nichwall:
This PR includes 2 new workflows:

- `i18n-auto-branch` which automatically runs the key-copying script on pushes to the main/master branch and opens a branch with the updated files if there were any changes
- `i18n-manual-branch` which can be manually triggered

The branch workflows are identical other than the trigger condition to allow for disabling the automatic workflow without removing the ability to run the manual workflow.

These two workflows will need write access to be able to create the branch and PR. These also require defining the USER_EMAIL and USER_NAME secrets for the git initialization. These values don't necessarily need to be secrets, but felt better to do that instead of just including a default value in the action.

These workflows make use of the https://github.com/audiobookshelf/audiobookshelf-i18n-updater GitHub action. The script this action runs ensures a newline at the end of each i18n file so the last line stops showing up as a change when viewing the diff when the number of lines changes in an i18n file.

### DIFF
--- a/.github/workflows/i18n-auto-branch.yml
+++ b/.github/workflows/i18n-auto-branch.yml
@@ -1,0 +1,65 @@
+name: Automatic update i18n files
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main # Only care about the main branch, don't care about any PRs or forks
+      - master
+    paths:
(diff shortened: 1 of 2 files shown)


nichwall:
This workflow is no longer needed due to Weblate being used to manage translations.

PR #2784, opened 2024-03-23
Fix custom metadata provider crash

rasmuslos:
The method `search(title, author, isbn, providerSlug, mediaType)` (`CustomProviderAdapter#17`) expects five arguments but only receives four. This causes a crash because no custom metadata provider `book` (passed as forth argument, not fifth) does exist


advplyr:
Thanks!

PR #2789, opened 2024-03-25
Fix book limit for the Continue Series shelf (with skip earlier books toggle active)

justcallmelarry:
Currently the Continue Series shelf (with skip earlier books toggle active) returns some weird results if you have more than 10 series which has a read book, but not all of them have a book that qualifies for the constrains of being of a higher sequence number than the ones already finished.

This is because the sql query returns the 10 series with most recent progress that fit for the untoggled library setting.

In order to solve the issues caused by this, I added the same sql query that calculates the `maxSequence` number to the original where-query, so that it will only return series that we know has books in them that are of a higher sequence than the ones already read.

The post-query code is still there, since the query still returns books that are below the `maxSequence` number.

Unfortunately the `maxSequenceQuery` is run twice against the db, but I did not figure out a better way of running the query in the current code base.

I am open to suggestions on improvements, though!


advplyr:
I can't think of a better way to do that either without refactoring that a bit. These queries are something I have planned on refactoring since they were put together to support the old data model. I am hesitant on making these any slower than they already are but I don't see another fix for now.
Thanks!

PR #2797, opened 2024-03-28
Add dir="auto" attribute where RTL display is needed

mikiher:
This adds dir="auto" attribute to some text elements, in which it makes sense to modify the display when the text is in a RTL language (e.g. Arabic, Hebrew, Urdu, etc.)

The change has no effect on LTR text. The main visible difference in RTL text is that it is naturally aligned to the right (and there are also some subtler differences in the rendering of mixed text containing both RTL and LTR characters).

Some examples after the change:

In the Latest Episodes page:
Screenshot 2024-03-29 000604

In the Recently Added row on the Homepage:
Screenshot 2024-03-29 000937

You can see in both examples how Hebrew text is right-aligned (as expected), while English text remains as-is.

*Note*: this is *not* an attempt at full RTL support for the Hebrew locale (which will likely require more work).

### DIFF
--- a/client/components/cards/LazyBookCard.vue
+++ b/client/components/cards/LazyBookCard.vue
@@ -6,7 +6,7 @@
     </div>
 
     <!-- Alternative bookshelf title/author/sort -->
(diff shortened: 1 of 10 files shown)


advplyr:
Great start, thanks!

PR #2803, opened 2024-04-01
OpenAPI Spec, try 2

nichwall:
This is another attempt to do the OpenAPI spec better than initially done in https://github.com/advplyr/audiobookshelf/pull/2671. That PR focused on the Collections endpoints, but this PR focuses on a stripped down version of the Author endpoints to make reviewing easier (the missing schemas will be added back in future PRs).

I changed from OpenAPI 3.1 to OpenAPI 3.0 due to better tool support as of March 2024. I still prefer 3.1, but migrating will be easier than writing it all from scratch twice.

I opted to create a new directory named `docs/` for the API spec files which matches the source code structure. While `swagger-jsdoc` did work well, it makes editing the schemas more difficult because VSCode cannot assist with any formatting in the block comments. Separating the documentation and code into their own files but keeping it in the same repository should still help keep the API docs up to date as unit/integration testing is added.

Initial research shows other projects tend to keep their OpenAPI spec in the same repository as the source code.

I am using the `vacuum` tool for bundling and linting the spec. Running the tool is discussed in `docs/README.md`.


advplyr:
Great start, thanks!

PR #2805, opened 2024-04-01
Add client name to possible device info lines

rasmuslos:
Right now the client name is not visible in the `Device Info` section of a listening session, this PR adds it a possible line


advplyr:
Thanks! This also made me notice that the android app wasn't sending the correct client version string.

PR #2820, opened 2024-04-03
Add option to set Signing Algorithm for OpenID Authentification

apocer:
Currently the ID Token returned from a OIDC Provider needs to be signed using the `RS256` algorithm. This is quite restrictive. I had trouble using my Let's Encrypt certificates for signing the ID token with Authentik, because Let's Encrypt uses `ES256`.

**This PR adds a setting for the signing algorithm, which is used to configure the Passport OpenID client.**

What changed:
- Added `authOpenIDTokenSigningAlgorithm` to `ServerSettings` with `RS256` being the default
- Added `Signing Algorithm` Field in the authentication settings of the WebUI
- The `Signing Algorithm` is also fetched by the `Auto-populate` button

grafik


apocer:
Documentation is updated in [audiobookshelf-web #86](https://github.com/audiobookshelf/audiobookshelf-web/pull/86)

advplyr:
I think it would be better if that was a dropdown of signing algorithms. We can also get the supported algorithms when users use the autopopulate button.
```
  "id_token_signing_alg_values_supported": [
    "RS256"
  ],
  ```
  @Sapd and I discussed this in Discord

apocer:
I added the following changes:

- A text input is shown as long as the "Auto-populate" button is not pressed
- A dropdown is shown when a list of supported signing algorithms is returned by the OIDC issuer
- A text input is shown when the OIDC server does not return a list of supported algorithms

@Sapd please have a look and make sure to check the Vue specific stuff. I never worked with Vue or nuxt before.

advplyr:
Thanks!

apocer:
Your welcome and thank you for this great project!

PR #2821, opened 2024-04-04
Separator between multiple series

lkiesow:
If a book is part of multiple series, this patch adds a separator between the series on the library item details page. With no separator, it is not immediately clear that they are separate series.

Screenshot from 2024-04-04 21-52-11


advplyr:
I'm not sure how I never noticed that, thanks!

PR #2877, opened 2024-04-24
add iTunes podcast regions for all ABS supported languages

v3DJG6GL:
This PR adds iTunes podcast regions for all languages that ABS currently supports.


advplyr:
Have you checked that these are supported by iTunes?

advplyr:
Thanks!

PR #2880, opened 2024-04-26
Added toggle button to upload files individually

### LINKED ISSUE #2766: [Enhancement]: Split upload filewise

Hey folks,

Just installed ABS and hit a snag with Cloudflare’s 100MB upload limit. I’m trying to upload a bunch of small files in a folder, but since ABS bundles them into one big upload, I’m stuck because it exceeds Cloudflare's cap.

Would it be possible to add a feature that splits uploads into smaller chunks? I get that turning off Cloudflare’s protection is a workaround, but I want the security benefits provided by Cloudflare.

This feature could make a difference for those of us using Cloudflare.

Kind regards.

ricardoboss:
May fix #2766

### DIFF
--- a/client/strings/en-us.json
+++ b/client/strings/en-us.json
@@ -89,6 +89,7 @@
   "ButtonSubmit": "Submit",
   "ButtonTest": "Test",
   "ButtonUpload": "Upload",
+  "LabelUploadChunked": "Upload Chunked",
   "ButtonUploadBackup": "Upload Backup",
   "ButtonUploadCover": "Upload Cover",
   "ButtonUploadOPMLFile": "Upload OPML File",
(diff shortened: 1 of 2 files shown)


nichwall:
Does it make more sense to just always split the files instead of adding a toggle?

client/strings/en-us.json
```diff
@@ -89,6 +89,7 @@
   "ButtonSubmit": "Submit",
   "ButtonTest": "Test",
   "ButtonUpload": "Upload",
+  "LabelUploadChunked": "Upload Chunked",
```
justcallmelarry, line 92: this needs to be reordered so that it's in the right place alåhabetically

PR #2891, opened 2024-04-29
Tweaks to custom metadata provider schema

ahobsonsayers:
When developing my [custom metadata provider](https://github.com/ahobsonsayers/abs-tract), I noticed a couple of issues with the provider schema that I thought I would fix. I also did a bit of minor tidy up (I hope this is ok, If not let me know and I will revert!)

Schema changes:
- `duration` type should be an integer not a number
- I have split out `SeriesMetadata` to its own component, as this will generate its own type for those who use generators with the schema, instead of nested anonymous objects.
- `series` under SeriesMetadata being required was specified in the incorrect way using `required: <bool>` instead of `required: <property array>`
- `sequence` under SeriesMetadata should not be a number, it should be a string to allow for sequences like `0.5`, `1-3` and `1,3,5`

Let me know if there is any changes you would like me to make!

### DIFF
--- a/custom-metadata-provider-specification.yaml
+++ b/custom-metadata-provider-specification.yaml
@@ -1,15 +1,16 @@
 openapi: 3.0.0
-servers: 
+
+servers:
   - url: https://example.com
-    description: Local server 
+    description: Local server
+
 info:
-  license: 
+  title: Custom Metadata Provider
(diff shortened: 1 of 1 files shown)


advplyr:
LGTM thanks!

PR #2896, opened 2024-04-30
Split the author call in the library stats page to 2 lighter functions

CoffeeKnyte:
Currently on the library stats page, the function getAuthorsWithCount() is called to get both the top 10 authors by number of books, and also the total number of authors in that library. It does this by querying for all authors and every related book. In a library of 30k authors, this query takes `567038ms`  or around 10 minutes.

This has been split into 2 calls - getAuthorsTotalCount() gets the total number of authors in the library in `1ms` and getAuthorsWithCount() only grabs the top 10 authors by number of books written in `35ms`. These values will vary based n the number of books in the library, but it represents a significant speed increase.

The results displayed on the stats page should be identical after the change.

### DIFF
--- a/server/utils/queries/authorFilters.js
+++ b/server/utils/queries/authorFilters.js
@@ -3,34 +3,47 @@ const Database = require('../../Database')
 
 module.exports = {
   /**
+   * Get authors total count
+   * @param {string} libraryId
+   * @returns {{id:string, name:string, count:number}}
+   */
+  async getAuthorsTotalCount(libraryId) {
+    const authorsCount = await Database.authorModel.count({
+      where: {
+        libraryId: libraryId
+      }
+    });
+    return authorsCount;
+  },
+
+  /**
    * Get authors with count of num books
    * @param {string} libraryId 
    * @returns {{id:string, name:string, count:number}}
    */
   async getAuthorsWithCount(libraryId) {
-    const authors = await Database.authorModel.findAll({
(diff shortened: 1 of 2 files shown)


server/utils/queries/authorFilters.js
```diff
@@ -2,35 +2,48 @@ const Sequelize = require('sequelize')
...
-        [Sequelize.literal('(SELECT count(*) FROM bookAuthors ba WHERE ba.authorId = author.id)'), 'count']
+        'authorId',
+        [Sequelize.fn('COUNT', Sequelize.col('authorId')), 'count']
       ],
-      order: [
-        ['count', 'DESC']
-      ]
+      group: ['authorId', 'author.id'], // Include 'author.id' to satisfy GROUP BY with JOIN
+      order: [[Sequelize.literal('count'), 'DESC']],
+      limit: 10
```
nichwall, line 40: Might be a good idea to make the `limit` an argument to the function in case the limit should be configurable somewhere? Could still default to 10 because that's what is on the stats page.

CoffeeKnyte: Sure. Just fixed that. I've added the "limit" parameter and also set the call from librarycontroller.js to be 10.

nichwall:
This PR is working for me when tested against my main database.

advplyr:
LGTM thanks!

Issue #2914, opened 2024-05-03
[Enhancement]: Description Read more | Read less button language translation for Enhancement #2614


Myticktack:
### Describe the feature/enhancement

The added read more button for big album/author description does not have a translation. #2614 
Currently it is only in english:
![grafik](https://github.com/advplyr/audiobookshelf/assets/7345086/bc1ef5a2-9e00-4f36-90b6-0d88e4278d57)
It would be nice to have it in the language for the user.

advplyr:
I added these translation strings for the next release. The translations are contributed by users https://github.com/advplyr/audiobookshelf/blob/724e06e9d25568b195cb13fc4aa7dc528f7cd12e/client/strings/de.json#L64-L65

advplyr:
Updated in [v2.10.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.10.0)

PR #2916, opened 2024-05-04
Update de.json, add translation for #2914

Myticktack:
Add Translation for "Read more" & "Read less" added by Issue #2914


advplyr:
Thanks!

Issue #1723, opened 2023-04-25
[Enhancement]: Add Whisper support


apiweb:
### Describe the feature/enhancement

Hi there!

I've been using AudioBookShelf for a while now, and I love the platform. I was thinking about how it could be improved, and I had an idea that I wanted to share with you all.

I think it would be great if AudioBookShelf could integrate with [Whisper](https://github.com/openai/whisper) speech-to-text model to automatically generate subtitles for audiobooks. This could be an external tool like Tone and ffmpeg that the user could enable or disable as needed.

With Whisper, it would be possible to transcribe speech across dozens of languages and even handle poor audio quality or excessive background noise. It would make it easier for people who are hard of hearing or have difficulty understanding accents to enjoy audiobooks.

Here are some tips on how to integrate this feature into the AudioBookShelf flow:

- Use the metadata language tag to automatically set the Whisper language.
- Automatically save the srt using the [Title Folder Naming](https://www.audiobookshelf.org/docs#book-title-folder-naming) structure.

I hope you will consider this suggestion for future updates to AudioBookShelf. Let me know if you have any questions or concerns.

Thank you for all your hard work making AudioBookShelf a great platform!

advplyr:
Maybe we first add support for an `srt` subtitle file.

The opposite of this was also requested for ebooks #601

advplyr:
There is also this issue https://github.com/advplyr/audiobookshelf/issues/189

damajor:
I tested Whisper on my setup and results are kind of good but far from perfect.

Using base model it took me around 4 minutes to transcribe 1 hour audiobook.
Using large model it took me a bit more than 1 hour to transcribe 1 hour audiobook.

Those tests were done on 5950X with 12 parallel threads (no GPU involved).

Ed1ks:
👍thumbs up. Would be very nice for self recorded audiofiles.

rounakdatta:
Should we consider working on this? I can volunteer to start contributing.
Currently [Snipd](https://www.snipd.com/) is the only app which gets AI-generated subtitles UX perfectly right, and this could be our chance to make Audiobookshelf the FOSS alternative to it.

advplyr:
Yeah I'm interested in this but I can't put much attention towards it now. We were talking about it in Discord the other day. If anyone wants to start putting something together or setup a proof of concept that would be great. We can chat about it in Discord

yuchen-lea:
[...]

PR #2918, opened 2024-05-04
WIP: Adding Transcription Support to the Web Player (VTT)

mfcar:
I have begun work on adding transcription support to the Web Player. 
I've used [Whisper](https://github.com/openai/whisper) to generate transcriptions for some audiobooks and podcasts. Many tools based on Whisper support exports in VTT and SRT formats. 
For this pull request, I'm only supporting [VTT](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API) as it is natively supported by browsers. Support for SRT can be added in a future pull request.

### How does it work?

A new endpoint, `api/items/:id/file/:fileid/transcript`, has been created on the backend. This endpoint attempts to return a transcription for each audio track. For instance, if there's an audio file named `adventuresherlockholmes_01_doyle_64kb.mp3`, this endpoint will attempt to return the file `adventuresherlockholmes_01_doyle_64kb.vtt`.

On the frontend, when an audio file is set as the source property of the `<audio>` HTML tag, a `<track>` is created and linked to that `<audio>`. The `source` property for the `<track>` HTML tag is populated with the link to the aforementioned endpoint.

### What does this PR support?

- Show/Hide transcription block
- Highlighting the current transcription line
- Clicking on a line to seek the player to that time
- Changing transcriptions when the audio file changes (supports audiobooks and podcasts)

### Demo

https://github.com/advplyr/audiobookshelf/assets/814828/3bd43148-6adc-48b7-8417-bc068be14c7b

### What is missing for the scope of this PR

- Hiding the "Show transcription" button when the transcription is not available for the audio file
- Known issues

### Known issues

- When playing an audio file with transcription, if you close the web player and reopen it, the transcription block is not displayed, even though the transcription is still available. Clicking on the "Show transcription" button to display the block again. I think this is related with the `MediaPlayerContainer.vue` component not reloading the `TranscriptionUi` component.

https://github.com/advplyr/audiobookshelf/assets/814828/c1aaef74-2f70-45a3-b0ce-b04053bf3bc5

- When playing an audio file with transcription, if you change the audio file, the active transcription line for the new audio file focuses on the first line. The focus shifts to the correct line only when the next line change occurs.

### Related
 -  #1723 - This PR can helps to implement the Whisper support on the Web Player

### DIFF
--- a/client/components/app/MediaPlayerContainer.vue
+++ b/client/components/app/MediaPlayerContainer.vue
@@ -101,9 +101,9 @@ export default {
     'playerHandler.playerState': function (newVal) {
       // Refresh the transcription UI when the audio track is changed
       if (newVal === 'LOADED') {
-        this.showTranscriptionUi = false;
+        this.showTranscriptionUi = false
         this.$nextTick(() => {
-          this.showTranscriptionUi = true;
+          this.showTranscriptionUi = true
         });
       }
     },
@@ -507,6 +507,7 @@ export default {
     this.$eventBus.$on('playback-time-update', this.playbackTimeUpdate)
     this.$eventBus.$on('play-item', this.playLibraryItem)
     this.$eventBus.$on('pause-item', this.pauseItem)
+    this.showTranscriptionUi = false
   },
   beforeDestroy() {
     this.$eventBus.$off('cast-session-active', this.castSessionActive)
--- a/client/components/app/MediaPlayerContainer.vue
+++ b/client/components/app/MediaPlayerContainer.vue
@@ -1,5 +1,6 @@
 <template>
-  <div v-if="streamLibraryItem" id="mediaPlayerContainer" class="w-full fixed bottom-0 left-0 right-0 h-48 lg:h-40 z-50 bg-primary px-2 lg:px-4 pb-1 lg:pb-4 pt-2">
+    <div v-if="streamLibraryItem" id="mediaPlayerContainer"  class="w-full fixed bottom-0 left-0 right-0 z-50 bg-primary px-2 lg:px-4 pb-1 lg:pb-4 pt-2"
+         :class="[showTranscriptionUi ? 'h-64' : 'h-48', showTranscriptionUi ? 'lg:h-64' : 'lg:h-40']">
     <div id="videoDock" />
     <div class="absolute left-2 top-2 lg:left-4 cursor-pointer">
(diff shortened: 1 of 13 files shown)


mfcar:
> The placement irks me for some reason. I think that this feature demands something like "Now Playing" screen. But even in this form I really really want this feature in. My sister is hearing impaired and this would really help her.


I also don't like the placement. I was thinking putting it on a side panel or a floating, movable modal. 
But, the side panel raises concerns about taking up too much space on the sidebar, especially if the user has a narrow display. 
The floating, movable modal adds more complexity to the JavaScript and CSS. 
I will make some tests with both behaviours and try to provide updates here.

Sidebar like Apple Music:

image

---

Floating transcription window:

image

ashwinm4friends:
Great job on the project! For UX improvement, please consider looking into word highlighting in Snipd, as shown in this video: https://www.youtube.com/watch?v=jBi-OId37Uw
https://www.youtube.com/watch?v=jzPekGpC4uw

mfcar:
> > Great job on the project! For UX improvement, please consider looking into word highlighting in Snipd, as shown in this video: https://www.youtube.com/watch?v=jBi-OId37Uw https://www.youtube.com/watch?v=jzPekGpC4uw
> 
> Snipd uses word level timestamps, while such subs are easy to generate the only sane format is ssa/ass (srt can blow into megabytes which is insane) afaik. Which is not natively supported by browsers. No idea if it's possible with vtt (and I'm guessing that snippd just uses raw JSON or something since the subs can be generated on the fly)

Look at the WebVTT, which supports something similar to the "[Karaoke Style](https://www.w3.org/wiki/VTT_Concepts#Timestamp_Tags_(Karaoke_Style_and_Paint_On_Caption_Text))" using [:past](https://www.w3.org/TR/selectors4/#past-pseudo) and [:future](https://www.w3.org/TR/selectors4/#future-pseudo) pseudo-classes. However, VTT files need to be adapted for this as well. I think it's not common to get a VTT file with this information. 
I was using Whisper to generate transcriptions, but I'm not sure if we can generate word-by-word transcriptions.

SSA/ASS and SRT support, I was checking what the best approach is. I was considering parsing to VTT to keep the implementation consistent with how we show the transcriptions, I'm not sure if this is the best way yet

image

ashwinm4friends:
In the past, I have used stable-ts to create VTT files. I generated word-level timestamps with Whisper’s base.en model.

Vito0912:
@mfcar Is there any progress made on this?
Would love to have some kind of this feature

PR #2920, opened 2024-05-05
Add item sessions endpoint

rasmuslos:
This endpoint allows API clients to request listing sessions for specific items. A possible use case for this would be to show the listing session of an audiobook on its page.


advplyr:
I doubt this would fulfill that FR because even pausing long enough isn't going to end the session. Sessions in Abs need to be refactored to handle that better. I guess it could be useful to see sessions for a specific audiobook but I don't think it will provide much use for tracking. The Abs android app has that feature and it is all tracked internally. It hasn't been added to the iOS app yet https://github.com/advplyr/audiobookshelf-app/issues/466.

server/controllers/MeController.js
```diff
@@ -33,6 +33,39 @@ class MeController {
     res.json(payload)
   }
 
+  // GET: api/me/item/listening-sessions/:libraryItemId/:episodeId
+  async getListeningSessions(req, res) {
```
advplyr, line 37: This is the same function name as above

server/controllers/MeController.js
```diff
@@ -33,6 +33,39 @@ class MeController {
...
+    const libraryItem = await Database.libraryItemModel.getOldById(req.params.libraryItemId)
+    const episode = (req.params.episodeId && libraryItem && libraryItem.isPodcast) ? libraryItem.media.getEpisode(req.params.episodeId) : null
```
advplyr, line 39: We can query using the new data model for this which will be much faster. There is a `podcastEpisode` model that can be queried directly

server/routers/ApiRouter.js
```diff
@@ -166,6 +166,8 @@ class ApiRouter {
     //
     this.router.get('/me', MeController.getCurrentUser.bind(this))
     this.router.get('/me/listening-sessions', MeController.getListeningSessions.bind(this))
+    this.router.get('/me/item/listening-sessions/:libraryItemId', MeController.getListeningSessions.bind(this))
+    this.router.get('/me/item/listening-sessions/:libraryItemId/:episodeId', MeController.getListeningSessions.bind(this))
```
advplyr, line 170: These can be combined to the same using a ? after the :episodeId

rasmuslos:
> I doubt this would fulfill that FR because even pausing long enough isn't going to end the session. Sessions in Abs need to be refactored to handle that better. I guess it could be useful to see sessions for a specific audiobook but I don't think it will provide much use for tracking. The Abs android app has that feature and it is all tracked internally. It hasn't been added to the iOS app yet [advplyr/audiobookshelf-app#466](https://github.com/advplyr/audiobookshelf-app/issues/466).

I think it's better than nothing and having play / pause show up is not really necessary (in my opinion). Also clients could just close the active listing session after x minutes of inactivity.

advplyr:
Thanks!

PR #2954, opened 2024-05-09
Fix series and collapsed series progress to be consistent and show average of book series progress

### LINKED ISSUE #2921: [Bug]:  Series and collapesed series progress bars are buggy
### Describe the issue

There are a couple of related issues here:

**1. Unituitive behavior for a series progress bar in the series page**
I feel like the series progress bar is not as sensible as it could be. The current behavior is:
1. each book in the series is alloted 1/seriesNumBooks of the progress bar full width
2. if the book is started or finished, it gets 100% of its alloted width.
3.  if a book is not started, it gets 0% of its alloted width.
4. if all books are finished, the progress bar is colored green, otherwise it's colored yellow.

So, for example:
  - If there are 2 books in a series, one finished, and one started, the progress bar will be yellow with 100% width
  - If there are 4 books, 3 started, one unstarted, the progress bar will be yellow with 75% width.

I find this behavior somewhat confusing an unintuitive.

I'd like to suggest that instead of the current 2. we should have:
&nbsp;&nbsp;&nbsp;2. if a book's progress is x%, it gets x% of its alloted width

So for example:
  - If there are 2 books in a series, one finished, and one 50% read, the progress bar will be yellow with 75% width (1x50 + 0.5x50)
  - If there are 4 books, one 80% read, one 40% read, and two unstarted, the progress bar will be yellow with 30% (0.8x25+ 0.4x25)

I think this is more intuitive and more consistent with the single book progress bars.

**2. inconsistent behavior for collapsed series progress bars**
I believe the collapsed series progress bar (in Library view when Collapse Series is checked) is just buggy.
Looking at the code:
```
<!-- Finished progress bar for collapsed series -->`
<div v-else-if="booksInSeries && seriesIsFinished" class="absolute bottom-0 left-0 h-1 shadow-sm max-w-full z-10 rounded-b bg-success" :style="{ width: width * userProgressPercent + 'px' }"></div>
```
According to the comment and the naming of the `seriesIsFinished` function, the progress bar for a collapsed series should only be shown when the whole series is finished. But then:
1. Why is the width set to `width * userProgressPercent`, and not just 100%?
2. `seriesIsFinished` actually _**always**_ returns `true`. This is because it is based on `libraryItemIdsInSeries`, which is always empty, because `collapsedSeries.libraryItemIds` is always undefined. It looks like the server never sets it (see [here](https://github.com/advplyr/audiobookshelf/blob/15ca3307bdc199680896c7bace608f0c8893ed8f/server/utils/queries/libraryItemsBookFilters.js#L604)), except in some special circumestances ([here](https://github.com/advplyr/audiobookshelf/blob/15ca3307bdc199680896c7bace608f0c8893ed8f/server/utils/libraryHelpers.js#L198)).
3. In general, why is the expected behavior different than the one in the series page? I think the collapsed series progress bar should be the same as the series progress bar.

Just set a up a library with a couple of books belonging to the the same series. 
1. Go to the Series page and play with different listening percentages.
2. Go to the Library page, check Collapse Series, and play with different listening percentages.

### Audiobookshelf version

v2.9.0

Built from source

mikiher:
This fixes #2921.

In a nutshell, the changes are:
- Correct series and collapsed series progress bars to show the same
  - The series card in the Series page and collapsed series book card display the same progress.
  - The progress displayed is now the average of the series books progress (i.e. the sum of all the series individual book progresses, divided by the number of books in the series)
- Added libraryItemIds to collapsedSeries objects return from the server
  - This fixes the bug in `seriesIsFinished` in LazyBookCard, which currently always returns true
- Simplified progress bar logic in LazyBookCard


mikiher:
Also added some component tests for LazyBookCard, especially around collapsedSeries cards.

advplyr:
Great, thanks!

Issue #2685, opened 2024-02-28
[Bug]: When editing chapters, hitting tab for next field skips the timestamp column


DavidPesticcio:
### Describe the issue

While editing the chapter details you are unable to use tab to access the timestamp column.

### Steps to reproduce the issue

1. From Home or Library view, select the "Pencil" of item cover
2. Select "Chapters" tab
3. Select "Edit Chapters"
4. Select first chapter "TITLE"
5. Attempt to tab into the START (timestamp) column either forward or backwards
6. Observe that you are unable to access in either direction with TAB or SHIFT-TAB

### Audiobookshelf version

2.8.0

### How are you running audiobookshelf?

Docker

advplyr:
Fixed in [v2.10.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.10.0)

Issue #1776, opened 2023-05-14
Cannot edit chapter times in mobile browsers


rkennedy:
### Describe the issue

When editing a chapter list on a phone or tablet, I can tap on the timestamp next to each chapter title, and its appearance changes as though it's received the input focus, but the on-screen keyboard doesn't appear, so I'm unable to type any digits to provide new start times of any chapters. I see this in Chrome on Android, and on Chrome and Safari on iOS.

Editing start times on a desktop browser works fine.

### Steps to reproduce the issue

1. Select "Edit Chapters" on an audiobook.
2. Select the start time of a chapter.
3. Observe that the on-screen keyboard doesn't appear.

### Audiobookshelf version

v2.2.19

### How are you running audiobookshelf?

Docker

NoWayOuy:
Workaround until this is fixed: change to show time in seconds and you should be able to edit and save

PR #3009, opened 2024-05-25
Time picker cleanup

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/2685

Original behavior was that the TimePicker.vue template was not targetable by a tab. I added a tab target and set the ones place of the seconds to be the default starting point. Currently, users can also tab out of the TimePicker and be focused on two elements, so that is also fixed in this PR.

Original behavior:

https://github.com/advplyr/audiobookshelf/assets/5686638/963ced40-ab9f-46d4-9042-8d1b35d83424

New behavior:

https://github.com/advplyr/audiobookshelf/assets/5686638/b352e903-503e-4715-934c-7ca377c1267d

### DIFF
--- a/client/components/ui/TimePicker.vue
+++ b/client/components/ui/TimePicker.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative">
+  <div tabindex="0" @focus="focusDigit('second0')" class="relative">
     <div class="rounded text-gray-200 border w-full px-3 py-2" :class="focusedDigit ? 'bg-primary bg-opacity-50 border-gray-300' : 'bg-primary border-gray-600'" @click="clickInput" v-click-outside="clickOutsideObj">
       <div class="flex items-center">
         <template v-for="(digit, index) in digitDisplay">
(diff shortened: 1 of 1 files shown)


nichwall:
I also attempted to resolve https://github.com/advplyr/audiobookshelf/issues/1776 by adding the `type="text"` to the element

client/components/ui/TimePicker.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative">
+  <div type="text" readonly tabindex="0" @focus="focusDigit('second0')" class="relative">
```
advplyr, line 2:
```suggestion
  <div tabindex="0" @focus="focusDigit('second0')" class="relative">
```
The `type` and `readonly` attributes aren't available on div elements. Unless there is some hacky thing I'm not aware of? Let me know if I'm missing something here

advplyr:
The difficulty is that I made the TimePicker a custom component and not an input. For #1776 I think the best way to resolve is to re-do the component so it is a native input instead, or a series of native inputs one for each number.
Alternatively we could fake it by calling a click() event on a hidden input and capture the inputs.

client/components/ui/TimePicker.vue
nichwall: It's a hacky thing. I was just trying to figure out how to make the keyboard show up on mobile.

advplyr: Hacky but does it work? Afaik those attributes aren't supported on div

Issue #2990, opened 2024-05-19
[Enhancement]: (Cosmetic) Add line about ensuring permission is provided to ereader provider when setting up ereader devices


ZLoth:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

In the Settings → Email Settings → Ereader Devices , there needs to be a disclaimer line stating that permission has to be provided in order for the ebooks to be emailed to the user. 

### Why would this be helpful?

While it is documented in https://www.audiobookshelf.org/guides/send_to_ereader#common-issues , this will actually call the issue out. 

### Future Implementation (Screenshot)

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/a2b6b5c5-b010-4fcf-92ae-a1ef1187ad59)

### Audiobookshelf Server Version

2.9.0

### Current Implementation (Screenshot)

I don't have a image, but this would be something that would be assist a new administrator.

advplyr:
Updated in [v2.10.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.10.0)

Issue #2991, opened 2024-05-19
[Enhancement]: List of Ereader devices should include access level


ZLoth:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

When setting up a e-reader device, there are several access levels available:
* Admin users only
* All users including guests
* All users excluding guests
* Specific Users

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/603d8b47-4b89-429a-9adf-272700e1c702)

However, when you display the list of devices, the associated access level isn't being displayed, and you have to edit the device to view the access level. 

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/ec0d84ff-479e-4ca4-8444-70393923c33e)

### Why would this be helpful?

It would make ereader administration easier in AudioBookShelf.

### Future Implementation (Screenshot)

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/76b39277-986b-45d7-8c6a-d895d1890a92)

### Audiobookshelf Server Version

2.9.0

### Current Implementation (Screenshot)

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/ec0d84ff-479e-4ca4-8444-70393923c33e)

advplyr:
Updated in [v2.10.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.10.0)

PR #3010, opened 2024-05-25
Ereader settings update

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/2990 and https://github.com/advplyr/audiobookshelf/issues/2991.

A comma separated list is generated if specific users have been given access to an e-reader. I thought about making this name the first two users and then "and N others", but decided against that for now.

I am not sure how to improve mapping user IDs to names, but it's something.

implemtation


client/pages/config/email.vue
```diff
@@ -146,6 +151,29 @@ export default {
...
+          console.error('Failed', error)
+          return []
+        })
+    },
+    getAccessibleBy(device) {
+      const user = device.availabilityOption
+      if (user === 'userOrUp') return 'Users (excluding Guests)'
+      if (user === 'guestOrUp') return 'Users (including Guests)'
+      if (user === 'specificUsers') {
+        this.loadUsers()
```
advplyr: For future ref, calling loadUsers here could result in multiple simultaneous calls.

advplyr:
Thanks!

ZLoth:
Nice, thank you.

PR #3014, opened 2024-05-26
Change Tags and Narrators to work the same as Genres on the Match page.

BrianCArnold:
Used the multi-select behavior that exists on Genres in Match to work on Narrators and Tags.


advplyr:
Thanks! I fixed an existing bug with the MultiSelect I found in testing.
I was going to update the Author input to also use the MultiSelect but it is shared with the podcasts which store authors as a string and I didn't want to spend on the time on that right now.

BrianCArnold:
That's exactly why authors isn't also fixed in this PR

PR #3015, opened 2024-05-26
Speed up playlist & collection query

### LINKED ISSUE #2852: [Bug]: Slow loading playlists
### Describe the issue

Playlists are taking 15+ seconds to load. This occurs when selecting the playlists tab to view the list of playlists and when clicking the button to add a podcast episode a playlist (both loading the list of playlist and after adding the episode to a playlist). I am seeing this when connecting to the web app from a desktop and from the mobile app with even just a single playlist created. 

I tested v2.7.2 and v.2.8.1 and have the issue with both versions. The podcast library does not have this slow loading issue. I reverted to v.2.3.3 which does not have this issue; playlists load almost instantly.

1. Audiobookshelf installed on docker (podcast and audiobook volumes on network share, config and metadata volumes local on SSD)
2. Click the "Your Playlists" button on a podcast episode or click the playlists tab

### Audiobookshelf version

2.8.1

Docker

BrianCArnold:
Should fix #2852

### DIFF
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -492,7 +492,7 @@ class LibraryController {
     }
 
     // TODO: Create paginated queries
-    let collections = await Database.collectionModel.getOldCollectionsJsonExpanded(req.user, req.library.id, include)
+    let collections = await Database.collectionModel.getOldCollectionsJsonExpanded(req.user, req.library.id, include, 2)
 
     payload.total = collections.length
 
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -513,7 +513,7 @@ class LibraryController {
    */
   async getUserPlaylistsForLibrary(req, res) {
     let playlistsForUser = await Database.playlistModel.getPlaylistsForUserAndLibrary(req.user.id, req.library.id)
-    playlistsForUser = await Promise.all(playlistsForUser.map(async (p) => p.getOldJsonExpanded()))
+    playlistsForUser = await Promise.all(playlistsForUser.map(async (p) => p.getLimitedOldJsonExpanded()))
 
     const payload = {
       results: [],
--- a/server/models/Collection.js
+++ b/server/models/Collection.js
@@ -40,9 +40,10 @@ class Collection extends Model {
    * @param {[oldUser]} user
    * @param {[string]} libraryId
    * @param {[string[]]} include
+   * @param {[number]} limitCollectionSize
    * @returns {Promise<object[]>} oldCollection.toJSONExpanded
    */
-  static async getOldCollectionsJsonExpanded(user, libraryId, include) {
+  static async getOldCollectionsJsonExpanded(user, libraryId, include, limitCollectionSize) {
     let collectionWhere = null
     if (libraryId) {
       collectionWhere = {
@@ -88,7 +89,7 @@ class Collection extends Model {
     })
     // TODO: Handle user permission restrictions on initial query
     return collections.map(c => {
-      const oldCollection = this.getOldCollection(c)
+      const oldCollection = this.getOldCollection(c, limitCollectionSize)
 
       // Filter books using user permissions
       const books = c.books?.filter(b => {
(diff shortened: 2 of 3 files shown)


BrianCArnold:
The reason this works is that when we load the `Collections` and `Playlists` section, it requests all of the Collections/Playlists, as well as _**ALL**_ of the metadata for every book in every Collection/Playlist.

If one has 100 playlists with 100 books each, that means we're loading the metadata for 10,000 books just to display the list of playlists, and then when a playlist is opened, we load those 100 books in that playlist again.

Instead, we only load 4 books for playlist, which should be sufficient to render the cover for each playlist, as the playlist consists of the covers of the first 4 books (if I understand correctly)

As for Collections, it appears that it loads the covers of the first 2 books, so the request for the list of collections only returns 2 books per collection.

Together, this vastly decreases the loading time when the total number of books in all collections/playlists is very large, in the hundreds to thousands of books, from multiple seconds to under 50ms

server/controllers/LibraryController.js
```diff
@@ -492,7 +492,7 @@ class LibraryController {
     }
 
     // TODO: Create paginated queries
-    let collections = await Database.collectionModel.getOldCollectionsJsonExpanded(req.user, req.library.id, include)
+    let collections = await Database.collectionModel.getOldCollectionsJsonExpanded(req.user, req.library.id, include, 2)
```
nichwall, line 495: Instead of making this 2 at the API level, it would probably be better to make this an optional request parameter, with a default of "get everything" so 3rd party clients and scripts can still get all items from the single API call.

Not sure what a good name would be, maybe `itemCount`?

server/controllers/LibraryController.js
```diff
@@ -513,7 +513,7 @@ class LibraryController {
    */
   async getUserPlaylistsForLibrary(req, res) {
     let playlistsForUser = await Database.playlistModel.getPlaylistsForUserAndLibrary(req.user.id, req.library.id)
-    playlistsForUser = await Promise.all(playlistsForUser.map(async (p) => p.getOldJsonExpanded()))
+    playlistsForUser = await Promise.all(playlistsForUser.map(async (p) => p.getLimitedOldJsonExpanded()))
```
nichwall, line 516: Same thing as the collections endpoint, but looks like this function is written backwards for doing the paging at the database request level, so that may need to be swapped for the paging to actually work.

nichwall: To be clear, it was already requesting everything from the database and then limiting the response in memory, so maybe the `getUserPlaylistsForLibrary` function needs to be moved around

advplyr:
The data model went through a migration from JSON objects in files to a relational db.

There are a bunch of places where more data is loaded than necessary because the new relational data model is mapping all the data onto the old JSON objects so that the entire API didn't have to get rewritten for the migration.

BrianCArnold:
I investigated the way that the query works, there's not really a way to limit the number of results in the included objects that I could see?

server/controllers/LibraryController.js
nichwall: For this one, I meant something like adding
`itemCount: req.query.itemCount && !isNaN(req.query.itemCount) ? Number(req.query.itemCount) : undefined,`
to the `payload` a few lines before, and then passing `payload.itemCount` to `getOldCollectionsJsonExpanded` instead of the constant value of 2.

server/controllers/LibraryController.js
advplyr: Thanks @nichwall you are right. I made a mistake initially it looks like and we are loading all the items twice.
I'm going to start fresh on this one instead of working off this PR to clean more of this up

advplyr: This explains the slow loading actually because we are iterating through every item and loading the playlistMediaItems when we should be getting them all in one go.
It would be helpful to also add an optional query param to limit the items returned in each playlist so I'll look into that also

advplyr: Just updated this https://github.com/advplyr/audiobookshelf/commit/157616421841b2535d108e5c4d402931429d4075
I wasn't able to limit the media items in that query because of a sequelize bug https://github.com/sequelize/sequelize/issues/12971
The only other way I can think to implement that is to filter out media items with an `order` < 5 or something like that.
It would be better to migrate that endpoint to return the new data model then to spend a lot more time trying to get the old one to perform better.

advplyr:
See https://github.com/advplyr/audiobookshelf/pull/3015#discussion_r1616377433

Issue #3016, opened 2024-05-26
[Bug]: Playlist Editing Requires "Can Update" Permission


lkjfdsaofmc:
### What happened?

When a user creates a playlist they are unable to use the edit function for renaming, adding a description, or deleting unless they have the "Can Update" permission.  They can add and remove books from the book itself but can't use the edit function on the playlist tab.

### What did you expect to happen?

Users should be able to use the edit function on playlists as playlists are user specific so editing them doesn't impact other users.  Since it requires the "Can Update" permission which would allow them to modify other content most users won't be able to fully use their own playlists.

### Steps to reproduce the issue

1. Have any user create a playlist without them having the "Can Update" permission for the server
2. Have the user go to the playlist tab and there will not be an edit button

### Audiobookshelf version

2.9.0

### How are you running audiobookshelf?

Windows Tray App

### What OS is your Audiobookshelf server hosted from?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Edge

advplyr:
Fixed in [v2.10.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.10.0)

PR #3017, opened 2024-05-26
Users can edit playlist in UI

nichwall:
Fixes: https://github.com/advplyr/audiobookshelf/issues/3016

Do not hide the "edit" or "delete" key on the playlist page even if users do not have "Can Update" permission because they are private structures, so editing does not affect other users.

User without edit permissions view:
image


advplyr:
Thanks!

Issue #2774, opened 2024-03-20
[Enhancement]: Toggle to display subtitle in library


GreatestFool:
### Describe the feature/enhancement

I think it'd be nice to be able to show the subtitles of books in the library, either selectively on some Libraries, or globally.

I have a Library dedicated to technical books and articles, and among those I have some ISO standards, which have pretty obtuse titles (like "ISO 3103:2019(E)"), and no covers that are easy to just eyeball at cover size 120. For those, it's usually the subtitle that is more humanly readable. And while I obviously _can_ just go by the ID, it'd be nice to be able to eyeball at least part of the subtitle in the Library and Home and stuff without having to click into the book details. Kind of like why #1755 is nice too.

PR #3037, opened 2024-06-03
Bookshelf and cards refactoring

mikiher:
This refactoring addresses a number of issues I found while trying to add display of an optional subtitle line (which I left out at the moment):
- The book card (and other cards) width and height are passed as props
  - This can (and does) cause incosistencies, where various card containers use different values. For example book cards currently have slightly different size on the home and library pages.
- The height of the card does not reflect the actual card height. the bottom book details use absolute positioning outside the card area.
  - This requires complex layout computation that also happens outside the card, to make the card containers have the correct height to display both the card image and the details below.
- `sizeMultiplier` is used *many* times in each of the cards. 
  - Wherever some size attribute is set, it is multiplied by `sizeMultiplier`. This clutters the code quite a bit.
  - The `sizeMultiplier` is computed differently in each of the cards, although the result should be the same
- `bookCoverAspectRatio` is passed as prop
  - This can also be computed in a single place.
- There's is a separate itemSlider component for each card type.
  - The Vue templates and code in most component are mostly duplicates.
- The bookshelf scaling of spacing and font-sizes is inconsistent
  - When the scaling changes (by increasing or decreasing the cover size), some of the spacing and font-sizes change, some don't

The following bookshelf fixes and refactoring were implemented (with the intention of leaving functionality as-is, except for bug fixes):
- The height and width of bookshelf cards is now determined by default inside the card
  - The height of the card graphics (cover image or material icon) has a default.
  - All other properties are computed.
  - the card computed height now includes the bottom description if it has one.
    -  one exception to this is the description in standard view, which needs to appear _on top_ of the shelf divider
- The shelf height is now determined by the computed height of the cards inside it
  - In lazyBookshelf, where the height needs to be determined in advance, one dummy card of the right type is mounted in order to get its width and height, and then removed.
  - This makes shelf height calculation straightforward and less error-prone.
- All the itemSliders are now merged into a single component
  - The card component is dynamically determined by the shelf type.
  - Also fixed some small edge case slider scrolling issue, and made it not dependent on the card width
- sizeMultiplier and bookCoverAspect ratio are now obtained from the store
- The scaling and font sizes is now made consistent by changing them to relative em units
  - This is mostly done transparently by changing the default tailwind definitions
  - A version of the original tailwind spacing is still available (by adding r to the spacing number, e.g. right-4r)
  - At the bookshelf level, the font-size is set to sizeMultiplier * 1rem
    - this makes the tailwind classes inside the bookshelf reactive to the sizeMultplier (since they are now specified in em units)
    - this also makes sizeMultiplier multiplication redundant in most places (by changing rem units to em), making the code less cluttered.
- Fixed some buggy booksPerFetch behavior
  - When cover size is set to smallest, lazyBookshelf doesn't get enough books to fill the first page, so something like this appears: 
Screenshot 2024-06-02 085102 This was due to booksPerFetch having static values. It was fixed to have a value depending on the the number of books per shelf and the number of shelfs per page.

I've tested this quite extensively in library, search, and home pages, in both stadard and detail view, while changing cover sizes back and forth. I believe all is working as intended now.


mikiher:
The last commit fixes #2774. 

If you don't want this now, I can remove it - I just wanted to demonstrate how it's quite easy now to make changes to a card that increase its height, by just adding the required html to the card template, and without needing to modify height calculations anywhere - all layout modification to the bookshelf or other card containers happen automatically!

mikiher:
OK, please hold on with reviewing, I found some issues with tailwind font-size and line-height replacements.

mikiher:
So, an update (sorry it took quite a bit, I've on vacation for most of the last couple of weeks): it looks like fully migrating to em-based spacing and font-sizing tailwind classes as a default has all sorts of side-effects on components unrelated to this refactoring (mostly due to know shrinking font issues). I tried fixing some of these but it just seems like too much changes all over the place.

I ended up reverting the the move to default em-based spacing and font-sizing. Instead:
- I extended tailwind spacing to add em-based classes (by adding an e to the class name, e.g. `px-2e` as the em-based version of `px-2`). 
- I modifed refactored components to use the em-based tailwind versions for spacing, and explicit style for em-based font-sizing.

I think this is ready for review now.

advplyr:
This is working great in my testing. I don't think we should include the show subtitles in this PR.
It doesn't make sense to show that setting on the regular bookshelf view. It is too tight on mobile for the library page. I think we should use a context menu dropdown if we add more checkboxes to the bookshelf toolbar. Showing series and genres there confused me at first and when it is empty it looks weird.

mikiher:
> I don't think we should include the show subtitles in this PR.
Sure, I'll revert that, and rework it in a different PR based on your comments.

advplyr:
Great, thanks!

devnoname120:
@mikiher Fantastic work, thanks a lot!

advplyr:
I just noticed that the LazyCollectionCard isn't updated for non-square covers anymore.

image

vs on the collection landing page
image

mikiher:
Thanks, will look at this when I get back home tomorrow.

Issue #3020, opened 2024-05-26
[Enhancement]: E-Reader Font Boldness Slider


lkjfdsaofmc:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

The ability to adjust font boldness with a slider similar to the font scaling and line spacing sliders.

### Why would this be helpful?

This would be useful for people who want their font to be more or less bold depending on their eyesight without necessarily making the entire font significantly larger.  This would also be significantly useful for the mobile apps since they rely on the same e-reader functionality.

### Future Implementation (Screenshot)

![FontScalingWithBoldness](https://github.com/advplyr/audiobookshelf/assets/7034251/ca727b80-1389-4fc3-97b9-bce73596bf6d)

### Audiobookshelf Server Version

2.9.0

### Current Implementation (Screenshot)

![FontScaling](https://github.com/advplyr/audiobookshelf/assets/7034251/7094ba72-f93f-4b60-9307-e98fc47d4873)

BimBimSalaBim:
@lkjfdsaofmc  hope this helps, I don't know how to make text thinner without changing the fonts but this should allow users to make text thicker

nichwall:
Resolved by https://github.com/advplyr/audiobookshelf/pull/3040

advplyr:
Added in [v2.11.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.11.0)

PR #3040, opened 2024-06-04
E-Reader Font Boldness Slider #3020

BimBimSalaBim:
Update font settings in EpubReader and Reader components to include a "Font Boldness" slider which applies a stroke to the text per issue #3020


advplyr:
Thanks!

PR #3086, opened 2024-06-19
bug: If !itemProgress unhandled exception syncing user progress

taxilian:
I'm not sure exactly the situation which creates this, but there is an if statement which checks if `itemProgress` -- but then uses it even if not. This caused my server to become useless recently =] Simple fix.


advplyr:
This works but the question I would have is how did that actually happen? Since the code directly above this creates the media progress record if it doesn't exist this should not be possible.

If you want to share any other info about this we may be able to find a deeper issue

taxilian:
Sadly I honestly don't know -- all I know is that suddenly all my instances were crashlooping due to hitting this. Might be worth adding some kind of logging?

PR #3094, opened 2024-06-21
Openapi middleware validation

nichwall:
This PR adds `openapi-openapi-validator` to compare API requests and responses against the OpenAPI specification located in `docs`. I am only adding this as a dev dependency because if the request/response does not match the spec, the web client may render incorrectly (I ran into issues with the "edit library item" modal sometimes presenting a blank tab when the spec was not met).

An example of errors given to help identify how a request or response does not match the spec:
Screenshot from 2024-06-21 14-52-02

This PR also updates some of the `requestBody` to `parameters` and creates some generic `parameter`s in response to changes made in https://github.com/advplyr/audiobookshelf/commit/d6438590d7010e9e9f512ecfdbf4cd7f0d149407

### DIFF
--- a/docs/controllers/AuthorController.yaml
+++ b/docs/controllers/AuthorController.yaml
@@ -74,17 +74,14 @@ paths:
       description: Get an author by ID. The author's books and series can be included in the response.
       tags:
         - Authors
-      requestBody:
-        required: false
-        description: The author object to create.
-        content:
-          application/json:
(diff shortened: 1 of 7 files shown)


nichwall:
Splitting this into 2 PRs, one with the spec updates and a standalone one with `express-openapi-validator` so it doesn't block more spec work.

PR #3095, opened 2024-06-21
Change: `requestBody` to `parameter`, allow commas in queries

nichwall:
This PR updates some of the `requestBody` to `parameters` and creates some generic `parameter`s in response to changes made in https://github.com/advplyr/audiobookshelf/commit/d6438590d7010e9e9f512ecfdbf4cd7f0d149407

This PR also allows for reserved characters (commas) to be included in select query parameters.


advplyr:
Thanks!

PR #3107, opened 2024-06-27
bug: if oldLibraryItem is null things crash

taxilian:
I'm not sure how it got into this state, but once in the state it keeps crashing. Might be because I have multiple instances running against the same sqlite database so two concurrent were scanning at the same time?

### DIFF
--- a/server/scanner/LibraryScanner.js
+++ b/server/scanner/LibraryScanner.js
@@ -176,9 +176,11 @@ class LibraryScanner {
 
             // TODO: Temporary while using old model to socket emit
             const oldLibraryItem = await Database.libraryItemModel.getOldById(existingLibraryItem.id)
-            oldLibraryItem.isMissing = true
-            oldLibraryItem.updatedAt = Date.now()
-            oldLibraryItemsUpdated.push(oldLibraryItem)
+            if (oldLibraryItem) {
+              oldLibraryItem.isMissing = true
+              oldLibraryItem.updatedAt = Date.now()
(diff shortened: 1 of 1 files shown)


nichwall:
SQLite does not support concurrent connections, so something is probably getting corrupted.

taxilian:
That is not actually correct -- though it's possible it is correct for the particular sqlite library being used, that I can't be certain of. I've used sqlite for years and it uses reader/writer locks to control access to the database. it's not very performant, but it absolutely works and is generally supported. (see https://sqlite.org/faq.html#q5)

The database is fine, but I suspect what happened is that two scans were happening at the same time and one updated something that the other then expected to be in the original state, thus setting off a chain reaction. Regardless any time you pull something that *might* be null you should probably check it, so I recommend using the fix =]

advplyr:
Thanks!

nichwall:
Thanks for the link, I was thinking the network mounted SQLite causing problems with multiple access since there are a lot of SQLite problems reported due to people mounting it over the network. I thought that also applied to multiple processes, but must just be Windows then.

taxilian:
I'm running it over ceph-fs which *might* have issues like that, but I've not so far seen anything that looks like corruption and everything I'm finding tells me that it should work fine, albeit slowly =] I definitely need to find some time to see if I can get this all working with mysql or pgsql, but I figured this is a good place to start and see where issues crop up.

So far this is the first issue I've hit which seems related to concurrency

PR #3117, opened 2024-07-04
New feature: Show Subtitles

### LINKED ISSUE #2774: [Enhancement]: Toggle to display subtitle in library

mikiher:
Fixes #2774.

This re-introduces a feature removed from #3037 after some comments from @advplyr :
>  I don't think we should include the show subtitles in this PR.
> It doesn't make sense to show that setting on the regular bookshelf view. It is too tight on mobile for the library page. I think we should use a context menu dropdown if we add more checkboxes to the bookshelf toolbar. Showing series and genres there confused me at first and when it is empty it looks weird.

This is a re-write of that feature taking those comments into account.
- The Show/Hide Subtitles setting has been moved to the Toolbar's context menu dropdown
- The subtitle field shows the subtitle from the metadata, if it exists, or the book's series and sequence if it doesn't.
  - No genres in this version, but I think series and sequence make sense if there's no subtitle.
  - This is only shown for books, not for podcasts
- For a collpased-series entry, the subtitle shows the number of books in the series.

I think it also makes sense to move the Collapse Series option to the same menu. Would you like me to implement that?


nichwall:
I like the dropdown menu to show/hide subtitles, and I think moving Collapse Series option to the same menu is a good idea.

Some comments:

If collapse series and show subtitles are both enabled, should the icon in the upper right of the cover be removed?
collapse series

Slightly related, extra spacing is not added when sorting by "Publish Year". Extra spacing is added for all of the other sort methods.
Publish year sort
Modified sort

nichwall:
In regards to the series count in the upper right being removed, maybe that should be it's own option? There have been some requests in the past to allow it to be removed, so maybe this dropdown is the best place to add that.

advplyr:
I was testing the card height issue and noticed that the shelf height isn't calculated correctly when sorting by title and having ignore prefixes when sorting enabled.

Highlighting the shelf div in inspector
image

mikiher:
Ah, thanks, I'll look into shelf height calculation bug.

advplyr:
I just figured out that bug and will push the fix

advplyr:
https://github.com/advplyr/audiobookshelf/commit/b01ef1c6913dd958d2361ff0e4d237c9db24f1bc

mikiher:
> [b01ef1c](https://github.com/advplyr/audiobookshelf/commit/b01ef1c6913dd958d2361ff0e4d237c9db24f1bc)

Good, nice catch!

mikiher:
> I like the dropdown menu to show/hide subtitles, and I think moving Collapse Series option to the same menu is a good idea.

OK, moved the Collapse/Expand Series to the context menu.

Note: I **_did not_** put it in the seriesContextMenu, since Collapse Series doesn't seem to have any effect in the single series page anyway (rightly so?).

mikiher:
One other thing I noticed - check boxes in ABS seem to be mostly localized, but menu items are not. Is that on purpose?

advplyr:
> One other thing I noticed - check boxes in ABS seem to be mostly localized, but menu items are not. Is that on purpose?

I see 2 menu items that aren't localized in BookShelfToolbar.vue, are you referring to those or are there other menu items you are seeing?

mikiher:
> > One other thing I noticed - check boxes in ABS seem to be mostly localized, but menu items are not. Is that on purpose?
> 
> I see 2 menu items that aren't localized in BookShelfToolbar.vue, are you referring to those or are there other menu items you are seeing?

Sorry, my bad - I generalized from seeing the ones I saw in BookShelfToolbar - I now see most are localized.
There's just one other partially localized one in Appbar.vue.

I'll fix the un-loclalized menu items.

mikiher:
Done.

advplyr:
Looks good, thanks!

Issue #2973, opened 2024-05-13
[Enhancement]: Modify path of backups


KellicTiger:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

The target for where the backups are located is on the same volume where the application is installed

![image](https://github.com/advplyr/audiobookshelf/assets/42510565/4428188e-b1e4-48d4-be05-62a4132af2ac)

Best practice for backups is to put them on a volume other than where you are running the application or where the dataset is located.  While it is possible for me to symlink a location to where my backup drive it, the better option is to allow the path to be modified.

As a temp work around document where the pathing is stored.  I'm assuming it is in a config file somewhere.

### Why would this be helpful?

Adhering to best practices for backups.  As in the event of the drive where the app is stored you would also lose the backup.

Allowing the changing of the path to another drive would allow for better data protection.

### Future Implementation (Screenshot)

![image](https://github.com/advplyr/audiobookshelf/assets/42510565/31831ad3-2eab-485f-9ef6-966cf14c4a44)

### Audiobookshelf Server Version

v2.9.0

nichwall:
Duplicate of https://github.com/advplyr/audiobookshelf/issues/199 but this uses the new format and looks nicer :)

noja11:
Amazing that I'm just finding this right now and the request is so new! 

I'm currently moving ABS from Docker to an LXC and I'm having a hard time linking it to my standard services backup share. Either changing it from within the app or allowing it to be set in the config file at /etc/default/audiobookshelf would be amazing. Thank you all for your effort!

ZLoth:
As I'm running ABS as a Docker container on TrueNAS Scale. I just set up a mapping from /metadata/backups to my backups folder. That folder is then uploaded nightly to a off-site backup.

advplyr:
I made an update so that the backup path can be overridden with an environment variable BACKUP_PATH.

It is now stored in server settings so that we can eventually make the path editable in the UI. That may not make it in the next release so for anyone that wants to modify it now will be able to pass in the env variable.

noja11:
> It is now stored in server settings so that we can eventually make the path editable in the UI.

Thank you so much for that change!

I just tried an LXC where I edited the audiobookshelf config file at /etc/default/audiobookshelf to include BACKUP_PATH.  As that didn't work yet, I'm assuming I'll need to wait for the next release after 2.10.1? 

Again, thank you so much for this quality of life upgrade, I really appreciate it!

[...]

PR #3122, opened 2024-07-05
Prevent backup path edits when ENV is set

nichwall:
This PR prevents the backup path from being edited in the web interface if the `BACKUP_PATH` environment variable is set. This is done to help prevent users from thinking their backup settings are not being applied correctly if they are not looking at the debug logs on server start as mentioned here https://github.com/advplyr/audiobookshelf/issues/2973#issuecomment-2179518347.

I changed the response for a `GET` request to `/api/backups` to include an additional value of whether this field can be editable, and set this value in the server settings in the same places where the ENV variable is checked.

Tested with setting and removing the ENV variable to make sure that the field is then editable.

View with environment variable not set:
backups_no_env

View with environment variable set:
backups_env

### DIFF
--- a/client/components/tables/BackupsTable.vue
+++ b/client/components/tables/BackupsTable.vue
@@ -171,7 +171,7 @@ export default {
       this.$axios
         .$get('/api/backups')
         .then((data) => {
-          this.$emit('loaded', data.backupLocation)
+          this.$emit('loaded', data)
           this.setBackups(data.backups || [])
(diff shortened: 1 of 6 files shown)


advplyr:
Thanks! I didn't see a reason to persist that variable in server config so I updated it to use the value returned from the API call.

nichwall:
Thanks! That makes more sense

PR #3126, opened 2024-07-06
Replace pkg with @yao-pkg/pkg and target node20

mikiher:
This replaces the use of pkg (which packs Node runtime and project code into a single binary) with the @yao-pkg/pkg fork. We do this since pkg is deprecated and in maintenance mode, and doesn't support packing a node20 runtime. Yao-pkg seems to be well maintained and used, and supports node20 targeting.

Changes made:
- Replace node18 pkg targets with node20
- Modify relevant workflows and scripts to install @yao-pkg/pkg instead of pkg
- Log Node version when server goes up for clarity

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches-ignore:
-      - 'dependabot/**'  # Don't run dependabot branches, as they are already covered by pull requests
+      - 'dependabot/**' # Don't run dependabot branches, as they are already covered by pull requests
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


mikiher:
One question: build/linuxpackager seems to call pkg, but I can't find any script which installs it. Should I add the npm install command to that script?

advplyr:
I updated that script. It is currently running on one of my ubuntu droplets that builds the PPA. Eventually that will become a github action hopefully.

The debian package built fine with this but I hit the same issue mentioned here https://github.com/yao-pkg/pkg/pull/55#issuecomment-2201899597

Currently audiobookshelf.service is setup to use audiobookshelf user and group.

https://github.com/advplyr/audiobookshelf/blob/3a2f786517533815b4ff71567de910ab2a898461/build/debian/lib/systemd/system/audiobookshelf.service#L12-L13

This user did not have a home directory so I had to create one and give it write access.
I'm going to try building again with 5.11.5

advplyr:
I installed https://github.com/yao-pkg/pkg/releases/tag/v5.11.5 and built the debian package again. No permission issues with the cache dir so I'll keep that version installed for building the debian package.
Great to be able to move to node20 for the binaries finally. Thanks!

Issue #3135, opened 2024-07-08
[Bug]: `index` column in `podcastEpisodes` table always has `-Inf` as value


nichwall:
### What happened?

The `index` column of the `podcastEpisodes` table seems to have been broken around server version `2.9.0`. Podcast episodes downloaded before February 2024 (not sure exactly when I upgraded my main server) have the `index` column set, but all of the rows have `-Inf` as their value if the `extraData` column has data other than `{}`. Screenshots from the database below.

Episodes from before `2.9.0`:
![index_normal](https://github.com/advplyr/audiobookshelf/assets/5686638/d632ce46-ab7b-4cef-8175-a896844478bb)

Episodes downloaded beginning with `2.9.0`:
![index_infinity](https://github.com/advplyr/audiobookshelf/assets/5686638/2c330df7-33cc-4f33-a822-e7350186c82d)

### What did you expect to happen?

Consistent data types for the OpenAPI spec, or some sort of database migration to keep everything consistent.
I'm not really sure how the "previous" and "next" logic works in the web client Episodes modal because the web client is behaving as expected for episodes with and without the `index` column set correctly. If the modal is just using the series and episode field, maybe this column can be removed?

### Steps to reproduce the issue

1. Download podcast episodes from before `2.9.0` and after upgrading to `2.9.0` and inspect the database.

### Audiobookshelf version

v2.10.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

advplyr:
I'm not sure if we should remove that column yet. Originally podcast episodes would set that index column and allow for sorting podcast episodes. Sorting was removed but the index column remained getting updated for a while even though it wasn't used for anything. I'm not sure why it is -Inf so that should be fixed regardless

nichwall:
That's fair. Removing the column from the database would also make rolling back to older versions more difficult. I mean if it isn't being used by a client then it can be removed from the API.

advplyr:
Yeah it can be removed from the API

PR #3012, opened 2024-05-25
Initial library endpoints

nichwall:
This PR includes:
- Some initial library endpoints
- Updates to the Author endpoints
- Fixes to existing schemas for exploded spec
- Adds series endpoints
- Adds workflow to lint the spec

Next steps are likely (in this general order):
1. Notification routes
2. Email routes
3. Podcast routes and schemas (used in lots of responses with the library)
4. Item routes and schemas
5. Collections routes and schemas
6. Playlists routes and schemas
7. RSS feeds
8. The rest of the Library routes

Preview of redocly generated docs:
image

PR #3140, opened 2024-07-09
Podcast endpoints

nichwall:
This PR continues on from https://github.com/advplyr/audiobookshelf/pull/3012.

This PR adds initial podcast endpoints and includes the `PodcastMetadata`, `PodcastEpisode`, and `Podcast` schemas. More work is needed, but these routes and schemas are as accurate as I can make it with my current understanding of the data models and API. This future work will likely take place while working on LibraryItems and as some functions are updated.

Each of the schemas has some failing validation, but I believe that is due to inconsistent data handling for podcasts when fetching new episode information from the internet or creating new podcast episodes. Specific examples are:
- explicit (sometimes boolean, sometimes "yes" or "no")
- duration (sometimes a string in "HH:MM:SS" format, sometimes an integer of total seconds)
- index (sometimes is an integer, sometimes is `-Inf` which causes validation to fail even though it technically is an integer https://github.com/advplyr/audiobookshelf/issues/3135)

This PR will have merge conflicts with https://github.com/advplyr/audiobookshelf/pull/3012, so once that PR is merged I can fix the bundled spec conflicts.

Next, I am planning to work on updating LibraryItems to include both book and podcast options.

### DIFF
--- a/docs/controllers/PodcastController.yaml
+++ b/docs/controllers/PodcastController.yaml
@@ -1,0 +1,331 @@
+paths:
+  /api/podcasts:
+    post:
+      summary: Create a new podcast
+      operationId: createPodcast
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '../objects/mediaTypes/Podcast.yaml#/components/schemas/Podcast'
+      responses:
+        200:
+          description: Successfully created a podcast
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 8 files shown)


nichwall:
Well, looks like there were not any merge conflicts. Assuming the previously mentioned validation issues, this is good to merge.

advplyr:
Thanks!

Issue #3157, opened 2024-07-13
[Enhancement]: Disable maximum size for backups


ZLoth:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

As currently implemented, the maximum size of backups is 1 GB, although it is administrator-adjustable. It would be niuce to set it to 0 and not worry about the maximum size of backups. 

### Why would this be helpful?

As currently configured, ABS sets a maximum size for backups at 1 GB, although that is a administrator-adjustable option. I've had to increase it to 5 GB for the size of my library. However, while I can understand the need to limit the size of backups, I would like to disable the maximum size as it is very unlikely that a backup would exceed the 17TB I have available in my TrueNAS server.

If the feature already exists (e.g. setting the maximum size to 0 to disable), it is not documented in https://www.audiobookshelf.org/guides/migration-and-backups nor in the hover-over help. 

### Future Implementation (Screenshot)

Not applicable. 

### Audiobookshelf Server Version

2.11.0

### Current Implementation (Screenshot)

![image](https://github.com/user-attachments/assets/c2a7654f-be28-42f8-a5af-3f8a888a1d2e)

nichwall:
Fixed by https://github.com/advplyr/audiobookshelf/pull/3163

ZLoth:
Thank you!

PR #3163, opened 2024-07-16
Adds support for allowing backups of unlimited size

ajyey:
This MR adds support for allowing backups of unlimited size (denoted by 0).

Addresses #3157

### DIFF
--- a/client/pages/config/backups.vue
+++ b/client/pages/config/backups.vue
@@ -64,7 +64,7 @@
         <ui-text-input type="number" v-model="maxBackupSize" no-spinner :disabled="updatingServerSettings" :padding-x="1" text-center class="w-10" @change="updateBackupsSettings" />
 
         <ui-tooltip :text="$strings.LabelBackupsMaxBackupSizeHelp">
-          <p class="pl-4 text-lg">{{ $strings.LabelBackupsMaxBackupSize }} <span class="material-symbols icon-text">info</span></p>
+          <p class="pl-4 text-lg">{{ $strings.LabelBackupsMaxBackupSize }} (0 for unlimited) <span class="material-symbols icon-text">info</span></p>
         </ui-tooltip>
       </div>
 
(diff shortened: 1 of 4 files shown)


client/pages/config/backups.vue
```diff
@@ -64,7 +64,7 @@
         <ui-text-input type="number" v-model="maxBackupSize" no-spinner :disabled="updatingServerSettings" :padding-x="1" text-center class="w-10" @change="updateBackupsSettings" />
 
         <ui-tooltip :text="$strings.LabelBackupsMaxBackupSizeHelp">
-          <p class="pl-4 text-lg">{{ $strings.LabelBackupsMaxBackupSize }} <span class="material-symbols icon-text">info</span></p>
+          <p class="pl-4 text-lg">{{ $strings.LabelBackupsMaxBackupSize }} (0 for unlimited) <span class="material-symbols icon-text">info</span></p>
```
nichwall, line 67: This should be added to the English string file in `client/strings` instead of adding a constant string value

ajyey: Done 👍

advplyr: @nichwall I wonder if changing that string in en-us will require the already translated strings in other languages to get reset

ajyey: Was wondering that as well. That might be outside my wheelhouse to be honest 😓

nichwall (commented):
Looks good to me. I put some random large files in `/metadata/items` (so my backups were about 7 GB) and everything is working as expected with both unlimited and maximum size set.
A new server did also correctly set the default maximum size to 1 GB.

advplyr:
Thanks!

client/pages/config/backups.vue
nichwall: Just checked Weblate, the translations were automatically marked as needing updates when the English string was updated

ZLoth:
Thank you!

Issue #2924, opened 2024-05-05
[Bug]: Language is not embedded


MarcoMue:
### Describe the issue

When filling out the Language Field and then using Embed Metadata => Quick Embed, no id3 tag for the language is added or filled.

I checked if its there with vlc, windows properties and Mp3Tag.
I was working with a m4b file.

### Steps to reproduce the issue

1. Fill the Language Field in the Book Details.
2. Navigate to Tools.
3. Use Quick Embed.

### Audiobookshelf version

v2.9.0

### How are you running audiobookshelf?

Docker

MarcoMue:
The Audiobook File Management Tools in ABS also show no Language embed.

I was expecting to see it under 'additionalFields' or a Language Field. Or ideally as TLAN:
https://id3.org/id3v2.3.0#TLAN

advplyr:
M4B files don't use ID3 tags, . This is one of the difficulties with meta tags. We'll be switching back to using ffmpeg for embedding meta tags so might be further limited on the data we can embed. https://wiki.multimedia.cx/index.php?title=FFmpeg_Metadata

PR #3170, opened 2024-07-17
Parse book language from NFO metadata source

ajvgwu:
This is a relatively simple change for parsing metadata from an NFO file to include a "language:" (or "lang:") field.

I tried to find the right files to edit by looking at PR #2305.

My motivation is that it seems hit-or-miss to embed the language in some audio files (and/or to have these tags picked up by audiobookshelf). For example, using `tone` to embed the language into a MP3 file creates additional metadata fields "language=eng" and "LANG=eng", while doing the same for an AAC/M4A file creates "lang=eng". Then, when imported into audiobookshelf, the language field is properly set for the MP3 file, while the language field for the AAC file is empty.

This could be somewhat related to Issue #2924. It wouldn't help with the embedding part, but could help anyone who uses NFO files as a way of keeping the metadata in a sidecar file. It also wouldn't matter what audio format/container is used, since the NFO sidecar is a simple text file. I would basically use this as a way to fill in missing language info when tags from the audio file are not properly picked up for whatever reason.


advplyr:
Thanks!

PR #3182, opened 2024-07-20
Add progress and fixes to m4b and embed tools

### LINKED ISSUE #3143: [Bug]: M4B conversion does not check if it has permissions to move/create files before proceeding
### What happened?

I have set my library to be RO, and the M4B conversion tries to move data from and to my directory after the very-long conversion process.

A check that the directories are actually writable.

1.  Put mp3's in a RO directory and try to convert them to M4B

### Audiobookshelf version

2.11.0

Debian/PPA

Linux

None

### Logs

_No response_

### Additional Notes

Mildly related to https://github.com/advplyr/audiobookshelf/issues/3073, as by resolving this issue, the conversion would still fail, just earlier.

mikiher:
This adds progress indicators to the embed and m4b tools, and fixes some issues in those tools.

The main changes:
- Both embed and m4b tools now show both task and per-track progress in percent
  - In the m4b tool, the per-track progress is calculated according to track durations and ffmpeg progress timemark.
  - The task progress is shown on the button instead of the spinner.
  - The track and task progress are persisted in the store, so they're rememberd when the user leaves and comes back to the tool page.
  - In the m4b tool, since there are two operations happening sequentially (m4b merging and embedding), I made the merging account for 95% of the progress (since it's usually takes much longer), and embedding acount for 5%. This was just just an arbitrary decision.
- Now checking writability of the directory (in m4b and embed) and write access to the audio files (in embed), before running ffmpeg. This fixes #3143
- Fixed issue that caused metadata not to be properly embedded after merging, in the m4b tool.
- Added a TaskFailed indicator to the embed tool UI (it always showed success up until now, even if embedding failed).
- Removed the redundant worker_threads logic from AbMergeManager, and moved the merging code mergeAudioFiles() in ffmpegHelpers.js
- Tightened error handling, especially around file system operations.


advplyr:
Working great, thanks!

kuldan5853:
Works indeed great - could the % complete also be added to the "ongoing tasks" view? (Top right dot circle icon) I think that place was missed.

advplyr:
I forgot about this in my review but for files in the root folder that are single-file books the `LibraryItem.path` will be the path to the audio file.
This is only handling cases where the `LibraryItem.path` is a folder. For single-file books the `LibraryItem.isFile` is set to true.

Bug reported https://github.com/advplyr/audiobookshelf/issues/3323

mikiher:
Thanks for reporting. I'm on vacation until tomorrow without access to my
dev machine. Will look when I get back.

PR #3185, opened 2024-07-21
Adds genres to gloabl search

### LINKED ISSUE #2937: [Enhancement] Include Genre in global library search
### What happened?

When I search for "Klassiker" which is used in "TAGS" (Schlagwörter)  all is fine:
image

But when I search for "AbenteuerRoman" which is used in "Genre" the system does not find the entry:
image

Under "Libray" the direct-filter on "AbenteuerRoman" works perfect:
image

search will also find libray-item with Search-Text in Genre like it works for TAGS

1. 

### Audiobookshelf version

2.8.1

Docker

Linux

Chrome

### Logs

_No response_

### Additional Notes

_No response_

mikiher:
Fixes #2937

This adds genres to the global search dropdown (for both Books and Podcast libraries).

Since the number of results in the dropdown is limited, I also added "ORDER By numItems DESC" to the tags and genres search queries, so we return the most useful ones first.

I chose "category" from the material symbols to denote genres. Please feel free to replace this with anything else.

### DIFF
--- a/client/components/cards/AuthorSearchCard.vue
+++ b/client/components/cards/AuthorSearchCard.vue
@@ -5,6 +5,7 @@
     </div>
     <div class="flex-grow px-2 authorSearchCardContent h-full">
       <p class="truncate text-sm">{{ name }}</p>
+      <p class="text-xs text-gray-400">{{ $getString('LabelXBooks', [numBooks]) }}</p>
     </div>
   </div>
 </template>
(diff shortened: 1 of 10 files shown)


mikiher:
Also fixed missing encodeURIComponent on the query (which caused ampersand in query not to be handled correctly).

advplyr:
Thanks! I think that icon makes sense for genres

PR #3188, opened 2024-07-22
New share - if existing share, then return the item

sevenlayercookie:
When attempting to create a share, currently request will fail returning “share already exists.” However, it would be more useful if the request returned the existing mediaitem, from which the requester can find the existing slug and use that instead of creating a new share.

Another option would be to create a new GET route that gets the existing share URL/slug by providing libraryitemid as parameter.

### DIFF
--- a/server/controllers/ShareController.js
+++ b/server/controllers/ShareController.js
@@ -272,7 +272,7 @@ class ShareController {
       })
       if (existingMediaItemShare) {
         if (existingMediaItemShare.mediaItemId === mediaItemId) {
-          return res.status(409).send('Item is already shared')
+          return res.status(200).json(existingMediaItemShare.?toJSONForClient())
         } else {
           return res.status(409).send('Slug is already in use')
         }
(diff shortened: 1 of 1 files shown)


advplyr:
This is the create media item share endpoint so it should return an error when trying to create. Can you be more specific on your use-case? Are you using the API directly or is this something with the UI/UX in the web client?

sevenlayercookie:
Yeah, I guess that makes sense. I'm new to APIs in general, but I have a working Amazon Alexa skill that I'd like to make public for users, but it requires bypassing the ABS auth. I think shares API would be a more secure way of handling it. Ideally a call to (1) check for and retrieve an existing share for a mediaitem, and (2) another to create a share if needed. I see (2) exists, but I didn't see (1) in the code.

sevenlayercookie:
Actually, I just realized the /api/items now has a share parameter, which will accomplish what I need. Will close this. Thanks!

emigrating:
@sevenlayercookie what kind of Echo skill? Is it for ABS, if so, care to share what you've got? :)

PR #3195, opened 2024-07-25
Simplify ItemSearchCard component by removing matching logic

mikiher:
This removes the mostly defunct matching logic from ItemSearchCard.vue, and also gets rid of the now unused matchKey and matchText fields in book and podcast search results.


advplyr:
Thanks!

PR #3198, opened 2024-07-27
WIP: feat: add more OpenAPI specs

Vito0912:
This PR is a work in progress and may never be "finished."

I will be updating it with commits periodically. This PR should not be merged until the last commit is ready for merging. If the last commit is not ready for merging, I am still adding new specifications. So at that time I did not tested it or generated the `OpenApi.json`. 
I try to test everything somewhat. Please review before merging.

Contributions and suggestions for better practices are appreciated, including naming conventions, references, etc. This is my first time writing OpenAPI specs, and I am doing my best to ensure each component is unique and references existing schemas when possible.

Currently, I am testing a bit, but I wanted to open this PR to show what I am working on so others can help or contribute.

### DIFF
--- a/docs/controllers/AuthController.yaml
+++ b/docs/controllers/AuthController.yaml
@@ -1,0 +1,73 @@
+paths:
+  /login:
+    post:
+      summary: Login to the server
+      description: Logs in a client to the server, returning information about the user and server.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                username:
+                  $ref: '../objects/entities/User.yaml#/components/schemas/username'
+                password:
+                  $ref: '../objects/entities/User.yaml#/components/schemas/password'
+              required:
+                - username
+                - password
+      responses:
+        '200':
+          description: Success
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  user:
+                    type: object
+                    properties:
+                     $ref: '../objects/entities/User.yaml#/components/schemas/user'
+        '401':
+          description: Unauthorized - Invalid username or password.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+  /logout:
+    post:
+      summary: Logout from the server
+      description: Logs out a client from the server. If the socketId parameter is provided, the server removes the socket from the client list.
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                socketId:
+                  $ref: '../objects/entities/User.yaml#/components/schemas/socketId'
+      responses:
+        '200':
+          description: Success
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  message:
+                    type: string
+                    example: Successfully logged out.
+        '401':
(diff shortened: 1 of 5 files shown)


Vito0912 (commented):
My language generator does not support the oneOf per https://github.com/OpenAPITools/openapi-generator. Therefore, I will focus on endpoints that do not use oneOf.

unl0ck:
@nichwall Looks good on my site, maybe push current state it is easier small pushes as big pushes

Vito0912:
@unl0ck I’m not sure this is ready for merging just yet.

I’ve encountered several issues with certain routes in my client. Additionally, I believe I added one or two routes to the OpenAPI spec for my client (if you want to take a look): [openapi-spec.json](https://github.com/Vito0912/abs_flutter/blob/main/lib/openapi/openapi-spec.json).

From what I recall, the search endpoint occasionally crashes. This seems to be related to the `series` attribute, which can either be a list of `Series` or just a single `Series` object. This inconsistency makes parsing tricky and less reliable. Moreover, I suspect there are some incorrect or inconsistent types. For instance, certain attributes (e.g., `backup`) can have multiple types—`null`, a boolean, or a string—which complicates handling.

Because of these challenges, I haven’t made much progress. To be honest, given the current state, it doesn’t seem logical to move forward if there are planned route changes. I recall nichwall mentioning the idea of establishing a solid foundation before introducing new endpoints (though I could be mistaken), but I feel that approach is no longer practical.

Lastly, we should avoid relying on `oneOf`. Nichwall’s PR appears to address this, which is a good direction. Most programming languages have limited or no support for `oneOf`, as outlined in the OpenAPI Generator documentation: [openapi-generator.tech/docs/generators](https://openapi-generator.tech/docs/generators). You will find near to none language supporting this. Dato: The goal of having autogenerated routes for thrid party devs does not exsits (despite having a proper doc ofc).
That just as a comment and my opinion.

I can help merging this into your repo so you can try plaing around with this and fix some things

Also, as some types are wrong (from the current docs) you can take a look at this models I made for my client. They should use proper types (but I merged many Objects together). I only have one Series Object that can have all nullable attributes (As there are even more distinct series models returned by the api as it currently states for example): https://github.com/Vito0912/abs_flutter/tree/main/lib/api

Vito0912:
Just scraped the docs:
| **Function**     | **Anzahl** |
|-------------------|------------|
| Simple           | 75         |
| Composite        | 74         |
| Polymorphism     | 39         |
| Union            | 2          |
| allOf            | 5          |
| anyOf            | 5          |
| oneOf            | 7          |
| not              | 0          |

**Generators that support `oneOf`:**
- cpp-restsdk
- dart-dio
- go
- julia-client
- python
- python-pydantic-v1
- rust

unl0ck:
@Vito0912 you are right swift is also an problem with oneOf and others

Vito0912:
@unl0ck I made a complete (excluding the LibraryitemController - As it should be near to impossible to do that as multiple specs fit in eachother) spec in the Discord. 
I will later or sooner close this PR in favour of a new I will start then. 
It should include all current docs and some new routes added in the meantime.

It has it's own issues though and will not replace the current version

PR #3204, opened 2024-07-29
Fix permission issues in embed/merge

### LINKED ISSUE #3146: [Bug]: Book file inaccessible outside ABS after using "embed metadata" due to permissions change
### What happened?

Prior to upgrading to v2.11.0, I could open book files from within ABS or the file share from Windows Explorer using VLC. I could also copy and/or delete the book file from Explorer. If I use the "embed metadata" option on a book, I can no longer open it in VLC and I lose rights to the file, meaning I cannot copy, move, or delete the file from Explorer any longer. I have tested this 3x and the issue continuously repeats itself. However, I can play the book in ABS Web.

To be able to open, copy, move, or delete book files from the file system like I could prior to upgrading, even after using "embed metadata" 

1. Access a book folder via Windows Explorer, e.g. Tim Schwab\2023 - The Bill Gates Problem {Tim Schwab}\The Bill Gates Problem.m4b
2. Open M4B file with VLC & it plays without issue. Open M4B with MP3tag and it opens the file without any issue. I can copy, move, and delete the M4B file without issue. 
3. Go to ABS Web and edit book from step 1
4. Go to Tools tab in editor and choose "Quick Embed" 
5. Verify log file does not report any errors 
6. Go back to Windows Explorer & attempt to open the M4B file with VLC. VLC throws an error "cannot open"
7. Attempt to open the M4B file with MP3tag, does not work. 
8. Attempt to delete the M4B file from the directory in Windows Explorer. Receive error 0x0007003b
image
10. Login to Synology NAS and delete corrupted M4B file, and copy backup of original file (prior to using "quick embed") to the book folder.
11. Open M4B file with VLC in Explorer without issue, and no issue with MP3tag either. 
12. Attempt to copy, delete, move the original M4B (no issues).
13. ***See notes below - I checked permissions & found rights are being stripped after using "embed metadata"

### Audiobookshelf version

v2.11.0

Docker

None

### Logs

```shell
2024-07-09 22:34:47.158 INFO 
[AudioMetadataManager] Starting metadata embed task Embedding metadata in audiobook "The Bill Gates Problem".

2024-07-09 22:35:03.347 INFO
[AudioMetadataManager] Successfully tagged audio file "/NonFiction/Tim Schwab/2023 - The Bill Gates Problem {Tim Schwab}/The Bill Gates Problem.m4b"
```

### Additional Notes

I am logged into ABS Web with the same admin account I've always used when doing updates to books. 

This same issue occurred with the previous 2 ABS versions I had installed when I used the "Make M4B Audiobook File" tool to convert MP3s to M4B, so I stopped using that feature and manually did it using auto-m4b. 

Well, I just decided to check file permissions to see if ABS was changing the rights on the file. Sure enough, it is changing the rights on the file to read only, write permissions removed. It was not doing that previously. I had files I used the "embed metatdata" option on yesterday and the permissions on those files did not get changed. 

Original file permissions (before using embed metadata option)
image

Modified file permissions (after using embed metadata option)
image

mikiher:
This fixes #3146.

This changes the way the output of the embedding/merging tools is written back to the original directory, so that the permissions of the original audio files are kept intact. 

Before this, we used fs.move(), which moved the ffmpeg output to the original file, effectively replacing the original permissions with whatever permissions the ffmpeg process assigned to its output file. Instead, we now pipe the contents of the ffmpeg output into the original file, so the contents of the original file are replaced but all file attributes (e.g. permissions) remain intact.

### DIFF
--- a/server/managers/AbMergeManager.js
+++ b/server/managers/AbMergeManager.js
@@ -7,7 +7,7 @@ const { writeConcatFile } = require('../utils/ffmpegHelpers')
 const ffmpegHelpers = require('../utils/ffmpegHelpers')
 const Ffmpeg = require('../libs/fluentFfmpeg')
 const SocketAuthority = require('../SocketAuthority')
-const fileUtils = require('../utils/fileUtils')
+const { isWritable, copyToExisting } = require('../utils/fileUtils')
 const TrackProgressMonitor = require('../objects/TrackProgressMonitor')
 
 class AbMergeManager {
@@ -64,7 +64,7 @@ class AbMergeManager {
 
   async runAudiobookMerge(libraryItem, task, encodingOptions) {
     // Make sure the target directory is writable
-    if (!(await fileUtils.isWritable(libraryItem.path))) {
+    if (!(await isWritable(libraryItem.path))) {
       Logger.error(`[AbMergeManager] Target directory is not writable: ${libraryItem.path}`)
       task.setFailed('Target directory is not writable')
       this.removeTask(task, true)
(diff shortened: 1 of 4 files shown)


advplyr:
Thanks!

Issue #2741, opened 2024-03-14
[Bug]: Server fails to start after successful build on *BSD


moozhub:
### Describe the issue

Server fails to start when trying to find ffmpeg / ffprobe and / or attempting to grab static ffmpeg.

dev.js:

 ```
// Using port 3333 is important when running the client web app separately
const Path = require('path')
module.exports.config = {
    Port: 3333,
    ConfigPath: Path.resolve('config'),
    MetadataPath: Path.resolve('metadata'),
    FFmpegPath: '/usr/local/bin/ffmpeg',
    FFProbePath: '/usr/local/bin/ffprobe'
}
```

npm run dev:

```
[root@audiobookshelf ~/audiobookshelf]# npm run dev
> audiobookshelf@2.8.0 dev
> nodemon --watch server index.js
[nodemon] 2.0.22
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): server/**/*
[nodemon] watching extensions: js,mjs,json
[nodemon] starting `node index.js`
Config /root/audiobookshelf/config /root/audiobookshelf/metadata
[2024-03-14 01:37:41.708] INFO: === Starting Server ===
[2024-03-14 01:37:41.712] INFO: [Server] Init v2.8.0
[2024-03-14 01:37:41.716] INFO: [Database] Initializing db at "/root/audiobookshelf/config/absdatabase.sqlite"
[2024-03-14 01:37:41.765] INFO: [Database] Db connection was successful
[2024-03-14 01:37:41.935] INFO: [Database] Db initialized with models: user, library, libraryFolder, book, podcast, podcastEpisode, libraryItem, mediaProgress, series, bookSeries, author, bookAuthor, collection, collectionBook, playlist, playlistMediaItem, device, playbackSession, feed, feedEpisode, setting, customMetadataProvider
[2024-03-14 01:37:41.964] DEBUG: Daily Log file found 2024-03-14.txt (LogManager.js:94)
[2024-03-14 01:37:41.965] INFO: [LogManager] Init current daily log filename: 2024-03-14.txt
[2024-03-14 01:37:41.965] DEBUG: [LogManager] Daily log file already exists - set in Logger (LogManager.js:71)
[2024-03-14 01:37:41.969] DEBUG: [DailyLog] 2024-03-14: Loaded 104 Logs (DailyLog.js:132)
[2024-03-14 01:37:41.977] INFO: [BackupManager] 0 Backups Found
[...]

PR #3210, opened 2024-07-30
Fix: Filter undefined entries from loadExtensions array

OISumeko:
This PR fixes an issue introduced in #3199, where the server would crash on startup if `SQLEAN_UNICODE_PATH` was not set. 

As I'm not sure if `loadExtensions` will be used in other locations in the future, I figured filtering out `undefined` entries within the function might be preferable to to simply checking for a valid `SQLEAN_UNICODE_PATH` value before calling it.

### DIFF
--- a/.devcontainer/dev.js
+++ b/.devcontainer/dev.js
@@ -5,6 +5,5 @@ module.exports.config = {
   ConfigPath: Path.resolve('config'),
   MetadataPath: Path.resolve('metadata'),
   FFmpegPath: '/usr/bin/ffmpeg',
-  FFProbePath: '/usr/bin/ffprobe',
-  SkipBinariesCheck: true
+  FFProbePath: '/usr/bin/ffprobe'
 }
--- a/index.js
+++ b/index.js
@@ -9,7 +9,6 @@ if (isDev) {
   if (devEnv.MetadataPath) process.env.METADATA_PATH = devEnv.MetadataPath
   if (devEnv.FFmpegPath) process.env.FFMPEG_PATH = devEnv.FFmpegPath
   if (devEnv.FFProbePath) process.env.FFPROBE_PATH = devEnv.FFProbePath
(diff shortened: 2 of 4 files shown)


mikiher:
Thanks for finding this!

That is not the proper fix, though. This extension is now required and BinaryManager should find it, fetch it if it's not found, or exit, and so the value of SQLEAN_UNICODE_PATH should always be set at this point.

I'm guessing that you have the env variable SKIP_BINARIES_CHECK=1. Instead of your fix, can you please remove the following lines from BinaraManager.js:
```
    // Optional skip binaries check
    if (process.env.SKIP_BINARIES_CHECK === '1') {
      Logger.info('[BinaryManager] Skipping check for binaries')
      return
    }
```
as the the Binary check is now not optional anymore.

Also, if you could also remove other references to SKIP_BINARIES_CHECK from the code.

advplyr:
I think we should still allow passing in SKIP_BINARIES_CHECK but I agree that it should not be the default for the devcontainer `dev.js`.

I'm not sure that the binaries we are downloading will cover every OS and some users may need to supply their own and bypass the check. That was brought to my attention in https://github.com/advplyr/audiobookshelf/issues/2741#issuecomment-1998488283

mikiher:
OK, but if we keep SKIP_BINARIES_CHECK, we also need to fail graecefully if any of the required environment variable are not set, like this:
```
    // Optional skip binaries check
    if (process.env.SKIP_BINARIES_CHECK === '1') {
      for (const binary of this.requiredBinaries) {
        if (!process.env[binary.envVariable]) {
          Logger.fatal(`[BinaryManager] Environment variable ${binary.envVariable} must be set`)
          process.exit(1)
        }
      }          
      Logger.info('[BinaryManager] Skipping check for binaries')
      return
    }
```
The code requires the unicode extension for any search to work now.

advplyr:
I pushed this up here https://github.com/advplyr/audiobookshelf/commit/4a5345dd5d0b4da9dff8addde118916a383c56ba
Thanks!

PR #3214, opened 2024-07-31
persist and show the advanced encoding options

faush01:
The changes allow two things:

1 - persist the advanced encoding settings so they can used as the default advanced settings
  - Saves the advanced bitrate, channels, codec to local storage
  - Populates the saved values if you select used advanced options

2 - Saves the advanced settings used for an encode task to the task.data so they can be shown in the in progress dialog
  - Saves the passed in advanced encode options in the Task data filed
  - When viewing an in progress encode task it uses these to show what the current encode is using


advplyr:
Nice addition, thanks!

Issue #3172, opened 2024-07-17
[Enhancement]: Book Club/Group Reading


lkjfdsaofmc:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

A book club or group reading feature allowing a group of users within an ABS server to be added together.  Possibly with its own tab in addition to library/series/etc or possibly just as a row on the home page.  The page or row would allow viewing of books assigned to the group by the groups "leader" (or possibly just requiring a server admin) so that any member of the group can see the books assigned to the group.  Ideally this would also allow marking what position (chapter) in the book they are currently supposed to be at.

### Why would this be helpful?

This would allow for groups of family or friends who are reading the book together to easily track the groups current book and position, such as a book club wanting to read one chapter a week being able to easily see what position in the book they are.  This could also allow for tracking previous reading of the group potentially having the "read progress" of the group show the same for all users, so if there was a separate library tab or filter to show group progress instead of self each member of the group would be able to see all books read by the group and to what percentage they've been read.  The read progress wouldn't need to necessarily be tracking what each user has read but simply what was marked as a "goal" for each week being marked as completed progress.

### Future Implementation (Screenshot)

![image](https://github.com/user-attachments/assets/571fcf67-2b37-4e08-aeef-2fd3376b7baa)

### Audiobookshelf Server Version

2.11.0

### Current Implementation (Screenshot)

Currently as far as I'm aware such a feature is completely unplanned/unrequested/unavailable.

nichwall:
Related to:
https://github.com/advplyr/audiobookshelf/issues/2725 (sync playback)
https://github.com/advplyr/audiobookshelf/issues/375 (progress visible to others)

If just wanting to share a "read these books" list, a Collection can be used.

Issue #375, opened 2022-02-19
[Enhancement]: Show book progress of other users


TheOnlyJonto:
### Describe the feature/enhancement

Show the progress of other users on each book. This could be activated via a new toggle on the user permissions for users who have the permission "see other users' progress". There could also be a second permission that would allow or disallow the sharing of progress for users that wish for their progress to be private. Below is an example of what this could look like. 

![audiobookshelf progress](https://user-images.githubusercontent.com/23222233/154819686-612a23cc-48b3-48aa-a91d-35ab7f6051d2.png)

PR #3223, opened 2024-08-03
Format the progress info into a table

faush01:
This change formats the progress information into a row structure so it is easier to read and is more consistent with the page layout.

I also change the confirmation question to use the internal app dialogue instead of the browser native one.
And moved some of the text to strings resources

All the existing logic for what and when it is shown is the same, it is just come clean up and formatting changes. 

Before Changes
image

After Changes
image

### DIFF
--- a/client/pages/item/_id/index.vue
+++ b/client/pages/item/_id/index.vue
@@ -72,18 +72,6 @@
             </div>
           </div>
 
-          <!-- Progress -->
-          <div v-if="!isPodcast && progressPercent > 0" class="px-4 py-2 mt-4 bg-primary text-sm font-semibold rounded-md text-gray-100 relative max-w-max mx-auto md:mx-0" :class="resettingProgress ? 'opacity-25' : ''">
-            <p v-if="progressPercent < 1" class="leading-6">{{ $strings.LabelYourProgress }}: {{ Math.round(progressPercent * 100) }}%</p>
-            <p v-else class="text-xs">{{ $strings.LabelFinished }} {{ $formatDate(userProgressFinishedAt, dateFormat) }}</p>
-            <p v-if="progressPercent < 1 && !useEBookProgress" class="text-gray-200 text-xs">{{ $getString('LabelTimeRemaining', [$elapsedPretty(userTimeRemaining)]) }}</p>
-            <p class="text-gray-400 text-xs pt-1">{{ $strings.LabelStarted }} {{ $formatDate(userProgressStartedAt, dateFormat) }}</p>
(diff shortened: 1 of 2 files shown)


nichwall:
I personally like the current progress bubble because it fits with the "play and read buttons" right below it, but for the sake of discussion my thoughts are below.

The progress should still be near the "play" button so it is in context with where it's used (pressing play or read to continue the media).

We may want to change up the column order so there isn't a large gap between progress and start date for ebook-only items. The current implementation looks like this, with just the progress and started date. There is a discussion somewhere about showing progress individually for both an ebook and an audiobook, but I'm can't find it right now.
image

However, the table could make viewing progress of other users easier (if sharing progress with other users is implemented).
https://github.com/advplyr/audiobookshelf/issues/375
https://github.com/advplyr/audiobookshelf/issues/3172

advplyr:
I also prefer the progress bubble but it could be that I'm just used to it.

faush01:
Yes UX preference is a tricky thing. 
Just to be clear I am not a fan of the bubble, I feel it is out of place on the page and interrupts visual flow.

I started off this trying to add current position to the position info and the bubble but the bubble then got bigger and that looked even more out of place. So I started experimenting with different approaches of displaying this info, the table felt like the best and fit with the page flow the best.

I also thought of an option to have the current position info added to the metadata at the top in the table (Option 1) in the screenshot. I moved the table of position into the just under the play button (Option 3) and also tried just adding the current position to the bubble (Option 2).

Here is a screenshot with some options.
If you don't like any of them that if cool, UX is very subjective and sometimes what you are used to is comfortable.

image

advplyr:
I'm not a fan of storing the progress info with the metadata because it is mixing user-specific data with the metadata.
I haven't had the need to see the current time there.
A lot of the UI/UX came from other users suggestions over the years. I'm not stuck on any particular design.

faush01:
> I'm not a fan of storing the progress info with the metadata because it is mixing user-specific data with the metadata.
Sorry I was unclear, not storing, just displaying it with the other metadata at the top, as shown in option 1 in the screenshot. That change does not require any storage changes, it is just some layout vue code changes.
> A lot of the UI/UX came from other users suggestions over the years. I'm not stuck on any particular design.
Yeah UI/UX is a very personal thing, there are some good rules of best practice but a lot of it comes down to what looks good or works for people and because people are diverse that is hard to cover everyone.
Overall I love the design and layout, its just the progress bubble, it looks so out off place to my eye, but again that is in my opinion.

PR #3235, opened 2024-08-05
Temporariliy disable sqlite3 unicode extension on linux-arm64

mikiher:
This fixes #3231.

I temporarily disabled the loading and usage of the unicode extension for linux-arm64, until I'm able to fix the underlying issue.
I'd appreciate if you could push a new release with this fix. I tested it as best as I could on my windows dev setup before setting the condition to linux-arm64 (`this.loadUnicodeExtionsion = process.platform === 'linux' && process.arch === 'arm64' ? false : true`).

This of course reverts the accent-insensitive search change for linux-arm64.

We'll at some point need to find a way to do an integration test on a linux-arm64 machine - I don't have one at my disposal.
I started to play with QEMU, but I haven't fully set it up yet.

Issue #3234, opened 2024-08-05
[Bug]: audiobookshelf didn't laucn on arm64 with 2.12


Machou:
### What happened?

I update as usual audiobookshelf, via docker compose, here is the file :

```yaml
services:
  audiobookshelf:
    image: ghcr.io/advplyr/audiobookshelf:latest
    container_name: audiobookshelf
    environment:
      - AUDIOBOOKSHELF_UID=99
      - AUDIOBOOKSHELF_GID=100
      - TZ=Europe/Paris
    ports:
      - 13378:80
    volumes:
      - /home/freebox/docker/audiobookshelf/audiobooks:/audiobooks
      - /home/freebox/docker/audiobookshelf/config:/config
      - /home/freebox/docker/audiobookshelf/podcasts:/podcasts
      - /home/freebox/docker/audiobookshelf/metadata:/metadata
    restart: unless-stopped
```

### What did you expect to happen?

n/a

### Steps to reproduce the issue

1. Update on ARM64 with docker-compose on arm64

### Audiobookshelf version

2.12

### How are you running audiobookshelf?

Docker

I have try with my config and without, new repertory with fresh install; on 2 différents Debian.

### What OS is your Audiobookshelf server hosted from?

Linux Debian

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
audiobookshelf  | Config /config /metadata
audiobookshelf  | [2024-08-05 10:31:11.066] INFO: === Starting Server ===
audiobookshelf  | [2024-08-05 10:31:11.071] INFO: [Server] Init v2.12.0
audiobookshelf  | [2024-08-05 10:31:11.072] INFO: [Server] Node.js Version: v20.16.0
audiobookshelf  | [2024-08-05 10:31:11.073] INFO: [Server] Platform: linux
audiobookshelf  | [2024-08-05 10:31:11.074] INFO: [Server] Arch: arm64
audiobookshelf  | [2024-08-05 10:31:11.292] INFO: [BinaryManager] ffmpeg not found or version too old
audiobookshelf  | [2024-08-05 10:31:11.395] INFO: [BinaryManager] ffprobe not found or version too old
audiobookshelf  | [2024-08-05 10:31:11.397] INFO: [BinaryManager] unicode not found or version too old
audiobookshelf  | [2024-08-05 10:31:11.401] INFO: [BinaryManager] Installing binaries: ffmpeg, ffprobe, unicode
[...]

Issue #3231, opened 2024-08-04
[Bug]: edge container giving SQLite error on startup


ic1415:
### What happened?

When using the following compose file, I am getting an SQLite error at startup:

```
services:
  audiobookshelf:
    image: ghcr.io/advplyr/audiobookshelf:edge
    container_name: audiobookshelf
    networks:
      proxy:
    ports:
      - 13378:80
    volumes:
      - /path/to/audiobookshelf:/books
      - /path/to/audio:/audiobooks
      - /path/to/Docker/AudioBookshelf/config:/config
      - /path/to/Docker/AudioBookshelf/meta:/metadata
     #- /var/run/docker.sock:/var/run/docker.sock:ro # optional, for docker integrations
    environment:
      - TZ=America/New_York
      - DOCKER_HOST=socket-proxy
    restart: always
  #logging service pushing to loki
  promtail:
    image: grafana/promtail
    container_name: abs-promtail
    hostname: abs-promtail
    restart: unless-stopped
    volumes:
      - /path/to/Docker/AudioBookshelf/meta/logs/daily:/var/log/abs:ro
      - /path/to/Docker/loki/promtail-config.yml:/etc/promtail-config.yml
    command:
      - '-config.file=/etc/promtail-config.yml'
networks:
  proxy:
    external: true
```

### What did you expect to happen?

Expected that ABS server would start 

### Steps to reproduce the issue

1. use above compose file to build edge container
2. see errors in log

### Audiobookshelf version

edge

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
Config /config /metadata
[2024-08-04 12:52:57.196] INFO: === Starting Server ===
[2024-08-04 12:52:57.197] INFO: [Server] Init v2.11.0
[2024-08-04 12:52:57.198] INFO: [Server] Node.js Version: v20.16.0
[2024-08-04 12:52:57.278] INFO: [BinaryManager] Found valid binary ffmpeg at /ffmpeg
[2024-08-04 12:52:57.278] INFO: [BinaryManager] Updating process.env.FFMPEG_PATH
[2024-08-04 12:52:57.337] INFO: [BinaryManager] Found valid binary ffprobe at /ffprobe
[2024-08-04 12:52:57.338] INFO: [BinaryManager] Updating process.env.FFPROBE_PATH
[...]

PR #3236, opened 2024-08-05
Fix unicode.so loading

devnoname120:
`unicode.so` is dynamically linked against glibc but Alpine only comes with musl. Installing the compatibility layer fixes it.

Fix https://github.com/advplyr/audiobookshelf/issues/3231
Fix https://github.com/advplyr/audiobookshelf/issues/3234
Close https://github.com/advplyr/audiobookshelf/pull/3235

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,7 @@ RUN apk update && \
     tzdata \
     ffmpeg \
     make \
+    gcompat \
     python3 \
     g++ \
     tini


mikiher:
Nice! have you tested this?

So, if I read your comment correctly, this is not about arm64, it's about systems that don't have glibc?

devnoname120:
@mikiher Yes! I'm not exactly sure why other architectures seem to be able to load it (for now!), but they could break and installing the compatibility layer is the safe choice for the future.

image

image

advplyr:
Thanks!

PR #3239, opened 2024-08-05
Allow skipping to next item in queue

IcyTv:
This commit is as simple as it sounds: When the last chapter of an audiobook is playing, and the queue contains at least another item, the "Next Chapter" button instead skips to the next book.

I'm not all too familiar with this codebase, so if there are naming or coding conventions I'm ignoring, please feel free to correct this.
Also, since I almost exclusively use Audiobookshelf for audiobooks, I'm not sure of the terminology etc. when listening to podcasts and there might need to be some changes to accommodate podcasts.

### DIFF
--- a/client/components/player/PlayerPlaybackControls.vue
+++ b/client/components/player/PlayerPlaybackControls.vue
@@ -20,8 +20,8 @@
           <span class="material-symbols text-2xl sm:text-3xl">forward_media</span>
         </button>
       </ui-tooltip>
-      <ui-tooltip direction="top" :text="$strings.ButtonNextChapter" class="ml-4 lg:ml-8">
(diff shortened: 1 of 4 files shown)


advplyr:
There are a few issues with this PR so I implemented this feature in a separate commit https://github.com/advplyr/audiobookshelf/commit/1320b6d785bd501693577bea50a94e24a205efe2
I also opened an issue for tracking https://github.com/advplyr/audiobookshelf/issues/3299

In case you are curious..
In this PR the next button is activated if there are any items in the queue when it should only be activated if there is a next item in the queue.
When going to the next item in the queue it will only go to the second item in the queue because the `1` index is hardcoded.
It doesn't take into account podcasts as you mentioned so "Next Audiobook" doesn't work.

I think this is a nice feature and makes sense so I went with it. Thanks!

PR #3262, opened 2024-08-09
lang/localization of "year in review"

### LINKED ISSUE #3260: [Enhancement]: Add Multi-Language Support for Yearly Overview Images in Statistics Section
### Type of Enhancement

Web Interface/Frontend

 would like to propose adding language support for the statistics section, particularly the yearly overview images. Currently, these images are generated in English only. It would be great if they could also be available in other languages, such as German.

Adding language support for the yearly overview images would make the statistics more accessible and intuitive for non-English speakers. It would enhance the user experience by allowing users to interact with their listening data in their native language, making the platform more relatable and encouraging regular use.

I have attached a screenshot to help visualize the current implementation of the yearly overview images. The feature could potentially work by detecting the language settings of the user’s device or allowing the user to manually select their preferred language for the statistics display. This would then generate the yearly overview images in the chosen language.

2.12.2

image
I have attached a screenshot to help visualize the current implementation of the yearly overview images. The feature could potentially work by detecting the language settings of the user’s device or allowing the user to manually select their preferred language for the statistics display. This would then generate the yearly overview images in the chosen language.

Vito0912:
This PR adds localization for the shareable image on the Stats page (also known as "YEAR IN REVIEW").

It's important to note that since this is an image/canvas and the text is "drawn" onto it, the length of the text is not truly flexible, as the boxes/containers do not adjust accordingly.

I suggest adding a note in Weblate to inform translators that this string should not be too long and setting a "soft limit" to proactively warn when a certain character limit is exceeded. Of course, some characters are longer or shorter, but this would serve as a guideline.

closes #3260


advplyr:
These need to pass in a max length so the text is truncated and doesn't overflow. If a string is added that is too long it should not overflow the card.

Vito0912:
See images for reference

Vito0912:
A big thanks to @Heisenberg9469 for sharing the additional pages that are not shown to me, due to lacking listening in 2023, so I could check the dimensions! :)

advplyr:
Thanks!

Issue #2785, opened 2024-03-23
[Bug]: Podcasts duplicate in UIs


BlackHoleFox:
### Describe the issue

Audiobookshelf seems to be generating, or somehow finding, duplicate episodes of multiple podcasts I have added and listen to. This is similar to https://github.com/advplyr/audiobookshelf/issues/2122 but doesn't create duplicate files on disk and happens in versions after that bugfix.

Weirdly this doesn't happen with every single podcast in my library, just a few. The ones I notice the most are:
- Security Cryptography Whatever (https://feeds.buzzsprout.com/1822302.rss)
- A More Civilized Age (http://amorecivilizedage.net/rss)

Nothing jumps out in my logs either, though let me know if I should enable debug logs for a week or so to help.

I've deleted duplicates from the library multiple times at this point, but they just seem to be coming back. For example AMCA's library entry has every new episode duplicated right now:
<img width="561" alt="Capture" src="https://github.com/advplyr/audiobookshelf/assets/20936452/1455b127-aec0-4f56-a70a-7d70146f1b3d">

One of each is not linked to an RSS episode: 
<img width="680" alt="Capture" src="https://github.com/advplyr/audiobookshelf/assets/20936452/0b126dd2-8809-4554-8349-2518c91bc18a">

And there are no duplicates on disk. They're purely database/UI located. They show up duplicated in both the web UI and Android app:
<img width="527" alt="Capture" src="https://github.com/advplyr/audiobookshelf/assets/20936452/d3e5ad54-9d62-4aca-b9fb-94ab22117004">

### Steps to reproduce the issue

1. Add https://amorecivilizedage.net/rss as a podcast in your library
2. Enable `Schedule Automatic Episode Downloads`. I have it running every day at midnight.
3. Wait a few days

### Audiobookshelf version

2.8.0

### How are you running audiobookshelf?

Docker

advplyr:
Can you disable the watcher in server settings as a test to see if you get duplicates on any episodes downloaded after that?

BlackHoleFox:
Turned it off and restarted audiobookshelf's container. I'll cleanup the duplicates in the two library items I mentioned and see if they come back.

BlackHoleFox:
[...]

PR #3269, opened 2024-08-11
Log actual object details when they appear in log arguments

BlackHoleFox:
I'm continuing to hunt down the root cause of #2785 and got tired of seeing `[object Object]` around relevant logging lines. Its impossible to tell what the server is actually seeing change today:
```
{"timestamp":"2024-08-08 00:00:23.785","source":"LibraryItem.js:305","message":"[LibraryItem] \"name\" audioFile was updated from \"[object Object]\" to \"[object Object]\"","levelName":"DEBUG","level":1}
{"timestamp":"2024-08-08 00:00:23.786","source":"LibraryItem.js:305","message":"[LibraryItem] \"name\" extraData was updated from \"null\" to \"[object Object]\"","levelName":"DEBUG","level":1}
```

The changes here serialize the objects into JSON to remove the opaqueness to make them an actual string. Should have no effect on most users since I haven't really seen any logs containing these at non debug levels.

Do let me know if this can be cleaner. I don't usually write JS :)

### DIFF
--- a/server/models/LibraryItem.js
+++ b/server/models/LibraryItem.js
@@ -11,6 +11,7 @@ const Book = require('./Book')
 const Podcast = require('./Podcast')
 
 const ShareManager = require('../managers/ShareManager')
+const { LogLevel } = require('../utils/constants')
 
 /**
  * @typedef LibraryFileObject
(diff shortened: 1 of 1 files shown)


advplyr:
The cleaner way to handle this is using `util.format`.

When looking at this I noticed the `libraryFiles` objects if they got expanded could cause issues so I had to break that out separately. I did this in commit https://github.com/advplyr/audiobookshelf/commit/bbf214fa4c3d46acb8ed7aab241d1f205b757992

I also came across the duplicate podcast episode issue the other day. I'm not sure how it is happening for your specific case but I'll add notes to the open issue.

PR #3278, opened 2024-08-13
Go back to downloading binaries from ffbinaries.com

mikiher:
This is in continuation to our discussion in #3199, in which we decided not download binaries from Github due to rate-limiting on unauthenticated API requests.

This re-implements downloading from ffbinaries (using roughly the same class and methods used for downloading from Github, and _not_ introducing back libs/ffbinaries), and removes support for Github downloading.


advplyr:
Thanks!

PR #3138, opened 2024-07-08
Update material icons

glorenzen:
This pull request adds font files for Material Symbols to replace the Material Icons fonts, which are deprecated.

The Material Icons files and classes are still there, but I have added new Material Symbols font files and replaced all material-icons CSS classes with new material-symbols classes.

PR #3291, opened 2024-08-15
Reduce size of Material Symbols font files

justcallmelarry:
Ever since https://github.com/advplyr/audiobookshelf/pull/3138 I have had some weird loading times on the web ui for the icons used for different things.

I tracked down the issue to the Material Symbol fonts being huge, and tried looking into how to circumvent having to use the entire file.

After some searching i found a [Stack Overflow thread](https://stackoverflow.com/questions/76347398/why-does-material-symbols-font-take-so-long-to-load-and-how-can-i-optimize-it) where there were some tips into how to get smaller font files.

I tried to follow the instructions to the best of my knowledge, and made two smaller font files with the symbols and only font weight 400, which seems to be the only one used, unless my meagre css skills fail me, which they very well might.

After trying it locally I do not see anything amiss, so I also build the docker image with the updated client on my server and am running it there currently.

I cannot see any icons looking weird or not working, but I do not generally have a great eye for this, and I do not know all of the places where the icons can show (I tried figuring it out from the files in the mentioned PR, but you never know).

Any possibility that someone with some more knowledge in the area could help me out?

Links used to generate the font files:
Outlined: https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:wght,FILL@400,0..1
Rounded: https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:wght,FILL@400,0..1


advplyr:
The `FILL` font-variant-setting isn't working for `material-symbols-outlined` for those font files.

In looking at this more I realized that we don't need to have 2 different font files anymore. The only reason that we had 2 was because we wanted to have outlined and filled but now we can accomplish that with one font file using the `FILL` setting.

I think we should remove `Material Symbols Outlined` font and only include the `Material Symbols Rounded` font.

justcallmelarry:
I can have a look at going for one font file only on Sunday. Going away tomorrow for a trip, but will have a poke around once I'm back home.

advplyr:
I ended up implementing this. I wanted to see about cleaning up the loading since showing the ligatures is ugly.
I replaced some of the ligatures with codepoints so it just shows a square while loading similar in size to the icon.

https://github.com/advplyr/audiobookshelf/commit/f9f89e1e51791dc80902c3dd7a6218c2dab10e58

Issue #2579, opened 2024-02-04
No failed login limit


sevenlayercookie:
### Describe the issue

It is possible to submit any number of failed login requests without limit, and without rate limit. **This makes Audiobookshelf susceptible to brute force login methods.**

as @nichwall noted, this appears to have broken in v2.6.0 with OIDC implementation.

related issue: #2533 no auth log entries.

### Steps to reproduce the issue

1. Submit incorrect password as many times as you like.
2. Login with your normal password immediately afterward without being locked out.

### Audiobookshelf version

V2.7.2

### How are you running audiobookshelf?

Docker

Sapd:
Probably was removed during the Auth/passport rerwite.

I think instead of implementing that directly, better would be a middleware:
https://www.npmjs.com/package/express-slow-down

Advantage is, it could be also used against other public API endpoints.

bytebone:
No need to solve problems the complicated way. Just fix the logging component so that IPs of unsuccessful login attempts are in logfiles, and let users figure out how to handle this with fail2ban / crowdsec / whatever else.

plague-doctor:
To enhance our security posture against brute-force attacks, I've integrated AudiobookShelf with [crowdsec](https://www.crowdsec.net) . This addition specifically addresses the challenge of protecting against unauthorized access attempts through repeated login failures.

The AudiobookShelf [collection](https://app.crowdsec.net/hub/author/PlagueDoctor/collections/audiobookshelf) now actively monitors and mitigates potential threats, ensuring a more robust defense against malicious actors attempting to gain unauthorized access to our systems through brute-force methods.

sevenlayercookie:
> Has anyone got fail2ban jail and filter working for audiobookshelf yet? 

I ended going for crowdsec because it's more sophisticated. It was complicated to set up because I use Cloudflare tunnels so Linux firewall doesn't apply. So I ended up setting crowdsec to parse my Caddy logs for Audiobookshelf subdomain for suspicious traffic and then crowdsec tells
Caddy to block further traffic from that IP address.

I haven't used fail2ban, but I imagine that aspect of setup is similar.

CTalvio:
@Guruleenyc 

Setting up fail2ban to watch any arbitrary log file is not difficult. You can find fairly simple guides for how to set up custom jails. If you have access to the log, and a regex string that matches with log entries for failed logins, you're set. Here's the filter I wrote for my system just now:

```
# Filter for audiobookshelf 
[Definition]
[...]

PR #3295, opened 2024-08-17
Log failed auth attempts

faush01:
Log failed auth attempts for local auth mode.
This should allow for better monitoring of brute force user/password attacks.

### DIFF
--- a/server/libs/passportLocal/strategy.js
+++ b/server/libs/passportLocal/strategy.js
@@ -3,7 +3,8 @@
  */
 const passport = require('passport-strategy')
 const util = require('util')
-
+const Logger = require('../../Logger')
+const requestIp = require('../../libs/requestIp')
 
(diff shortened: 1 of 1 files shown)


nichwall:
Partially fixes https://github.com/advplyr/audiobookshelf/issues/2579 based on follow up comments

advplyr:
This is better implemented in Auth than in the local auth lib. I added it here https://github.com/advplyr/audiobookshelf/commit/1b914d5d4fbb8ac53b0bf760986ceb201237269b along with logs for successful local auth logins.

faush01:
Thanks, makes sense. 
I was not sure how best to get the req into the call, I did see the option passReqToCallback but it was weird as sometimes a function was passed in to options. I am not really a fan of JS with its loose var types, and having logic use detected var types makes my head hurt.

Issue #3242, opened 2024-08-05
[Enhancement]: missing translation 2.12.1+


Machou:
#### 1

Go on an item > on file list > More infos > button **Probe Audio File**

![1](https://github.com/user-attachments/assets/e6f9608c-19e6-4cc8-be3d-acb6dfdb92d6)

-----

#### 2

Go to an item (need be View) > Remove the tooltip > Alert: **Are you sure you want to reset your progress?**

![3-min(1)](https://github.com/user-attachments/assets/ce6898cf-12d1-4235-9023-14e8c8174257)

-----

#### 3

Start Reading an Audio Book > Sleep timer > Button **Set**

![Capture d’écran 2024-08-10 023135](https://github.com/user-attachments/assets/a2a1872e-1962-4127-8efa-b999d268561a)

-----

#### 4

Start Reading an Audio Book > Click on reset timer > Tooltip Popup: **Your progess was reset**

![a](https://github.com/user-attachments/assets/34459402-c813-43ad-9da1-344059d186be)

-----

#### 5

Upload a file: **Updated** & **Added**

![Capture d’écran 2024-08-11 234335](https://github.com/user-attachments/assets/85a4f847-9fc6-4468-990a-eff3a509cf07)

-----

#### 6

On eBook > Go to **Find** > Tooltip **Please wait**

![6](https://github.com/user-attachments/assets/98fc205f-ccc5-43fe-af47-accdc63b1c6c)

-----

#### 7

Go to **Library** > Sort by **File creation** > Below a book > **Born**

Same fr **Modified**
Same for **Duration**
Same for **Size**
Same for **Added** 
Same for **Published**

![Capture d’écran 2024-08-12 001533](https://github.com/user-attachments/assets/5baa116e-ccf5-4129-8b2b-64b01885a1a1)

-----

#### 8

Go To Authors > if 1 book, book have a **s**

1 book
2 book**s**

![Capture d’écran 2024-08-20 124812](https://github.com/user-attachments/assets/4c5b07ce-ff2b-44ad-8a08-c13db47bbb48)

MikeGold:
Same in Polish.
Potential translations for your convenience:

1. 
Zbadaj plik audio

2. 
Czy jesteś pewien, że chcesz zresetować swój postęp?

3. 
5 minut
15 minut
30 minut
45 minut
60 minut
90 minut
2 godziny
Koniec rozdziału

4. 
Twój postęp został zresetowany

[...]

Issue #3303, opened 2024-08-19
[Enhancement]: Server generated strings to be localized


nichwall:
### Type of Enhancement

None

### Describe the Feature/Enhancement

Client strings are able to be translated, but strings generated by the server are not able to be translated right now. This was originally mentioned in https://github.com/advplyr/audiobookshelf/issues/3242 under point 5, where the Tasks are not translated.

Currently, Weblate is only looking at the string resource in `client/strings/`. We can add the localization for the client include these strings for the server, but having the server reference a resource file as `../client/strings` feels a little clunky. Not sure what the best way to handle that is.

### Why would this be helpful?

Ability to localize tasks and other server generated strings.

### Future Implementation (Screenshot)

N/A

### Audiobookshelf Server Version

v2.12.3

### Current Implementation (Screenshot)

Task strings are generated by the server

![image](https://github.com/user-attachments/assets/6ecc443e-ed43-400e-bb20-3f778955bff0)

advplyr:
Since the client is where users are selecting their language (besides the server default language) the translation should remain client side.

The way I solved this for the home page shelves is with an additional `labelStringKey` for each shelf and that key maps to a string in `client/strings/..`.
3rd party clients can optionally continue using the english only label or localize.

https://github.com/advplyr/audiobookshelf/blob/27b3a4414726a6a2778fb19e6038ce3ac1195aee/server/models/LibraryItem.js#L622-L629

It won't be exactly the same for these task descriptions since they also include a variable in the string, but a similar solution can be used.

nichwall:
Ah, great. Yeah, I specifically mean things like the watcher where it's hard coded, but something like you shared for the shelves should work.

https://github.com/advplyr/audiobookshelf/blob/27b3a4414726a6a2778fb19e6038ce3ac1195aee/server/Watcher.js#L287

advplyr:
All the tasks made with the TaskManager are hard coded so whatever solution we come up with should work for all tasks and be an update to the Task object.

PR #3304, opened 2024-08-19
Translation strings added

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/3242

This PR updates the strings in the client mentioned in the above issue, as well as updating other strings in those files which were not translatable.

This PR does *not* update the task strings which are generated/updated by the server because right now only the client has i18n functionality. I opened https://github.com/advplyr/audiobookshelf/issues/3303 to track this separately.


nichwall:
Got some more strings translated. I also cleaned up a few similar strings that were already in the translation files.

This PR should be good to merge. There are about 100 toast strings left based on the number of search results, but it will probably be a few days before I get to them so I can open a new PR for those.

nichwall:
Okay, I'll stop editing the strings here. There are still about 50 more toast messages remaining, but this PR just kept getting longer. More similar strings were combined.

Most of the remaining strings are related to podcast episodes and are not all in toasts, so that will take a bit to go through the files and update everything.

advplyr:
Thanks! I wanted to remove a lot of those success toasts because it is redundant when the change can be seen.
Most of the "failed to get.." error toasts can be consolidated in the generic failed to load toast. There is always a more specific console log in the rare case it has to be distinguished.

nichwall:
Makes sense, thanks for cleaning those up.

PR #3305, opened 2024-08-19
Update api linting workflow

nichwall:
This PR fixes the trigger for the OpenAPI linting workflow, as well as limits the permissions for the workflow to read-only of `contents`.


advplyr:
Thanks!

Issue #3272, opened 2024-08-11
[Enhancement] Clear image cache when restoring from backup


nichwall:
### Type of Enhancement

Server Backend

### Describe the Feature/Enhancement

When restoring from backup, the image cache should probably be cleared to allow for covers to be generated again.

Originally reported on Discord (message and image copied), resolved by clearing the server cache.

"Hello, i recently moved to a new NAS and used the backup and restore function to move the installation, settings etc. It seemed to work fine and on the web interface everything is normal, if I want to connect via the Android app however I get the displayed issue you can see in the screenshot (both for podcasts and audiobooks). I also can't play anything since it says "server request failed". After closing the app and opening it again it also doesn't remember the server, so I have to login again. I have uninstalled and reinstalled the app, but that has not helped. I'd really appreciate some support. Thanks"

![Screenshot_20240811-082907](https://github.com/user-attachments/assets/15831ab2-abc7-482d-a84b-452eacadd00f)

https://discord.com/channels/942908292873723984/981321213882282035/1272080567885893663

### Why would this be helpful?

Some confusing behavior in clients after restoring backup

### Future Implementation (Screenshot)

N/A

### Audiobookshelf Server Version

v2.12.3

### Current Implementation (Screenshot)

Database can include references to images which do not exist when moving servers

PR #3311, opened 2024-08-20
Backup restore clear cache

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/3272

This PR purges the entire cache in `/metadata/cache` after restoring from a backup. The main reasoning for this is that backups may have different images or be done when migrating to a new server, so the cache should be cleared regardless.

I also updated the `CacheManager` to use `ensureDir` to be a little more readable.


advplyr:
The reason `ensureDir` wasn't used there initially is because we used to set the UID/GID from the env variables but now that we aren't doing that `ensureDir` is better. Thanks!

PR #3319, opened 2024-08-21
Fix Content-Type header when browser user-agent is from an Apple mobile device

### LINKED ISSUE #3310: [Bug]: iOS Safari incorrectly interprets file download content-type headers and adds incorrect file extensions
### What happened?

This isnt' a bug with audiobookshelf, but I'm filing this issue because I'd like to implement a workaround for bad iOS Safari behavior. 

Basically, when downloading a book such as `my-favorite-book.m4b` from the "audio tracks" section of audiobookshelf, Safari on iOS will prompt if you wish to view/download `my-favorite-book.m4b.mp4`, adding the `.mp4` extension. Based on some testing, Safari on iOS (and apparently Chrome too) will interpret the `Content-Type` header and add an "appropriate" file extension to the existing file name being downloaded if needed.

Audiobookshelf is correctly returning `Content-Type: audio/mp4` however, Safari interprets this to mean it should be an "mp4" file, causing it to add the `.mp4` extension (even though "mp4" and "m4b" are _technically_ the same, and m4b itself is a convention that Apple itself created... 🙃).

Now, this isn't the end of the world, but it means for certain audiobook readers, it will not open the file unless it's renamed, as these apps interpret the file as a video file based on the file extension. This means you must rename the file in the iOS files app before it can be imported into the audiobook reader app.

Now, this is something I can manage, but it's a lot to ask for my wife and my mom. What I'd like, is to see if you would be willing to accept a PR to add a work-around for this behavior. I've tested, and if you use `Content-Type: audio/m4b` (yes: which isn't standard) it doesn't add the file extension.

So I'd like to ask:

Would you accept a PR that:
- Unconditionally returns `Content-Type: audio/m4b` for m4b files
OR:
- Checks the user-agent to see if the browser is Safari and returns `Content-Type: audio/m4b` for m4b files. This is probably more complex/complicated then simply using `audio/m4b` though.

References:
- https://discussions.apple.com/thread/250904310?sortBy=rank
- https://github.com/epoupon/fileshelter/issues/30
- https://www.redmine.org/issues/31275
- https://discussions.apple.com/thread/6493440

There's a ton of references if you search "Safari unwanted file extension" in Google.

The file is downloaded without the `.mp4` extension.

On an iPad/iPhone:
1. Open Safari
2. Browse to audiobookshelf
4. Open an audiobook with an m4b file in audiobookshelf
5. Open the "Audio Tracks" section 
6. Click on the ⋮
7. Click download
8. Notice the prompt to download the file adds an extra extension

### Audiobookshelf version

v2.12.3

Docker

Linux

Safari on iOS

### Logs

_No response_

### Additional Notes

I've temporarily made a work-around in my reverse proxy to simply hardcode the content-type header to `audio/m4b` but that will cause issues for any other file types I try to download.

chancez:
Fixes #3310

### DIFF
--- a/server/controllers/LibraryItemController.js
+++ b/server/controllers/LibraryItemController.js
@@ -1,13 +1,14 @@
 const { Request, Response, NextFunction } = require('express')
 const Path = require('path')
 const fs = require('../libs/fsExtra')
+const uaParserJs = require('../libs/uaParser')
 const Logger = require('../Logger')
 const SocketAuthority = require('../SocketAuthority')
 const Database = require('../Database')
 
 const zipHelpers = require('../utils/zipHelpers')
 const { reqSupportsWebp } = require('../utils/index')
-const { ScanResult } = require('../utils/constants')
+const { ScanResult, AudioMimeType } = require('../utils/constants')
 const { getAudioMimeTypeFromExtname, encodeUriPath } = require('../utils/fileUtils')
 const LibraryItemScanner = require('../scanner/LibraryItemScanner')
 const AudioFileScanner = require('../scanner/AudioFileScanner')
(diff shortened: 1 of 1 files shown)


advplyr:
Tested on multiple browsers + native app and it works. Thanks!

PR #3328, opened 2024-08-24
Update series and collection width to account for book aspect ratio

mikiher:
This addresses @advplyr's [comment](https://github.com/advplyr/audiobookshelf/pull/3037#issuecomment-2303164251) on PR #3037

I fixed cardWidth() on both Series and Collection cards to account for the aspect ratio correctly.

I checked against the Series and Collections pages as they were in v2.10.1 (before PR 3037 was integrated), and now they look the same. Sorry for missing this on that PR.


advplyr:
Thanks!

PR #3330, opened 2024-08-24
Feature - New endpoint /libraries/stats

Vito0912:
A new endpoint has been added at `/libraries/stats` to retrieve the statistics for all libraries (Same as `/libraries/:id/stats`). The endpoint returns a list of individual statistics for each library.

This is part of the preparations for a new page.

Moved logic from `stats` to Helper class


Vito0912:
United with #3335

Issue #2538, opened 2024-01-20
[Enhancement]: Warning: connect.session() MemoryStore is not designed for a production environment


nichwall:
### Describe the feature/enhancement

There have been a few reports in recent weeks around express-rate-limit using `memory-store`. This issue is to provide context.

`MemoryStore is not
designed for a production environment, as it will leak
memory, and will not scale past a single process.`

This warning does not matter for ABS, but has been causing some confusion for users.
https://express-rate-limit.mintlify.app/reference/stores

The MemoryStore is only used to provide rate limiting and does not end to be very robust since it is designed for a single server and rebooting does not affect the rate limiting feature for ABS (ABS only tracks for the past 60 seconds).

Potential changes:
1. Suppress the warning so it doesn't show up at all (don't like that idea)
2. Use a different storage, such as `rate-limit-memcached`. That seems to be the lightest replacement but adds a dependency

zzyzx-dc:
Thank you for posting this. I am troubleshooting why Audiobookshelf crashes every week or two and I saw this error message. Good to know it is not related to my issue.

nichwall:
> Thank you for posting this. I am troubleshooting why Audiobookshelf crashes every week or two and I saw this error message. Good to know it is not related to my issue.

Glad it helped!

If you can provide logs around the crash, that will help narrow it down. If you haven't already done so, change your server log level to Debug. You may also need to look at the Docker or systemctl logs (depending on how you are running the server) because those logs can catch more information than ABS itself does for crashes.

Please post those logs/description in a new issue or on Discord so this issue doesn't get a ton of unrelated comments. :)

Eschguy:
Having this issue as well. Based on info logs, mine seems to crash after updating podcasts. I've set the logger to debug and will post results when it crashes next. 

INFO logs from crash/startup
```
{"timestamp":"2024-04-01 15:00:03.285","source":"LibraryItem.js:389","message":"[LibraryItem] Library item \"0246f2a8-adfd-4912-ac7e-8e62049b29b1\" updated","levelName":"INFO","level":2}
[...]

Issue #3251, opened 2024-08-07
[Bug]: ABS Server becomes unreachable 


PulsarFTW:
### What happened?

This morning (07:00 local time) my audiobookshelf instance was unreachable via the app and browser. I checked using the internal IP and external address. 

There are no entries in the debug log or the server crash log and I don't know what time the issue occurred.

Additional Information:
-My automatic backup scheduled at 06:00 did not run.
-I changed the time and ran an automatic backup after this incident which was successful so crash unlikely to be related.
-I upgraded to 2.12.1 yesterday and this was the first night this version was running.
-No new books/files have been added for two days.

### What did you expect to happen?

I expected to be able to reach the ABS install via website/app. This is the first crash I've noticed on this install in 6+ months.

### Steps to reproduce the issue

Currently unknown, sorry.

### Audiobookshelf version

2.12.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
{"timestamp":"2024-08-07 00:41:18.666","source":"PlaybackSessionManager.js:304","message":"[PlaybackSessionManager] syncSession \"aa20c784-8b42-4d7c-b852-a282b31b27d0\" (Device: SM-N976B SDK 31 / v0.9.74-beta) | Total Time Listened: 596","levelName":"DEBUG","level":1}
{"timestamp":"2024-08-07 00:41:18.681","source":"ApiCacheManager.js:21","message":"[ApiCacheManager] Array.afterUpsert: Clearing cache","levelName":"DEBUG","level":1}
{"timestamp":"2024-08-07 00:41:18.681","source":"SocketAuthority.js:58","message":"[SocketAuthority] clientEmitter - no clients found for user edd38149-4318-4100-b8de-9434198f65bd","levelName":"DEBUG","level":1}
{"timestamp":"2024-08-07 00:41:18.693","source":"ApiCacheManager.js:21","message":"[ApiCacheManager] Object.afterBulkUpdate: Clearing cache","levelName":"DEBUG","level":1}
[...]

PR #3332, opened 2024-08-24
memorystore: simplify, refactor, re-enable

itzexor:
Removes a lot of unused (in ABS) functionality, refactors to ES6 style class, and re-enables this custom implementation with check period and ttl of 1 day, and 1000 max entries.

The class now only implents the required (as per express-session docs) methods and removes optional methods, except touch() which allows the TTL of an entry to be refreshed without affecting its LRU recency.

There is no longer a way to stop the prune timer, but I don't believe the function was ever being called beforehand. The session store's lifetime is the same as the application's, and since it is unref()'d should not cause any shutdown issues.

ref: https://github.com/advplyr/audiobookshelf/issues/3251
fixes: https://github.com/advplyr/audiobookshelf/issues/2538

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -41,6 +41,7 @@ const LibraryScanner = require('./scanner/LibraryScanner')
 //Import the main Passport and Express-Session library
 const passport = require('passport')
 const expressSession = require('express-session')
+const MemoryStore = require('./libs/memorystore')
 
 class Server {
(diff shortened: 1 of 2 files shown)


itzexor:
Store API reference: https://www.npmjs.com/package/express-session#session-store-implementation

advplyr:
I tested with lower prune/ttl values and it is working great. Thanks for fixing that!

PR #3334, opened 2024-08-25
feat: Add support for "Open Current Chapter" button

BimBimSalaBim:
Added a button to the playerUI if there is an ebook that goes along with the audiobook. If the file is not an epub it will just open like normal but if it is an epub the file will open to the current chapter of the audiobook that is being played.

### DIFF
--- a/client/strings/en-us.json
+++ b/client/strings/en-us.json
@@ -48,6 +48,7 @@
   "ButtonNextChapter": "Next Chapter",
   "ButtonNextItemInQueue": "Next Item in Queue",
   "ButtonOk": "Ok",
+  "ButtonOpenCurrentChapter": "Open Current Chapter",
   "ButtonOpenFeed": "Open Feed",
   "ButtonOpenManager": "Open Manager",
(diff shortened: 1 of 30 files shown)


client/strings/en-us.json
```diff
@@ -48,6 +48,7 @@
   "ButtonNextChapter": "Next Chapter",
   "ButtonNextItemInQueue": "Next Item in Queue",
   "ButtonOk": "Ok",
+  "ButtonOpenCurrentChapter": "Open Current Chapter",
```
nichwall, line 51: We only want to add the string to the English file, not all of the other files unless you know the translation (Weblate will handle the other languages)

BimBimSalaBim: How do i undo all the changes? do i just undo locally and commit again?

nichwall: If you did all of the changes in one commit you can do `git revert [SHA]`, or checkout the files from a specific commit and then do another commit, or undo it manually.

Thanks for doing that, it makes it easier to review the changes. :)

PR #3336, opened 2024-08-25
Feature - new view: Activity Summary

Vito0912:
Adds a view to display the total listening time and the number of items the user has listened to in this library. Additionally, it recommends books that the user can listen to for further progress.

Currently, podcasts are not supported, as there is no efficient method to retrieve all episodes in a library without individually requesting each podcast (at least not to me). Redirects have been implemented to prevent access when a library is a podcast.

image

(The name of the page has been changed to "Activity Summary" as I think it fits the scope more and "Current Progress" was removed)


Vito0912:
Not suitable for large use as it would need to load all items

PR #3348, opened 2024-08-31
Config issue workflow

nichwall:
This PR adds a workflow to automatically comment on issues when labels are applied to the issue. The two included with this PR are `help wanted` and `config-issue`, but they can be renamed as needed.

When an issue has the `config-issue` label applied to it, the issue will automatically be commented on using the template message and then closed as "not planned".


nichwall:
An example of this workflow running can be seen here https://github.com/nichwall/issue_testing/issues/6

advplyr:
Great, thanks!

PR #3350, opened 2024-08-31
Delete untranslated strings

nichwall:
This PR deletes all of the keys from other language files that are the same as the English source string. This is because there are some languages that incorrectly show as mostly translated, but many of their strings have obviously not been translated. This will require translators to mark some strings as the same in their language if the English string was identical, but Weblate provides a single button to map the string for untranslated strings.

I pulled the latest strings from Weblate, so once everything is merged I will fix the Weblate repo.


advplyr:
Makes sense, thanks!

Issue #3353, opened 2024-09-01
[Bug]: Items created through web client causes crash


nichwall:
### What happened?

Things that are currently causing a crash:
- Adding new podcasts through web UI
- Uploading books through web UI

Scanning books added through the file system is working for creating new library items.

This may be related to some of the migrations away from the old data model.

### What did you expect to happen?

Podcasts and books to be created and server to not crash

### Steps to reproduce the issue

1. Create new server of version 2.13.0
2. Create a single podcast library
3. Search for a new podcast. I tested with "Up First from NPR" and "Business Wars"
4. Attempt to add podcast. Crash occurs after clicking "Submit" button on the pop-up
![image](https://github.com/user-attachments/assets/79b6705a-998a-484b-8bb6-3fc53dded887)

### Audiobookshelf version

v2.13.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Other (list in "Additional Notes" box)

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
Config /config /metadata
[2024-08-31 20:16:46.029] INFO: === Starting Server ===
[2024-08-31 20:16:46.031] INFO: [Server] Init v2.13.0
[2024-08-31 20:16:46.032] INFO: [Server] Node.js Version: v20.17.0
[2024-08-31 20:16:46.032] INFO: [Server] Platform: linux
[2024-08-31 20:16:46.032] INFO: [Server] Arch: x64
[2024-08-31 20:16:46.039] INFO: [Database] absdatabase.sqlite not found at /config/absdatabase.sqlite
[2024-08-31 20:16:46.040] INFO: [Database] Initializing db at "/config/absdatabase.sqlite"
[2024-08-31 20:16:46.137] INFO: [Database] Db connection was successful
[2024-08-31 20:17:00.317] INFO: [Database] Db initialized with models: user, library, libraryFolder, book, podcast, podcastEpisode, libraryItem, mediaProgress, series, bookSeries, author, bookAuthor, collection, collectionBook, playlist, playlistMediaItem, device, playbackSession, feed, feedEpisode, setting, customMetadataProvider, mediaItemShare
[2024-08-31 20:17:00.350] INFO: [LogManager] Init current daily log filename: 2024-08-31.txt
[2024-08-31 20:17:00.678] INFO: [BackupManager] 0 Backups Found
[...]

PR #3354, opened 2024-09-01
Fix libraryFolder not existing on new `libraryModel`

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/3353, or at least the cause of two crashes I was able to identify in that issue. I am not sure if there are other places where a similar problem occurs.

This PR uses the `libraryFolders` table to ensure the folder exists. This was just my attempt at a quick fix, so no worries if there's a better/easier way to resolve the issue.


advplyr:
Thanks for quickly identifying that one. There was a better way that I implemented which is to include the libraryFolders when fetching the library. https://github.com/advplyr/audiobookshelf/commit/fc276b330a602fe6f4728d612ffd434b39b740f0

Issue #3366, opened 2024-09-02
[Bug]: Cannot set email for root user


nichwall:
### What happened?

Attempt to set email for root user when logged in as the root user causes an error of `Invalid account type`.

It looks like this is due to the validation not including the `root` account type due to groups validation from OAuth and the `root` account being a specific type that cannot be overridden.
https://github.com/advplyr/audiobookshelf/blob/48f232790a5026ac886564ef57660a338a168187/server/controllers/UserController.js#L228-L230

### What did you expect to happen?

Email to be set for root user.

### Steps to reproduce the issue

1. Log in as root user
2. Try to change the root user

### Audiobookshelf version

2.13.2

### How are you running audiobookshelf?

Built from source

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
No logs generated
```

PR #3367, opened 2024-09-02
Fix: root user cannot be edited

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/3366

Check that the account type is not equal to `root` before returning an error.


nichwall:
Fixed by https://github.com/advplyr/audiobookshelf/commit/60e348fcc188d150b5452cde56ddc941d949daf6 instead

Issue #3365, opened 2024-09-02
[Bug]: Tag Invert No Longer Sticks 


darthrater78:
### What happened?

Tag "invert" no longer sticks for user. 

### What did you expect to happen?

When I click on tag invert it should block users from seeing content tagged as such. 

### Steps to reproduce the issue

1. Set and mark as inverted
2. go back into account and the setting has reverted. 

### Audiobookshelf version

 v2.13.2

### How are you running audiobookshelf?

Debian/PPA

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Chrome

### Logs

```shell
[UserController] update: Invalid permission key: selectedTagsNotAccessible
```

### Additional Notes

Same issue when trying to set inversion as root. Settings tags normally works fine.

PR #3368, opened 2024-09-02
Fix tag permissions

nichwall:
Fixes https://github.com/advplyr/audiobookshelf/issues/3365

This PR updates the default User model to include `selectedTagsNotAccessible` on the default user creation. The `patch` request for `/api/users/{id}` is also updated to compare the payload against the default permissions to allow for values to be updated by clients on existing users which do not have the key in the `permissions` array. I'm not sure if that's the best way to do this, but figured it was more secure to validate against the default properties instead of allowing arbitrary keys to be created by a client.

This does not go through and update existing users which do not have `selectedTagsNotAccessible`.

### DIFF
--- a/server/controllers/UserController.js
+++ b/server/controllers/UserController.js
@@ -270,8 +270,10 @@ class UserController {
       const permissions = {
         ...user.permissions
       }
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks!

Issue #3398, opened 2024-09-10
[Bug]: Cannot Download Metadata content 


HeckerSec:
### What happened?

When attempting to use the automatic matching system in ABS an error message will appear after pressing the "submit" button. 

In The cover section of an audiobook, when searching for a cover, pressing on that cover thumbnail in the window results in the following error message to pop up on the screen "Failed to Download Image from URL: Unknown Error"

When attempting to use the match tab inside the audio book details page, all the appropriate data will display in the window, but upon pressing the "submit" button the error "Failed to update item details" pops up on the screen. 

Finally when attempting to quick match an author, the message "Author not found" pops up on the screen.  

### What did you expect to happen?

It is expected that when using the metadata matching tools that the data displayed in the interface is going to be applied to the book, and that tools like the author quick match will update the author details to contain their bio/photo/etc

### Steps to reproduce the issue

1. Setup Audiobookshelf as normal from the guide on the site
2. Attempt to use any of the aforementioned tools.

### Audiobookshelf version

v2.13.4

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
{"timestamp":"2024-09-09 00:46:28.017","source":"SocketAuthority.js:108","message":"[SocketAuthority] Socket Connected e-Ux55f15-JUi8vzAAAf","levelName":"INFO","level":2}
{"timestamp":"2024-09-09 00:56:52.192","source":"SocketAuthority.js:135","message":"[SocketAuthority] Socket e-Ux55f15-JUi8vzAAAf disconnected from client \"Erik\" after 624226ms (Reason: ping timeout)","levelName":"INFO","level":2}
{"timestamp":"2024-09-09 04:00:34.518","source":"SocketAuthority.js:108","message":"[SocketAuthority] Socket Connected 3sJN0G3Q7wwoJV-YAAAh","levelName":"INFO","level":2}
{"timestamp":"2024-09-09 04:00:51.936","source":"SocketAuthority.js:135","message":"[SocketAuthority] Socket 3sJN0G3Q7wwoJV-YAAAh disconnected from client \"Erik\" after 17531ms (Reason: transport close)","levelName":"INFO","level":2}
[...]

PR #3405, opened 2024-09-11
Log non-strings into log file like console.log does

mikiher:
This was inspired by #3398, in which the reporter attaches the log file, which doesn't contain a detailed logging of AggregateError.

Examining the Logger, I found that when logging to file and emiting log events to listeners, the log object is built like this:
```js
    const logObj = {
      timestamp: this.timestamp,
      source: src,
      message: args.join(' '),
      levelName: this.getLogLevelString(level),
      level
    }
```

However, `Array.join` calls `toString` on each arg, which, in the case of `Error` args, just returns the name and message of the error, without the stack trace or any additional fields. 

`console.log` on the other hand, uses `util.inspect` on non-string arguments, which returns a more detailed object representation, and also has special handling for Error objects.

I fixed this issue by mimicking how `console.log` builds the message.

I also made a couple of other minor changes:
- Moved the logging logic into a single private method `\#log`
- Before my changes, `fatal` and `note` always logged, regardless of `this.logLevel`, but file logging still checked against `this.logLevel`. I fixed the inconsistency (now file logging also disregards `this.logLevel` for `fatal` and `note`). I didn't know if I should also disregard it for log event emitting? currently left that unchanged.

In addition, I added unit tests for Logger.

### DIFF
--- a/server/Logger.js
+++ b/server/Logger.js
@@ -86,7 +86,7 @@ class Logger {
 
     // Emit log to sockets that are listening to log events
     this.socketListeners.forEach((socketListener) => {
-      if (socketListener.level <= level) {
+      if (level >= LogLevel.FATAL || level >= socketListener.level) {
         socketListener.socket.emit('log', logObj)
       }
     })
--- a/server/Logger.js
+++ b/server/Logger.js
@@ -1,5 +1,6 @@
 const date = require('./libs/dateAndTime')
 const { LogLevel } = require('./utils/constants')
+const util = require('util')
 
 class Logger {
   constructor() {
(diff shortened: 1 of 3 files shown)


mikiher:
I also needed to resolve a seemingly unrelated flakiness in the BookFinder unit test. 

Not sure exactly what happened there. I modified the test code slightly and it went away (the many changes are due to prettier formatting).

mikiher:
> We can disregard those for the log event emitting also.

OK, I'll fix the code then.

mikiher:
Done. This is ready for review.

advplyr:
Thanks!

PR #3407, opened 2024-09-12
Default deny explicit content to users

agraubert:
[Documentation](https://www.audiobookshelf.org/guides/users/#creating-a-new-user) specifies that new users are created without access to explicit content by default. It looks like in the recent change to the new User model, this behavior was changed. 

Personally, I think defaulting to not allowing explicit content is a safer 'fail closed' kind of mentality. Admins can manually grant explicit content access directly in ABS or through Oauth grants etc


advplyr:
Works for me, thanks!

Issue #3406, opened 2024-09-11
[Bug]: Socket Disconnected on Scan


GhostTownCowboy:
### What happened?

I've added a library that links to a directory that is all comics (CBZ/CBR). When I scan the comic folder (issue not present in EBook/Audiobook directory) it starts, finds about 4 then gives the "SOCKET DISCONNECTED" popup. The scan seems to continue with notifications about the socket connecting and disconnecting. No other items are found. 

### What did you expect to happen?

The scan should add comics similar to other books unless I'm misunderstanding something. 

### Steps to reproduce the issue

1. Unsure without a Synology Nas and several hundred comic files.

### Audiobookshelf version

2.13.4

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
Crash Log: 
{"timestamp":"2024-09-11 09:44:43.583","source":"Server.js:185","message":"[Server] Unhandled rejection: [object Object], promise: Promise {\n  <rejected> {\n    message: 'Truncated RAR file data',\n    name: 'Error',\n    stack: 'Error: Truncated RAR file data\\n' +\n      '    at ArchiveReader.entries (/server/libs/libarchive/wasm-module.js:110:17)\\n' +\n      '    at entries.next (<anonymous>)\\n' +\n      '    at MessagePort.<anonymous> (/server/libs/libarchive/libarchiveWorker.js:43:20)\\n' +\n      '    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:820:20)\\n' +\n      '    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)'\n  }\n}","levelName":"FATAL","level":5}
[...]

PR #3410, opened 2024-09-12
Handle library scan failure gracefully

mikiher:
This is in partial response to #3406

Library scan is a complex process and can fail in unexpected ways.
This tries wrap it up with a try-catch block, do the sensible things on error, save the scan log, and not crash the server.

There are a couple of questions I still have regarding the code:
- In the following code:
  ```js
      if (canceled && !libraryScan.totalResults) {
        task.setFinished('Scan canceled')
        TaskManager.taskFinished(task)

        const emitData = libraryScan.getScanEmitData
        emitData.results = null
        return
      }
  ```
  what is the last part trying to achieve? it seems it has no effect. Also, the premature return prevents some cleanups from happening.
- In the following code:
  ```js
    if (libraryScan.totalResults) {
      libraryScan.saveLog()
    }
  ```
  Why do we check library.Scan.totalResults? don't we want to save the scan log anyway?

### DIFF
--- a/server/scanner/LibraryScan.js
+++ b/server/scanner/LibraryScan.js
@@ -75,13 +75,14 @@ class LibraryScan {
     return date.format(new Date(), 'YYYY-MM-DD') + '_' + this.id + '.txt'
   }
   get scanResultsString() {
-    if (this.error) return this.error
     const strs = []
     if (this.resultsAdded) strs.push(`${this.resultsAdded} added`)
     if (this.resultsUpdated) strs.push(`${this.resultsUpdated} updated`)
     if (this.resultsMissing) strs.push(`${this.resultsMissing} missing`)
-    if (!strs.length) return `Everything was up to date (${elapsedPretty(this.elapsed / 1000)})`
-    return strs.join(', ') + ` (${elapsedPretty(this.elapsed / 1000)})`
+    const changesDetected = strs.length > 0 ? strs.join(', ') : 'No changes detected'
+    const timeElapsed = `(${elapsedPretty(this.elapsed / 1000)})`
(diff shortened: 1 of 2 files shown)


advplyr:
Way back when scans showed persistent toast and when the scan emit data results were null that was when it would show "canceled" in the toast. It looks like that last part was accidently leftover when I refactored. It doesn't do anything.

If the user were to start a scan and cancel it quickly I don't think we want to update the `lastScan`, `lastScanVersion` and `lastScanMetadataPrecedence`. I don't think it would matter much if we did though.

I'm not sure why I setup the scan log to only save when it has updates. I think we can save it regardless and we can also save it if the scan was canceled.

mikiher:
Thanks for the clarifications, I'll make the necessary changes.

mikiher:
> If the user were to start a scan and cancel it quickly I don't think we want to update the lastScan, lastScanVersion and lastScanMetadataPrecedence. I don't think it would matter much if we did though.

Updating `lastScan` and `lastScanVersion` probably don't matter much, but if we update `lastScanMetadataPrecedence` and this scan had `forceRescan=true` (which happens when metadata precedence changes), then the next scan will have `forceRescan=false` (assuming data precedence hasn't changed) - is that the intended behavior? I'm not sure. What do you think?

It seems to me that for the sake of simplicity, we should only update all those fields after a complete full scan was run - that's what I implemented.

This is ready for review now.

advplyr:
Yeah you're right. We don't want to update `lastScanMetadataPrecedence` on a cancel because when the metadata precedence changes we want the next scan to do a full re-scan.

Thanks!

PR #3411, opened 2024-09-12
Add duration to local sessions on creation

justcallmelarry:
If session is not sent in by the clients it defaults to 0. This uses `libraryItem.duration` when creating the session.

Duration seems like something that shouldn't have to be provided by the clients, and can safely be used from the data already present in the server. Additionally this is how it already works for the non-local sessions.

I could switch to only use it if either `!session.duration` or `session.duration !== null`, but I personally feel like there is no need to use anything other than `libraryItem` as the source of truth for this attribute in particular.


advplyr:
That was setup so long ago I can't remember why I had the client sending duration. It isn't necessary.

There is a lot I'd like to refactor with sessions once the old data model is phased out.

Thanks!

Issue #2541, opened 2024-01-21
[Bug]: Author/Series with non-ASCII names are listed once for every book when updated


thatguy7:
### Describe the issue

When the author's name is in cyrillic letters, it seems like ABS does not reliably process it.

![error](https://github.com/advplyr/audiobookshelf/assets/5512286/6175aeb1-399e-4b46-8831-207677676b22)

As can be seen, e.g. I have 10 books by Анджей Сапковский (Andrzej Sapkowski, red), but instead of listing the author once with 10 books, the author is listed 10 times, once for each book.

I set the author's name using multple-selection and mapping the authors field, just to be sure it's the same for all of them.
When I changed the author Василий Маханенко (Vasily Mahanenko, blue) to the latin-lettered version, all his books were linked to a single author entry.

Interestingly, this bug does not seem to always happen - there clearly are multiple authors with more than 1 book in the screenshot.

The books by Sapkowski and Mahanenko are a single series each, wheres the books by Dostoevsky (green) are not part of a series.

### Steps to reproduce the issue

1. I guess just adding multiple books with the same author name in cyrillic letters.
2. Go to Authors tab.

#### UPDATE:
This error occurs only, when the tag is edited post import in ABS:
Pick any number of books and add the same cyrillic name to the authors tag, it will create a new author for every book. 

### Audiobookshelf version

v2.7.2

### How are you running audiobookshelf?

Docker

nichwall:
Have you verified that every author is identical and there are no differences in the Unicode or hidden characters?

This issue came up a few months ago in Discord, and was resolved once the user fixed the Unicode.

"Might be that ABS handles that fine, but й (U+0439) and й (U+0306 and U+0438) look the same for example"

Beginning of conversation: https://discord.com/channels/942908292873723984/942908295579062335/1176217664918663229

thatguy7:
Thanks for pointing me to the conversation, I will report back once I had a look.

nichwall:
One way you can check is by comparing the `metadata.json` for the library items. That will show you what ABS is using, depending on whether you got author names from metadata tags, the filesystem, or an online match, or a combination of sources

thatguy7:
I checked the metadata.json. They seem to be the same for all of the books.

```
rg "Фёдор Михайлович Достоевский"
9551d32c-3b00-4864-a4b2-f58bf9077496/metadata.json
254:    "Фёдор Михайлович Достоевский"

[...]

PR #3414, opened 2024-09-13
Improved handling of Authors and Series with names containing non-ASCII characters

thatguy7:
This PR fixes: https://github.com/advplyr/audiobookshelf/issues/2541

In summary: SQLite does not support the `LOWER`/`UPPER` function (unless it is compiled to do so, which it usually is not).

When checking if a Series/Author is already in the DB, these functions are used. Any name containing a capital non-ASCII character would always test as false and thus a new entry in the DB was created every time.

This checks is the name contains a non-ASCII character. If it does, the `lower` function is not used and the issue is avoided.

Should we not use the `lower` function at all to avoid this issue?

### DIFF
--- a/server/models/Author.js
+++ b/server/models/Author.js
@@ -53,14 +53,26 @@ class Author extends Model {
    * @returns {Promise<Author>}
    */
   static async getByNameAndLibrary(authorName, libraryId) {
-    return this.findOne({
-      where: [
-        where(fn('lower', col('name')), authorName.toLowerCase()),
-        {
-          libraryId
+    const containsOnlyASCII = /^[\u0000-\u007f]*$/.test(authorName)
+
+    // SQLite does not support lower with non-Unicode chars
+    if (!containsOnlyASCII) {
+      return this.findOne({
+        where: {
+          name: authorName,
+          libraryId: libraryId
         }
-      ]
-    })
+      })
+    } else {
+      return this.findOne({
+        where: [
+          where(fn('lower', col('name')), authorName.toLowerCase()),
+          {
+            libraryId
+          }
+        ]
+      })
+    }
   }
 
   /**
--- a/server/models/Author.js
+++ b/server/models/Author.js
@@ -1,5 +1,6 @@
 const { DataTypes, Model, where, fn, col } = require('sequelize')
 const parseNameString = require('../utils/parsers/parseNameString')
+const { asciiOnlyToLowerCase } = require('../utils/index')
 
 class Author extends Model {
   constructor(values, options) {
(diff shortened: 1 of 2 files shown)


nichwall:
Maybe we can use the `asciionlytolowercase` function? Not sure if it's still in the code base.

https://github.com/advplyr/audiobookshelf/commit/b447cf5c1ccb273fd45af8a7e8b11f3844c28fe6

thatguy7:
The function is still in the code base and solves my issue just as well.

In your comment on my issue I set out to fix, you mentioned https://github.com/advplyr/audiobookshelf/issues/2678.

It seems like the goal is to get SQLite to properly handle non-ASCII characters. If done, `asciiOnlyToLowercase` will again not match what SQLite's `lower` function does. This fix will later become the cause of the issue it fixes now.

Thus again, the question: Should we here perhaps not try to do anything fancy and just match char for char. Then it's up to the user to properly set the value of the tag.

As it is, If I mistype the author to  e.g. "sTephen King", all subsequent books by the author will be tagged with the typo, and there is no way to fix it. If I edit the name to the correct spelling, it will still use the first one it saw.

advplyr:
> As it is, If I mistype the author to e.g. "sTephen King", all subsequent books by the author will be tagged with the typo, and there is no way to fix it. If I edit the name to the correct spelling, it will still use the first one it saw.

It should be fixable by editing the author on the authors page. That issue would be there regardless of if we get proper unicode support.

Before Sqlite it used to also strip punctuation. I believe there are some issues currently open to go even further in trying to match with existing authors by removing middle initials, stripping accents, checking against nicknames, and removing prefixes/suffixes.

I'm not sure what the best solution is but this is the right temporary fix I think

advplyr:
Thanks!

thatguy7:
PR https://github.com/advplyr/audiobookshelf/pull/3468 already breaks this temporary fix.

I checked with the most recent state of master, and the fix now causes the issue it was set out to fix.

@advplyr @nichwall, what is the protocol to undo the changes of this PR?

nichwall:
Probably open a new PR with whatever fix is needed. Having an issue to go with the PR is usually a good idea, but if it's just a PR to fix issues from other PRs you can just open the PR and include details about what was broken.

Are you meaning just get rid of `asciionlytolowercase`?

thatguy7:
> Are you meaning just get rid of asciionlytolowercase?

Exactly. With https://github.com/advplyr/audiobookshelf/pull/3468, the behaviour of SQLite's `lower` function was modified to behave more (or perhaps exactly) like `String.toLowerCase()`. So `asciionlytolowercase` now breaks things.

PR #3416, opened 2024-09-13
Set intitial database version in migrationsMeta to 0.0.0

### LINKED ISSUE #3415: [Bug]: Migration Manager not recognizing migrations from before current version
### What happened?

I am working on getting my first migration written (for the series duplication bug). Upgrading a database from `2.12.3` to `2.13.4` with a test migration occurring in `2.13.2` does not run the migration script. I assume this is because the `migrationMeta` table is being created with the current server version, so the migration manager sees the newly created table as having the same value as the current server version.

This may cause problems with future upgrades, or if people are upgrading from older server versions to a version after a migration occurred.

I am basing the migration script soley off of the README in the migrations folder, so I could be doing something wrong with the version numbers in my testing.

All migration scripts to run between previous and current version when creation of `migrationsMeta` table. When creating the table, the "old version" can be assumed to be `2.13.4` or similar.

Manually editing the versions in the table and then running the server again does correctly run the migration.

1. Use older database
2. Create migration script that targets a version between old database and current server version, without having upgraded to a server version that has the `MigrationManager`
3. Observe migration script does not run

### Audiobookshelf version

v2.13.4-edge

Built from source

Linux

None

### Logs

```shell
[2024-09-13 10:58:04.236] INFO: [MigrationManager] Database is already up to date.
[2024-09-13 10:58:04.346] INFO: [Database] Db initialized with models: user, library, libraryFolder, book, podcast, podcastEpisode, libraryItem, mediaProgress, series, bookSeries, author, bookAuthor, collection, collectionBook, playlist, playlistMediaItem, device, playbackSession, feed, feedEpisode, setting, customMetadataProvider, mediaItemShare
[2024-09-13 10:58:04.351] DEBUG: Set Log Level to DEBUG
[2024-09-13 10:58:04.373] INFO: [Database] Server upgrade detected from 2.12.3 to 2.13.4
```

### Additional Notes

_No response_

advplyr:
fixes #3415

The issue is that the migrations get copied because the initial maxVersion is 0.0.0 but the migrations don't run because the databaseVersion matches the serverVersion.

Setting the databaseVersion to 0.0.0 on create will have all the migrations be run.

I'm not sure if I'm missing anything here so would appreciate if you can take a look @mikiher 

I tested this by going back to a database state before the migrationsMeta and SequelizeMeta.

Create a migration file for v2.13.4 and update the package.json version to v2.13.5.

This now runs the migration file for v2.13.4

### DIFF
--- a/server/managers/MigrationManager.js
+++ b/server/managers/MigrationManager.js
@@ -42,6 +42,7 @@ class MigrationManager {
 
     await this.fetchVersionsFromDatabase()
     if (!this.maxVersion || !this.databaseVersion) throw new Error('Failed to fetch versions from the database.')
+    Logger.debug(`[MigrationManager] Database version: ${this.databaseVersion}, Max version: ${this.maxVersion}, Server version: ${this.serverVersion}`)
 
     if (semver.gt(this.serverVersion, this.maxVersion)) {
       try {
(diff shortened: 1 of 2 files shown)


mikiher:
I acknowledge the bug - sorry about that. 

I think setting databaseVersion to 0.0.0 is fine, but then you also need to modify the code in Database.js to the following:
```js
      if (!this.isNew) {
        await migrationManager.runMigrations()
      } else {
        await migrationManager.updateDatabaseVersion()
      }
```
so that databaseVersion is set correctly for new databases (because migrations are not run for new database).

mikiher:
Actually, it's best if this isNew logic is inside the MigrationManager class, and then I can also write appropriate unit tests.

If you're OK with it, I'll send out another PR to fix this.

PR #3417, opened 2024-09-14
Add: series migration to be unique

nichwall:
This PR resolves https://github.com/advplyr/audiobookshelf/issues/3207 and is a continuation of https://github.com/advplyr/audiobookshelf/pull/3326.

A migration script is added for `2.13.5`, so it will work for both a minor and patch release (not sure what the plans are for the next release). I added some unit tests for this migration, but also tested it manually before writing the unit tests. I also updated the `Series` model to have an index for series name and libraryId, because series can have the same name in different libraries.

I am not sure if I did the QueryInterface parts of the migration script correctly, so if y'all can check it that would be great. Basically, I am using raw SQL instead of Sequelize, so I'm not sure if this is the correct way to do things.

PR #3418, opened 2024-09-14
Fix MigrationManager initial run behavior

### LINKED ISSUE #3415: [Bug]: Migration Manager not recognizing migrations from before current version

mikiher:
This fixes #3415 more comprehensively.

- The initial value for databaseVersion is fixed
- The isNewDatabase logic was pushed inside MigrationManager, with new tests added
- Usage of MigrationManager becomes simpler (construct, init, runMigrations unconditionally)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
   "pkg": {
     "assets": [
       "client/dist/**/*",
-      "node_modules/sqlite3/lib/binding/**/*.node"
+      "node_modules/sqlite3/lib/binding/**/*.node",
+      "server/migrations/*.js"
     ],
     "scripts": [
       "prod.js",
--- a/server/Database.js
+++ b/server/Database.js
@@ -171,9 +171,9 @@ class Database {
     }
 
     try {
-      const migrationManager = new MigrationManager(this.sequelize, global.ConfigPath)
+      const migrationManager = new MigrationManager(this.sequelize, this.isNew, global.ConfigPath)
       await migrationManager.init(packageJson.version)
-      if (!this.isNew) await migrationManager.runMigrations()
+      await migrationManager.runMigrations()
     } catch (error) {
       Logger.error(`[Database] Failed to run migrations`, error)
       throw new Error('Database migration failed')
--- a/server/managers/MigrationManager.js
+++ b/server/managers/MigrationManager.js
@@ -11,11 +11,13 @@ class MigrationManager {
 
   /**
    * @param {import('../Database').sequelize} sequelize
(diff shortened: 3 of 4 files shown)


mikiher:
Please don't review yet. I'm doing some more testing to reproduce the integration test failure that @nichwall had in #3417

mikiher:
OK, I added another fix which addresses the integration test failure in #3417.

- Add migration scripts as pkg assets.

This popped up since this was the first time a migration script was checked in. 

Oddly enough, I tested migration scripts with pkg on Windows, and it had no trouble with those, so it looks like there's some difference in pkg's behavior on Linux and Windows. I didn't think I needed to add the scripts to `pkg.assets` since they were already included in `pkg.scripts`, but obviously I was wrong.

mikiher:
This is now ready for review.

advplyr:
Thanks!

PR #3425, opened 2024-09-15
added postcssOptions to remove npm warning

wommy:
npm run generate had a warning related to the different format that postcss expected

edited nuxt.config.js to fix it


advplyr:
Thanks!

PR #3454, opened 2024-09-25
Add: abs databases from before v2.13.4

nichwall:
This PR adds test databases for each ABS server version to `test/server/testdata`, using the naming scheme of `schema_v[version].sql`. These databases were created by:
- Checking out each release beginning with 2.3.0 and ensuring the correct version of node was used (node 16 prior to v2.8.0, node 20 afterwards)
- Deleting the 'config/` directory to ensure we started from a clean state
- Starting the server
- Running `sqlite3 config/absdatabase.sqlite ".schema" > "schema_v[version].sql"`

This inspiration was taken from comments in the PR about series migrations, specifically https://github.com/advplyr/audiobookshelf/pull/3417#issuecomment-2351115265. A different directory could be used for these databases, but I didn't have a better suggestion than was mentioned there.

We don't *really* need every database beginning with 2.3.0, but am including them all for completeness.

### Follow-up
Would it be useful to have a GH workflow which can be manually triggered (probably after the version bump commit, but before release?) to run the sqlite dump command and open a PR with the new SQL file? This would be merged manually, but then is a button press instead of needing to remember to clean out `config/`, start the server, and move/rename the file.

If so, I will do that in a separate PR.


advplyr:
I'm not understanding what the benefit is of having snapshots of the db for versions that don't have a migration.
Is the unit test going to use these?

nichwall:
The main reason was to make databases available for future unit tests. The old database schemas aren't as important, I was just including them for completeness (include every SQLite database that has been created).

Issue #2228, opened 2023-10-19
[Enhancement]: Add notifications for Audiobook downloads.  Enhance the notify feature.


NF8C:
### Describe the feature/enhancement

The current notification system is great.  I had no trouble setting it up to send notifications through Home Assistant.  Here is the basic URL format if anyone needs it.  There is a ton of stuff that can be done once a notification is sent to Home Assistant.
json://{IP address}:8123/api/webhook/{webhook ID}

Unfortunately I only see the option for podcasts and test notifications.  It would be awesome if we had access to more events.  Particularly new books.  There was some great talk on the podcast notification request that was already fulfilled.
https://github.com/advplyr/audiobookshelf/issues/996

PS: If anyone gets into this I just wanted to let them know there is a small typo "notificataion" they could correct in the body of the test notification. 

Great work everyone!

jcfalcone:
It would be nice if the Notification system could send a URL to the Cover art as well; I do most of my alerts though telegram ( I send to a webhook I created ) and it looks nicer with the cover image.

I more or less figured out how to remake the Cover URL inside my webhook but I could receive it in the same way I get a URL to the podcast

advplyr:
@jcfalcone There is a feature request open for this https://github.com/advplyr/audiobookshelf/issues/1812. The only hesitation I have on this is we would have to open up a URL accessible without authentication.

With this particular request of sending a notification for newly added books the issue is how to handle 100 books added at the same time.

NF8C:
Could you add some more variables to the test notification? Let me play around with it and I'll propose some default templates.  Once I have a nice looking default template we can iterate.  Give me one variable with the basic details of the last 5 or 10 books, when they were added, and a variable with how many books were added in the last scan.  I've been testing and the form is basic jinja/yaml just like what I use in Home Assistant.  I can do a lot with a jinja template.

emigrating:
> With this particular request of sending a notification for newly added books the issue is how to handle 100 books added at the same time.

I'd say, just add a pause/stop/disable notifications for X minutes option. If this isn't toggle isn't hi before starting a large import you'll just have to live with the mass of notifications and you're likely to remember next time around.

I also want more variables accessible, for both current types of notifications. Image URL for sure, but also direct url to whatever's just added. Plan on having a look at the notification system but still haven't got my dev environment setup via the container using my code server :/

jcfalcone:
[...]

Issue #996, opened 2022-09-19
[Enhancement]: Add notifications for Podcast downloads


SAS-1:
### Describe the feature/enhancement

Hi

It would be good to have notifications when ABS downloads a podcast so we know there is a new episode available. 

This could be done by adding an option for people to use something like Apprise (https://github.com/caronc/apprise) I would be happy with a solution like NZBHydra (https://github.com/theotherp/nzbhydra2/) uses that you need to provide an Apprise API to use and it's not included in the app itself just a page to control the notifications.

Thanks

advplyr:
We've discussed this a few times in the discord server. Apprise has been mentioned a few times so I just tried it out.

The way I was thinking of using it before is exactly how nzbhydra implemented it. I haven't run nzbhydra so I'm still curious how they are getting the apprise api info. Any chance you can send me a screenshot of the config page for apprise in nzbhydra?

SAS-1:
No worries 2 shots below in the first one the blanked out bit is the apprise URL in the format of http://IP/notify the second one is the URLs to notify that are configured in Apprise

the third one is the options you can select for notifications

![image](https://user-images.githubusercontent.com/47122073/191178492-c40eb2a5-18d5-4177-bfe4-d8def52437ce.png)
![image](https://user-images.githubusercontent.com/47122073/191178604-4450270f-5b3e-434b-8c3d-ebbcbb005a29.png)
![image](https://user-images.githubusercontent.com/47122073/191178859-cff63a92-615c-4915-9f63-6290053f9c97.png)

advplyr:
There have been other requests to have notifications for when books are added and I'm trying to think this through before building out the notification system.

I'm going to dump my thoughts on notification events here since we don't have another discussion for this.

## 'Library item added' notification
This would include an additional dropdown to select which library you want this notification to apply to. It can be a book or podcast library so for podcasts this would be triggered when you add a new podcast.

The problem with this is sometimes you may scan in a hundred new library items and obviously we don't want to send 100 notifications. We will need to send a notification that can say something like "X books have been added to Y library" and the message body would list each new book title (or "X podcasts have been added.."). 

If we go that route then I don't think we would be able to allow changing the title and body template. It would basically be 2 templates for this notification; one for a single item added and one for multiple items added. That would be complicated to build in the UI and confusing to the user but it may also be confusing as to why they can't edit this template.

## 'Podcast episode downloaded' notification

I think this should also have you select the podcast library this would apply to.

[...]

PR #3464, opened 2024-09-27
Added notification support for new items being added to a library

silvercolt:
Supports the client upload method (one or more items), or folder changes.  I tested each scenario locally through docker and using Pipedream webhooks to receive the notification event.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3225,6 +3225,7 @@
         "type": "string",
         "description": "The name of the event the notification will fire on.",
         "enum": [
+          "onItemsAdded",
           "onPodcastEpisodeDownloaded",
           "onBackupCompleted",
           "onBackupFailed",
--- a/server/scanner/LibraryScanner.js
+++ b/server/scanner/LibraryScanner.js
@@ -16,7 +16,8 @@ const LibraryItemScanData = require('./LibraryItemScanData')
 const Task = require('../objects/Task')
 
 class LibraryScanner {
-  constructor() {
+  constructor(notificationManager) {
+    this.notificationManager = notificationManager
     this.cancelLibraryScan = {}
     /** @type {string[]} - library ids */
     this.librariesScanning = []
@@ -284,6 +285,7 @@ class LibraryScanner {
             'items_added',
             newOldLibraryItems.map((li) => li.toJSONExpanded())
           )
+          this.notificationManager.onItemsAdded(newOldLibraryItems)
           newOldLibraryItems = []
         }
 
@@ -296,6 +298,7 @@ class LibraryScanner {
           'items_added',
           newOldLibraryItems.map((li) => li.toJSONExpanded())
         )
+        this.notificationManager.onItemsAdded(newOldLibraryItems)
       }
     }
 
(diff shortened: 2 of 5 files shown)


docs/openapi.json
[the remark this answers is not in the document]
nichwall: Oh, whoops. I misread the previous edit, disregard.

server/scanner/LibraryScanner.js
```diff
@@ -647,14 +650,15 @@ class LibraryScanner {
...
+        this.notificationManager.onItemsAdded([oldNewLibraryItem])
       }
       itemGroupingResults[itemDir] = newLibraryItem ? ScanResult.ADDED : ScanResult.NOTHING
     }
 
     return itemGroupingResults
   }
 }
-module.exports = new LibraryScanner()
+module.exports = new LibraryScanner(new NotificationManager())
```
advplyr, line 661: The server doesn't run because `NotificationManager` is not defined.
We wouldn't want to instantiate a new instance of NotificationManager though because it is already instantiated in Server.js.
The NotificationManager could be made a singleton like the LibraryScanner so it can be imported anywhere

advplyr: I updated it to a singleton so you can merge master to fix
https://github.com/advplyr/audiobookshelf/commit/7cd8d7f44d999565e6ca2941704158519c78b869

silvercolt: Interesting, I ran the server just fine with the new instantiation of `NotificationManager` here. Maybe I missed something when migrating my changes from a branch off master to my fork. Agreed, a singleton would have been better, but I didn't feel comfortable making a sweeping change like that, so I appreciate the enhancement. I will update accordingly and resubmit the PR.

advplyr:
https://github.com/advplyr/audiobookshelf/issues/2228

The main reason I hesitated on adding this feature is what I described here: https://github.com/advplyr/audiobookshelf/issues/996#issuecomment-1254108583

When testing this I scanned in a new library and it queued up dozens of emails. For some users it could be 10k+ emails. I'm not sure the best way to handle that.

The other issue I have with this is the notification is named `onItemsAdded` when it doesn't handle podcast library items and it is only one library item so should be `onItemAdded`

silvercolt:
> #2228
> 
> The main reason I hesitated on adding this feature is what I described here: [#996 (comment)](https://github.com/advplyr/audiobookshelf/issues/996#issuecomment-1254108583)
> 
> When testing this I scanned in a new library and it queued up dozens of emails. For some users it could be 10k+ emails. I'm not sure the best way to handle that.
> 
> The other issue I have with this is the notification is named `onItemsAdded` when it doesn't handle podcast library items and it is only one library item so should be `onItemAdded`

These are valid concerns, I appreciate referencing the previous conversation for added context. Let me think about this and see if I can propose a potential solution that will help address these. There will be some UI changes needed to give the user some added control over the library, notification thresholds/limits (n > items consolidate into a single notification, restrict notification event to specific library(ies), be more specific about what kind of item: book, podcast, disable option when adding a new library, allow users to un/subscribe to notifications, potentially an option as an Admin to manually trigger a notification via the book's details page, etc.).

I do think this is worth solving as there are many out there looking for this feature like I am - hence the motivation to try and contribute to its solution.  I am currently wanting to send notification embeds to my private Discord server where my users are listening.  

I will see if I can document a full Feature specification for this that User Stories could be groomed from for design, estimation, impact/risk, and prioritization. Maybe others could assist in the development.

PR #3468, opened 2024-09-29
Nunicode integration

### LINKED ISSUE #2678: [Enhancement]: Diacritic-insensitive search

mikiher:
This is my second attempt at integrating an SQLite extension that provides unicode aware case-folding and unaccenting, which fixes #2678 properly.

After numerous attempts and discussions with @devnoname120, I settled on [Nunicode](https://bitbucket.org/alekseyt/nunicode/src/master/) SQLite extension, which, compared to the previous ([SQlean/unicode](https://github.com/nalgeon/sqlean/blob/main/docs/unicode.md)) extension I tried:
- Does not override the SQLite NOCASE collation (which lead to crashes with the previous extension, because indices were not properly re-indexed).
- Provides superior case-folding support which wasn't provided by the previous extension (e.g. `Maße` case-folded to `masse`)
- Supports unicode-aware unaccent(), upper() and lower() functions, and a unicode aware case-insensitive LIKE, like the previous extension
- Like the previous extension, Nunicode doesn't have any dependencies, and generates a very small library (~300Kb)

I set up a separate Github project [nunicode-binaries](https://github.com/mikiher/nunicode-binaries), which builds the extension from source and allows to download the binaries for all the platforms Audiobookshelf server supports (linux-x64, linux-arm64, win-x64, and osx-arm64). The build and release are implemented as Github workflows.

This PR has the following changes:
- BinaryManager now downloads the extension from nunicode-binaries (the download is done directly from the latest release page, not through the API, which has rate limitations).
- BinaryManager now also saves version info for downloaded libraries (and also checks it against the versions defined as valid)
- BinaryManager was also tightened against various failures.
- DockerFile was modified to download and install the extension from nunicode-binaries
- The extension is loaded at Database initialization
- The extension has been defined as _optional_ -  if it is not found or cannot be downloaded for any reason, the server will start without it.
  - This is done in order to deal with any unexpected failures during the initial rollout. After this runs in prod for a while, I'd like to make it required, so we don't have different search behaviours.
- Searches now use the following algorithm (slightly modified from the previous attempt)
  - If the search query does not contains accents, it is matched against an unaccented colum
    - `unaccent(column) LIKE '%query%'`
  - If the search query contains accents, it is matched against the original column 
    - `column LIKE '%qüery%'`
      - this provides a way to match without accent normalization.
  - All searches are unicode-aware case-insensitive, due to the re-implementation of the LIKE operator

The extension integration was tested on Windows, and on linux-x64 and linux-arm64 Docker (and passes our integration test).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,9 @@ ENV NUSQLITE3_PATH="${NUSQLITE3_DIR}/libnusqlite3.so"
 
 RUN case "$TARGETPLATFORM" in \
   "linux/amd64") \
-  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.0/libnusqlite3-linux-x64.zip" ;; \
+  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.1/libnusqlite3-linux-x64.zip" ;; \
   "linux/arm64") \
-  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.0/libnusqlite3-linux-arm64.zip" ;; \
+  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.1/libnusqlite3-linux-arm64.zip" ;; \
   *) echo "Unsupported platform: $TARGETPLATFORM" && exit 1 ;; \
   esac && \
   unzip /tmp/library.zip -d $NUSQLITE3_DIR && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,19 +11,35 @@ FROM node:20-alpine
 ENV NODE_ENV=production
 
 RUN apk update && \
-    apk add --no-cache --update \
-    curl \
-    tzdata \
-    ffmpeg \
-    make \
-    gcompat \
-    python3 \
-    g++ \
-    tini
+  apk add --no-cache --update \
+  curl \
+  tzdata \
+  ffmpeg \
+  make \
+  gcompat \
+  python3 \
+  g++ \
+  tini \
+  unzip
 
 COPY --from=build /client/dist /client/dist
 COPY index.js package* /
 COPY server server
+
+ARG TARGETPLATFORM
+
+ENV NUSQLITE3_DIR="/usr/local/lib/nusqlite3"
+ENV NUSQLITE3_PATH="${NUSQLITE3_DIR}/libnusqlite3.so"
+
+RUN case "$TARGETPLATFORM" in \
+  "linux/amd64") \
+  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.0/libnusqlite3-linux-x64.zip" ;; \
+  "linux/arm64") \
+  curl -L -o /tmp/library.zip "https://github.com/mikiher/nunicode-sqlite/releases/download/v1.0/libnusqlite3-linux-arm64.zip" ;; \
+  *) echo "Unsupported platform: $TARGETPLATFORM" && exit 1 ;; \
+  esac && \
+  unzip /tmp/library.zip -d $NUSQLITE3_DIR && \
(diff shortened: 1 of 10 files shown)


advplyr:
There is a crash on mac.

```bash
[2024-09-30 16:40:29.697] ERROR: [Database] Failed to load extension /audiobookshelf/libnusqlite3.dylib [Error: SQLITE_ERROR: dlopen(/audiobookshelf/libnusqlite3.dylib.dylib, 0x000A): tried: '/audiobookshelf/libnusqlite3.dylib.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/audiobookshelf/libnusqlite3.dylib.dylib' (no such file), '/audiobookshelf/libnus] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
```

It is appending `.dylib` on whatever path is passed in to `loadExtension`. For testing I updated the path to remove `.dylib` before calling `loadExtension` and it gives: 
`mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64h' or 'x86_64'`.

```bash
[2024-09-30 16:44:56.405] ERROR: [Database] Failed to connect to db [Error: SQLITE_ERROR: dlopen(/audiobookshelf/libnusqlite3.dylib, 0x000A): tried: '/audiobookshelf/libnusqlite3.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64h' or 'x86_64')), '/System/Volumes/Preboot/Cryptexes/OS/audiobookshelf/libnusqlite3.dylib' (n] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
```

As a side note before release we will want to catch the loadExtensions exceptions to allow proceeding without the extension. Then once it's stable we make it required like you said.

mikiher:
OK, I believe the osx library issue is now fixed. 
Can you please refresh your code (switched to nunicode-binaries v1.1), and try again?

As I said above, there should be no need now for removing the `.dylib` extension from the `NUSQLITE3_PATH` env variable.

advplyr:
It's working great. This looks good to be in 2.14.0. I'll do some more testing later.
Thanks for figuring that out

mikiher:
> As a side note before release we will want to catch the loadExtensions exceptions to allow proceeding without the extension. Then once it's stable we make it required like you said.

After I fixed the osx issue, I think it is very unlikely that loadExtension will fail (given that it is already found or downloaded successfully, and leading to setting of the `NUSQLITE3_PATH` envireonment variable). I tested on all other platforms, and I would actually like to know if any failures occur at this point in the code rather than ignoring them, so we can quickly address them. 

So if you don't object, I'd like leave the loadExtension code as-is and get bug reports if they happen. But if you insist, we can surround it with a try-catch block.

advplyr:
Yeah that makes sense. I've tested on my ubuntu server, windows and mac running locally

devnoname120:
Great work everyone!

PR #3487, opened 2024-10-06
Move authors to LazyBookshelf

mikiher:
This moves the authors page to use LazyBookshelf. The main gain here is reduction in page load time. I have about 2750 authors in my library, and loading the page used to take 4-5 seconds on a pretty powerful PC. Now it takes less than half a second (it is somewhat hard to measure exactly due to all the hidden updates to the DOM).

The following client changes were made:
- Modifed `AuthorCard ` to be "lazy" (i.e. loaded late by `bookshelfCardsHelpers`)
  - Stopped requiring to provide an author upfront
  - Stopped relying on `author` availability in computed methods 
  - Added `setEntity`, `destroy`, and `setSelectionMode` (required by `bookshelfCardsHelpers` and `LazyBookshelf`)
- Changed `bookshelfCardHelpers.js` to be able to load `AuthorCard` components.
- Modified `BookShelfToolbar`
  - Removed the authors prop (since it is not available to caller anymore)
  - Added number of authors on the left (to be consistent with other bookshelf views)
  - Changed `matchAllAuthors` to fetch all authors from server (since authors are not passed as a prop anymore)
- Modified `LazyBookshelf` to handle author editing, sorting, and loading
  - Calls to fetch authors are now paginated like other entity types
- Changed all `/library/_library/authors` references to `/library/_library/bookshelf/authors`
- Removed `/library/_library/authors/index.vue`

The following server changes were made:
- Added support for pagination and server-side sorting in `getAuthors` (in `LibraryController.js`)


advplyr:
Working great and much faster, thanks!

PR #3488, opened 2024-10-07
Use musl-based libnusqlite3 in Docker

mikiher:
Following @devnoname120's advice, I am switching to a musl-based nunicode sqlite extension on Docker, and doing away with installing and relying on gcompat (which is no longer needed). The new musl-based binaries are obtained from the new v1.2 version of nunicode-binaries.

Other nunicode binaries remain unchanged, but I switched to v1.2 in BinaryManager as well, to keep the versions consistent.

Note: BinaryManager doesn't check the exact linux version and always installs the standard glibc-based version. This means that running if you run on a non-Docker Alpine (probably as a developer), you will have to install gcomapt manually for the program to load. I think this is OK, but let me know if you want me to add a special use-case for Alpine in BinaryManager.

I tested on Docker on both architectures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,6 @@ RUN apk update && \
   tzdata \
   ffmpeg \
   make \
-  gcompat \
   python3 \
   g++ \
   tini \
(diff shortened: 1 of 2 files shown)


advplyr:
I think it's okay. Tested on all my machines. Thanks!

PR #3489, opened 2024-10-07
Add published decade filter option

### LINKED ISSUE #3448: [Enhancement]: "Published > Decade" filter option
### Type of Enhancement

Server Backend

I'm trying to avoid asking for "smart collections," something I have in Plex I use to group my audiobooks like this:
image
Instead, I think I might get by with an additional Filter for "Published > Decade ####" where #### is a decade like 1840 or 1970 or 2020. 

It would--to an extent--replace functionality I'm losing in migrating from Plex.

image
image

v2.13.4

_No response_

glorenzen:
This adds a library filter option for Published Decade and resolves #3448 

It doesn't follow the same format as the images in the issue, but I thought "Published Decade" was more descriptive and user-friendly than just "Published" in the top list with "Decade XXXX" in the sublist.

Everything seems to be working on my end, but there may be something I'm missing.

image

image


advplyr:
Working great, nice and clean, thanks!

The projects formatting options are setup for vscode and should automatically remove semicolons on save. If you're not using vscode the .prettierrc and .editorconfig files might be able to work with your code editor to auto apply our formatting settings.

glorenzen:
I am using VS Code, so that's weird it didn't remove the semicolons automatically. I'll have to check that there isn't something that I misconfigured.

pwinnski:
Thank you, @glorenzen! I totally agree that your solution is better than my proposal. I wasn't sure how "Published Decade" would look with all the other single-word items in that last, but I clearly chose poorly, as it looks great!

pwinnski:
Minor nit, I have a single book for which the published year is `0400`, and while `400` appears as an option in the Published Decade list, selecting that option tells me there are no matches, so apparently it doesn't like the three-digit year.
Still, this is great!

glorenzen:
@pwinnski I just opened a new pull request with a fix for the issue you're having: #3518

PR #3492, opened 2024-10-09
Use object-cover for author images unless AR is really high or low

mikiher:
While working on lazy loading the Author's page, I felt that we're using object-contain on author images too much - while this works nicely for covers, I feel (subjecively) that the author images look unpolished when we use object-contain. So I modified the current decision algorithm for using object-contain to only use it when the AR is very high or low.

Here's my author page before:
Screenshot 2024-10-09 150942

And after:
Screenshot 2024-10-09 151103

Some of the more significant changed are marked on the After screenshot.

I guess it's really subjective, but I find that I like the After much more, even though in some cases (e.g. John Scalzi), it results in a slightly cut image - but I think in those cases the original looks quite bad as well. 

I guess I should also work on allowing users to upload their own author image...

### DIFF
--- a/client/components/covers/AuthorImage.vue
+++ b/client/components/covers/AuthorImage.vue
@@ -65,10 +65,6 @@ export default {
   },
   methods: {
     imageLoaded() {
-      var aspectRatio = 1.25
-      if (this.$refs.wrapper) {
-        aspectRatio = this.$refs.wrapper.clientHeight / this.$refs.wrapper.clientWidth
-      }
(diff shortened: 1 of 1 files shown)


advplyr:
I agree it looks better for my authors also. 
We can probably remove that functionality altogether and use `object-cover`. An image with an aspect ratio < 0.5 or > 2 will probably display worse when it uses `object-contain`.
Maybe after the next release if users like using object-cover

Issue #3495, opened 2024-10-09
[Enhancement]: Add missing strings to weblate for translating


thehijacker:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

To have every possible default English string in Web and phone application translated to desired language.

### Why would this be helpful?

Unification of GUI in one selected language.

### Future Implementation (Screenshot)

Few examples where there are still English strings:

![image](https://github.com/user-attachments/assets/a40339ca-9492-4122-8a0a-4c8827cd00da)

![image](https://github.com/user-attachments/assets/ad070643-907d-4875-9154-f8ba0f8348f7)

![image](https://github.com/user-attachments/assets/2b31ce6a-0eb2-453a-9387-cb73c4b197bb)

![image](https://github.com/user-attachments/assets/21c86ced-3308-4109-aa0c-23505b094a56)

### Audiobookshelf Server Version

2.14.0

### Current Implementation (Screenshot)

See examples above.

thehijacker:
Two more locations I have found that are untranslated:

![image](https://github.com/user-attachments/assets/fdc1da52-6a30-48bf-85da-f9f51e6aaca6)

![image](https://github.com/user-attachments/assets/310efe4f-ede0-48ae-9554-ac1f88c85b7b)

PR #3500, opened 2024-10-11
2.14.0 string localization

nichwall:
This PR mostly fixes https://github.com/advplyr/audiobookshelf/issues/3495. This does not change the notification translations. I took a look at https://github.com/advplyr/audiobookshelf/commit/1dec8ae12289638ebbc3d9c763071f7804b92935 for how to add localization for server generated strings, but not sure how to get that to play nice with https://github.com/advplyr/audiobookshelf/blob/master/server/utils/notifications.js.

I did not translate any of the logs. I don't think the logs need to be localized because their primary purpose is to provide debugging information to contributors, where English is the common language.


advplyr:
I added the strings for notification descriptions. I don't think we need to translate anything more than the description since the default values are changeable by the user and the test data is just test data. The trigger names and variables also shouldn't be updated.
I agree on not localizing the logs.

Thanks!

Issue #3499, opened 2024-10-11
[Bug]: Unknown column NaN when opening "Add" podcast page, crashes server


nichwall:
### What happened?

Using the `edge` image from commit d258b42e0144ca6d63e950d9547d363ad075bc7d (after 2.14.0), I am encountering crashes with a newly created server when navigating to the "add" page in a podcast library as the root user. The crashes occur regardless of whether a podcasts exist in the library or not.

Looks like this may be related to https://github.com/advplyr/audiobookshelf/commit/64b78b5822f5c5410975dfab524c6f3ab09ad5ba moving the `limit` and `page` validation to the middleware.

### What did you expect to happen?

Able to navigate to "add" page. Functionality is correct in 2.14.0.

### Steps to reproduce the issue

1. Create new server
2. Navigate to add page

### Audiobookshelf version

edge, d258b42e0144ca6d63e950d9547d363ad075bc7d

### How are you running audiobookshelf?

Built from source

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
[2024-10-10 20:45:48.821] FATAL: [Server] Unhandled rejection: Error
    at Database.<anonymous> (/home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/dialects/sqlite/query.js:185:27)
    at /home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/dialects/sqlite/query.js:183:50
    at new Promise (<anonymous>)
    at Query.run (/home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/dialects/sqlite/query.js:183:12)
    at /home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/sequelize.js:315:28
    at async SQLiteQueryInterface.select (/home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/dialects/abstract/query-interface.js:407:12)
    at async podcast.findAll (/home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/model.js:1140:21)
    at async Promise.all (index 1)
    at async podcast.findAndCountAll (/home/chell/data/Documents/git/audiobookshelf/node_modules/sequelize/lib/model.js:1322:27)
[...]

PR #3501, opened 2024-10-11
Library middleware positive check

nichwall:
This PR fixes https://github.com/advplyr/audiobookshelf/issues/3499.

If one of the pagination values is undefined, it now defaults to 0 to prevent a crash in Sequelize.


advplyr:
I don't think we want to put the limit and page query string on every endpoint. I fixed this by
a) replacing the bad API call on the podcast add page
b) updating the 3 paginated API endpoints that could hit this by having them fall back to 0.

When we re-work the API we may want to include a "paginated" middleware or use one of those libraries that handles validation for us.

PR #3508, opened 2024-10-13
Fix next/previous chapter behavior on public share player

mikiher:
The `Next/Prev Chapter` player buttons on the new public share page were not functioning correctly.
This fixes this issue (there's no open bug report, I think).


advplyr:
This hasn't been reported yet. Thanks!

Issue #825, opened 2022-07-10
[Bug]: Error HTTP 413 at upload attempt


stemy2:
### Describe the issue

I can't upload anything. No matter the file i use, it results by a "error 413: Request entity too large".

### Steps to reproduce the issue

1. Create a library
2. Try to upload a file

### Audiobookshelf version

2.0.23~ynh1

### How are you running audiobookshelf?

Other

Thovi98:
> 2.0.23~ynh1

It seems you are using Yunohost, right ? It's better to report the issues there because it's an Yunohost issue : https://github.com/Yunohost-Apps/audiobookshelf_ynh

You have to change the nginx config.

Thovi98:
It’s already solved by https://github.com/YunoHost-Apps/audiobookshelf_ynh/pull/19 thanks to @ericgaspar.

dit7ya:
Since this is a nginx setting thing actually, I am posting a link to the issue explanation here for the next person stumbling upon this 

[Nginx: 413 - Request Entity Too Large Error and Solution - nixCraft](https://www.cyberciti.biz/faq/linux-unix-bsd-nginx-413-request-entity-too-large/)

You essentially have to change `client_max_body_size` in your nginx config.

ekasprzak:
How to change it when audiobookshelf is deployed using Docker?
Is nginx (and its conf) part of docker image?

PR #3514, opened 2024-10-14
chore(docs): add client_max_body_size parameter in nginx config

koralowiec:
Added additional parameter in the example nginx configuration that prevents 413 Request Entity Too Large error (https://github.com/advplyr/audiobookshelf/issues/825#issuecomment-1211867159)

Additionally, I changed the indentation of the config


advplyr:
Thanks!

PR #3518, opened 2024-10-14
Fix and simplify filter logic for publishedDecades

glorenzen:
This adds a small change to the publishedDecades filter query to allow for decades before 1000 and simplifies it a bit as well.


glorenzen:
Ah, got it. I see why the updates you made were needed, but now I'm confused how it worked when I tested it.

Oh well. Thanks for the fixes!

pwinnski:
Thank you both!

Issue #3513, opened 2024-10-14
[Enhancement]: Add missing strings for translation


thehijacker:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Now checked the Podcasts menus and noticed some parts that are still in English even if my Slovenian translation is at 100% on Weblate.

![image](https://github.com/user-attachments/assets/48af0fba-dd32-43ae-b565-d875f6655891)

![image](https://github.com/user-attachments/assets/23290e0a-fc69-402d-84ec-e4b9e6ef099e)

![image](https://github.com/user-attachments/assets/6d34902e-5504-466b-9c8d-fc468422d83b)

### Why would this be helpful?

Goal is to have everything in UI translatable.

### Future Implementation (Screenshot)

It would look translated.

### Audiobookshelf Server Version

2.15.0

thehijacker:
Noticed one more:

![image](https://github.com/user-attachments/assets/5f2d4c21-5947-43c6-badf-44ecb8a91182)

nichwall:
Thanks for helping find these. I went through and got as many other strings as I could find and added them to the PR, but probably still haven't gotten everything.

thehijacker:
Localization is a big project. It is completely understandable that there are parts not yet handled.

PR #3520, opened 2024-10-15
Localization updates for 2.15.0

nichwall:
This PR fixes https://github.com/advplyr/audiobookshelf/issues/3513

I went through and updated the rest of the strings in the episode table and row files as well.


nichwall:
Went through and added a lot more strings for localization.

advplyr:
Thanks!

PR #3521, opened 2024-10-15
Fixes and cleanup for subdirectory serving support

mikiher:
This PR contains some cleanup and a few fixes to allow for serving Audiobookshelf with a non-empty ROUTER_BASE_PATH, in both development and production environments.

The following modifications are included:
- Cleanup: removed of `/dev` routing on the client
  - I'm not sure what it was used for - it seems completely unnessesary.
- Cleanup: defined of `routerBasePath` and `serverHostUrl` constants in nuxt.config.js
- Cleanup: removed unnessary if(development) blocks
- Cleanup: removed all explicit server host references (`localhost:3333`) from the code
- Fix: added proxies for `public` and `hls`
  - These are required for session cookies to be passed correctly in development environment
- Fix: add $config.routerBasePath in some places where it was required.
- Fix: fixed socket.io path on both client and server to use the router base path.
- Fix: when router base path is non-empty, added the base path to all requests that don't start with the base path.
- Fix: use `/healthcheck` on the main router rather than the app.

I have tested quite extensively with empty and non-empty ROUTER_BASE_PATH, and with development and production client setups, and it seems to work fine on all setup now.


advplyr:
That's much better, thanks for cleaning that up. Working well in my tests and with base path set on the server side.

advplyr:
Found an issue in dev with the endpoints not `/api/` `/public/` and `/hls/`.

If you logout it will show the init screen because the request to `/status` doesn't get proxied.

I realize this is why did the `/dev` proxy with the axios middleware modifying the url. There is probably a better solution though

mikiher:
ok, I'll take a look.

PR #3526, opened 2024-10-16
prettier: client

wommy:
i have spent too much time undoing git blocks from autosave prettier 
so i just ran prettier on the whole client


wommy:
btw i think,

``` diff
.prettierrc
---
{
  "semi": false,
  "singleQuote": true,
- "printWidth": 400,
+ "printWidth": 120,
  "proseWrap": "never",
- "trailingComma": "none",
+ "trailingComma": "es5",
  "overrides": [
    {
      "files": ["*.html"],
      "options": {
        "singleQuote": false,
        "wrapAttributes": false,
        "sortAttributes": false
      }
    }
  ]
}
```

would be nice

it sucks when youre rearrange props and the commas always jumping

also some amount of print width just helps scanning larger functions,
some of these vue.methods have like 6 destructured arguments


i also would like `arrowParens: "avoid"` but im not pushy 😉

Issue #2316, opened 2023-11-16
[Bug]: [FfmpegHelpers] Extract Cover Error Error


dmuiX:
### Describe the issue

Scanning new m4a files tagged with mp3tag throws:
[FfmpegHelpers] Extract Cover Error Error: ffmpeg exited with code 1: Error while decoding stream #0:2: Invalid data found when processing input Cannot determine format of input stream 0:2 after EOF Error marking filters as finished Conversion failed!

### Steps to reproduce the issue

1. put a file in the audiobookfolder and let audiobookshelf scan new files

### Audiobookshelf version

v.2.5.0

### How are you running audiobookshelf?

Docker

advplyr:
Can you share the full ffprobe output?

dmuiX:
Where can i find it? Or do you mean the whole log that I can find under logs?

advplyr:
![image](https://github.com/advplyr/audiobookshelf/assets/67830747/218571ae-8a45-4a99-be17-5c59c71f6eea)


Then you will see a "Probe Audio File" button

dmuiX:
{
  "programs": [],
  "streams": [
    {
      "index": 0,
      "codec_name": "aac",
      "codec_long_name": "AAC (Advanced Audio Coding)",
      "profile": "HE-AAC",
      "codec_type": "audio",
      "codec_tag_string": "mp4a",
      "codec_tag": "0x6134706d",
      "sample_fmt": "fltp",
      "sample_rate": "44100",
      "channels": 2,
      "channel_layout": "stereo",
      "bits_per_sample": 0,
      "initial_padding": 0,
      "id": "0x1",
      "r_frame_rate": "0/0",
      "avg_frame_rate": "0/0",
      "time_base": "1/22050",
      "start_pts": 1056,
      "start_time": "0.047891",
      "duration_ts": 26463232,
      "duration": "1200.146576",
      "bit_rate": "82570",
      "nb_frames": "25843",
      "extradata_size": 5,
      "disposition": {
        "default": 1,
        "dub": 0,
        "original": 0,
        "comment": 0,
        "lyrics": 0,
        "karaoke": 0,
        "forced": 0,
        "hearing_impaired": 0,
        "visual_impaired": 0,
        "clean_effects": 0,
        "attached_pic": 0,
        "timed_thumbnails": 0,
        "captions": 0,
        "descriptions": 0,
        "metadata": 0,
        "dependent": 0,
        "still_image": 0
      },
[...]

PR #3529, opened 2024-10-17
Fix "Extract Cover Error" for files with multiple embedded covers

asoluter:
I've encountered the same issue as in https://github.com/advplyr/audiobookshelf/issues/2316 for files with multiple embedded covers (ex. Front and Back covers for disks)

```
2024-10-17 00:02:33.221455-07:00[2024-10-17 00:02:33.221] ERROR: [FfmpegHelpers] Extract Cover Error Error: ffmpeg exited with code 234: frame=    1 fps=0.0 q=11.2 Lq=6.8 size=N/A time=00:00:00.00 bitrate=N/A speed=   0x    
2024-10-17 00:02:33.221494-07:00Conversion failed!
```

The issue comes from the input video stream selection which is not limited to a single video stream for ffmpeg, leading to the following error:
```
[image2 @ 000001c697da0440] The specified filename 'out.jpg' does not contain an image sequence pattern or a pattern is invalid.
[image2 @ 000001c697da0440] Use a pattern such as %03d for an image sequence or use the -update option (with -frames:v 1 if needed) to write a single image.
[image2 @ 000001c697da0440] Cannot write more than one file with the same name. Are you missing the -update option or a sequence pattern?
```

The proposed change limits the input selection to a single video stream, fixing the error.  (Good explanation here https://stackoverflow.com/a/63928441)

### DIFF
--- a/server/utils/ffmpegHelpers.js
+++ b/server/utils/ffmpegHelpers.js
@@ -55,7 +55,7 @@ async function extractCoverArt(filepath, outputpath) {
   return new Promise((resolve) => {
     /** @type {import('../libs/fluentFfmpeg/index').FfmpegCommand} */
     var ffmpeg = Ffmpeg(filepath)
-    ffmpeg.addOption(['-map 0:v', '-frames:v 1'])
+    ffmpeg.addOption(['-map 0:v:0', '-frames:v 1'])
     ffmpeg.output(outputpath)
 
     ffmpeg.on('start', (cmd) => {


asoluter:
There's an alternative option to fix this using `%03d` as part of output file path, which will copy all the embedded images in a enumerated format, but I didn't dig deep enough into the following processing, as it'll need to support multiple outputs from the cover extraction.

advplyr:
In testing this I found out that this is not an issue with ffmpeg v5.1 which is what Abs is currently using. That's why there aren't more reports of this error and why I couldn't reproduce this error before.
I tested with ffmpeg v7 and was able to reproduce it.
This solution works for both versions.

Are you running an old version of Abs or using another install method with SkipBinaries env flag passed in?

advplyr:
Thanks!

asoluter:
I've got this using the latest version of TrueNAS 24.04 App, which just uses v2.15.0 Docker image from what I could gather.

asoluter:
@advplyr Just checked the image contents and it looks like it uses `alpine/ffmpeg 6.1.1-r8` and not v5.1

advplyr:
Ah yes, thanks. I forgot we switched back to that.

PR #3530, opened 2024-10-17
Add server proxies for all server paths

mikiher:
Your recent [Add in /dev proxy for development](https://github.com/advplyr/audiobookshelf/commit/9d1f51c6ba0201f42e72f56a3b25e539a19a8c66) is incomplete since it doesn't take care of all url references. Since the `/dev' prefix is added only for axios calls, it doesn't cover references to the server that are not through axios calls.

For this to work properly without having to modify individual references in the client code, we need to proxy all routes that need to go to the server. 

Unfortunately there's no top-level namespace (like `/server`) that identifies server paths from client paths, so we just need to specify all server paths that are referenced by the client individually.

This is what's implemented in this PR. 
- `/login`, `/logout`, `/status`, and `/init` now work properly. 
- Also added proxies for `/feed/` and `/auth/` which probably also did not work properly before.
- There are no references to `/ping` or `healthcheck` so I left them out of the list
- Made the proxy definition more compact

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -2,6 +2,8 @@ const pkg = require('./package.json')
 
 const routerBasePath = process.env.ROUTER_BASE_PATH || ''
 const serverHostUrl = process.env.NODE_ENV === 'production' ? '' : 'http://localhost:3333'
+const serverPaths = ['api/', 'public/', 'hls/', 'auth/', 'feed/', 'status', 'login', 'logout', 'init']
+const proxy = Object.fromEntries(serverPaths.map((path) => [`${routerBasePath}/${path}`, { target: process.env.NODE_ENV !== 'production' ? serverHostUrl : '/' }]))
 
 module.exports = {
   // Disable server-side rendering: https://go.nuxtjs.dev/ssr-mode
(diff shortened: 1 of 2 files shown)


advplyr:
I believe all the other URL references were taken care of by the other proxys listed

mikiher:
ok, something didn't work for me after patching your commit. I'll see if I
can reproduce it.

advplyr:
It's probably better to handle everything in nuxt.config.js. I'm pretty sure though all cases were handled

mikiher:
Ah, I see the issue - did you actually check this with the debug client (`npm run dev`) on port 3000? 

It can't work since the pathRewrite rule is incorrect: it's currently tries to match `'^/dev/'` but this doesn't match since the path now begins with `/audiobookshef/dev`, not with `/dev`. There's also a superfluous / there.

The rule has to be either: 
```js
pathRewrite: { '/dev': '' }
```
or, if you want to be safer,
```js
pathRewrite: { [`^${routerBasePath}/dev`]: `${routerBasePath}` }
```

When you fix the pathRewrite rule, it does seem like the other proxy rules I added are not necessary.

Howver, I still like my version better. IMO The `/dev` trick is hacky and less readable. 

Let me know which way you'd like to go.

PR #3533, opened 2024-10-18
Add a REINDEX NOCASE v2.15.1 migration and update v2.15.0 migration

mikiher:
This is a migration aimed at fixing potential hidden old database index corruptiuons due to the faulty introduction of the unicode extension in v2.12.0.  It perform a reindexing of all NOCASE indices in the database.

Migration v.2.15.0 is also updated to include the same fix, since we've seen that migration fail due these corruptions.

### DIFF
--- a/server/migrations/changelog.md
+++ b/server/migrations/changelog.md
@@ -2,6 +2,7 @@
 
 Please add a record of every database migration that you create to this file. This will help us keep track of changes to the database schema over time.
 
-| Server Version | Migration Script Name        | Description                                       |
-| -------------- | ---------------------------- | ------------------------------------------------- |
-| v2.15.0        | v2.15.0-series-column-unique | Series must have unique names in the same library |
+| Server Version | Migration Script Name        | Description                                                                          |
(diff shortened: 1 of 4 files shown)


advplyr:
Since we've identified a few indexes that should be added I wonder if we should try to put them in a migration or just put them in to sequelize like https://github.com/advplyr/audiobookshelf/pull/3528

From my experience I have seen that sequelize will create the index if it doesn't exist but I'm not sure how full-proof that is. What do you think?

I'm leaning towards putting it in a migration to be safe but might be unnecessary.

mikiher:
I would prefer a migration. 

It's true that additional indices usually do not change the functionality of the database, just its performance characteristics, but it is still considered a good practice to have a migration - his keeps the database schema consistent with code versions, which is part of the point of using migrations.

For example, you know that when server version 2.17.0 is running, function X should be fast because it uses an index added in version 2.16.0 migration, but when the server version is 2.15.0, you expect X to be slow, because either the index was not yet added then, or because it was reverted due to a downward migration.

advplyr:
My thoughts as well.
Tested this migration and looks good. Thanks!

PR #3546, opened 2024-10-22
API PATCH /me/progress/:id - allow providing createdAt and respect provided finishedAt when syncing progress

justcallmelarry:
I was trying to make a script to set some of my old read books to done, but to avoid screwing up the stats i wanted to set them on a previous year.

I noticed some issues however, and during my research I found that there are some issues when creating and updating `mediaProgress`
* `createdAt` (or `startedAt`, as the api documentation states) was not respected at all at either during creation or update.
* `finishedAt` was not respected during creation, but can be updated after the initial call.

For the first issue, this allows `createdAt` to be set on _creation_ but I could not figure out how to let it be updatable. My suspicion is that there is some magic going on which disallows the field `createdAt` to be updated due to [sequelize default timestamps on their models](https://sequelize.org/docs/v6/core-concepts/model-basics/#timestamps).

For the second issue I just found where the `finishedAt` timestamp was overwritten on creation, and fixed the issue.


advplyr:
Thanks!

PR #3553, opened 2024-10-28
Add proper error handing for file downloads

mikiher:
Fixes #3344

This PR adds proper error handling for file the following file download APIs:
- `/api/items/:id/download`
- `/api/items/:id/file/:fileid/download`
- `/api/items/:id/ebook/:fileid`

Issue #3560, opened 2024-10-29
[Bug]: Time left is disproportionate


JFPCreations:
### What happened?

Since last update, when I start a podcast and switch or stop it, the time left is very disproportionate. 

I tried and it seems ok with version 2.15.1 but after that it goes bad.

### What did you expect to happen?

Showing the right time left like before.

### Steps to reproduce the issue

1. Start a podcast.
2. Listen to it for some times
3. Switch to another one, or just stop and leave.
4. Come back to see time left.
![brave_4OxDajewXy](https://github.com/user-attachments/assets/1b7d43b9-bfa7-4c26-9179-32dd0d38815c)

### Audiobookshelf version

2.16.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Other (list in "Additional Notes" box)

### Additional Notes

Brave browser version 1.71.118

advplyr:
These will get fixed on the next sync

PR #3565, opened 2024-10-29
Fix incorrect call to handleDownloadError

mikiher:
Sorry, this is to fix the crash reported [here](https://github.com/advplyr/audiobookshelf/pull/3553#issuecomment-2445031131).


advplyr:
I can push a patch now. I also had to patch https://github.com/advplyr/audiobookshelf/issues/3560

Are you still reviewing this one?

mikiher:
Sorry, forgot to push the Ready for Review button

PR #3568, opened 2024-10-30
Update: user directive in docker compose file

nichwall:
Updates the incorrect use of environment variable for running as a non-root user.


advplyr:
Thanks!

devnoname120:
Related: https://github.com/AJGranowski/docker-user-mirror

PR #3574, opened 2024-10-31
'mpg' and 'mpeg' added as supported audio-type/file-extension

4ch1m:
Hi there,

please add `mpeg` and `mpg` to the supported audio-types/file-extensions.

I have tons of `*.mp(e)g` podcast files in my library; which - at the moment - aren't detected by the scanner at all.
This change will fix this.

Thanks a lot!


advplyr:
I tested some mpeg 1/2 files and they play but have to be transcoded for chromium browsers. I didn't look into firefox.
I think it is always best to use file formats that can be direct played but we can support this because they do transcode fine.

Issue #3444, opened 2024-09-23
[Enhancement]: Keyboard navigation


pwinnski:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Many elements in ABS cannot be handled with keyboard alone. (If this is not the case, I'd love to learn otherwise!)

For example, if a book has been matched, many of the elements on the Details screen (Authors, Series, Genres, Tags, and Narrators) are "pills," and cannot be edited or deleted with the keyboard alone. 

### Why would this be helpful?

If a book is matched incorrectly, it currently takes a lot of clicking to remove the pills before typing in correct information.

While editing books after adding them, keyboard shortcuts for things like "Save" or "Re-Scan" would speed things up considerably.

Generally it would be really nice to be able to move around the interface without having to take hands off the keyboard.

### Future Implementation (Screenshot)

No visible difference, but tabbing into a field with a pill should select the pill.

### Audiobookshelf Server Version

v2.13.4

### Current Implementation (Screenshot)

![image](https://github.com/user-attachments/assets/d7ed7f0f-59af-4fee-946c-25e4c0ea5db0)

PR #3575, opened 2024-10-31
Multi select keyboard navigation

glorenzen:
This adds some updates to the MultiSelectQueryInput component to allow for better navigation with a keyboard. I plan to add similar updates to the MultiSelectDropdown as well.

Related issue: #3444


glorenzen:
Here is a short clip of the updated MultiSelectQueryInput and the MultiSelect components with keyboard inputs.
I realized the MultiSelectDropdown wasn't used in the edit item modal, so I didn't add the keyboard input to it.
https://github.com/user-attachments/assets/faaeb7b8-03b8-4357-a0bb-e9cdebeafdc0

glorenzen:
I also added a state variable called "inputFocused", but I realized the naming might be confusing since there is also an "isFocused" boolean. Maybe "iconFocused" would be better?

advplyr:
This is great, thanks! I've been working on similar updates recently with accessibility. Now when you open modals they automatically get focus.

PR #3585, opened 2024-11-04
Avoid parsing first and last names in Chinese, Japanese and Korean la…

snakehnb:
There's no need to parse first and last names in Chinese, Japanese and Korean languages. For example, parsing names from "张三, 李四" or "张三 & 李四" should return ["张三", "李四"], not ["李四 张三"].


advplyr:
Thanks!

Issue #3588, opened 2024-11-07
[Bug]: Migration can crash if `migrationsMeta` table exists but is empty


nichwall:
### What happened?

A user in Discord reported a crash when upgrading from server version `2.12.3` to `2.16.2`. Upon request, they tried upgrading to a few different versions, including `2.13.3` and `2.14.0`. They were able to upgrade to all versions before `2.14.0`, but the migration failed and crashed for all servers version at `2.14.0` and above.

The user provided their database to me and I was able to reproduce the crash. Upon investigation, it looks like they somehow got an empty `migrationsMeta` table. I am not sure if this is a result of the failed upgrade to `2.16.2` or something else.

### What did you expect to happen?

Migration to be handled when the table is empty. It appears to be a problem with how `isDatabaseNew` is handled. Instead of only checking if the table exists, we should also check that the `version` row exists.

https://github.com/advplyr/audiobookshelf/blob/0c244cbf952efd8e0b63a1c73d93ea4584bc7ff4/server/managers/MigrationManager.js#L190-L217

### Steps to reproduce the issue

1. Create `migrationsMeta` table in an existing database with no rows
2. Attempt to upgrade to a version greater than or equal to `2.14.0`

### Audiobookshelf version

v2.16.2

### How are you running audiobookshelf?

Built from source

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
[2024-11-06 21:53:19.790] ERROR: [Database] Failed to run migrations TypeError: Cannot read properties of undefined (reading 'version')
    at MigrationManager.fetchVersionsFromDatabase (/home/chell/data/Documents/git/audiobookshelf/server/managers/MigrationManager.js:178:14)
    at async MigrationManager.init (/home/chell/data/Documents/git/audiobookshelf/server/managers/MigrationManager.js:46:5)
    at async Database.init (/home/chell/data/Documents/git/audiobookshelf/server/Database.js:183:7)
    at async Server.init (/home/chell/data/Documents/git/audiobookshelf/server/Server.js:128:5)
    at async Server.start (/home/chell/data/Documents/git/audiobookshelf/server/Server.js:191:5)
[2024-11-06 21:53:19.793] FATAL: [Server] Unhandled rejection: Error: Database migration failed
    at Database.init (/home/chell/data/Documents/git/audiobookshelf/server/Database.js:187:13)
[...]

PR #3589, opened 2024-11-07
Check that `migrationsMeta` table is well formed instead of just existing

nichwall:
If the rows do not exist, the migration manager crashes.

This fixes https://github.com/advplyr/audiobookshelf/issues/3588 by checking that both `version` and `maxVersion` are defined in the `migrationsMeta`. This PR drops the table and triggers the creation of the `migrationsMeta` table again if either row is missing.


advplyr:
Thanks!

Issue #3525, opened 2024-10-15
[Bug]: Performance of querying for Newest Authors on Homepage > 2mins for big E-Book Library


friral:
### What happened?

The Newest Author load time on the Homepage of is still (v2.15) very high for my 40K E-Book Library. Smaller libraries (1-2K audiobooks) are fine. The other similar 6 queries (see log) on the same page take only very few seconds or ms.
Tested on my desktop Firefox & the Android app with same results

### What did you expect to happen?

1. Load as fast as the other queries. For some reason the "Recently Added" query runs much quicker than "Newest Authors" although there are twice as many books/items than authors in the library
2. For now, quickfix to make the library usable again: Maybe make the query for the newest authors disable-able in the settings

### Steps to reproduce the issue

1. Load 40k books into a book-library
2. Navigate to Homepage of that library
3. Watch the App freeze for a couple of minutes
4. Check logs for the actual time

### Audiobookshelf version

v 2.15.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
2024-10-14 08:49:38.557
DEBUG
Loaded 8 of 8 items for "Continue Listening/Reading" in 0.72s

2024-10-14 08:49:43.263
DEBUG
Loaded 0 of 0 items for "Continue Series" in 4.71s

2024-10-14 08:49:43.672
DEBUG
Loaded 10 of 41478 items for "Recently Added" in 0.41s

2024-10-14 08:49:43.720
DEBUG
Loaded 5 of 2254 series for "Recent Series" in 0.05s

2024-10-14 08:49:47.160
DEBUG
Loaded 10 of 34816 items for "Discover" in 3.44s

2024-10-14 08:49:47.625
DEBUG
Loaded 4 of 4 items for "Listen/Read Again" in 0.47s

2024-10-14 08:52:07.970
DEBUG
Loaded 10 of 18309 authors for "Newest Authors" in 140.34s

2024-10-14 08:52:07.974
DEBUG
Loaded 6 personalized shelves in 150.13s
```

nichwall:
Same as https://github.com/advplyr/audiobookshelf/issues/3237, but for authors instead of series. It is the same sort of query causing the slowdown. The backend data model likely needs to be changed to fix this.

[...]

PR #3594, opened 2024-11-08
Increase cache time for `filterdata` in library

nichwall:
This PR adds some additional logic to the generation of `filterData` to check if any changes have occurred to the database since the data was cached to reduce the cold-start time of requesting library information. The current behavior is to cache the filter data for 30 minutes, and then clear the cache for `filterData` regardless of database state. Ideally, we will need to update the data model to generate `filterData` more efficiently, but this should account for most "normal" usage of browsing/using the server without needing large changes to the data model.

This issue is more pronounced with large libraries, such as mentioned in https://github.com/advplyr/audiobookshelf/issues/3525#issuecomment-2442787507, but is still present in smaller libraries (a cold start on my ~500 book test server takes about ~2.3 seconds to generate).

I experimented with using individual queries, such as `unique` queries on the `narrators` column instead of having a loop in JS, and while each individual query was slightly faster than the loop, having 7 individual queries was much slower than just loading all of the data and iterating over it in javascript. The `tags` and `genres` were also slower when iterating over individual JSON elements in the SQL query, but it is possible to do directly using a query.

In the end, I determined that during normal usage, this `filterdata` doesn't really change. The main reason there is a 30 minute timeout is so that when metadata changes, the filter list in the library view is updated in a reasonable amount of time.
image

To simplify this, I added a check to get any rows in the relevant tables which have updated since the cache was created. If no results are returned, then the cache creation time is updated to validate the cache for an additional 30 minutes, and if any of the tables had a more recent `updatedAt`, then we regenerate the `filterData` as was already done.

## Further work before it's ready:

This does *not* handle deleted items due to not having a row to check the `updatedAt` column.

The solutions I am investigating:
- Add a migration for these tables to add the [`paranoid` attribute ](https://sequelize.org/docs/v6/core-concepts/paranoid/), then clean up any soft-deleted rows when the filterdata is checked. This will likely complicate other logic due to the soft-deleted rows still showing up in raw queries.
- Add a migration to the `library` table to add `entityLastUpdated` columns or similar to show the last time something was updated/deleted. This is a bit more brittle and requires updating another table whenever a change is made.
- Adding more fields to the `filterData` array to store the count of books, authors, series, and podcasts for the library, and then checking this value to ensure no items were deleted (most promising, local change only)

### DIFF
--- a/server/utils/queries/libraryFilters.js
+++ b/server/utils/queries/libraryFilters.js
@@ -462,7 +462,42 @@ module.exports = {
       numIssues: 0
     }
 
+    const lastLoadedAt = cachedFilterData ? cachedFilterData.loadedAt : 0
+
     if (mediaType === 'podcast') {
+      // To reduce the cold-start load time, first check if any podcasts
+      // have an "updatedAt" timestamp since the last time the filter
+      // data was loaded. If so, we can skip loading all of the data.
+      // Because many items could change, just check the count of items instead
+      // of actually loading the data twice
+      const changedPodcasts = await Database.podcastModel.count({
+        include: {
+          model: Database.libraryItemModel,
+          attributes: [],
+          where: {
+            libraryId: libraryId,
+            updatedAt: {
+              [Sequelize.Op.gt]: new Date(lastLoadedAt)
+            }
+          }
+        },
+        where: {
+          updatedAt: {
+            [Sequelize.Op.gt]: new Date(lastLoadedAt)
+          }
+        },
+        limit: 1
+      })
+
+      if (changedPodcasts === 0) {
+        // If nothing has changed, update the cache to current time for 30 minute
+        // cache time and return the cached data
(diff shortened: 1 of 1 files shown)


nichwall:
I ended up going with my 3rd idea for handling the case of deleted items, where I just store how many items were returned by the database query to see if the number has changed.

This has worked well in my testing to reduce the number of times the `filterData` needs to be constructed.

advplyr:
This is a nice solution. 

Right now the library filter data is getting reset anytime a library item is updated (in `Database.updateLibraryItem`). When the old library item object gets phased out that logic will get improved to keep the cache up-to-date.

The library cache is staying up-to-date for some functionality. For example, editing/removing genres & tags on the metadata utils config pages keeps the library cache up-to-date.

It is mainly library item updates that aren't keeping the cache up-to-date due to the complexity of the old library item object still being used.

Thanks!

PR #3595, opened 2024-11-08
Only run CodeQL and Integration actions if code changed

nichwall:
This PR updates the triggers for the CodeQL and Integration Test actions to not include changes to the README or other unrelated files. For example, d7e810fc2f873e05eedb8ec41919656e21496412 took 4 minutes to complete actions when there was only a Readme change. The CodeQL workflow will still run every week regardless of changes.

This is a very minor thing and probably not worrying about it but something I noticed when going through action logs.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,11 +1,25 @@
-name: "CodeQL"
+name: 'CodeQL'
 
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks!

Issue #2324, opened 2023-11-19
[Bug]: Data type mismatch in database


lkiesow:
### Describe the issue

I've noticed that in the SQLite database `libraryItems.mediaId`, `books.id` and `podcasts.id` actually have different column types. Given that ABS constantly joins on these tables, they should probably match:

```
sqlite> PRAGMA table_info(books);
  cid|name|type|notnull|dflt_value|pk
  0|id|UUID|0||1
sqlite> PRAGMA table_info(podcasts);                                                                                    
  cid|name|type|notnull|dflt_value|pk
  0|id|UUID|0||1
sqlite> PRAGMA table_info(libraryItems);                                                                                
  cid|name|type|notnull|dflt_value|pk
  …
  4|mediaId|UUIDV4|0||0
```

Not sure if that's a problem right now, but this could easily become a problem at any time.

The relevant code is:

https://github.com/advplyr/audiobookshelf/blob/e07d17c472f9a799c08e28d17861e27588991a80/server/models/LibraryItem.js#L771

https://github.com/advplyr/audiobookshelf/blob/e07d17c472f9a799c08e28d17861e27588991a80/server/models/Book.js#L246-L250

https://github.com/advplyr/audiobookshelf/blob/e07d17c472f9a799c08e28d17861e27588991a80/server/models/Podcast.js#L140-L144

But, this might also need additional data migration.

### Audiobookshelf version

> 2.5.0 (d2aea869)

PR #3597, opened 2024-11-09
`MediaId` UUID migration

nichwall:
This PR fixes https://github.com/advplyr/audiobookshelf/issues/2324.

This adds a migration to convert the `mediaId` column in the `libraryItems` table to be a UUID instead of a UUIDv4. After performing the migration on my test server, everything still seemed to be linked correctly between `mediaProgress`, `libraryItem`, `books`, and `podcasts`, where deleting a book did cascade correctly. I was also able to scan my library and saw new rows created correctly in each table.

The only performance difference I saw was in the noise (less than 3 ms difference on any given query).

I am a little hesitant about this change if anything goes wrong during the migration due to SQLite copying the table, creating a new table, then moving data over. From my reading, this can trigger delete cascades, but I did not experience that in my testing. Additional reading at https://sequelize.org/docs/v6/other-topics/query-interface/#changing-and-removing-columns-in-sqlite and https://stackoverflow.com/a/70486686


advplyr:
There are 5 other models that I added into the migration still using UUIDv4.
I tested this thoroughly and didn't come across any issues.
Updated it for v2.17.0.

Thanks!

PR #3600, opened 2024-11-10
set library id lowercase make more compatible

unl0ck:
iOS swift will send with openapi generated client lib the UUID to string not lowercased and so it is not possible the find the correct library

### DIFF
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -1314,12 +1314,12 @@ class LibraryController {
    * @param {NextFunction} next
    */
   async middleware(req, res, next) {
-    if (!req.user.checkCanAccessLibrary(req.params.id)) {
+    if (!req.user.checkCanAccessLibrary(req.params.id.toLowerCase())) {
       Logger.warn(`[LibraryController] Library ${req.params.id} not accessible to user ${req.user.username}`)
       return res.sendStatus(403)
     }
 
-    const library = await Database.libraryModel.findByIdWithFolders(req.params.id)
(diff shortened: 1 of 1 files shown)


nichwall:
Huh, I didn't realize that parsers are required to accept uppercase UUID, but software is required to generate UUID lowercase. Looks like it's a thing of Apple not confirming to the spec, shortest summary here) https://stackoverflow.com/a/48333432/3422368

Anyway, I think we want to modify the Sequelize models to just handle the lowercasing of the UUID field automatically instead of manually changing everywhere a UUID is used, since this will affect every model, not just libraries.

unl0ck:
So I should change it everywhere ?

unl0ck:
@nichwall soll ich die Models alle anpassen ? Beziehungsweise darf ich die anpassen ?

nichwall:
I'm not sure. I haven't had time to look into it more. Some of my reading made it sound like Postgres uses uppercase UUID, but they use a 128-bit binary representation instead of the string representation for SQLite. We are not currently supporting other database backends, but that is planned eventually and why we are using Sequelize.

If you can do some more research to show that using `lowercase` everywhere is better than just updating the model to do it automatically, that would be good.

nichwall:
To clarify, I think changing all of the middleware to convert to lowercase is fine, but we want to make sure this won't cause a different problem somewhere.

It may not be simple to change the model to handle the lowercase logic, and there might be other server logic which expects it to be a lowercase UUID, in which case we do want to convert it to lowercase like you've done here.

PR #3607, opened 2024-11-12
Add LibraryItemController

unl0ck:
Added Missed LibraryItemController
Added media to LibraryItemBase

### DIFF
--- a/docs/objects/LibraryItem.yaml
+++ b/docs/objects/LibraryItem.yaml
@@ -55,6 +55,12 @@ components:
           type: boolean
         mediaType:
           $ref: './mediaTypes/media.yaml#/components/schemas/mediaType'
+        media:
+          type: object
+          description: The media of the library item.
+          oneOf:
+            - $ref: './mediaTypes/Book.yaml#/components/schemas/bookBase'
+            - $ref: './mediaTypes/Podcast.yaml#/components/schemas/Podcast'
     libraryItemMinified:
       type: object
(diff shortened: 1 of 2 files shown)


nichwall:
I haven't sat down and gone through it.

There is also https://github.com/advplyr/audiobookshelf/pull/3198 which adds a lot of new things to the OpenAPI spec for the current API.

I have been working on defining a new API in https://github.com/advplyr/audiobookshelf/pull/3388 off and on, which fixes issues we found with defining the current API as an OpenAPI spec due to generators having issues with the many different levels of `oneOf`, `anyOf`, and `allOf` that Vito found in their PR.

unl0ck:
@nichwall  so is it possible to release a current state of #3198  ? also maybe it is better to create an own repository for the openapi ?

nichwall:
> @nichwall  so is it possible to release a current state of #3198  ? also maybe it is better to create an own repository for the openapi ?

It will be a few days before I can look at it again. If you want to look at it and help clean up any issues with the PR to simplify the review process that could be good. I did not finish going through everything since there were issues that Vito mentioned.

Eventually the OpenAPI docs will replace https://github.com/audiobookshelf/audiobookshelf-api-docs for hosting as a website, but the OpenAPI spec will remain in this repo once it's fully defined so the spec can remain up to date with the server.

All of this has been in flux for multiple months since there are multiple moving parts and dev time is limited.

PR #3614, opened 2024-11-15
fix memory spike and high cpu usage when getting playlists with podcasts

drwggm:
This might be related to https://github.com/advplyr/audiobookshelf/pull/3015 

- I have a large number of podcasts, and some podcasts with a large number of episodes.  
- The current way podcasts are queried in playlists is very slow. 
- As far as I can tell, after looking up the `libraryItemIds` it grabs all episodes for any `podcastId`in the playlist.
- It only then filters the podcasts episodes contained in the playlist causing
-  Having so many podcasts episodes loaded everytime something is done with a playlist creates big memory spikes and cpu usage.

I do not know if this works for audiobooks, because I do not have a library to test it on, but it does seem to work for podcasts, so even if it breaks something hopefully it will give you some ideas on how to fix.

PS. I don't code JS, so apologies for messing up your codebase.

### DIFF
--- a/server/models/LibraryItem.js
+++ b/server/models/LibraryItem.js
@@ -10,6 +10,7 @@ const { filePathToPOSIX, getFileTimestampsWithIno } = require('../utils/fileUtil
 const LibraryFile = require('../objects/files/LibraryFile')
 const Book = require('./Book')
 const Podcast = require('./Podcast')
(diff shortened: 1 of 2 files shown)


advplyr:
As of v2.18.0 playlists no longer pull all episodes for podcasts of playlist items

PR #3604, opened 2024-11-12
Remove unnecessary episode_download_queue_updated socket event causing OOM

mikiher:
This fixes #3601.

The first commit removes an unnecessary socket event with a large payload that was emitted every time an episode download was queued. If a large number of episodes `n` were queued, this caused the creation of `O(n^2)` large objects to be created in quick sucession and queued for sending through the web socket, which caused an OOM for n=~2000.

The second commit resolves an unrelated bug in the podcast item page, in which if you pressed on the button to clear the download queue for a library item, the button and the message `{0} Episode(s) queued for download` were not removed after the download queue was emptied.

PR #3615, opened 2024-11-16
Use a simpler database fetch in fullUpdateFromOld

mikiher:
This is followup to #3604.

As discussed privately, the database fetch in `LibraryItem.fullUpdateFromOld` causes a memory spike because of the query complexity. This uses a simpler query to achieve the same result.

### DIFF
--- a/server/models/LibraryItem.js
+++ b/server/models/LibraryItem.js
@@ -237,35 +237,7 @@ class LibraryItem extends Model {
    * @returns {Promise<boolean>} true if updates were made
    */
   static async fullUpdateFromOld(oldLibraryItem) {
-    const libraryItemExpanded = await this.findByPk(oldLibraryItem.id, {
-      include: [
-        {
-          model: this.sequelize.models.book,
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

server/models/LibraryItem.js
```diff
@@ -237,35 +237,7 @@ class LibraryItem extends Model {
...
-            {
-              model: this.sequelize.models.author,
-              through: {
-                attributes: []
-              }
-            },
-            {
-              model: this.sequelize.models.series,
-              through: {
-                attributes: ['id', 'sequence']
```
advplyr, line 254: Turns out we were using the `bookSeries.id` to handle updating series sequence. https://github.com/advplyr/audiobookshelf/issues/3636
I looked for this specifically and didn't see it. Just an fyi

mikiher: Good catch. I didn't notice the difference.

Issue #3623, opened 2024-11-18
[Bug]: `filterData` crashes on new server


nichwall:
### What happened?

When initializing a new server and creating a library, the `filterData` can crash. This is due to the `bookCount`, `authorCount`, and `seriesCount` not being defined under `Database.libraryFilterData[libraryId]`.

This was introduced by https://github.com/advplyr/audiobookshelf/pull/3594

### What did you expect to happen?

No crash

### Steps to reproduce the issue

1. Create new server
2. Create library

### Audiobookshelf version

2.17.0

### How are you running audiobookshelf?

Built from source

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
[2024-11-17 22:14:35.150] FATAL: [Server] Unhandled rejection: TypeError: Cannot read properties of undefined (reading 'bookCount')
    at Object.getFilterData (/server/utils/queries/libraryFilters.js:616:77)
    at async ApiRouter.findOne (/server/controllers/LibraryController.js:198:26) 
promise: Promise {
  <rejected> TypeError: Cannot read properties of undefined (reading 'bookCount')
      at Object.getFilterData (/server/utils/queries/libraryFilters.js:616:77)
      at async ApiRouter.findOne (/server/controllers/LibraryController.js:198:26)
}
```

nmatton:
> ```shell
> [2024-11-17 22:14:35.150] FATAL: [Server] Unhandled rejection: TypeError: Cannot read properties of undefined (reading 'bookCount')
>     at Object.getFilterData (/server/utils/queries/libraryFilters.js:616:77)
>     at async ApiRouter.findOne (/server/controllers/LibraryController.js:198:26) 
> promise: Promise {
>   <rejected> TypeError: Cannot read properties of undefined (reading 'bookCount')
>       at Object.getFilterData (/server/utils/queries/libraryFilters.js:616:77)
>       at async ApiRouter.findOne (/server/controllers/LibraryController.js:198:26)
> }
> ```

This is exactly well described and reproduced on my side too

luparas95:
I have the same problem

```shell
2024-11-18 13:43:23.192

ERROR

[...]

PR #3624, opened 2024-11-18
Add: check fields exist in `filterData`

nichwall:
This PR fixes https://github.com/advplyr/audiobookshelf/issues/3623

I do not know if this is the best way to fix this crash (where `filterData` is missing fields), but I just added a check before comparing against the cached data. This fixes the crash for a new server.

### DIFF
--- a/server/utils/queries/libraryFilters.js
+++ b/server/utils/queries/libraryFilters.js
@@ -507,13 +507,15 @@ module.exports = {
       })
 
       if (changedPodcasts === 0) {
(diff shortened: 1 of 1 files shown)


advplyr:
We can just add a `?` there. Not to say it is a better way, just simpler. https://github.com/advplyr/audiobookshelf/commit/778256ca165fb1248cdb5463146ac4e0561f2c82

nichwall:
You are correct, I forgot you can do that in JS. Thanks!

Issue #2690, opened 2024-02-29
iOS upload picker


sevenlayercookie:
### Describe the issue

When attempting to upload an audiobook from an iOS device, the iOS file picker does not allow selection of .m4b files.

.mp3s are selectable.

I have not tested if .m4as are selectable or not.

### Steps to reproduce the issue

1. Using iOS web browser, click upload in Audiobookshelf Web app.
2. Tap “choose file”
3. Browse for an .m4b file
4. It is greyed out and not selectable.

### Audiobookshelf version

2.8.0

### How are you running audiobookshelf?

Docker

PR #3632, opened 2024-11-21
on iOS, do not restrict file types for upload

sevenlayercookie:
## Brief summary

1. On iOS, in the upload media screen of ABS, .m4b's are not selectable when using file extension or MIME-type accept filters for the file input HTML element. By disabling these filters on iOS-based browsers, m4b's are now selectable.
2. In addition, iOS webkit does not support selecting folders for upload. Thus, on iOS, UI related to folders should be hidden on iOS browsers.

## Which issue is fixed? 
#2690 

## In-depth Description

On iOS, in the upload media screen of ABS, .m4b's are not selectable when using file extension or MIME-type accept filters for the file input HTML element. This holds true even when .m4b and audio/* are added to the accept filter. From what I can tell, iOS Webkit browser cannot determine MIME-type of .m4b files (related issue to #3310 ?), and iOS webkit file input only supports MIME-type filtering, **not** file extension filtering ([caniuse.com](https://caniuse.com/?search=file%20input)). By disabling accept filters on iOS-based browsers, m4b's are now selectable.

1. This PR includes a check for an iOS-based User Agent. If found, it disables file type check for upload. Otherwise, it filters file types per usual.
2. second commit hides UI elements related to folder selection on iOS browsers. Adds a dictionary string to split "files" from "folders"

## How have you tested this?

I have tested on iOS browsers (Chrome, Safari), as well as on my MacBook browsers (Chrome, Safari).
## Screenshots
**Before / non-iOS browsers:**

**After on iOS browsers:**

## Possible issues

This does allow iOS users to select any file type for upload, which will need to be handled by ABS. However it seems that ABS handles this already by ignoring incompatible files that are selected.

## Room for improvement

There may be a different/more standard way of determining user agent (such as method used in #3319 ).
There may be a way to keep filters on iOS browsers, but I was unable to find a simple way to do this.


advplyr:
We should try to avoid concatenating translation strings since the concatenation may not make sense in other languages.
Other than that I just updated the UI/UX so that those inputs wrap on mobile.
Thanks!

PR #3655, opened 2024-11-26
Fix player settings modal on share page

glorenzen:
## Brief summary

I noticed that the player settings modal did not work when playing a book from the share page so I moved the player settings modal state and component to the PlayerUi component to fix this.

## Screenshots

image


advplyr:
The jump forward/backward times wasn't functional so I updated that.
Thanks!

PR #3664, opened 2024-11-28
Add migration to fix dropped foreign key constraints dropped in v2.17.0 migration

advplyr:
## Brief summary

The migration in v2.17.0 caused the modified tables to lose their foreign key constraints. This is a [known bug](https://github.com/sequelize/sequelize/issues/12741) in sequelize.

## Which issue is fixed?

This resolves server crashes when deleting records in the affected table that have an association. 

## In-depth Description

This migration does the following for each table affected (follows the outlined method in [sqlite docs](https://www.sqlite.org/lang_altertable.html#otheralter)):
1. Creates a new table with a temp name with the correct fk constraints
2. Populates the new table with the original tables records
3. Drops the old table
4. Renames the old table to the original

## How have you tested this?

Testing in progress.

Will most likely add an additional check to see if this migration is necessary for users who have downgraded and are upgrading again.

### DIFF
--- a/server/migrations/changelog.md
+++ b/server/migrations/changelog.md
@@ -2,9 +2,10 @@
 
 Please add a record of every database migration that you create to this file. This will help us keep track of changes to the database schema over time.
 
-| Server Version | Migration Script Name        | Description                                                                          |
-| -------------- | ---------------------------- | ------------------------------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


advplyr:
I added an additional check for if the fk constraints need to be updated for each table. If not, the table is unmodified.

This can happen if a user is migrating from a server version < v2.17.0 OR if the user never had to run the v2.17.0 migration (they first installed v2.17.0 - v2.17.2)

PR #3665, opened 2024-11-29
Subdirectory support for OIDC and SocketIO

mikiher:
## Brief summary

This is another PR in preparation for switching to subdirectory support in accordance with #3535 

This covers two issues:
1. SocketIO support for both new and existing clients
2. Support for new and existing OIDC authentication setups

## Which issue is fixed?

There are no existing bugs for this PR. 

There should be no visible effect on existing (no-subdirectory) setups.

### SocketIO support
After the planned switch to subdirectory support, we will have both existing (no subdirectory) clients and new clients (either using a subdirectory or not). Since our SocketIO lives outside of the normal express middleware handlers, we need to support both subdirectory and no-subdirectory paths to access it. 

This is achieved by bringing up two identical SocketIO servers that only differ in their paths. One uses the standard `/socket.io` path, and the other uses `${subdirectory}/socket.io` (e.g. `/audiobookshelf/socket.io`). This way, legacy clients can still connect to the no-directory SocketIO server.

### OIDC Auhentication support
OIDC authentication setups need to register a couple of authorized callback URLs in the OAuth provider. After the provider performs the authentication, it checks the callback URL it got from the server against the authorized callback URLs, and only calls back the server if it found a match.

So, in order for ABS servers to work after switching to subdirectory support, they need to remember which URLs were registered with the provider:
- For existing OIDC setups, we use no-subdirectory URLs.
- For new OIDC setups, we let the user choose between no-subdirectory and subdirectory URLs

This way, exising OIDC setups will continue to work without any manual setup required after the planned switch to subdirectory support.

To achieve this, we introduce a new server setting, `authOpenIDSubfolderForRedirectURLs`. 
For existing OIDC setups it would be set to None (empty string). For new setups after the planned switch, it could be set to either None, or `${subdirectory}` (e.g. `/audiobookshelf`).  `Authentication.vue` was modified to enable setting it.

## How have you tested this?

This was extensively tested using existing (no-subdirectory) and new (subdirectory) clients.

The server migration to add the new OIDC subfolder server setting was unit-tested as well as run on a test database with and without an existing OIDC setup.


advplyr:
I tested this with authentik OIDC and it is working well for both cases.

Do you think we should comment out the setting in `authentication.vue` until this is ready? I'm guessing it won't be long after release for someone to ask about that setting.

mikiher:
Yes, that's possible, although I had two purposes in mind with this setting:
1. Setting the value (where currently only `None` is available)
2. Making it explicit in the settings that the user needs to authorize the URLs in their IDP settings (this currently only appears in the last sentence of the documentation) I'm sure there are some users who want this feature but don't RTFM.

advplyr:
It was helpful having the redirect URLs in the settings there.
Working great, thanks!

advplyr:
Any idea about this https://github.com/advplyr/audiobookshelf/issues/3823#issuecomment-2585431882?

mikiher:
I'll check this. I don't recall ever seeing this happen.

PR #3666, opened 2024-12-01
Feat/download via share link

### LINKED ISSUE #3606: [Enhancement]: Download via share link
### Type of Enhancement

Web Interface/Frontend

When you share a book with someone via the share button I would love the ability for the user of the link to download the book. 

This will be useful when you want to share the audio file with someone without needing to add the file to some other file sharing service.

for example 

1. click the burger menu
2. select share
3. toggle downloadable radio button
4. get link

then from the link user's side 

1. open link 
2. click download button 
3. starts downloading the book

screenshot-2024-11-12-16-31-42

v2.16.2

screenshot-2024-11-12-16-20-23

glorenzen:
This feature adds an option when sharing a media item to allow users to download the zip file.

This resolves #3606 

This is still a work in progress, but I've added the toggle to the share modal, updated the MediaItemShare model to include a field for whether or not the item is downloadable, added the migration file for the updated column, and updated the ShareController.

image


glorenzen:
@advplyr I think this should be ready to go, but let me know if you have any thoughts or think anything should be changed. I added to the download method in the LibraryItemController to handle share items, but there might be a better way to handle it that I didn't think of.

Also, I'm not sure if the migration script name needs to be changed since the server version was updated after I started this PR.

glorenzen:
Here's a screenshot of the download button on the share page (top left corner): 
image

server/controllers/LibraryItemController.js
```diff
@@ -157,33 +157,49 @@ class LibraryItemController {
    * @param {Response} res
    */
   async download(req, res) {
```
advplyr, line 159: Using this endpoint isn't going to work because all of the endpoints in `/api/` are authenticated. The reason it would have worked in your testing is because you're still authenticated via the session. You can test it by using the share link in incognito.

The download functionality should be in the `ShareController` which gets served from the unauthenticated `PublicRouter`. Most of those functions are authenticating using the share_session_id cookie.

server/migrations/v2.17.3-share-add-isdownloadable.js
```diff
@@ -0,0 +1,32 @@
...
+ * @property {MigrationContext} context - an object containing the migration context.
+ */
+
+/**
+ * This migration script adds the isDownloadable column to the mediaItemShares table.
+ *
+ * @param {MigrationOptions} options - an object containing the migration context.
+ * @returns {Promise<void>} - A promise that resolves when the migration is complete.
+ */
+module.exports = {
```
advplyr, line 20: Migrations should keep the same format of using individual functions `up` and `down` for consistency. The JSDoc types also aren't going to work when they are above the export like that.
If you're not using an editor that can use JSDocs I highly recommend it.

The other migrations have a log for BEGIN/END on up and down which we should keep consistent

server/migrations/v2.17.3-share-add-isdownloadable.js
```diff
@@ -0,0 +1,32 @@
...
+  up: async ({ context: { queryInterface, logger } }) => {
+    await queryInterface.addColumn('mediaItemShares', 'isDownloadable', {
```
advplyr, line 22: Since the `mediaItemShares` table was added later someone may be upgrading from a version that doesn't have the table yet which will cause this to crash. See `v2.17.3-fk-constraints` migration on wrapping this in a `tableExists` check.

Also, we want these migrations to be idempotent. If this migration were to run and the `isDownloadable` column already exists, the migration should still run.

advplyr:
> Also, I'm not sure if the migration script name needs to be changed since the server version was updated after I started this PR.

Yeah if you can merge master and update the name of the migration file to v2.17.5

advplyr:
There is also a changelog.md in the migrations folder that should have a short description of each migration

server/controllers/LibraryItemController.js
glorenzen: Ah, got it. I didn't test in an incognito and didn't realize all endpoints in `/api/` were authenticated, but that makes sense! I'll move the download functionality to the ShareController.

glorenzen:
@advplyr I think all the updates should be complete now, but let me know if I missed something or need to change anything.

server/controllers/ShareController.js
```diff
@@ -210,6 +211,75 @@ class ShareController {
...
+    const mediaItemShare = await ShareManager.findBySlug(slug)
+    // If the file isDownloadable, download the file
+    if (mediaItemShare.isDownloadable) {
+      // Get mediaItemId and type
+      const { mediaItemId, mediaItemType } = mediaItemShare
+
+      // Get the library item from the mediaItemId
+      const libraryItem = await Database.libraryItemModel.findOne({
+        where: {
+          mediaId: mediaItemId
```
advplyr, line 236: This isn't going to work when podcast episodes get supported because `mediaId` would be the podcast id and not the podcast episode id.
This is why I distinguish between `mediaId` and `mediaItemId`.

This works for books though since `mediaId` and `mediaItemId` is the same.

I'll update this part

advplyr:
I also removed the `require` from the migration file because we should be able to keep those functions self-contained.

Thanks!

Issue #3649, opened 2024-11-25
[Bug]: Deleted Genres & Tags re-appear after re-scan of the library


b3nd1k:
### What happened?

Through the WebUI, I navigated to Settings -> Item Metadata Utils. Under the section _Manage Tags_ and _Manage Genres_ I deleted a large amount of unnecessary tags, as well as merging tags and genres that have similar names. For example, Fiction Books, Fiction, Fictional was merged into Fiction.

### What did you expect to happen?

I expected the changes and removal of genres and tags I made to remain after a re-scan of the library. I have the "Store metadata with item" option checked under Settings, but I assume what happens is that the changes I made are not written to the metadata file, so that when the re-scan appears it adds the information back. 

### Steps to reproduce the issue

To reproduce the issue, go to Settings -> Item Metadata Utils and either remove or rename any tags. After re-scanning the library, you will notice that the removed or renamed tags reappear.

### Audiobookshelf version

v.2.17.2

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

advplyr:
I'm not able to reproduce this issue. I can see that the tag gets removed from the `metadata.json` file.

Maybe the tags are getting pulled from another file during the scan. Do you have an OPF file or an NFO file with that book?

b3nd1k:
> Do you have an OPF file or an NFO file with that book?

You're probably on to something. The path of the my eBook library in ABS is pointed to my Calibre eBook library. Under _Scanner_, I have given the _Audiobookshelf metadata file_ a higher priority than _OPF file_. The thought was that ABS would pick up the metadata from the Calibre library on the initial scan, but then ignore changes to the OPF-file on additional scans. 

Am I correct to assume that when I delete a tag from _Item Metadata Utils_, the _Audiobook metadata field_ for tags goes blank and re-import the tags from the _OPF file_, rather than locking that field to be blank? Since I have Audiobook metadata field_ set higher than the OPF file, the expected behaviour would be that ABS prevents the tags from getting re-added.

<img width="799" alt="Screenshot 2024-12-02 at 08 19 31" src="https://github.com/user-attachments/assets/e0885af5-289d-42cd-9042-98fe0787c3ce">

b3nd1k:
[...]

PR #3671, opened 2024-12-02
Fixing 'Narrators' and 'Series' update on re-scan

warnwar:
## Brief summary

When re-scanning an existing audiobook, I noticed that the 'Narrators' and 'Series' fields will not update if you change the audio tags from empty to something else

## Which issue is fixed?

I did not create an issue for this, instead I figured I would fix it myself

## In-depth Description

Steps to repro:

- Add an audiobook to your library. Leave the 'composer' and 'grouping' audio tags empty.
- Check that audiobookshelf did not somehow pick up data for 'Narrators' and 'Series' fields
- Outside of audiobookshelf, fill in the empty tags
- Inside audiobookshelf, click the 'Re-Scan' button
- See that the two fields are not updated

## How have you tested this?

I followed my repro steps listed above, debugging through the code. I found the problem was the AbsMetadataFileScanner.js was overwriting the new 'Narrator' and 'Series' information with empty data.

### DIFF
--- a/server/scanner/AbsMetadataFileScanner.js
+++ b/server/scanner/AbsMetadataFileScanner.js
@@ -40,6 +40,8 @@ class AbsMetadataFileScanner {
         if (key === 'genres' && !abMetadata.genres?.length) continue
         if (key === 'tags' && !abMetadata.tags?.length) continue
         if (key === 'chapters' && !abMetadata.chapters?.length) continue
+        if (key === 'narrators' && !abMetadata.narrators?.length) continue
+        if (key === 'series' && !abMetadata.series?.length) continue
 
         bookMetadata[key] = abMetadata[key]
       }


advplyr:
This is an issue we have to decide how to handle.

It may be the case that you have some incorrect data in the compose field, so in Abs you manually remove it.
If the `Audiobookshelf metadata file` is set to the highest priority then it is fair to assume that future scans would keep the narrators empty instead of continue to fill it with the bad data in the compose field.
image

I never figured out how best to handle this.
Oddly enough this scenario in the opposite direction just came up: https://github.com/advplyr/audiobookshelf/issues/3649#issuecomment-2510783811

In that case they don't want the bad tags from the OPF source.

Right now the code is inconsistent so it should be adjusted to be consistent. But what is the best way?

warnwar:
@advplyr I think rewriting the code to better handle metadata preference is a little beyond my abilities. Would it make sense to merge this (for consistency) and deal with that at a later time?

advplyr:
I'm not suggesting that you write the code. It is a question of how best to handle the problem I outlined. We try to implement solutions that best serve the userbase as a whole instead of pushing changes for specific use-cases.

I think we go the other direction because it is more intuitive to have the metadata file take precedence, even for empty values. We can also consider adding an additional scan option (maybe only for a "Re-Scan" in the modal to start) to ignore the metadata file for that scan. Alternatively, we add an additional library setting in the metadata precedence tab that specifies how to handle empty values.

Since you want the audio file metadata to take precedence over the metadata file you can adjust the order of precedence in library settings.

CLHatch:
> This is an issue we have to decide how to handle.
> 
> Consider the scenario where you have some incorrect data in the compose field, so after the scan you manually remove the narrator in Abs.
> 
> If the `Audiobookshelf metadata file` is set to the highest priority then it is fair to assume that future scans would keep the narrators empty instead of continue to fill it with the bad data in the compose field.
> 
> image
> 
> I never figured out how best to handle this. Oddly enough this scenario in the opposite direction just came up: [#3649 (comment)](https://github.com/advplyr/audiobookshelf/issues/3649#issuecomment-2510783811)
> 
> In that case they don't want the bad tags from the OPF source.
> 
> Right now the code is inconsistent so it should be adjusted to be consistent. But what is the best way?

It seems to me that the order of precedence a user would want would depend on the source of the file ("source" being ripped from Audible, made themselves from a CD and metadata embedded by them, some download site where they like the metadata embedded, another site where they don't like the metadata embedded...).

Given that, wouldn't it make sense to have separate precedence settings that apply to a folder rather than one for the entire library?  For example, I have a library where I add I add a folder for my Audible books, another folder my wife's Audible books, and a few other folders listed as "Other".  I'll generally trust the metadata added by Libation for my Audible books, but may not for the "Other" ones.  I want them all in the same library, but I don't want them all to have the same metadata preference settings.

advplyr:
I'm not sure how many people have that use-case (I don't personally) but changing the metadata precedence to be folder specific isn't a solution to the issues I mentioned. If the metadata precedence did become folder specific the issues mentioned would still need to be addressed

CLHatch:
> I'm not sure how many people have that use-case (I don't personally) but changing the metadata precedence to be folder specific isn't a solution to the issues I mentioned. If the metadata precedence did become folder specific the issues mentioned would still need to be addressed

Ah, true, I suppose it wouldn't fix this.

PR #3689, opened 2024-12-07
Resolved a server crash when a playback session lacked media metadata

Vito0912:
## Brief Summary

This PR introduces optional operators to allow stats to function without media metadata for playback Sessions. This change prevents server crashes when media metadata is absent.

## In-depth Description

The logic for handling authors, narrators, genres has been updated to handle cases where `mediaMetadata` is absent. This ensures that the server can generate responses without requiring metadata, thereby preventing crashes.

A follow-up PR should be made to pull the `mediaMetadata` when missing from request.

## How Have You Tested This?

Tested on a personal server.

### DIFF
--- a/server/utils/queries/userStats.js
+++ b/server/utils/queries/userStats.js
@@ -127,20 +127,20 @@ module.exports = {
           bookListeningMap[ls.displayTitle] += listeningSessionListeningTime
         }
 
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #3690, opened 2024-12-07
Improved Metadata Handling and PlaybackSession Metadata Robustness

Vito0912:
## Brief summary

This appends the server's existing `mediaMetadata` to the `playbackSession` if the client does not supply it.

## Which issue is fixed?

This resolves server crashes caused by missing `mediaMetadata` (even though they are addressed separately) and corrects inaccurate stats when the client does not provide `mediaMetadata`. Additionally, it slightly reduces bandwidth usage during local session syncing.

## In-depth Description

If `mediaMetadata` is absent in the payload sent by the client, the server will use the metadata it already has (`libraryitem.media.metadata`). 
Only essential attributes are included in the `playbackSession` to avoid unnecessary data, such as descriptions, which are often lengthy and typically unneeded. 
Additionally, `displayTitle` and `displayAuthor` will fall back to the corresponding `mediaMetadata` values when not provided, ensuring they are displayed even if the client does not sends them.

## How have you tested this?

This has been tested with both my own client and the web client.

----
### Copilot summary
This pull request includes changes to the `PlaybackSessionManager` class to ensure that client-provided metadata is preferred over library metadata when available. This change aims to make the system more robust and user-friendly by prioritizing user-specific information.

### Improvements to metadata handling:

* [`server/managers/PlaybackSessionManager.js`](diffhunk://#diff-1c399b59a75a97b2f162fed8967242975e2f87b54d0730701e4b74bbf966dcdeR178-R189): Added logic to prefer client-provided metadata (such as title, subtitle, narrators, authors, series, genres, and author) over library metadata if available. This ensures that important metadata is synchronized and displayed correctly.


nichwall:
This makes sense to me, along with the fix for the crash in https://github.com/advplyr/audiobookshelf/pull/3689.

Taking a brief look at the database, we probably want to remove the `coverPath`, `mediaMetadata`, `displayTitle`, and `displayAuthor` from the table entirely and just fetch that information from the relevant `podcastEpisodes` or `books` table row when requesting playback sessions. I have opened https://github.com/advplyr/audiobookshelf/issues/3691 to track that separately. (this may have been mentioned before but I am forgetting if so)

Vito0912:
Is there anything that needs to be changed? I remember advplr saying that there needs to be something done for podcasts. I tested some, and it seemed to work for them (although I only had one sample podcast).

Vito0912:
@advplyr, can you check this PR? The stats images are also generated from the session metadata. Plappa seems to only provide the title, for example, so it does not include full metadata too.
(At least from what I saw from a chat with someone having wrong stats while using plappa and their sessions do not include any more mediaMetadata)

advplyr:
I updated this to populate any metadata fields that are undefined. I didn't want to go too far with validating and all of that because I'd like to update local syncing to not include metadata at all.
Thanks!

PR #3692, opened 2024-12-07
Remove serverAddress from Feeds and FeedEpisodes URLs

### LINKED ISSUE #3563: [Bug]: RSS Feeds created using a different server address are inaccessible
### What happened?

RSS feeds created using a different server address (but inside the same database) always refer to the address they were created with. 

This causes them to be inaccessible when the server address changes. 

There are other implications on the web client related to this, like broken images in RSS Feeds config page:
Screenshot 2024-10-29 093936

RSS feeds should dynamically refer to the server under which they are currently running (or more precisely, to the host of feed request)

1. Create an RSS feed in the web client using e.g. localhost:3333
2. Bring up the server using a different port, or access it using a reverse proxy using some external domain
3. Access the updated feed URL (with just the server address changed to the new one)
4. Feed contains links pointing to the old server address.

### Audiobookshelf version

v2.16.1

Built from source

Windows

None

### Logs

_No response_

### Additional Notes

_No response_

mikiher:
## Brief summary

This fixes #3563, and also makes the necessary preparation for subdirectory support for RSS feeds.

## Which issue is fixed?

Fixes #3563

## In-depth Description

This removes the protocol and host from all URLs in the Feeds and FeedEpisodes tables (including a migration that changes existing records). The host and protocol are now dynamically determined at runtime based on the request.

The web client is also modified to handle the shortened URLs.

This achieves 2 goals:
1. It fixes the issues described in #3563
2. It enables subdirectory support by properly accounting for the subdirectory in generated URLs.

## How have you tested this?

The migration (both up and down) has been unit tested and also tested against a real ABS database.

Modifications to server and client were tested with both existing (non-subdirectory) and subdirectory builds. The server and client were accessed through different server addresses and behaved correctly:
- RSS feed modals displaying the correct URL
- URLs inside the RSS feed are pointing at the right server, including subdirectory (if exists).
- Images on the RSS Feeds config page are all displayed correctly

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -251,18 +251,17 @@ class Server {
 
     const router = express.Router()
     // if RouterBasePath is set, modify all requests to include the base path
-    if (global.RouterBasePath) {
-      app.use((req, res, next) => {
-        const host = req.get('host')
-        const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
-        const prefix = req.url.startsWith(global.RouterBasePath) ? global.RouterBasePath : ''
-        req.originalHostPrefix = `${protocol}://${host}${prefix}`
-        if (!req.url.startsWith(global.RouterBasePath)) {
-          req.url = `${global.RouterBasePath}${req.url}`
-        }
-        next()
-      })
-    }
+    app.use((req, res, next) => {
+      const urlStartsWithRouterBasePath = req.url.startsWith(global.RouterBasePath)
+      const host = req.get('host')
+      const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
+      const prefix = urlStartsWithRouterBasePath ? global.RouterBasePath : ''
+      req.originalHostPrefix = `${protocol}://${host}${prefix}`
+      if (!urlStartsWithRouterBasePath) {
+        req.url = `${global.RouterBasePath}${req.url}`
+      }
+      next()
+    })
     app.use(global.RouterBasePath, router)
     app.disable('x-powered-by')
 
--- a/server/Server.js
+++ b/server/Server.js
@@ -253,6 +253,10 @@ class Server {
     // if RouterBasePath is set, modify all requests to include the base path
     if (global.RouterBasePath) {
       app.use((req, res, next) => {
+        const host = req.get('host')
+        const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
+        const prefix = req.url.startsWith(global.RouterBasePath) ? global.RouterBasePath : ''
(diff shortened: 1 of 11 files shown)


mikiher:
_Note_: I have not deleted the serverAddress column in the feeds table, and it is still set when creating new RSS feeds. While it's not used anymore in the code, it is stored in order to be used in downward migrations. If there are no issues with the new code, we can decide to remove it in some future migration.

server/Server.js
```diff
@@ -253,6 +253,10 @@ class Server {
     // if RouterBasePath is set, modify all requests to include the base path
     if (global.RouterBasePath) {
       app.use((req, res, next) => {
+        const host = req.get('host')
+        const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
+        const prefix = req.url.startsWith(global.RouterBasePath) ? global.RouterBasePath : ''
+        req.originalHostPrefix = `${protocol}://${host}${prefix}`
```
advplyr, line 259: We need to define this even if `global.RouterBasePath` is not set.
Testing this without setting the `RouterBasePath` results in undefined passed to `buildXml`
image

server/objects/Feed.js
```diff
@@ -402,12 +402,12 @@ class Feed {
     this.xml = null
   }
 
-  buildXml() {
+  buildXml(originalHostPrefix) {
     if (this.xml) return this.xml
```
advplyr, line 406: It may be worthwhile to check if the cached `xml` is using the same url. 

Otherwise you could open an RSS feed on your local network to test, that caches the XML, then when accessing the RSS feed on your reverse proxy it still uses the cached version of the XML with your local ip address.

server/Server.js
mikiher: Sorry about this. I made some changes after switching to non-empty ROUTER_BASE_PATH and haven't properly checked them. This is fixed now.

_Note_: `urlStartsWithRouterBasePath` is true if `global.RouterBasePath` is empty, so the I removed the `if (global.RouterBasePath)` check altogether

server/objects/Feed.js
mikiher: I removed the xml cache - It is never used anyway, since a new Feed Object is created for each `getFeed` request. 

Caching for this needs to be handled by the API cache, which I need to revisit to make it more useful.

mikiher:
I fixed the two issues you mentioned. PTAL again.

server/objects/Feed.js
advplyr: Oh right, I forgot that I updated that to pull from the database on every request instead of keeping the feeds in memory. It would be better to handle that cache in a more standardized way anyway.

advplyr:
Working well with & without router base path. Thanks!

Issue #3081, opened 2024-06-16
[Bug]: Downloading audiobooks - poor performance on local LAN


codhopper:
### What happened?

I've been fighting with downloading audiobooks to my android phone for a while. Initially I thought it might have been my wifi, router, virtualisation layers, bug with pfsense, haproxy slowness etc. 

Attempted to download them from a directly connected machine with 10gbit connection. iperf gets around 9.37gbit. Audiobook gets around 100mbit.

The CPU usage is also very high. Out of 2 or 4 cores allocated on the i5-10400 it seemed to use around 100% of a single core any way.

### What did you expect to happen?

Audiobooks download in less than about 10 minutes. CPU load is lower. 

### Steps to reproduce the issue

1. Download large audiobook
2. Grab some coffee
3. Finish download

### Audiobookshelf version

v2.10.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
zlib set to 9
user@server-on-lan:~$ curl -o /dev/null http://192.168.70.2:10228/api/items/4f7088d8-3a90-4b95-b9df-8e2c13c7ba0a/download?token=eyJiwiatruncatedforj85I
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1265M    0 1265M    0     0  14.3M      0 --:--:--  0:01:28 --:--:-- 15.4M

zlib set to 1
user@server-on-lan:~$ curl -o /dev/null http://192.168.70.2:10228/api/items/4f7088d8-3a90-4b95-b9df-8e2c13c7ba0a/download?token=eyJiwiatruncatedforj85I
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1279M    0 1279M    0     0  30.5M      0 --:--:--  0:00:41 --:--:-- 30.5M

zlib set to 0
user@server-on-lan:~$ curl -o /dev/null http://192.168.70.2:10228/api/items/4f7088d8-3a90-4b95-b9df-8e2c13c7ba0a/download?token=eyJiwiatruncatedforj85I
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #3714, opened 2024-12-14
Change: no compression when downloading library item as zip file

nichwall:
## Brief summary

This PR speeds up the endpoint for downloading the zip of a library item by removing compression.

## Which issue is fixed?

Partially fixes https://github.com/advplyr/audiobookshelf/issues/3081

## In-depth Description

Removing compression on downloads for library items greatly reduces CPU usage (removes costly compression, just streams directly from file to response). This does increase file size by about 2-3%, but removes the limit of 10-15 MBps (depending on hardware) for downloading a zip file. Compressing already compressed files does not really save any bandwidth. More discussion in https://github.com/advplyr/audiobookshelf/issues/3081#issuecomment-2526359501

This can increase download time if the network connection is slower than 10 MBps by about 2-3% due to the file size being slightly larger.

I did investigate adding the `Content-Length` header to give users a better "estimated time remaining", but this ended up being a bit difficult due to the length of the zip file not being a direct sum of each file size. https://stackoverflow.com/questions/10927442/calculate-size-of-zip-file-with-compression-level-0/19380600#19380600

## How have you tested this?

I have tested downloading over Ethernet, WiFi, and cellular data to benchmark each connection.

## Screenshots

N/A

### DIFF
--- a/server/utils/zipHelpers.js
+++ b/server/utils/zipHelpers.js
@@ -7,7 +7,7 @@ module.exports.zipDirectoryPipe = (path, filename, res) => {
     res.attachment(filename)
 
     const archive = archiver('zip', {
-      zlib: { level: 9 } // Sets the compression level.
+      zlib: { level: 0 } // Sets the compression level.
     })
 
     // listen for all archive data to be written


advplyr:
Thanks!

PR #3717, opened 2024-12-14
DropDown for Year in Review

Vito0912:
## Brief summary
Adds a dropdown to select the year to view in Year in Review.

## Which issue is fixed?

Discord request

## In-depth Description

This adds a dropdown displaying the years from the user's creation year up to the current year. The default year remains unchanged. It allows users to view both past years and the current year. Some of the displayed data is quite interesting, making it a valuable option to explore even before the year ends.

The button is shown only when there is more than one year. Additionally, I included a prop in the dropdown to enable adding custom classes to the dropdown button.

## How have you tested this?

Devserver with production database

## Screenshots

grafik
grafik

### DIFF
--- a/client/components/stats/YearInReview.vue
+++ b/client/components/stats/YearInReview.vue
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <div v-if="processing" class="max-w-[800px] h-80 md:h-[800px] mx-auto flex items-center justify-center">
+    <div v-if="processing" role="img" :aria-label="$strings.MessageLoading" class="max-w-[800px] h-80 md:h-[800px] mx-auto flex items-center justify-center">
       <widgets-loading-spinner />
     </div>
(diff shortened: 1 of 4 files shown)


advplyr:
I thought it would be better to put the year dropdown at the top since changing it affects both personal and server stats. After making that change we don't need the exact height override on the dropdown component, so I removed that.

image


I updated the `Share` button so that it shows on larger screen sizes. I'm not sure why I originally made that hidden. Browsers have built-in share functionality.

I also made accessibility improvements to the year in review components.

I noticed that you aren't using the editor config files that do things like remove trailing commas and update spacing. If you're using VSCode this should be applied automatically on save as long as you have the recommended extensions. If you're not using VSCode there is a `.pretterrc` and `.editorconfig` file that could be used by other editors.

Thanks!

PR #3727, opened 2024-12-17
feat: Added Australia and New Zealand podcast regions

brinlyau:
## Brief summary

Added Australia and New Zealand podcast store regions

## Which issue is fixed?

Added AU/NZ podcast store regions

## In-depth Description

Pretty self explanatory


brinlyau:
Need to compile it again and test it, but tested on a old branch a while ago.

advplyr:
Thanks!

PR #3732, opened 2024-12-18
check for mrss item media:content when extracting item enclosures

### LINKED ISSUE #3695: [Enhancement]: Support media:content in addition to enclosure.url for podcast episodes
### Type of Enhancement

Server Backend

Some podcasts do provide URLs to their content not within the enclosure -> url attribute, but within a media:content attribute. Audiobookshelf currently only checks for enclosure.url and thus ignores episodes with an url within a media:content attribute. I think this format is known as Mrss, I don't know for sure though. That seems to be the main way of exporting podcasts from rss-bridge https://github.com/RSS-Bridge/rss-bridge.
An example podcast can be found here: https://rss.nixnet.services/?action=display&bridge=ARDAudiothekBridge&path=https%3A%2F%2Fwww.ardaudiothek.de%2Fsendung%2Fkalk-welk%2F10777871%2F&limit=&format=Mrss

Rss Bridge is a pretty useful tool, and most likely not the only one using Mrss. It'd be great to be able to use rss bridge together with ABS.

Just a backend change, no visual change required.

v2.17.5

_No response_

Timtam:
## Brief summary

Some RSS feeds use the enhanced media RSS format (Mrss), particularly the ones created by Rss Bridge. This format stores its enclosures in media:content tags, which ABS didn't yet understand. This PR adds support for those tags as a fallback from traditional enclosures.

## Which issue is fixed?

Fixes #3695 

## In-depth Description

The function responsible for extracting item episode info will test for enclosure urls first, and if it doesn't find one, it'll search for media:content entries that do have a type that starts with "audio". This is necessary since Mrss also supports storing images within the media:content attributes. If an entry is found, its used as an enclosure, since it seems to have the same tags. If no media:content with audio/something type is found, the function will abort with the usual error.

## How have you tested this?

I tested with regular RSS feeds and an Mrss feed. I particularly tested this one: https://rss.nixnet.services/?action=display&bridge=ARDAudiothekBridge&path=https%3A%2F%2Fwww.ardaudiothek.de%2Fsendung%2Fkalk-welk%2F10777871%2F&limit=&format=Mrss


advplyr:
Thanks!

Issue #3748, opened 2024-12-26
[Bug]: config/rss-feeds display distorted feed images


PhantomPhoton:
### What happened?

When I go to config/rss-feeds, I see distorted cover images that make it difficult to use the list

![Screenshot 2024-12-26 at 12 03 51 AM](https://github.com/user-attachments/assets/c78e8855-f771-43a1-ac26-64eda9902a37)

### What did you expect to happen?

I expected to see the normal cover images

### Steps to reproduce the issue

1. Have active RSS feeds
2. Go to settings -> Rss Feeds

### Audiobookshelf version

v2.17.5

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Safari

nichwall:
I'm not able to reproduce this. Do you see images with weird dimensions like this anywhere else?

Have you adjusted the zoom level of the browser or within ABS? Do you have any browser plugins installed, and if so does this go away when you disable all plugins?

nichwall:
I do not have any Apple devices, I was testing with Firefox.

Thanks for helping narrow it down to the h-full property.

nichwall:
Does adding a second RSS feed make the images half as tall? So the same vertical space is used, but split between two images?

PR #3751, opened 2024-12-27
Change: height of RSS feed preview to match aspect ratio

nichwall:
## Brief summary

This PR forces the height of the image preview in the RSS feed to be square and match the size of the width.

## Which issue is fixed?

This should fix https://github.com/advplyr/audiobookshelf/issues/3748

## In-depth Description

This PR replaces the `h-full` with `h-auto`. I did play with changing this to be a `div` to force the images to be square, but the current behavior allows for rectangular aspect ratios so I stuck with `h-auto`.

The `h-full` attribute only seems to be an issue in Safari, as per #3748.

## How have you tested this?

Tested with square and rectangular covers. Tested with series and individual books.

Only tested in Firefox, because I do not have access to a Mac for testing on Safari.

## Screenshots

image

### DIFF
--- a/client/pages/config/rss-feeds.vue
+++ b/client/pages/config/rss-feeds.vue
@@ -25,7 +25,7 @@
           <tr v-for="feed in feeds" :key="feed.id" class="cursor-pointer h-12" @click="showFeed(feed)">
             <!--  -->
             <td>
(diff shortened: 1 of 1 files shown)


advplyr:
I tested this on Safari and it is working. Thanks!

PR #3761, opened 2024-12-30
Remove old Collection object + remove unnecessary toasts

advplyr:
Refactors the Collection model and CollectionController to not use the old Collection object in order to remove it.

Similar to https://github.com/advplyr/audiobookshelf/pull/3721 where Feeds were refactored


advplyr:
This also removes 3 toast messages that were unnecessarily being shown.
Adding items, removing items, and creating a new collection.

PR #3771, opened 2025-01-01
Adds a configuration for podcast feed and episode download timeout

### LINKED ISSUE #3662: [Enhancement]: add either global or podcast-dependent timeout for reading RSS feed
### Type of Enhancement

Server Backend

I'd love to be able to control the timeout when ABSfetches RSS updates from podcasts. Maybe even both, allow a global control and the ability to override it locally for any podcast if necessary.

I'm currently listening to podcasts which get auto-generated from websites which don't offer their own RSS feeds (namely the german ARD Mediathek). The server in-between, which translates incoming requests to RSSfeeds by processing the web page and gathering the info on-the-fly, takes some time, which gets longer the more episodes / the more paginated the web page for the project is. I've got a podcast here which takes longer than a few seconds to load, which causes ABS to show an error and abort. Calling the same URL via the browser works just fine, it just takes between 10 and 20 seconds, which seems to be too long for ABS. I guess it'd already be enough to just allow a longer timeout, or make it configurable in the first place.

I'm blind and sadly can't come up with something like that, sorry.
But adding an additional tab to the global settings and an additional settings to the podcast edit dialog would suffice I think.

v2.17.2

_No response_

sbyrx:
## Brief summary

Added a new global configuration PODCAST_DOWNLOAD_TIMEOUT to configure the axios timeout for Podcast feed and episode downloads. The value defaults to the same 30000ms as before if not set, but can now be overridden, or set to 0 to disable timeouts all together.

## Which issue is fixed?

Fixes #3662 

## In-depth Description

I've been experimenting with [podly_pure_podcasts](https://github.com/jdrbc/podly_pure_podcasts) to remove ads from Podcast episodes. This results in really long download times as the podly server processes the episode, causing downloads in Audiobookshelf to timeout.

## How have you tested this?

Tested by running locally, configuring the timeout to 0, downloading an episode through the podly server, and confirming the download does not timeout.

All tests pass when running `npm run test`

## Screenshots

N/A


advplyr:
The env variable should be specific to podcast file downloads unless it is also used for cover image downloads.

sbyrx:
Sure, I can rename the env variable to PODCAST_DOWNLOAD_TIMEOUT if you would prefer.

advplyr:
Thanks!

Issue #29, opened 2021-09-09
Certain Folder Structures "hide" audiobooks from Scanner.


Kudjo-Grand:
There are 4 books in this particular collection (Dennis Taylor's Bobiverse Series). 

They are in the folders like this:
![image](https://user-images.githubusercontent.com/25695143/132606709-9e3ace3c-c8a8-43b9-94cc-0fd57a423d91.png)

Notice the three .m4b files and a folder.

In AudioBookShelf, the only thing detected by the scan were the contents of that last folder
![image](https://user-images.githubusercontent.com/25695143/132609715-4c813542-6871-44dc-b2ee-425346c5939d.png)

View from ABS:
![image](https://user-images.githubusercontent.com/25695143/132609958-c3d84ae4-b8a3-435f-aee5-064f816cbaff.png)

advplyr:
There were some major flaws in the scanner when it came to nested paths, so I had to rewrite it again. I'm calling it scanner version 3.

#31 I explained the difference with scanner v3 in a discussion. It should resolve the case you mentioned.

There is another downside though, which I point out in the second example of the discussion post. There are some workarounds that can be added later to make that better.

I really appreciate your help with testing all these features and your suggestions, you've played a significant part in taking this project out of pre-release, so thank you!

Budlyte:
I'm mainly posting so I can get an updated when v3 is released, because I now had the opposite happen, Book 1 in the series folder was discovered and everything else was ignored. And until we can get Series tokens for file naming in Readarr, messes like this are going to be common.  :(

![image](https://user-images.githubusercontent.com/4451365/133548522-b3ea3fe7-fc38-4d8a-a8b9-0df2418f21f6.png)

advplyr:
Scanner v3 is released already. In your picture all of those Book audio files should be detected by the scanner.
Are the audio files being ignored or what is happening?

Budlyte:
It had only detected Book 1 - Off To Be The Wizard in that folder.  So I went back and fixed the mess that Readarr had made of the folder, deleted the book from Audiobookshelf, and did a rescan.  This time it detected all the files but piled them under Book 1.

Here's the folder structure I have used throughout my collection, which the scanner did well with until I deleted a book and re-scanned.

`Author/Series Name/Series Name - Book X - Title/ Series Name - Book X - Title.ext`
[...]

PR #3776, opened 2025-01-02
Fix ffmpeg concat file escaping

mikiher:
## Brief summary

This fixes an escaping issue causing UNC paths not to be read correctly in ffmpeg concat files generated by ABS

## Which issue is fixed?

Fixes the re-opened [audiobookshelf-windows issue #29](https://github.com/mikiher/audiobookshelf-windows/issues/29#issuecomment-2547238819).

## In-depth Description

The issue was that the \\\\ chatacters in the beginning of the UNC path were not escaped in the ffmpeg concat file, and therefore the path was mangled. 

Rather than escaping those, I fixed this by surrounding file paths in the concat file with single quotes (and consequently also modifying `escapeSingleQuotes`)

This way is safer, since now the only character that needs escaping inside a file name is a single quote, which the aptly named `escapeSingleQuotes` function now really does - no other characters (like spaces or backslashes) need to be escaped.

## How have you tested this?

I tested with a combination of UNC paths that also contain single quotes.
I tested both codepaths that call `writeConcatFile` (encode m4b, and stream).

### DIFF
--- a/server/utils/ffmpegHelpers.js
+++ b/server/utils/ffmpegHelpers.js
@@ -8,8 +8,8 @@ const { filePathToPOSIX, copyToExisting } = require('./fileUtils')
 const LibraryItem = require('../objects/LibraryItem')
 
 function escapeSingleQuotes(path) {
-  // return path.replace(/'/g, '\'\\\'\'')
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #3777, opened 2025-01-02
Fixed Day Calculation Logic for Stats Page

Vito0912:
## Brief summary

The calculation of the days listened was based on the days with a displayed value. On a smaller scale, not all days of the last year were rendered. This PR changes the calculation of days to include those outside of the rendered days.

While I tried to incorporating the logic into the current loop, I failed because the day's position depends on "i," which would need to be expanded. Maybe I missed something, but the added burden of either going from back to front or adding logic that calculates the columns and rows based on the viewed days would significantly complicate the logic. Therefore some overhead is added by this PR

## Which issue is fixed?

Discord

## How have you tested this?

Mobile and Desktop, on a "Production" server

### DIFF
--- a/client/components/stats/Heatmap.vue
+++ b/client/components/stats/Heatmap.vue
@@ -197,6 +197,18 @@ export default {
       let maxValue = 0
       let minValue = 0
 
+      const wholeYearDays = (52 * 7 + this.dayOfWeekToday);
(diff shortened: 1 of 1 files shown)


Vito0912:
Fixed in a commit in main

PR #3780, opened 2025-01-03
API Cache Manager route uses case-insensitive match

nichwall:
## Brief summary

Fix case-sensitive matching for API Cache Manager.

## Which issue is fixed?

CodeQL issue from PR #3335.
Workflow output: https://github.com/advplyr/audiobookshelf/pull/3335/checks?check_run_id=35084757148

## In-depth Description

https://github.com/advplyr/audiobookshelf/pull/3335#issuecomment-2567188052
This PR changes the regex path matching to be case insensitive for the API Cache Manager. Two methods I found were to force all request URLs to be lowercase, or to ensure that the regex matches against a case-insensitive string. I chose to use case-insensitive matching for now due to some routes potentially being case sensitive, such as:
`DELETE /api/tags/:tag`
`DELETE /api/genres/:genre`

## How have you tested this?

I changed various API calls in the web client to use various capitalization for library API endpoints and verified that everything was still routed correctly. The API Cache Manager continued to have cache hits and return data correctly.

## Screenshots

N/A

### DIFF
--- a/server/managers/ApiCacheManager.js
+++ b/server/managers/ApiCacheManager.js
@@ -42,6 +42,8 @@ class ApiCacheManager {
         Logger.debug(`[ApiCacheManager] Skipping cache for random sort`)
         return next()
       }
+      // Force URL to be lower case for matching against routes
+      req.url = req.url.toLowerCase()
       const key = { user: req.user.username, url: req.url }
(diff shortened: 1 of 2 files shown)


advplyr:
Thanks!

PR #3800, opened 2025-01-05
Migrate to new library item in scanner

advplyr:
Continuation of https://github.com/advplyr/audiobookshelf/pull/3779

Also consolidates the series/author updating in the `Book` model


advplyr:
This removes the old `Podcast`, `Book`, `PodcastEpisode`, `BookMetadata`, `PodcastMetadata` objects.

The update library item endpoint at PATCH `/items/:id` is removed. It wasn't used by Abs clients and wasn't documented.

I found an issue where the `Continue Reading` and `Read Again` home page shelves weren't being populated. These are now populated for library items that are ebook only.

Issue #3775, opened 2025-01-02
[Bug]: Updating a publish date without time fails silently


PhantomPhoton:
### What happened?

I was trying to set a item's publish date to a date and the update appeared successful but didn't update the item's date

### What did you expect to happen?

I expected the item's date to be set

### Steps to reproduce the issue

1. Attempt to update an item's publish date without setting the time components in the dialog box
2. Save
3. Notice item was not updated

### Audiobookshelf version

v2.17.7

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Safari

### Logs

```shell
There is no logged attempt at updating the value, even with debug logs enabled.
```

nichwall:
I assume you are talking about the publish date of an individual podcast episode, not full podcasts or book items?
![image](https://github.com/user-attachments/assets/d78686ad-cbb6-45e2-a65d-df59b69aad90)

Is the time in the component shown as dashes?

Issue #1519, opened 2023-02-19
[Enhancement] Support podcast episode Pub Date without specifying time


apiweb:
### Describe the issue

When I try to edit a podcast episode. If I don't input a time on the Pub Date field, the form is submitted with a null PubDate value.
Some episodes I don't hour the hour date, but the ABS required a time when I try to edit a podcast file.
If I input only the date info, my request is sent with a null/empty value as PubDate.

![image](https://user-images.githubusercontent.com/5881027/219977127-db5f6acc-08bb-4a96-98d7-06c89def65a2.png)

### Steps to reproduce the issue

1. Edit a podcast episode
2. Click on the pub date and insert only the date
3. Save the episode
4. The pub date is saved as null

This happens if you are using the tab key to jump between the inputs
When you insert a date (typing, not choosing by the calendar component), if you press the tab key, the form will jump to the Title field, just with the Date inserted on the PubDate field.
And if you try to save just the date, a null value is sent instead of the Date only.

<img width="409" alt="ErrorOnTheRequest" src="https://user-images.githubusercontent.com/5881027/219977510-af0af0cd-4b88-42ee-892a-7c8c72b77f19.png">

### Audiobookshelf version

v2.2.15

### How are you running audiobookshelf?

Docker

chrishoage:
To me this is a data loss bug. The form should be in an "invalid" state and not let you save at the very least. 

I just spent a lot of time filling out podcast details not knowing I was not saving my work.

advplyr:
The next release will include an error message preventing submission and a red border showing the date is invalid. It won't support leaving the time blank so I'll leave this open.

The issue with leaving the time blank is that we are using the browsers native datetime input by passing `datetime-local` in the type attribute of the HTML input. When the date and time is either incomplete or invalid the value is left empty on the input.

For example, if you type "25/01/2015" and leave the time blank there is no way (that I know of) to get the "25/01/2015". We have to build a custom input to support that.

I wrote a bit more about that here https://github.com/advplyr/audiobookshelf/pull/3805#issuecomment-2576172888

PR #3805, opened 2025-01-07
Text input date validation

nichwall:
## Brief summary

This PR adds a check for a non-null value for date-time fields in the web client to ensure `null` is not incorrectly submitted.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/3775 and partially fixes https://github.com/advplyr/audiobookshelf/issues/1519

## In-depth Description

The submit button now checks that the date-time field to ensure that it is non-null. The datetime will be null if any field (day, month, year, hour, minute) is not set. I added a new error toast to inform the user of the date not being formatted correctly.

We could have the UI element default to 12:00 AM if only the time is not specified, such as is mentioned in #1519, as well as to ensure that different browsers render the "default" time correctly, such as Safari in #3775.

## How have you tested this?

Tested the web client with fully valid datetimes, each individual field missing, and no change to the date.
Tested using Firefox.

## Screenshots

error toast on edit screen


nichwall:
I spent some more time trying to figure out how to get the hours and minutes to automatically set to 12:00 if not set, but could not figure it out, so just leaving it as a submission error if everything is not set.

advplyr:
There was an issue with this in that you couldn't remove the pub date after it was already set. I'm not sure why you would want to do that but it should still be possible.

I made some updates using the validation set on the datetime input.

I don't think it is possible to automatically set the time to 12:00 AM unless we create our own datetime input.
 
As far as I know it is not possible to get the value of an incomplete or invalid date in a native HTML datetime input. The "value" on the HTML input is only set when the date is valid. If you enter "6/31/2024 12:00 AM" it will still not populate the value.

The `validity`  object is set though which we can use to check for any bad input. When the date is not set at all `validity.badInput` is false, so we can use this to check if we should show the error toast message or not.

image

I also added a red border anytime `badInput` is true as well as updated the toast message to cover both invalid and incomplete.

image

image

advplyr:
Thanks!

nichwall:
Thanks for cleaning that up! I figured there was a better way to do it, I was mostly just trying to get the framework sorted out.

PR #3808, opened 2025-01-07
Merge prod.js into index.js

mikiher:
## Brief summary

Merges prod.js and index.js.

## Which issue is fixed?

No existing issue. I thought that having two different entry points into the code was not healthy and could lead to bugs.

## In-depth Description

The new index.js has the commandline parameters and default values that prod.js had, as well the dev.js config functionality from the old index.js. The merge has a couple of logic changes:
- `--dev` or `-d` flag to force `NODE_ENV=development`
- Logging of the `NODE_ENV` value and of all of the arguments passed to the `Server` constructor, for better debuggability.

Consequently a couple of additional changes in other places were required:
- Defining the required default environment variables explicitly in `dockerfile` (which is also good for readabilty)
- Switcing from `prod.js` to `index.js`, and adding the `--dev` flag to the `dev` script in `package.json`

_Note:_ in the new code, you need to run with NODE_ENV=development or `--dev` to force development mode, otherwise production is assumed.

## How have you tested this?

- Ran with `--dev`/`-d`
- Built and run docker image with new ENV settings


advplyr:
I'm thinking that we leave prod.js in for a release and add in the release notes a message that it will be removed soon.

The reason is because I'm sure there are some 3rd party install methods using `prod.js`. I found one https://github.com/YunoHost-Apps/audiobookshelf_ynh/blob/master/conf/systemd.service

I'm not sure if there are any other breaking changes we should make a notice of before requiring the change

advplyr:
Thanks!

Thovi98:
Thanks @advplyr for taking third party packages into account and for writing the breaking changes in the release note, it's verry helpful!

index.js
```diff
@@ -17,14 +32,19 @@ if (isDev) {
   process.env.ROUTER_BASE_PATH = devEnv.RouterBasePath || ''
 }
 
-const PORT = process.env.PORT || 80
```
advplyr, line 20: I think we should switch back to using port 80 as default for production to fix https://github.com/advplyr/audiobookshelf/issues/3868.

And keep 3333 as default for development.

What do you think?

mikiher: I'm not sure how come this happens? I also changed dockerfile in the same commit.

advplyr: I assumed maybe they were using a custom Dockerfile since they are all Synology users but now I'm not so sure after looking into Synology tutorials online.

jrhedman: Also running on Synology and had the same issue. It seems the port we need to expose moved from port 80 to 3333. I checked the Dockerfile both in the repo and docker hub and it's still showing as port 80 so I'm not sure if that's whats causing the issue.

I know I explicitly defined port 80 to be mapped. Alternative might be to update dockerfile to expose port 3333 instead of 80 and add in the release notes that going forward, audiobookshelf will use port 3333.

Switching my mapping to port 3333 addressed my issues and it looks like a few others as well.

advplyr: The problem with changing the port now is that would break the vast majority of users that are not using Synology. As I mentioned [here](https://github.com/advplyr/audiobookshelf/issues/3868#issuecomment-2605153972) this is a small subset of users using Synology and we should figure out why that is happening before making any changes.

jrhedman: Sure thing, I've shared some details of my docker inspect and docker logs (currently set up for failure state) over on the issue page so we're not jumping back and forth.

Let me know how I can bug test for you since I'm able to reproduce.

advplyr: Not an Abs issue as far as we can tell. Something with Synology caching is the [best guess](https://github.com/advplyr/audiobookshelf/issues/3868#issuecomment-2605327783).

Issue #1417, opened 2023-01-20
Collections RSS feed repeats the first book


benonymity:
### Describe the issue

When I open an RSS feed of a collection, though every book is shown with proper title, description, and length, trying to play any will only play the audio file of the first book.

### Steps to reproduce the issue

1. Create a collection with multiple books
2. Open its RSS feed and add it to your RSS app.
3. Try to listen to any of the “episodes”

### Audiobookshelf version

v2.2.12

### How are you running audiobookshelf?

Docker

advplyr:
I'm not able to reproduce this. It is possible I fixed it in the latest release.

Let me know if you can reproduce and we will re-open

PR #3754, opened 2024-12-28
Experimental proxy support by manually following redirects

maxlajoie99:
## Brief summary
Make downloading podcast episodes work when `HTTP_PROXY`/`HTTPS_PROXY` environment variable are present.

## Which issue is fixed?

#1313 

## In-depth Description

Mannually following redirects seems to fix the bug in axios when using a proxy. It is merely a workaround while we wait to upgrade axios / replace pkg like mentionned [here](https://github.com/advplyr/audiobookshelf/pull/2666#issuecomment-1974083316).

## How have you tested this?

Ran a forward proxy locally using docker, and added both `HTTP_PROXY` and `HTTPS_PROXY` to the start command. Then, I searched and downloaded a few podcast episodes.

PR #3810, opened 2025-01-09
Enable subdirectory support by default

### LINKED ISSUE #385: [Enhancement]: Add custom path settings

mikiher:
## Brief summary

Switch to enabling subdirectory support with `/audiobookshelf` on the server and web client, according to the plan laid out in discussion #3535.

## Which issue is fixed?

This finally fixes #385

Users will now be able to set their reverse proxies to proxy `/audiobookshelf` subdirectory requests

## In-depth Description

This is the last step in supporting subdirectory with the existing static site generation approach. 

The plan is described in detail in #3535.

As a reminder, this will set a fixed subdirectory of `/audiobookshelf` on the web client (non-subdirectory paths will get rewritten by the server), and the server will support both subdirectory requests (with `/audiobookshelf` prefix) and non-subdirectory requests (with no prefix).

_Note_: Mobile app support for subdirectory is already implemented and checked in with audiobookshelf-app PR [#1417](https://github.com/advplyr/audiobookshelf-app/pull/1417), but is pending release of a new version. Other mobile clients that would like to support subdirectory access **will likely need to make some changes** in the spirit of that PR.

## How have you tested this?

This PR just switches the feature on by default. 

The many changes required to make this happen were previously tested in respective previous PRs, and were tested again to make sure they still worked now.

One thing I haven't tested properly is the reverse proxy part, since my reverse proxy (on my Synology NAS) doesn't support support subdirectory proxying. I am sure this works well once you set it up on a supporting reverse proxy (like NGINX), but haven't tested this myself.

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -1,6 +1,6 @@
 const pkg = require('./package.json')
 
-const routerBasePath = process.env.ROUTER_BASE_PATH || ''
+const routerBasePath = process.env.ROUTER_BASE_PATH || '/audiobookshelf'
 const serverHostUrl = process.env.NODE_ENV === 'production' ? '' : 'http://localhost:3333'
 const serverPaths = ['api/', 'public/', 'hls/', 'auth/', 'feed/', 'status', 'login', 'logout', 'init']
 const proxy = Object.fromEntries(serverPaths.map((path) => [`${routerBasePath}/${path}`, { target: process.env.NODE_ENV !== 'production' ? serverHostUrl : '/' }]))
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ const CONFIG_PATH = inputConfig || process.env.CONFIG_PATH || Path.resolve('conf
 const METADATA_PATH = inputMetadata || process.env.METADATA_PATH || Path.resolve('metadata')
 const SOURCE = options.source || process.env.SOURCE || 'debian'
 
-const ROUTER_BASE_PATH = process.env.ROUTER_BASE_PATH || ''
+const ROUTER_BASE_PATH = process.env.ROUTER_BASE_PATH || '/audiobookshelf'
 
 console.log(`Running in ${process.env.NODE_ENV} mode.`)
 console.log(`Options: CONFIG_PATH=${CONFIG_PATH}, METADATA_PATH=${METADATA_PATH}, PORT=${PORT}, HOST=${HOST}, SOURCE=${SOURCE}, ROUTER_BASE_PATH=${ROUTER_BASE_PATH}`)
--- a/prod.js
+++ b/prod.js
@@ -25,7 +25,7 @@ const CONFIG_PATH = inputConfig || process.env.CONFIG_PATH || Path.resolve('conf
 const METADATA_PATH = inputMetadata || process.env.METADATA_PATH || Path.resolve('metadata')
 const SOURCE = options.source || process.env.SOURCE || 'debian'
 
-const ROUTER_BASE_PATH = process.env.ROUTER_BASE_PATH || ''
+const ROUTER_BASE_PATH = process.env.ROUTER_BASE_PATH || '/audiobookshelf'
 
 console.log(process.env.NODE_ENV, 'Config', CONFIG_PATH, METADATA_PATH)
 


advplyr:
Thanks!

4ch1m:
The new release (Docker container `2.18.0`) - which includes/activates this feature - works perfectly fine for me in both the browser and the Android app.

I'm super happy that @mikiher made the effort to implement it. :+1:
Again... thank you very much. :bow: 

However, I think there are few things that should be revised:

* The environment variable `ROUTER_BASE_PATH` is (in it's current state) a bit confusing, since it doesn't seem to have any effect at all; or can't actually be used to change the path value.  
(I'm aware that @mikiher explained that the `/audiobookshelf` path is hardcoded now due to several technical reasons.)  
So why exactly does it still exist?

* The README still says:  
  > Note: Subfolder paths (e.g. /audiobooks) are not supported yet. See https://github.com/advplyr/audiobookshelf/issues/385`  

  This could/should be removed now, I guess.

* The README section containing all reverse proxy setup examples needs to be updated.  
I'm using Apache2 and got everything working with this adjustment:
```
RewriteRule (.*)/audiobookshelf/?(.*) "ws://127.0.0.1:<audiobookshelf_port>/audiobookshelf/$2" [P,L]
```
(Not sure if there's a simpler solution though.)

mikiher:
> The new release (Docker container `2.18.0`) - which includes/activates this feature - works perfectly fine for me in both the browser and the Android app.
> 
> I'm super happy that @mikiher made the effort to implement it. 👍 Again... thank you very much. 🙇
> 
> However, I think there are few things that should be revised:
> 
> * The environment variable `ROUTER_BASE_PATH` is (in it's current state) a bit confusing, since it doesn't seem to have any effect at all; or can't actually be used to change the path value.
>   (I'm aware that @mikiher explained that the `/audiobookshelf` path is hardcoded now due to several technical reasons.)
>   So why exactly does it still exist?

It is still possible to use the environment variable to change the `/audiobookshelf` path, but _only if you build the code yourself_ (i.e. set the environment variable then run `npm run client`). This isn't something most of our users do, hence my use of term "hardcoded".

> * The README still says:
>   > Note: Subfolder paths (e.g. /audiobooks) are not supported yet. See [https://github.com/advplyr/audiobookshelf/issues/385`](https://github.com/advplyr/audiobookshelf/issues/385%60)
>   
>   
>   This could/should be removed now, I guess.

You're right. We'll remove this.

> * The README section containing all reverse proxy setup examples needs to be updated.
>   I'm using Apache2 and got everything working with this adjustment:
> 
> ```
> RewriteRule (.*)/audiobookshelf/?(.*) "ws://127.0.0.1:<audiobookshelf_port>/audiobookshelf/$2" [P,L]
> ```
> 
> (Not sure if there's a simpler solution though.)

Great! We should definitely provide at least one or two examples of modifying reverse proxy setups for subdirectory support (the existing setups should still work for subdomain). I'm not sure we can be responsible for providing details for every reverse proxy out there, but we definitely should encourage users to share their setups with us and publish them.

You're also welcome to submit a PR to change the documentation, if you like!

alternativesurfer:
FWIW, I am running through HAProxy, not Apache.

It looks like this is what broke it: https://github.com/search?q=repo%3Aadvplyr%2Faudiobookshelf%20%20EXP_PROXY_SUPPORT&type=code


I added the env flag: EXP_PROXY_SUPPORT=1 and now its working.
So, seems the fix to this: https://github.com/advplyr/audiobookshelf/pull/3754 broke it for those of us whom reverse proxy's were already working for.

beschoenen:
Shouldn't the default be `'audiobookshelf'`? I'm trying to host this application on `/:user/audiobookshelf` and I'm having a hard time since the Docker is precompiled with `/audiobookshelf`. Removing the slash would make it relative to the current subfolder in the url.

client/nuxt.config.js
```diff
@@ -1,6 +1,6 @@
 const pkg = require('./package.json')
 
-const routerBasePath = process.env.ROUTER_BASE_PATH || ''
+const routerBasePath = process.env.ROUTER_BASE_PATH || '/audiobookshelf'
```
devnoname120, line 3: When the `ROUTER_BASE_PATH` env variable is set to an empty string in order to restore the old behavior then it overrides it to `/audiobookshelf`.

Is that expected? I would have thought that the `/audiobookshelf` fallback should only have been only used in the case where the `ROUTER_BASE_PATH` env variable is undefined.

devnoname120: https://github.com/advplyr/audiobookshelf/pull/3958

PR #3816, opened 2025-01-10
Fix Trix to use paragraphs and break on return

mikiher:
## Brief summary

This fixes a couple of issues with the Trix text editor

## In-depth Description

- Trix translates `p` tags to `div` by default, and inserts redundant `br` tags on paragraph breaks. This fixes this.
- Trix by default always puts a `br` tag on Enter, and never breaks a paragraph (there's literally no way to start a new paragraph). This changes the behavior to break paragraphs on Enter (you can still use Shfit+Enter for line breaks, like in other normal editors)

## How have you tested this?

This was tested on various podcast episode descriptions in the Episode Details tab.

### DIFF
--- a/client/components/ui/VueTrix.vue
+++ b/client/components/ui/VueTrix.vue
@@ -249,11 +249,33 @@ export default {
         }
       }
       return target
+    },
+    enableBreakParagraphOnReturn() {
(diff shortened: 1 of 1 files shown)


advplyr:
Working great, thanks!

Issue #3809, opened 2025-01-07
[Bug]: Port no longer showing in podcasts generated in 2.17.5 and newer


andrewinmoco:
### What happened?

After upgrading to the current audiobookshelf I'm unable play RSS feeds generated ion my podcast player.  When investigating this I found that the port for the web server is no longer included in the generated RSS feed which means that the MP3 files can't be downloaded. 

Working item from RSS feed generated in 2.17.4
![image](https://github.com/user-attachments/assets/c53edab9-7db1-4ef3-8b46-9fc3e555728b)

Non working item from RSS feed generated in 2.17.5
![image](https://github.com/user-attachments/assets/f71d52d9-3fd1-4c65-96e3-82d597b076e4)

### What did you expect to happen?

I expect the port to be included in the URL to download the JPG and the MP3. 

### Steps to reproduce the issue

1. Generate a RSS feed for a book while not on port 443
2. Attempt to load the RSS feed in a podcast reader

### Audiobookshelf version

v2.17.7

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

advplyr:
I see that this is a bug with express stripping the port.

https://expressjs.com/en/guide/migrating-5.html#req.host
> In Express 4, the req.host function incorrectly stripped off the port number if it was present. In Express 5, the port number is maintained.

We can see about migrating to express 5 to resolve this. It doesn't strip the port number in all cases though so I haven't reproduced the issue.

Relevant PR https://github.com/advplyr/audiobookshelf/pull/3692 @mikiher

mikiher:
I'll look at this. Maybe we can work around this. It would be also good switch to express 5.

mikiher:
Actually, looking at the code, it's not using `req.host`, but rather `req.get('host')`, (equivalent to `req.headers.host`) which should not suffer from the bug you referred to.
```js
      const urlStartsWithRouterBasePath = req.url.startsWith(global.RouterBasePath)
      const host = req.get('host')
      const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
      const prefix = urlStartsWithRouterBasePath ? global.RouterBasePath : ''
      req.originalHostPrefix = `${protocol}://${host}${prefix}`
```

@andrewinmoco can you please do the following:
- Copy the feed URL
- Open a Chrome tab
[...]

PR #3828, opened 2025-01-12
recommend using $http_host for ngnix

mikiher:
## Brief summary

Changes our recommendation for Nginx reverse proxy setup to set the `Host` header to `$http_host` instead of `$host`

## Which issue is fixed?

This helped in fixing #3809, which turned out to be caused by Ngnix misconfiguration.

## In-depth Description

In Ngnix, `$host` and `$http_host` are usually the same, except when Nginx is configured to a non-standard HTTP port (i.e. not 80 or 443). When Nginx uses a non-standard HTTP port, `$host` is just the hostname without the port, while `$http_host` is the actual 
`Host` header from the original request, which has both hostname and port.

Since ABS now uses the host to construct RSS feed entries dynamically, it's important to get the actual host from the original request, so that feed links are properly constructed.

## How have you tested this?

Changing the config solved the issue for #3809 reporter.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,8 +111,8 @@ server {
 
    location / {
       proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;
-      proxy_set_header  X-Forwarded-Proto $scheme;
-      proxy_set_header  Host              $host;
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

Issue #3806, opened 2025-01-07
[Enhancement]: Information for book series.


ErikHallmark:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Include more information about a series of books, such as genre(s) author(s) and a description. 

### Why would this be helpful?

It would make it easier for users to familiarize themselves with the books in an ABS library. 

### Future Implementation (Screenshot)

(Excuse my poor photo editing skills)
Ideally the series page would be expanded to include a description, genres, and author information. 
![ABS](https://github.com/user-attachments/assets/653a9485-05ca-4a21-8906-01dd2e0a9c87)

### Audiobookshelf Server Version

v2.17.7

### Current Implementation (Screenshot)

![Screenshot_20250106_192424](https://github.com/user-attachments/assets/44fc3311-c03b-4f66-91a4-3be8dc2f55b4)

advplyr:
Related https://github.com/advplyr/audiobookshelf/issues/711

Issue #711, opened 2022-06-10
[Enhancement]: Series page similar to author page


hobesman:
### Describe the feature/enhancement

I think it'd be nice to have the option of adding series metadata similar to how author metadata is its own entry. So it would be possible to designate a picture (or fall back to the default collage picture) and be able to add a description and maybe some tags or genre info for the series.

Also, when a series page is brought up, it'd be nice to have a generated list of all authors that appear in the metadata of the books in the series. Most series that would just be one author, but that's ok to show "the" author on the series page too.

Also perhaps auto-generate a date range based on the oldest and newest book in the series. Just some ideas but I think being able to add more info onto the "series" page would be nice.

chyron8472:
I like the collage. Maybe if series art is added, it can be displayed as the top artwork in the collage.

advplyr:
Possibly adding easy updating of series sequence on this page as mentioned here #969

b3nd1k:
It would be cool if in addition this, you could also add custom artwork for each series rather than the auto generated one. (This would be great for collections as well).

LeeThompson:
I'd love to see this too!

ElDubsNZ:
Would love series to be treated as objects with their own metadata. Many of my series have their own descriptions and artworks. Being able to detail that would be such an improvement in experience.

PR #3839, opened 2025-01-15
WIP: Feat/book series info

glorenzen:
I've seen a number of requests for enhancements to Series pages, so I'm working on some simple changes, starting first with showing a Series description. I've added a new context menu item on the series pages, which displays a modal for editing the series, but so far it's just the UI.

Is there an API endpoint for updating a series? I've been looking, and have found the routes for getting all series in a library or a specific series in a library, but nothing for updating.

Related issues:
#711 #3806

### DIFF
--- a/client/pages/library/_library/series/_id.vue
+++ b/client/pages/library/_library/series/_id.vue
@@ -17,6 +17,16 @@
             </h2>
             <div>{{ series.description }}</div>
           </div>
+          <div class="mb-6">
+            <h2 class="font-semibold">
+              {{ authors.length > 1 ? $strings.LabelAuthors : $strings.LabelAuthor }}
+            </h2>
+            <div>{{ authors.join(', ') }}</div>
+          </div>
+          <div class="mb-6">
+            <h2 class="font-semibold">{{ this.$strings.LabelTotalDuration }}</h2>
+            <div>{{ totalDuration.hours }}<span>hrs</span> {{ totalDuration.minutes }}<span>min</span></div>
+          </div>
         </div>
 
         <app-lazy-bookshelf page="series-books" :series-id="seriesId" />
@@ -41,7 +51,7 @@ export default {
       return redirect(`/library/${libraryId}`)
     }
 
-    const series = await app.$axios.$get(`/api/libraries/${library.id}/series/${params.id}?include=progress,rssfeed`).catch((error) => {
+    const series = await app.$axios.$get(`/api/libraries/${library.id}/series/${params.id}?include=progress,rssfeed&expanded=1`).catch((error) => {
       console.error('Failed', error)
       return false
     })
@@ -62,6 +72,22 @@ export default {
   computed: {
     streamLibraryItem() {
       return this.$store.state.streamLibraryItem
+    },
+    totalDuration() {
+      const totalSeconds = this.series.books.reduce((acc, book) => acc + book.duration, 0)
+      const hours = Math.floor(totalSeconds / 3600)
+      const minuteRemainder = totalSeconds % 3600
+      const minutes = Math.floor(minuteRemainder / 60)
+
+      return { hours, minutes }
+    },
+    authors() {
+      // Get nested array of authors
+      const nestedAuthors = this.series.books.map((book) => book.authors.map((author) => author.name))
+      // Flatten to one array
+      const authors = nestedAuthors.flat(1)
+      // Remove duplicates
+      return [...new Set(authors)]
     }
   },
   methods: {


glorenzen:
I was thinking of just adding the information to the existing series pages. Is there a reason we can't do that? My thought was the description and other metadata can go above the items in the series as text blocks.

I added an Edit Series option and a modal:
image
image

advplyr:
That was the part I couldn't get to look right. Just adding the description above the list of books. I tried it a while back and it looked strange and couldn't think of a good design for it.

glorenzen:
This layout is pretty simple, but I was just thinking it could look something like this:
series-info-mockup-v1
We could also add an option where users can choose whether the metadata is displayed or not.

glorenzen:
@advplyr Also, not sure if you saw my note earlier, but are the API endpoints for Series? I couldn't find any endpoints for updating Series.

advplyr:
There isn't series endpoints yet.

I think this should be built similar to the authors page but use the lazy bookshelf instead of the horizontal scrolling cards. I could have the sort/filter options above the bookshelf like is done for podcast episodes.
image

glorenzen:
Okay, sounds good. Do you think this should replace the existing template for individual series, or should this be a new page?

glorenzen:
> There isn't series endpoints yet.

I saw there are two endpoints in the ApiRouter file, but they look like they are deprecated and I don't believe they are being used.
`this.router.get('/series/:id', SeriesController.middleware.bind(this), SeriesController.findOne.bind(this))
this.router.patch('/series/:id', SeriesController.middleware.bind(this), SeriesController.update.bind(this))`

Should I use that SeriesController for adding the new methods? Or should I add the logic to the LibraryController, similar to the "/api/libraries/:id/series/:seriesId" route?

nichwall:
> > There isn't series endpoints yet.
> 
> I saw there are two endpoints in the ApiRouter file, but they look like they are deprecated and I don't believe they are being used.
> `this.router.get('/series/:id', SeriesController.middleware.bind(this), SeriesController.findOne.bind(this))
> this.router.patch('/series/:id', SeriesController.middleware.bind(this), SeriesController.update.bind(this))`
> 
> Should I use that SeriesController for adding the new methods? Or should I add the logic to the LibraryController, similar to the "/api/libraries/:id/series/:seriesId" route?
> 
> 

The original method was using the `/api/libraries/:id/series/:seriesId`, but routes will be replaced by `/api/series/:id` and similar for everything else.

The library being part of the path is not necessary and will eventually be removed to help simplify things. From an old message on Discord for more context:

"""
The `/api/libraries/{id}/series/{seriesId}` should be removed now and `/api/series/{seriesId}` used.
The reason I deprecated /api/series/{seriesId} originally is because early version of series didn't store the libraryId with the series so books across multiple libraries could have had the same series id. When migrating to sqlite I put a libraryId with series and so now the series id is unique to each library.
I never switched the mobile app over to use the libraries/{id}/series endpoint so now I only have to switch the web client back to using `/api/series/{seriesId}`
"""

glorenzen:
> > > There isn't series endpoints yet.
> > 
> > I saw there are two endpoints in the ApiRouter file, but they look like they are deprecated and I don't believe they are being used.
> > `this.router.get('/series/:id', SeriesController.middleware.bind(this), SeriesController.findOne.bind(this))
> > this.router.patch('/series/:id', SeriesController.middleware.bind(this), SeriesController.update.bind(this))`
> > 
> > Should I use that SeriesController for adding the new methods? Or should I add the logic to the LibraryController, similar to the "/api/libraries/:id/series/:seriesId" route?
> > 
> > 
> 
> The original method was using the `/api/libraries/:id/series/:seriesId`, but routes will be replaced by `/api/series/:id` and similar for everything else.
> 
> The library being part of the path is not necessary and will eventually be removed to help simplify things. From an old message on Discord for more context:
> 
> """
> The `/api/libraries/{id}/series/{seriesId}` should be removed now and `/api/series/{seriesId}` used.
> The reason I deprecated /api/series/{seriesId} originally is because early version of series didn't store the libraryId with the series so books across multiple libraries could have had the same series id. When migrating to sqlite I put a libraryId with series and so now the series id is unique to each library.
> I never switched the mobile app over to use the libraries/{id}/series endpoint so now I only have to switch the web client back to using `/api/series/{seriesId}`
> """

Got it, that makes more sense. Thanks for the clarification! I'll use the `/api/series/{seriesIs}` routes then.

glorenzen:
Here is what I have so far on the Series page:
image
Clicking the edit icon brings up the modal for editing metadata, similar to the Author pages.
image
Thoughts?

shadowbq:
I also like the author page layout. The Series could have a specific Image or default to the first book in the series as the icon or use "The stack" in the series selection.

saltedlolly:
I'm very pleased to see this being worked on, as it has a feature I have really missed having moved from Plex a few months ago.

Would it be possible to make sure the series description includes paragraph support, and better yet, basic html support (i.e. paragraphs, bold, italics, underline, links). The formatting is obvious, but links can also be helpful to link to external material such as maps etc. (given that these are not currently supported in ABS.) 

Taking it a step further, being able to embed images would be a really nice bonus feature. And perhaps markdown support. Being able to insert `[Map of Westeros](./westeros-map-1.jpg)` directly into the description, that displays a clickable thumbnail with the larger map could have so many potential uses. It would go some way to solving issue #4897. 

(Aside: It's worth noting that standard paragraphs (non-html) are not currently displayed in the ABS UI when included in "Collection" descriptions. However, they do display in Prologue.)

I was also wondering if the metadata tool could also be updated to look for a file called series-description.txt located in the series folder and import its contents as the series description. e.g. /George R.R. Martin/A Song of Ice and Fire/series-description.txt

saltedlolly:
Can I also suggest that in the web UI, for very long descriptions,  there simply be a "More..." link after the first few lines of description, that reveals the rest when clicked. It would stop the books themselves being pushed too far down the page in this situation, when the page first loads.

Issue #3013, opened 2024-05-26
[Bug]: Podcast episode rows disappear while listening for a long period of time.


tymscar:
### What happened?

After listening to an hour or so of podcasts, some of the episode tiles disappear. I have managed to take a screenshot:
<img width="875" alt="image" src="https://github.com/advplyr/audiobookshelf/assets/3742502/deb41fe2-7fdd-4bde-9b61-c2276d61b1b8">

### What did you expect to happen?

I expect to always see all the episodes in the list, depending on the filters applied.

### Steps to reproduce the issue

1. Open audiobookshelf
2. Go to the podcasts tab
3. Start an episode that takes a long time (1+ hours), or just listen to multiple shorter episodes
4. Scroll around and notice how some of the tiles are blank (sometimes 20+ tiles, sometimes all, sometimes one)

### Audiobookshelf version

V2.8.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Other (list in "Additional Notes" box)

### Logs

```shell
I have noticed it on Chrome and Firefox on Linux, on Arc (Chromium), Firefox, and Safari on macOS, and on Firefox on Windows.
```

### Additional Notes

This issue has been present in every version I've tried for the past a year and a half, so it's not a bug with a recent commit. I have also noticed it on every desktop OS, and browser. The bug is not present on the iOS application, so I suspect its a CSS issue, but I can't confirm.
The console is clean of any errors when the bug happens, here is a screenshot:
<img width="417" alt="image" src="https://github.com/advplyr/audiobookshelf/assets/3742502/2587b8d9-615a-4b74-b99c-763b73a033f9">

nichwall:
Looks like the websocket is closing and the lazy loading of new podcast episodes takes a bit to connect and start loading new items.

Related behavior https://github.com/advplyr/audiobookshelf/issues/2950#issuecomment-2105358205

tymscar:
The websocket might've closed there just out of inactivity. Im not sure if that's the issue, as there aren't that many audiobooks in this library. Maybe 10?

advplyr:
Are new episodes being downloaded? Do you have automatic library scans enabled?

tymscar:
I do have automatic library scans enabled, no new episodes where downloaded when that issue happened. A lot of time I listen to super old podcasts that dont even launch new episodes anymore and I still find this bug.

[...]

PR #3842, opened 2025-01-15
Dynamically calculate episode row height on LazyEpisodeTable init

### LINKED ISSUE #3511: [Bug]: When scrolling through episodes in a podcast the listing of items disappears
### What happened?

* Scroll down through the episodes of a podcast
* The number of episodes diminishes until none are visible

* Be able to see the text of all the episoides

1. Select a podcast from Library
2. Begin to scroll down through the episodes
3. Visible episodes become fewer and fewer until completely gone
4. At that point it's not possible to select an older episode  
  
(see attachments)

### Audiobookshelf version

2.15.0

Docker

Linux

Chrome

### Logs

_No response_

### Additional Notes

Thorium (Chromium) 126.0.6478.231
Mercury (Firefox) 129.0.2 
This has been going on for a long time, and I've seen the same behavior across all browsers and versions, extensions disabled or not.

mikiher:
## Brief summary

Instead of using a fixed size for episodeRowHeight, it is given a default value but recalculated when the episode table is mounted.

## Which issue is fixed?

This fixes #3511. While this is marked as related to #3013, I believe they're not actually related.

## In-depth Description

The bug was reproduced by manually setting the root font-size to a larger value (I suspected that this might be reason, because the reporter mentioned system scale changes).

The original code hardcodes episodeRowHeight value to 176 px (which is indeed the row height when the root font size is the default 16px). episodeRowHeight is used to make scroll calculations to decide which episodes to mount, and when it is different from the actual value, it results in mounting incorrect episodes.

The fix determines episodeRowHeight from the actual dom element of the first episode (or from the "no-episodes" element, if no episodes exist), which fixes the issue above.

## How have you tested this?

Used Chrome settings to change the root font-size value, and ensured scrolling worked OK after reloading the page.

_note:_ this fix doesn't automatically react to changes in the root font-size - it only recalculates episodeRowHeight in the init() function. I can make it reactive, but it seems unnecessary, as this is not something that users play with frequently.

### DIFF
--- a/client/components/tables/podcast/LazyEpisodesTable.vue
+++ b/client/components/tables/podcast/LazyEpisodesTable.vue
@@ -30,7 +30,7 @@
         <ui-text-input v-model="search" @input="inputUpdate" type="search" :placeholder="$strings.PlaceholderSearchEpisode" class="flex-grow mr-2 text-sm md:text-base" />
       </form>
     </div>
-    <div class="relative min-h-[176px]">
+    <div class="relative min-h-44">
       <template v-for="episode in totalEpisodes">
         <div :key="episode" :id="`episode-${episode - 1}`" class="w-full h-44 px-2 py-3 overflow-hidden relative border-b border-white/10">
           <!-- episode is mounted here -->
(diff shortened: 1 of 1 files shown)


advplyr:
I don't think it is necessary to make this reactive. Works great, thanks!

Issue #891, opened 2022-08-04
Stream URL


eximo84:
### Describe the feature/enhancement

We have a YotoPlayer for our kids. They have the ability to play audio streams via a url

Would it be possible to create a stream url to an audio book or series so that can be used instead of uploaded mp3 files to the Yoto website. 

Details here https://us.yotoplay.com/pages/makeyourown#

advplyr:
What is it expecting for a stream URL? An HLS playlist?

eximo84:
Maybe? https://support.yotoplay.com/en-US/make-a-card-from-radio-stations-that-are-not-on-the-yoto-app-21682

Example would be similar to radio stations I guess. I'm not really sure.

batonac:
So actually the RSS feeds generated by Audiobookshelf works, but only if the files are MP3. The enhancement we'd need, however, is to reverse the order of the files in the feed.

PR #3890, opened 2025-01-24
Feat: RSS Feed Reverse Order

batonac:
## Brief summary

This PR adds the ability to choose to reverse the order of the episodes in a generated RSS Feed.

## Which issue is fixed?

A workaround for #891

## In-depth Description

Some players that support RSS playback (in this case the Yoto portable players) play Audiobookshelf-generated RSS feeds from back to front instead of front to back. This feature allows users to optionally reverse the order of the episodes in the feed in order to work with this behavior.

## How have you tested this?

I've tested the code in a local dev container. The server and client work fine with these changes. I've verified that the episode order is reversed when the option is applied, by searching for the `length` of the first two `item` nodes without the reverse order option, and verifying that those nodes are sent to the bottom in reverse order when `reverseOrder` is enabled.

## Screenshots

reverseOrder


server/models/Feed.js
```diff
@@ -58,6 +60,8 @@ class Feed extends Model {
     this.explicit
     /** @type {boolean} */
     this.preventIndexing
+    /** @type {boolean} */
+    this.reverseOrder
```
nichwall, line 64: If the database is changing, we will need to have an accompanying database migration script to support rolling back to previous database versions.

More information can be found here https://github.com/advplyr/audiobookshelf/blob/master/server%2Fmigrations%2Freadme.md

I'm not sure if this table has anything referencing it or if we can just use `changeColumn`. We have had some issues with `changeColumn` in the past.

client/strings/bg.json
```diff
@@ -426,6 +426,7 @@
   "LabelRSSFeedCustomOwnerName": "Потребителски собственик Име",
   "LabelRSSFeedOpen": "RSS Feed Оптворен",
   "LabelRSSFeedPreventIndexing": "Предотврати индексиране",
+  "LabelRSSFeedReverseOrder": "Обратен ред",
```
nichwall, line 429: Where did you get all of the translations? Generally translations should be handled through Weblate so native speakers can ensure translation accuracy.

nichwall (changes requested):
I haven't tested any of this yet, but it looks good to me overall from glancing over it. My main concern is there needs to be a migration script for the database.

client/strings/bg.json
batonac: I wondered about that. I just used GPT-4o. I thought that adding the string to all the translations would be necessary, and I figured this was simple enough that a LLM should do fine, but then I realized later that the existing translations aren't all complete anyway. We can just drop them.

server/models/Feed.js
batonac: `v2.17.6-share-add-isdownloadable` looks like a perfect template. Added!

client/strings/bg.json
batonac: Reverted LLM translations!

advplyr:
I think it may be better to do what apple does and utilize the podcast type to determine the order. Episodic or serial

I haven't thought about all the details of how this would work but I've had this bookmarked for a while

https://podcasters.apple.com/support/3143-how-to-set-the-order-of-podcast-episodes

batonac:
> I think it may be better to do what apple does and utilize the podcast type to determine the order. Episodic or serial

Gotcha. [It appears that the technical details are here.](https://help.apple.com/itc/podcasts_connect/#/itcb54353390:~:text=%3Citunes%3Atype%3E,type%20of%20show.). I can give it a try with Yoto.

batonac:
Hey @advplyr, I ran my own test using manually-injected channel-level `<itunes:type>serial</itunes:type>` and item-level `<itunes:episode>` tags into an audiobookshelf RSS feed. Unfortunately these don't seem to be supported by Yoto. That said, I did find another reasonable workaround in the Yoto app itself that I could just document for other Yoto users. So I suppose it does make sense to let this go and look towards implementing proper type and episode support in the future to cover the widest array of use-cases.

advplyr:
I hadn't really thought about it for RSS feeds since we are setting up RSS feeds similar to how itunes does. What I have wanted to do is have our web client and mobile app use the podcast type for sorting.
I'm not sure what you mean with manually-injected. The podcast type and episode/season # can be set inside Abs before opening the RSS feed.

advplyr:
Ah I just realized you were opening RSS feeds for audiobooks. For audiobooks I set podcast type to `serial` so that the order is based on the pub date for apps that are following itunes guidelines.

batonac:
Yeah the issue is that Yoto players don't respect that tag. I found a workaround I'm happy with though. 

The Yoto app pulls in the tracks in the correct order when you import a playlist from an RSS feed, but the playlist plays by date, and therefore in reverse, on the actual player. The track numbers also don't show on the display. 


If you edit the playlist on the Yoto app and reorder one of the tracks, save, and then return to the proper order and save again, both issues are resolved.

Issue #3556, opened 2024-10-28
[Enhancement]: 1.15x speed


DankMemeGuy:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Right now the lowest increase is 1.25x. Would be nice to have 0.15x

### Why would this be helpful?

1.25x is sometimes too much, but 1.0x is too slow. 1.15 is ideal. Many podcasting software use 1.15x option

### Future Implementation (Screenshot)

Screenshot isnt needed

### Audiobookshelf Server Version

2.16

glorenzen:
Could you clarify this a little bit? Do you want to be able to increment by 0.05 for the playback rate? Right now, you can increment the playback rate by .1. Are you saying 1.1x is too slow and 1.2x is too fast?

![image](https://github.com/user-attachments/assets/7f666450-b039-4057-a252-7bff2791f800)

DankMemeGuy:
> Could you clarify this a little bit? Do you want to be able to increment by 0.05 for the playback rate? Right now, you can increment the playback rate by .1. Are you saying 1.1x is too slow and 1.2x is too fast?
> 
[...]

PR #3892, opened 2025-01-25
Add adjustable increment and decrement value for playback rate

glorenzen:
This adds a new user setting that allows changing how much the playback rate increments or decrements (0.1 or 0.05).
Related to a request on #3556 

image


advplyr:
Thanks!

PR #3899, opened 2025-01-26
Allows setting of some pragma values through environment variables

### LINKED ISSUE #3750: [Enhancement]: Allow adjustment of SQLITE PRAGMA values
### Type of Enhancement

None

I would like to adjust some [PRAGMA](https://www.sqlite.org/pragma.html) values for the sqlite database.

My podcast library is fairly large and it takes approximately 20 seconds to load each "page" on the library view. Adjusting [mmap_size](https://www.sqlite.org/pragma.html#pragma_mmap_size) to `268435456` (256 megabytes) reduces the latency of the query to approximately 3 seconds per "page". My `absdatabase.sqlite` file is approximately 150 megabytes.

I've tested this via using the `sqlite3` command line which running the query before the change matches pretty closely the latency being reported by AudioBookShelf.

Optionally a list of values that would be able to be adjusted or a freeform text box allowing a list of values to be set that is run on every newly created connection. The majority of the PRAGMA values are specific to the connection and won't persist once the connection is closed.

v2.17.5

_No response_

mikiher:
## Brief summary

Adds support for setting some whitelisted SQLite PRAGMA values through environment variables.

## Which issue is fixed?

Fixes #3750

## In-depth Description

A whitelist of allowed pragma settings is defined,  which currently contains `mmap_size`, `cache_size`, and `temp_store`.
Users can set the values for those pragmas, by setting the environment variable `SQLITE_<UPPERCASE_PRAGMA_NAME>` (e.g. SQLITE_MMAP_SIZE) to the requested value.

These helped the reporting user optimize their database, and could be used by us in the future to set reasonable default values.

## How have you tested this?

Set up all the supported pragmas through the corresponding env variables.

### DIFF
--- a/server/Database.js
+++ b/server/Database.js
@@ -226,6 +226,28 @@ class Database {
 
     try {
       await this.sequelize.authenticate()
+
+      // Set SQLite pragmas from environment variables
+      const allowedPragmas = [
+        { name: 'mmap_size', env: 'SQLITE_MMAP_SIZE' },
+        { name: 'cache_size', env: 'SQLITE_CACHE_SIZE' },
+        { name: 'temp_store', env: 'SQLITE_TEMP_STORE' }
+      ]
+
+      for (const pragma of allowedPragmas) {
+        const value = process.env[pragma.env]
+        if (value !== undefined) {
+          try {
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #3906, opened 2025-01-28
server/podcast: stabilize random ID

tharvik:
## Brief summary

when two podcast's episode advertise the same filename, a random ID will get added to ensure that theses do not overlap.

## Which issue is fixed?

sadly, the generation of this random ID is redone everytime `PodcastEpisodeDownload.targetFilename` is accessed and so the filename is changing, making ffmpeg outputting to a different path than excepted.

## In-depth Description

there is already an `id` field in this class, which is, as far as my IDE told me, not used for anything else. it was probably meant to be used for that exact purpose.

## How have you tested this?

I actually found the issue when investigating why [this podcast](https://podcasts.apple.com/ch/podcast/id311604979) wasn't updating anymore. I tried to download two episodes with the same name and it failed with
```
DEBUG [FfmpegHelpers] downloadPodcastEpisode: Cmd: ffmpeg -i pipe:0 -y -loglevel debug -c:a copy -map 0:a -metadata podcast=1 [...]  -metadata podcast-type=episodic /[...]/Le 12h30 ‐ La 1ère/Le 12h30 - Présenté par Pauline Rappaz (0264cf14-702d-4277-a580-f3bda69e9c6c).mp3
ERROR [PodcastManager] Podcast Episode downloaded but failed to probe "/[...]/Le 12h30 ‐ La 1ère/Le 12h30 - Présenté par Pauline Rappaz (991253f7-0ad8-460c-a6f1-52174bee2288).mp3" No such file or directory
```
the changing ID gave it away.
then, I dwelled in code and tried out my patch and it worked.

haven't added a testcase, as I didn't saw an already existing test for episode downloads.

### DIFF
--- a/server/objects/PodcastEpisodeDownload.js
+++ b/server/objects/PodcastEpisodeDownload.js
@@ -74,7 +74,7 @@ class PodcastEpisodeDownload {
     return this.rssPodcastEpisode.title
   }
   get targetFilename() {
-    const appendage = this.appendRandomId ? ` (${uuidv4()})` : ''
+    const appendage = this.appendRandomId ? ` (${this.id})` : ''
     const filename = `${this.rssPodcastEpisode.title}${appendage}.${this.fileExtension}`
     return sanitizeFilename(filename)
   }


advplyr:
Thanks!

PR #3907, opened 2025-01-29
Add: workflow to close blank issues

nichwall:
## Brief summary

Adds a workflow to close issues which do not use the issue templates

## Which issue is fixed?

No issue number

## In-depth Description

To help encourage users to use the issue templates, this workflow closes any issues which are opened and have less than 3 headings in them. This does not prevent users from opening issues with their own headings, but should help prevent the blank issue template from being used.

## How have you tested this?

Tested with my issue templates repo

## Screenshots

Screenshot from 2025-01-28 20-25-27

### DIFF
--- a/.github/workflows/close_blank_issues.yaml
+++ b/.github/workflows/close_blank_issues.yaml
@@ -1,0 +1,42 @@
+name: Close Issues not using a template
+
+on:
+  issues:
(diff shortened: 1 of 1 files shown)


advplyr:
Awesome thanks!

PR #3929, opened 2025-02-04
Add resize to trix editor

### LINKED ISSUE #3928: [Bug]: Resize Handle for Descriptions is No Longer Visible
### What happened?

With the update to rich text (phenomenal addition btw), the resize handle for descriptions in the edit modal is no longer available. I believe this was a feature before the most recent update (2.19.0). Now that I'm going back through all of my books to update the descriptions, this feature is more necessary than ever! 

Resize handle should be available in the bottom right corner of the description box while editing books to make the text box larger or smaller. 

1. Click the edit button on a book
2. Check the bottom right corner of the description box where the resize handle was previously

### Audiobookshelf version

2.19.0

Windows Tray App

Windows

Chrome

### Logs

```shell

```

### Additional Notes

Tell me if I'm crazy and this feature was never there, I could very easily be misremembering.

mikiher:
## Brief summary

Adds a resize handle to the rich text editor.

## Which issue is fixed?

Fixes #3928 

## In-depth Description

- Added `resize: vertical` to text-editor style
- Set `height` and `min-height` properties (instead of `max-height`)

## How have you tested this?

Resize handle now shows up where the rich text editor appears, and works as expected.

### DIFF
--- a/client/components/ui/VueTrix.vue
+++ b/client/components/ui/VueTrix.vue
@@ -351,8 +351,10 @@ export default {
   background-color: white;
 }
 trix-editor {
-  max-height: calc(4 * 1lh);
+  height: calc(4 * 1lh);
+  min-height: calc(4 * 1lh);
   overflow-y: auto;
+  resize: vertical;
 }
 
 trix-editor * {


advplyr:
The white resize handle doesn't look great but it works. 

image

I remember a long time ago this used be an issue with Textarea but I think the browsers all use transparent handles now. I'm not sure what Trix is doing since there are no docs about it. It's not worth spending any additional time on right now.

Thanks!

mikiher:
I don't think trix is doing anything special. As far as I've seen this is the same as with the old textarea. 
Here's a screenshot from v2.18.1:
Screenshot 2025-02-05 080548

I tried tinkering with it for a little while but couldn't succeed in making it look right. I think it might have something to do with the non-standard scrollbar we use.

nichwall:
What browser are y'all using? From the referenced issue, the screenshot I shared was using Firefox and it does not have the little white box.

advplyr:
Ah yeah it only happens when the textarea is scrollable. If you remove all the contents it is good.
image

And this is the same with our plaintext textarea.

Issue #3038, opened 2024-06-03
[Bug]: Anthologies with many authors result in player controls being displaced


EloquenceWinters:
### What happened?

Whenever I have an anthology with a large-ish number of authors, the list of authors in the "player" strip at the bottom pushes the progress bar, chapter information, and controls down until they're no longer visible. This can be fixed by lowering the browser scale, but this isn't an optimal solution. 

![anthology](https://github.com/advplyr/audiobookshelf/assets/74223202/9ed9adf9-5d7a-47c9-92eb-e12af8ab9c50)

In the attached image, I've had to reduce the scale to 80% to even see the controls, and getting the progress bar and chapter information is impossible.

### What did you expect to happen?

The progress bar, chapter information, and controls should be static at the bottom of the player strip, so they are not lost when there are 10-20 authors in an anthology. 

### Steps to reproduce the issue

1. Add an anthology (e.g., The Big Book of Victorian Mysteries) with many authors (Otto Penzler, Waters, Wilkie Collins, Charles Dickens, James M'Govan, Eden Phillpotts, Moll Bourne, Dick Donovan, Arthur Conan Doyle, C. L. Pirkis, Max Pemberton, L. T. Meade, Clifford Halifax, Arthur Morrison, Headon Hill, George Griffith, M. McDonnell Bodkin, Fergus Hume, Emmuska Orczy, J. Sheridan Le Fanu, Catherine Crowe, Mary Elizabeth Braddon, Robert Louis Stevenson, Charles Reade, Thomas Hardy, Oscar Wilde, Robert Eustace, H. G. Wells, E. W. Hornungh, Grant Allen, George R. Sims, Guy Boothby, Pedro de Alarcón, Erckmann-Chatrian, Leo Tolstoy, Alfredo Oriani, Émile Gaboriau, Luigi Capuana, Guy de Maupassant, Anton Chekhov, Dietrich Theden, Edgar Allan Poe, Mark Twain, Ambrose Bierce, Frank Stockton, Melville Davisson Post, Anna Katharine Green, L. Frank Baum).
2. Try to play the book. 

### Audiobookshelf version

2.10.1

### How are you running audiobookshelf?

Windows Tray App

### What OS is your Audiobookshelf server hosted from?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Edge

### Additional Notes

[...]

PR #3944, opened 2025-02-06
prevent long author strings from pushing the player controls down by truncating

smithmd:
## Brief summary

Prevent long author strings from pushing the player controls down by truncating the text.

## Which issue is fixed?

There's no issue number. I just noticed this and thought it could be cleaned up.

## In-depth Description

I added the truncate class to the `div` containing the author information.

## How have you tested this?

I've tested the change in the browser to make sure it makes the player controls look better.

## Screenshots

Before:
image

After:
image

### DIFF
--- a/client/components/app/MediaPlayerContainer.vue
+++ b/client/components/app/MediaPlayerContainer.vue
@@ -11,9 +11,9 @@
           </nuxt-link>
           <widgets-explicit-indicator v-if="isExplicit" />
         </div>
-        <div class="text-gray-400 flex items-center w-1/2 sm:w-4/5 lg:w-2/5 truncate">
+        <div class="text-gray-400 flex items-center w-1/2 sm:w-4/5 lg:w-2/5">
           <span class="material-symbols text-sm">person</span>
-          <div v-if="podcastAuthor" class="pl-1 sm:pl-1.5 text-xs sm:text-base">{{ podcastAuthor }}</div>
+          <div v-if="podcastAuthor" class="pl-1 sm:pl-1.5 text-xs sm:text-base truncate">{{ podcastAuthor }}</div>
           <div v-else-if="authors.length" class="pl-1 sm:pl-1.5 text-xs sm:text-base truncate">
             <nuxt-link v-for="(author, index) in authors" :key="index" :to="`/author/${author.id}`" class="hover:underline">{{ author.name }}<span v-if="index < authors.length - 1">,&nbsp;</span></nuxt-link>
           </div>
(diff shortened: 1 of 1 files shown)


smithmd:
One thing I thought of, but don't know how to do the right way in Vue.js is that it could maybe use some hover text when the text is truncated.

client/components/app/MediaPlayerContainer.vue
```diff
@@ -11,7 +11,7 @@
           </nuxt-link>
           <widgets-explicit-indicator v-if="isExplicit" />
         </div>
-        <div class="text-gray-400 flex items-center w-1/2 sm:w-4/5 lg:w-2/5">
+        <div class="text-gray-400 flex items-center w-1/2 sm:w-4/5 lg:w-2/5 truncate">
           <span class="material-symbols text-sm">person</span>
           <div v-if="podcastAuthor" class="pl-1 sm:pl-1.5 text-xs sm:text-base">{{ podcastAuthor }}</div>
           <div v-else-if="authors.length" class="pl-1 sm:pl-1.5 text-xs sm:text-base truncate">
```
advplyr, line 17: truncate is already being added here and works for me in brave browser.
Were you running the latest server version or are you using a different browser?

Fixed in v2.11.0 as reported here: https://github.com/advplyr/audiobookshelf/issues/3038

smithmd: ah, maybe I hadn't updated yet. I'll double check that. But I had branched off of master and didn't see truncate in there.

smithmd: I still see the problem in firefox, and I've updated to the latest version: 
image

smithmd: I think it's worth noting that the fix you linked seems to have been for multiple authors, but this podcast only has one very long author. So truncating the author container fixes all possible author types being too long.

smithmd:
I just pushed perhaps a better fix to make it more specific to truncate the single author field, as I do see what you're saying about truncate being there for multiple authors.

client/components/app/MediaPlayerContainer.vue
smithmd: resolving this since it's not really relevant since my last push.

smithmd:
@advplyr any chance this can get another look? If you add the podcast Better Offline or Nocturno you can see what I'm saying about a single author string that breaks the layout of the player. This isn't related to multiple authors, just a very long single author. I fix it with the developer tools every time I open the web app.

advplyr:
Thanks. As a side note, I added those podcasts and neither of them have a long author for me. Maybe you imported from OPML and the data was weird?

smithmd:
> Thanks. As a side note, I added those podcasts and neither of them have a long author for me. Maybe you imported from OPML and the data was weird?

Could be. I did move my server by exporting from one machine and importing into the other. Maybe something weird happened there. I'll check my data.

Thanks again.

PR #3945, opened 2025-02-06
Adding support for 'minified' parameter when fetching a library's items

### LINKED ISSUE #2123: [Bug]: Endpoint /api/libraries/<id>/items always returns minified results
### Describe the issue

The `/api/libraries/<id>/items` endpoint always returns minified results, regardless of whether `?minified=0` is passed in the URL or not.

In the response, the `minified` key is correctly set, but the `results.media` items are always the minified versions of the podcast/book data.

```jsonc
{
    // ...
    "minified": true,
    "results": [
        {
            // ...
            "media": {...}  // <-- still minified!
        },
        // ...
    ]
}
```

1. Query `/api/libraries/<id>/items?minified=0`
2. Receive results with minified podcast/book `media` items.

### Audiobookshelf version

v2.4.3

Debian/PPA

r0adkll:
## Brief summary

This PR attempts to fix the issue where the `/api/libraries/<id>/items` endpoint only returns minified library items.

## Which issue is fixed?

Fixes #2123 

## In-depth Description

This endpoint currently just maps all library items `toOldJSONMinified()` regardless of the `minified` parameter sent in the request. This change checks the parameter and maps to either the minified, or the `toOldJSONExpanded()`.

## How have you tested this?

Tested by running locally through API test client with both response prior, and response after (with `minified=1`) and then with `minified=0`.

Also verified that change didn't affect web / mobile clients

## Screenshots

n/a

### DIFF
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -619,6 +619,7 @@ class LibraryController {
       filterBy: req.query.filter,
       mediaType: req.library.mediaType,
       minified: req.query.minified === '1',
+      expanded: req.query.expanded === '1',
       collapseseries: req.query.collapseseries === '1',
       include: include.join(',')
     }
--- a/server/models/LibraryItem.js
+++ b/server/models/LibraryItem.js
@@ -297,65 +297,43 @@ class LibraryItem extends Model {
     const { libraryItems, count } = await libraryFilters.getFilteredLibraryItems(library.id, user, options)
     Logger.debug(`Loaded ${libraryItems.length} of ${count} items for libary page in ${((Date.now() - start) / 1000).toFixed(2)}s`)
 
-    let items = []
-    if (minified) {
-      items = libraryItems.map((li) => {
-        const oldLibraryItem = li.toOldJSONMinified()
-        if (li.collapsedSeries) {
-          oldLibraryItem.collapsedSeries = li.collapsedSeries
-        }
-        if (li.series) {
-          oldLibraryItem.media.metadata.series = li.series
-        }
-        if (li.rssFeed) {
-          oldLibraryItem.rssFeed = li.rssFeed.toOldJSONMinified()
-        }
-        if (li.media.numEpisodes) {
-          oldLibraryItem.media.numEpisodes = li.media.numEpisodes
-        }
-        if (li.size && !oldLibraryItem.media.size) {
-          oldLibraryItem.media.size = li.size
-        }
-        if (li.numEpisodesIncomplete) {
-          oldLibraryItem.numEpisodesIncomplete = li.numEpisodesIncomplete
-        }
-        if (li.mediaItemShare) {
-          oldLibraryItem.mediaItemShare = li.mediaItemShare
-        }
-
-        return oldLibraryItem
-      })
-    } else {
-      items = libraryItems.map((li) => {
-        const oldLibraryItem = li.toOldJSONExpanded()
-        if (li.collapsedSeries) {
-          oldLibraryItem.collapsedSeries = li.collapsedSeries
-        }
-        if (li.series) {
-          oldLibraryItem.media.metadata.series = li.series
-        }
-        if (li.rssFeed) {
-          oldLibraryItem.rssFeed = li.rssFeed.toOldJSON()
-        }
-        if (li.media.numEpisodes) {
-          oldLibraryItem.media.numEpisodes = li.media.numEpisodes
-        }
-        if (li.size && !oldLibraryItem.media.size) {
-          oldLibraryItem.media.size = li.size
-        }
-        if (li.numEpisodesIncomplete) {
-          oldLibraryItem.numEpisodesIncomplete = li.numEpisodesIncomplete
-        }
-        if (li.mediaItemShare) {
-          oldLibraryItem.mediaItemShare = li.mediaItemShare
-        }
-
-        return oldLibraryItem
-      })
-    }
-
     return {
-      libraryItems: items,
+      libraryItems: libraryItems.map((li) => {
+        let oldLibraryItem = {}
+        if (minified) {
+          oldLibraryItem = li.toOldJSONMinified()
+        } else {
+          oldLibraryItem = li.toOldJSONExpanded()
+        }
+  
+        if (li.collapsedSeries) {
+          oldLibraryItem.collapsedSeries = li.collapsedSeries
+        }
+        if (li.series) {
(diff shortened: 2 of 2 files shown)


r0adkll:
@advplyr Hello 👋 Would it be possible to get a review of this PR?

Vito0912:
Just want to add that this could potentially break clients that currently use minified, because they think it's working correctly and then miss specific values from response.
This should be announced at least one version before imo

advplyr:
What data specifically are you needing that is not provided in minified? We will be improving the API at some point and I put in this "minified" concept in the API when it was a small project. Some users have hundreds of thousands of books and books with thousands of audio files, etc so fully expanded and fetching all items isn't practical in a real application

jbhul:
> Just want to add that this could potentially break clients that currently use minified, because they think it's working correctly and then miss specific values from response. This should be announced at least one version before imo

Clients that currently omit minified=1 as a URL param could definitely break if they are using content present in minified but not in the regular output (in the minified - regular - extended paradigm). If this fix were to be implemented a way to prevent downstream issues would be to default to minified. Obviously this wouldn't be consistent with other endpoints but just a thought.

---

The way this is discussed in the docs is "regular" is the default but in reality the default is minified presently and regular doesn't work (even if minified=0 is passed).

I agree with @advplyr that such a large output may not be needed and could present issues if a user has a very large library with this much information being sent in a single call. I do think that the API docs should be updated to include that minified is the default and only output option. I'm also guessing the user who did the PR might want/need the full output and adding an option, even with a warning that this may not be performant or require pagination, might be a good option too.

Vito0912:
The API documentation is currently not being maintained, is outdated, and more or less end-of-life.

It doesn’t make much sense to make major changes to it right now, since the backend is being overhauled anyway.

I don’t want to presume to decide whether it should be updated for this specific case, but updating the docs might send the wrong signal - it would suggest the documentation is current, which it’s not. Many endpoints are missing from the site, and some of the response and request information is outdated (as in missing, to much or just wrong). 

---

But the question regarding why one would need a non minified output outside of one item still stands. What are the reasons you would want such an output?

jbhul:
@Vito0912 Ah gotcha, didn't know that about the API docs. The docs themselves don't mention it and just say WIP which communicates the inverse of EOL lol.

In any case if they are EOL I'd think no reason to update to correct or notate, since it's one of many issues and there's no real reason to address this one while ignoring others.

---

I am not the original person who requested so I'd defer @deanishe from the [original issue](https://github.com/advplyr/audiobookshelf/issues/2123) or the PR creator.

Vito0912:
> Ah gotcha, didn't know that about the API docs. The docs themselves don't mention it and just say WIP which communicates the inverse of EOL lol.
I see. Yeah, you can go to the Repo and actually see a PR that removes minfied from the docs from 2024. Maybe considerable to add a notice @advplyr ? WIP indeed is a bit misleading.

r0adkll:
> But the question regarding why one would need a non minified output outside of one item still stands. What are the reasons you would want such an output?

@Vito0912 @advplyr My need in getting the full output was to provide a smoother UX in the client that I am building by having the more detailed information stored by the time a user opens a detail page for any given title. Rather, than hitting the detail page with minified information, then it updating when the network request finishes.  

The original driver behind this PR was a bug in my assumption that data would come back as expanded and accidentally wiping out the relational (i.e. expanded) data related to a given item. I've since fixed this assumption and the experience works as expected sans waiting on a network load.

I was unaware of 
> The API documentation is currently not being maintained, is outdated, and more or less end-of-life.

at the time of building this and the `minified` field still being part of the documentation and sought to restore the functionality here.

---

> Just want to add that this could potentially break clients that currently use minified, because they think it's working correctly and then miss specific values from response.

> Clients that currently omit minified=1 as a URL param could definitely break if they are using content present in minified but not in the regular output

This is a great callout / concern and given the existing usage and assumptions around this endpoint the better strategy would be to some sort of opt-in parameter for this (an `expanded = 1` maybe?) since `minified` isn't really a part of this request anyway. 

---

> Some users have hundreds of thousands of books and books with thousands of audio files, etc so fully expanding all items isn't practical in a real application

I would say that this is also true of the minified response as well given that level of volume. I believe that the expanded format has a place here especially with pagination, but I do agree that using the `minified` parameter (in a way that is default expanded) is not the best path for this change. If you are agreeable to enabling expanded support on this endpoint I can update this PR to instead use `expanded` as an opt-in parameter ensuring that we don't disrupt existing API usage.

deanishe:
> What data specifically are you needing that is not provided in minified?

Speaking as the author of #2123, the issue is syncing up a locally-cached copy of the data. Loading data view-by-view from the server ruins the experience of a native application.

When `minified=0` worked, syncing up meant pulling a few pages from the `/libraries/<ID>/items` endpoint. With minified results, it's necessary to pull each updated podcast individually, which may be many hundreds of requests.

> We will be improving the API at some point

Then I would ask you to give some consideration to applications that want to sync up a local database, not pull data for a specific view.

r0adkll:
Ah sorry for the delay on this but I have updated this PR to require an explicit `expanded = 1` request parameter making the return of expanded information an opt-in experience so as to not break any existing implementations that don't send a `minified = 1` 

@advplyr @Vito0912 If I could get your input once more as I believe I have addressed initial concerns and several others have chimed in on the use cases of returning expanded information in this endpoint.

Vito0912:
Thanks for pinging, but it's not up to me to decide. I am just a contributor.  

I can only share my opinion, and I already mentioned this before on Discord. Imo this should not be merged.

> Loading data view-by-view from the server ruins the experience of a native application.

!!!My opinion!!!:
Loading all expanded data should never be the ~~default~~ option for a client. Receiving all data really means ALL. In my opinion, this should never be a client task. 
A client should never need all the data. If it should work offline, you still would need the media files to actually use it. And then you can just also call the item point while downloading the book.
Having this as an easy option just leads to unnecessary network traffic, wasted cache space, and unused data.  
I have built tools that depend on expanded data, and if you truly need it, you can always fetch it using other endpoints (and I don't mean fetching items one by one). Especially for podcasts that would be a no-go. A single podcast can be hundreds of MB in size. Now, if multiple are loaded at once (this is of course an extreme example), it becomes a serious issue - and many people actually have thousands of podcast episodes.

But as I said, this is just my opinion. Adding this would only encourage clients to use this endpoint. And there are clients missinusing the API enough. Espeically due to cached data and overwriting other apps states.
!!!End!!!

Issue #3874, opened 2025-01-21
[Bug]: Subfolder path being forced on after update v2.18.0


DieselTech:
### What happened?

After the v2.18.0 update that added subdirectory support, ABS defaults to it being turned on for existing installs. 

![Image](https://github.com/user-attachments/assets/5d28533a-e737-4401-91fe-10f5acc61310)

Because it is defaulting on, it is causing problems with setups that are configured for a specific path. 

### What did you expect to happen?

Not to be forced into using subdirectories after the update

### Steps to reproduce the issue

1. Update to v2.18.0

### Audiobookshelf version

v2.18.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

mikelococo:
For anyone looking to come up to speed on the history of this feature:

* [Design discussion](https://github.com/advplyr/audiobookshelf/discussions/3535): Frustratingly, it appears that comments from sapd discussing why some users would likely not be happy about a mandatory subpath were deleted.
* [Github Issue and PR](https://github.com/advplyr/audiobookshelf/issues/385): The issue and PR have additional history and design discussion.
* v2.8.1 fixed two problems subpathing broke in v2.8.0:
  * Epub ebooks fail to load when using subdirectory https://github.com/advplyr/audiobookshelf/issues/3865
  * 404 not found for texture image when using subdirectory https://github.com/advplyr/audiobookshelf/issues/385 by @mikiher in https://github.com/advplyr/audiobookshelf/pull/3864

I haven't upgraded to v2.8.x yet because it felt like a strange choice to add a mandatory subpath rather retain consistency for existing installs via a disabled-by-default optional subpath and I was concerned about interactions with my proxy setup.

nichwall:
> [Design discussion](https://github.com/advplyr/audiobookshelf/discussions/3535): Frustratingly, it appears that comments from sapd discussing why some users would likely not be happy about a mandatory subpath were deleted.

The comments are still there. https://github.com/advplyr/audiobookshelf/discussions/3535#discussioncomment-10993141

> I haven't upgraded to v2.8.x yet because it felt like a strange choice to add a mandatory subpath rather retain consistency for existing installs via a disabled-by-default optional subpath and I was concerned about interactions with my proxy setup.

[...]

PR #3958, opened 2025-02-10
Fix `ROUTER_BASE_PATH` override for empty string

devnoname120:
## Brief summary

When the `ROUTER_BASE_PATH` env variable is set to an empty string it's mistakenly overriden to `/audiobookshelf` instead.
The `/audiobookshelf` fallback should only be used when the `ROUTER_BASE_PATH` env variable is undefined, not just an empty string.

Regression introduced in https://github.com/advplyr/audiobookshelf/pull/3810
See also: https://github.com/advplyr/audiobookshelf/pull/3810#discussion_r1948790937

## Which issue is fixed?

Related: https://github.com/advplyr/audiobookshelf/issues/3874 (requires frontend recompilation to fix that one)

## How have you tested this?

Running the following before starting Audiobookshelf:

```shell
export ROUTER_BASE_PATH=''
```

## Screenshots

N/A

### DIFF
--- a/client/nuxt.config.js
+++ b/client/nuxt.config.js
@@ -1,6 +1,6 @@
 const pkg = require('./package.json')
 
-const routerBasePath = process.env.ROUTER_BASE_PATH || '/audiobookshelf'
+const routerBasePath = process.env.ROUTER_BASE_PATH ?? '/audiobookshelf'
 const serverHostUrl = process.env.NODE_ENV === 'production' ? '' : 'http://localhost:3333'
 const serverPaths = ['api/', 'public/', 'hls/', 'auth/', 'feed/', 'status', 'login', 'logout', 'init']
 const proxy = Object.fromEntries(serverPaths.map((path) => [`${routerBasePath}/${path}`, { target: process.env.NODE_ENV !== 'production' ? serverHostUrl : '/' }]))
--- a/index.js
+++ b/index.js
@@ -29,7 +29,7 @@ if (isDev) {
   if (devEnv.AllowIframe) process.env.ALLOW_IFRAME = '1'
   if (devEnv.BackupPath) process.env.BACKUP_PATH = devEnv.BackupPath
   process.env.SOURCE = 'local'
-  process.env.ROUTER_BASE_PATH = devEnv.RouterBasePath || ''
+  process.env.ROUTER_BASE_PATH = devEnv.RouterBasePath ?? ''
 }
 
 const inputConfig = options.config ? Path.resolve(options.config) : null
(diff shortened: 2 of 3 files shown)


mikiher (approved):
Thanks, approved.

Note, though, that it's not enough to just set the environment variable and run the server - you have to rebuild the client (`npm run client`), as it is statically served.

mikiher:
Oh, and it definitely does not fix #3874, because of what I said above.

devnoname120:
@mikiher I just edited the commit message accordingly :)

devnoname120:
@mikiher Btw, what's the reasoning behind hard-coding the path in the frontend source code rather than setting the standard `<base>` HTML element at runtime?

mikiher:
I don't know if you've read discussion #3535. 

The reason we switched to `/audiobookshelf` by default was due to the current architectural constraints of having a statically built web client, and @advplyr request to support subdirectory while keeping a single docker/debian image. 

There has been quite some backlash to this from the discord dev channel, but @advplyr has decided to keep it this way.

devnoname120:
@mikiher Thank you for the link. I don't see [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) mentioned anywhere though. Was it considered?

mikiher:
> @mikiher Btw, what's the reasoning behind hard-coding the path in the frontend source code rather than setting the standard `<base>` HTML element at runtime?

I haven't considered using `<base>`, but note that our code doesn't currently support server side rendering - I can only serve static html and Javascript. Also, there are other things beside html links that need to be considered for subdirectory support (Javascript fetch requests etc.).

If you have more comments/questions on the design, let's discuss in #3535.

mikiher:
> **Edit 2**: this merged PR adds support for runtime `base` and `publicPath`: [nuxt/nuxt#8520](https://github.com/nuxt/nuxt/pull/8520)

Note in the caveats:
> This does not affect statically generated sites...

Our setup is a statically generated site.

advplyr:
Even though you may not be utilizing this given the discussion in #3535 it makes sense to merge incase anyone wants to build from source

Vito0912:
Just validated. Still works if ENV is passed correctly. @aidivas Please make sure that the process does have the ENV as context.
If you want a simple and command that works on all systems use something like: https://www.npmjs.com/package/cross-env.

aidivas:
Thanks for replying.Setting env in source doesn’t change anything in my unix or windows tests. I’ve hardcoded the empty spaces into the relevant files so all good now.

PR #3980, opened 2025-02-14
Fix count cache by stringify Symbols #3979

advplyr:
## Brief summary

Book library queries are returning the incorrect count because the cached key is matching when it shouldn't. 

Sequelizes uses Symbols in the queries and JSON.stringify ignores these. This adds a separate function that sanitizes the query before Stringify.

## Which issue is fixed?

Fixes #3979

## In-depth Description

Reproduce by filtering by Issues and clearing the filter. Those 2 cache keys will be the same because the library item where option is not stringified.

PR #3993, opened 2025-02-15
fix stringifySequelizeQuery and add tests

mikiher:
## Brief summary

I believe the implementation of stringifySequelizeQuery in PR #3980 was buggy (it didn't pass the test I wrote).

## Which issue is fixed?

Same as PR #3980

## In-depth Description

I wrote a slightly simpler implementation using by implementing a `replacer` function which you can pass to `JSON.stringify`
The replacer function deals with two values that are ignored by JSON.stringify by default:
- Symbols
- Classes (mostly for e.g. `model: Database.libraryItemModel`)

## How have you tested this?

Wrote a unit test, and made sure the correct key is generated for real queries in the code.


advplyr:
Thanks!

PR #3952, opened 2025-02-08
Improve book library page query performance on title, titleIgnorePrefix, and addedAt sort orders.

mikiher:
## Brief summary

Significantly improves book library Sequelize page queries for the following sort orders:
- title (with or without ignorePrefix)
- addedAt

## Which issue is fixed?

This partially fixes #2073 (resolving the book library load times, but not the podcast library load times)

## In-depth Description

After digging more into the details of the issues people were complaining about in #2073 and doing additional performance analysis in Sequelize query to bring the page [here](https://github.com/advplyr/audiobookshelf/blob/25ae6dd59a06f9f6f62c97d98ead908b23d5bc37/server/utils/queries/libraryItemsBookFilters.js#L563), I made the following observations:
1. Due the the current architecture of the database (specifically, the separate libraryItems and books tables), SQLite cannot sort the results of the main query efficiently, even when an index on the sorted book column exists (more details below)
2. The main query performs a join with the feeds table, which somehat complicates the query (especialy if there are many feeds)
3. A query for the total count of results is run every time, even though it is usually unnecessary as the total count doesn't change during a scrolling over the library.

1 is by far the most serious problem, and also causes significant degradation in query performance as the offset becomes larger. 
When the main query is sorting by title:
```
SELECT book.*, libraryItem.*, [libraryItem->feeds].*
  FROM books AS book
       INNER JOIN
       libraryItems AS libraryItem ON book.id = libraryItem.mediaId AND 
                                      (libraryItem.libraryId = 'b3bf8580-edba-4d7d-90eb-3b54206d16c3' AND 
                                       libraryItem.mediaType = 'book') 
       LEFT OUTER JOIN
       feeds AS [libraryItem->feeds] ON libraryItem.id = [libraryItem->feeds].entityId AND 
                                        [libraryItem->feeds].entityType = 'libraryItem'
 ORDER BY book.title ASC
 LIMIT 0, 35;
```
it's evident in the query plan that the query engine cannot use the existing book.title index, and needs to build a temporary tree for sorting.
```
SEARCH libraryItem USING INDEX library_items_library_id_media_type_size (libraryId=? AND mediaType=?)
SEARCH book USING INDEX sqlite_autoindex_books_1 (id=?)
SCAN libraryItem->feeds LEFT-JOIN
USE TEMP B-TREE FOR ORDER BY
```

Even when you remove the `feeds` table join from the query:
```
SELECT book.*, libraryItem.*
  FROM books AS book
       INNER JOIN
[...]

PR #3996, opened 2025-02-16
Improve podcast library page query performance on title, titleIgnorePrefix, and addedAt sort orders

### LINKED ISSUE #3965: [Bug]: Podcast library page load time is very high
### What happened?

I am opening this as a sub-issue of #2073 (in which the general complaint was about slow loading library items post v2.4.0).

This issue should specifically address **podcast library page load** performance problems.

People who complained about have Audiobookshelf setups with the following characteristics:
- Large podcast libraries (hundreds of podcasts, tens of thousands of episodes)
- One user reports running on docker on relatively weak hardware (Synology 720+)
- Load times go as high as 20-40 seconds per page load
- Similar or higher load times are reported when scrolling down the library

Average and mean page load time for a typical page (~50 podcast per-page) should not exceed 200 ms.

1. Open a large podcast library (100s of podcasts, 10000s of podcast episodes).
2. Click on the Library page
3. Page load slowly
4. Scroll down
5. Subsequent page also load slowly

### Audiobookshelf version

v2.19.0

Docker

### Logs

```shell

```

### Additional Notes

Reported on Synology hardware.

Not browser-specific.

mikiher:
## Brief summary

This PR tries to optimize some of the podcast library page load and scrolling database queries, following up on what's been done for the book library in #3952

## Which issue is fixed?

Fixes #3965 

## In-depth Description

Podcast library page queries are more complex than book library page queries, because they aggregate data from podcast episodes, specifically `numEpisodes` (the number of episodes that the podcast has), and `numEpisodesIncomplete` (the number of episodes that the current user has not yet finished listening to).

Before this change, these two virtual columns were obtained via per-row subqueries, making the podcast queries very inefficient, leading to very high library page query latency on large podcast libraries. In addition, the queries suffered from the same main issue that caused high latency in the book library page queries, namely the separation between the `libraryItems and `podcasts` tables.

### Resolution
The following changes affecting the main podcast library page query were made:
- Similar changes to the ones made for books in #3952 were implemented for podcasts
   - Copied the title/titleIgnorePrefix columns to `libraryItems`
   - Set the title/titleIgnorePrefix when a podcast library item is created
   - Added triggers to update those when the original columns change.
   - Separated the feed query
   - Added a count cache in `libraryItemsPodcastFilters`
   - Added environment-variable controlled profilling
- Added a `numEpisodes` column to the `podcasts` table
   - Populated during database migration
   - Kept updated as podcast episodes are added and removed
   - The per-row subquery for `numEpisodes` could now be removed
- Added a `podcastId` column to the `mediaProgresses` table
   - Populated during database migration
   - Set when the mediaProgress record is created
   - The per-row subquery for `numEpisodesIncomplete` could now be removed
      - Instead, `numEpisodesComplete` is calculated in-memory using the cached user record
      - `numEpisodesIncomplete = numEpisodes - numEpisodesComplete`
   - The cached user record is updated whenever mediaProgress is created or destroyed

In addition, a couple of other small changes were implemented:
- Added environment-variable controlled profilling for podcast episodes queries (used mainly in the podcast library homepage)
- Added an `ANALYZE` database query at database initialization
  - This query updates statistics about the database tables, indices, and their sizes
  - It's recommended to run it mostly after large bulk updates and schema changes
  - Usually takes no more than a second to run, so opted to run it during server startup.
     - After scanning a very large podcast library, running it caused the query plan for some of the database queries to change, improving those queries significantly

### Performance

The issues described above tend to be more noticable in large podcast libraries, so I tested this on a synthetic podcast library containing 1000 podcast and 130,000 podcast episodes. Like in the previous PR, I ran tests on a Synology 920+ NAS, which has a relatively weak hardware.

No non-default pragma values were applied.

I tested on the same sort orders as in #3952
- title
- titleIgnorePrefix
- addedAt

Other sort orders were not optimized.

I'll post more detailed results later, but the overall effect on latency is very large:
- When podcast library page is sorted by title:
   - Before this PR (edge image) - the first few 35-podcast-per-page page took 15-20 seconds per page to load. the next few took more than 2 minutes per-page.
   - After the change - First 35-podcast-per-page page took ~150 milliseconds to load. the next few pages took between 50-150 milliseconds.

This is an overall drop of ~99%  or more in podcast library page load time (roughly similar to what we see in #3952).

### Correctness

- Migration unit tested and tested on a real database
- Played a lot with podcast, podcast episode, and media progress addition and removal
  - Made sure the new column values are updated correctly
  - Made sure count cache is invalidated as expected
  - Made sure `numPodcasts` and `numPodcastsIncomplete` are properly calculated/updaed.

### DIFF
--- a/server/utils/queries/libraryItemsPodcastFilters.js
+++ b/server/utils/queries/libraryItemsPodcastFilters.js
@@ -84,9 +84,9 @@ module.exports = {
       return [[Sequelize.literal(`\`podcast\`.\`author\` COLLATE NOCASE ${nullDir}`)]]
     } else if (sortBy === 'media.metadata.title') {
       if (global.ServerSettings.sortingIgnorePrefix) {
-        return [[Sequelize.literal('`podcast`.`titleIgnorePrefix` COLLATE NOCASE'), dir]]
+        return [[Sequelize.literal('`libraryItem`.`titleIgnorePrefix` COLLATE NOCASE'), dir]]
       } else {
-        return [[Sequelize.literal('`podcast`.`title` COLLATE NOCASE'), dir]]
+        return [[Sequelize.literal('`libraryItem`.`title` COLLATE NOCASE'), dir]]
       }
     } else if (sortBy === 'media.numTracks') {
       return [['numEpisodes', dir]]
--- a/server/utils/queries/libraryItemsPodcastFilters.js
+++ b/server/utils/queries/libraryItemsPodcastFilters.js
@@ -1,6 +1,9 @@
 const Sequelize = require('sequelize')
 const Database = require('../../Database')
 const Logger = require('../../Logger')
+const stringifySequelizeQuery = require('../stringifySequelizeQuery')
+
+const countCache = new Map()
 
 module.exports = {
   /**
@@ -96,6 +99,28 @@ module.exports = {
     return []
   },
 
+  clearCountCache() {
+    countCache.clear()
+  },
+
+  async findAndCountAll(findOptions, model, limit, offset) {
+    const cacheKey = stringifySequelizeQuery(findOptions)
+    if (!countCache.has(cacheKey)) {
+      const count = await model.count(findOptions)
+      countCache.set(cacheKey, count)
+    }
+
+    findOptions.limit = limit
+    findOptions.offset = offset
+
+    const rows = await model.findAll(findOptions)
+
+    return {
+      rows,
+      count: countCache.get(cacheKey)
+    }
+  },
+
   /**
    * Get library items for podcast media type using filter and sort
    * @param {string} libraryId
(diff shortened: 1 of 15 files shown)


revilo951:
Thanks! This is epic, well done!

D3r3k23:
This made a huge improvement with my library. Thanks!

server/utils/queries/libraryItemsPodcastFilters.js
```diff
@@ -169,10 +194,12 @@ module.exports = {
         }
       ],
       order: this.getOrder(sortBy, sortDesc),
-      subQuery: false,
-      limit: limit || null,
```
TunaLobster, line 173: I think this change broke some API behavior. A few days ago I was able to make a get library items call (GET https://abs.example.com/api/libraries/<ID>/items) with the limit parameter set to null and get back a full list of all of the library items in a single call (albeit this might not be a great default behavior for libraries with 1000s of items). Now I have to provide a non-zero value for limit to get a response on the API from the same call. I'm far from being versed in reading JS so I might be in the wrong spot. This line stood out to me in the recent changes when it stopped working.

mikiher: Thanks for reporting. I'll look into this.

mikiher: Ah, it seems like this is a duplicate of #4014 and @advplyr already fixed that.

PR #4005, opened 2025-02-19
Add title triggers in new databases

### LINKED ISSUE #4003: [Bug]: Sort books by title does not work properly for new databases
### What happened?

After a new database is created with the latest version (v2.19.3), changing the title of books does not affect their order on a title-sorted library page.

When a book title is changed, its position should changed according to its new title.

1. Create a new database
2. Create a book library and scan it
3. Open the book library page
4. Change to sort by "Title"
5. Edit the last book
6. Change the title so that its position in the title sort order should change to first
7. Click Save & Close
8. Reload the library page

Expected: the edited book's position should change from last to first.
Actual: the edited book's position remains at it's original position.

### Audiobookshelf version

v2.19.3

Windows Tray App

Windows

None

### Logs

```shell

```

### Additional Notes

_No response_

mikiher:
## Brief summary

Adds the new title[IgnorePrefix] update triggers (introduced in v2.19.1 and v2.19.4 migrations) if they don't exist, to handle new databases (for which migrations do not run).

## Which issue is fixed?

Fixes #4003 

## In-depth Description

When triggers were added in migrations v2.19.1 and v2.19.4, I forgot to create those triggers in new databases, for which migrations don't run. Therefore, the new libraryItems columns title and titleIgnorePrefix were not updaed after a title was changed in a book or podcast, and the sort order wasn't updated.

I added code in Database.js to create the triggers if they don't exist.

## How have you tested this?

Made sure that books/podcasts were correctly sorted after a title change in a newly created database (using edge code).

### DIFF
--- a/server/Database.js
+++ b/server/Database.js
@@ -190,6 +190,8 @@ class Database {
     await this.buildModels(force)
     Logger.info(`[Database] Db initialized with models:`, Object.keys(this.sequelize.models).join(', '))
 
+    await this.addTriggers()
+
     await this.loadData()
 
     Logger.info(`[Database] running ANALYZE`)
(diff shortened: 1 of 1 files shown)


mikiher:
Note: v2.15.0 migration test is still failing sporadically.

I don't think this has anything to do with my change.

advplyr:
Thanks!

PR #4008, opened 2025-02-19
Re-sort title-sorted bookshelf after title change

### LINKED ISSUE #3998: [Bug]: When sort order is "Title", changing the title of a book does not cause it to relocate to its proper position
### What happened?

When I set the sort order to title, then change a book's title and save, the book remains at the same location on the library page.

The book should have changed its position according to its title.

1. Set sort order to "Title" (ascending or descending)
2. Open a book details
3. Change the title
4. Click "Save & Close"

Expected: the book moves to a new position according to the new title
Actual: the book stays at the same position

### Audiobookshelf version

v2.19.3

Built from source

Windows

Chrome

### Logs

```shell

```

### Additional Notes

_No response_

mikiher:
## Brief summary

When a bookshelf is sorted by title and the title of a book/podcast changes (e.g. after being edited), the bookshelf is rebuilt so the book appears at its proper position (based on its new title)

## Which issue is fixed?

Fixes #3998

## In-depth Description

Up until now, when `LazyBookshelf` got an `item_updated` message from the server, it only replaced the old entity with the updated entity. With this change, if the current sort order is by title, and a change of title is detected, `resetEntities()` is called, and the item is moved to its new position.

## How have you tested this?

Changed title of books/podcasts and verified that after saving the change, the book/podcast moved to its new position.

### DIFF
--- a/client/components/app/LazyBookshelf.vue
+++ b/client/components/app/LazyBookshelf.vue
@@ -419,7 +419,7 @@ export default {
 
       this.postScrollTimeout = setTimeout(this.postScroll, 500)
     },
-    async resetEntities() {
+    async resetEntities(scrollPositionToRestore) {
       if (this.isFetchingEntities) {
         this.pendingReset = true
         return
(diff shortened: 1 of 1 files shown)


advplyr:
I made a small update so that the scroll position is kept for title change re-sorts. I figured because we know the page size will be the same on an update we might as well keep the scroll pos.

Thanks!

PR #4020, opened 2025-02-21
Invalidate count cache on entity update

### LINKED ISSUE #4018: [Bug]: When filtering by Missing ASIN, Updaing books with ASIN causes blank entries to appear at the end
### What happened?

When setting the filter to Missing ASIN, then updating the ASIN for one or few books, the new books are dropped the list (only after a page reload), but the count at the top is not updated, and blank entries appear at the end.

Image

Ony after server restart the count gets correct and the blank entries disappear.

Expected count to change and no blank entries to appear.

1. Open a book library
2. Go to Library page
3. Filter by Missing ASIN
4. Edit a book
5. Fill in the ASIN
6. Click Save & Close
7. **The book does not disappear from the list**
8. Reload the page
9. The book disappears, but the count does not change and now a blank entry appears in the end.

Expected: Book count changes immediately, and no blanks entries appear

### Audiobookshelf version

v2.19.4

Built from source

Windows

Chrome

### Logs

```shell

```

### Additional Notes

_No response_

mikiher:
## Brief summary

Invalidates the count cache in libraryItems(Book|Podcast)Filters when a book/podcast/episode is updated.

## Which issue is fixed?

Fixes #4018 partially

## In-depth Description

Up until now, the cache count was invalidated only when an item was created or destroyed. However, when a filter is applied to the library page and an item metadata is updated, the number of items returned by that filter might change. It's not easy to identify which of the cached counts change or not, so the whole count cache needs to be invalidated in this case as well.

I added hooks to invalidate the count cache after book/podcast/episode updates.

_Note:_ an immediate update of the count and removal/addition of the updated book in the filtered view is preferred, but it is hard to identify when such a change requires `resetEntities()` to be called without actually running the count query again. So I left the client code as is (it is not a regression). That's why I said above that if fixes the bug partially

## How have you tested this?

Set a filter on the library page, changed an item's metadata so that it should affect the filtered items count, and made sure that the count changes after the library page is reloaded, and no blank entries appear at the end.

Did this for both podcast and book libraries.

### DIFF
--- a/server/models/Book.js
+++ b/server/models/Book.js
@@ -201,10 +201,6 @@ class Book extends Model {
     Book.addHook('afterCreate', async (instance) => {
       libraryItemsBookFilters.clearCountCache('afterCreate')
     })
-
-    Book.addHook('afterUpdate', async (instance) => {
-      libraryItemsBookFilters.clearCountCache('afterUpdate')
-    })
   }
 
   /**
--- a/server/models/Book.js
+++ b/server/models/Book.js
@@ -201,6 +201,10 @@ class Book extends Model {
     Book.addHook('afterCreate', async (instance) => {
       libraryItemsBookFilters.clearCountCache('afterCreate')
     })
+
+    Book.addHook('afterUpdate', async (instance) => {
+      libraryItemsBookFilters.clearCountCache('afterUpdate')
+    })
   }
 
   /**
--- a/server/models/Podcast.js
+++ b/server/models/Podcast.js
@@ -157,6 +157,10 @@ class Podcast extends Model {
     Podcast.addHook('afterCreate', async (instance) => {
       libraryItemsPodcastFilters.clearCountCache('podcast', 'afterCreate')
     })
+
+    Podcast.addHook('afterUpdate', async (instance) => {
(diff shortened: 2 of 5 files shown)


advplyr:
I found a few other places where the cache needs to be invalidated.

1. Filter by series, edit a book and remove a series.
2. Filter by author, edit a book and remove an author.
3. Filter by "Share Open", click into a book and delete the share.

A separate issue I found is filtering by "RSS Feed Open" doesn't work anymore. I believe it has to do with the addition of `separate: true` to the feedModel query in #3952. The `required` property wouldn't work anymore.

advplyr:
I think it makes sense to only use `separate: true` when the feed is not required so I made that change and it is working.

This adds another place where we need to invalidate the cache though.

3. Filter by "RSS Feed Open", click into a book and close the RSS Feed

mikiher:
Thanks for spotting the issues. I'm on it. 

With all the different cases, it may be safer not to cache any counts where filterGroup != null

advplyr:
I just realized it may also occur with media progress filters but haven't tested.

We could start with skipping the count cache for those specific filters

mikiher:
OK, I changed it so the count cache is not used when there's a filter.

advplyr:
Thanks!

PR #4030, opened 2025-02-22
new sort option for podcasts view (-> sort by filename)

### LINKED ISSUE #3637: [Enhancement]: Podcast Episode Table Sort by Filename
### Type of Enhancement

Web Interface/Frontend

In addition to pub date, title, season, and episode, there should be an option to sort by filename.

I have a handful of podcasts in ABS, and almost none of them sort right currently. Some don't sort right under any of the available options. Podcasts are not a very standardized medium, and there's wide variation in how podcast hosts choose to upload their files. Taking a few examples from my collection:
- One podcast was uploaded by the creators with no metadata in some episodes; one episode is missing the episode number tag; the title field has different spelling in one episode, and another with a totally different title format. I downloaded the files from their RSS feed, so the file dates are all the same. None of the current sort options in ABS work correctly here.
- One has title tags in completely different formats across different platforms and episodes. All of the files are listed as episode 1. Again, nothing works here!
- Another similarly has every episode listed as episode 1, and also has some episodes whose title tags look like "Episode X: Title" and some that just look like "Title".

It's very easy for a user to fix the filenames if they need fixing. It's more annoying to fix the metadata, and in some cases that might not be desirable anyway. I think most people who download lots of media are going to standardize the filenames anyway, so sorting that way is much less effort.

image

v2.17.2 windows

image

4ch1m:
## Brief summary

Adds a new sorting option (-> "Filename") to the podcasts view/table.

## Which issue is fixed?

Fixes #3637

## In-depth Description

Locally stored podcast files often don't contain proper meta-data ("published date", "episode number", etc.); which is needed for reasonable sorting.
 
However, files that contain suitable name-prefixes (e.g. "20250213-podcast-xyz.mp3") can provide the necessary information that way.

## How have you tested this?

Tested via dev-container setup.

## Screenshots

image

### DIFF
--- a/client/components/tables/podcast/LazyEpisodeRow.vue
+++ b/client/components/tables/podcast/LazyEpisodeRow.vue
@@ -10,8 +10,13 @@
         <div class="h-10 flex items-center mt-1.5 mb-0.5 overflow-hidden">
           <p class="text-sm text-gray-200 line-clamp-2" v-html="episodeSubtitle"></p>
         </div>
+
         <div class="h-8 flex items-center">
-          <div class="w-full inline-flex justify-between max-w-xl">
+          <p v-if="sortKey === 'audioFile.metadata.filename'" class="text-sm text-gray-300 truncate font-light">
+            <strong className="font-bold">{{ $strings.LabelFilename }}</strong
+            >: {{ episode.audioFile.metadata.filename }}
+          </p>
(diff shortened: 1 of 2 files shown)


advplyr:
I updated the episode row to show the filename if it is being used for sorting.
image

Thanks!

Issue #2399, opened 2023-12-08
[Enhancement]: Ignore/remove temp files from database


nichwall:
### Describe the feature/enhancement

Similar to how the scanner and watcher ignores `.part` files (https://github.com/advplyr/audiobookshelf/issues/2063), also ignore temporary files (common ones end with `.tmp`, `.temp`, or `~`)

Related to https://github.com/advplyr/audiobookshelf/issues/1641

Issue #4029, opened 2025-02-22
[Enhancement]: avoid scanning / monitoring temporary files


aureq:
### Type of Enhancement

Server Backend

### Describe the Feature/Enhancement

Hi,

I just recently installed audiobookshelf and it works like a charm 🌿.

I'm now using [`podcast-dl`](https://github.com/lightpohl/podcast-dl) to download my podcasts and have them available in audiobookshelf.

I noticed though, audiobookshelf tracks/monitors files ✅ but also temporary files (including `*.tmp`) that are temporarily created by `podcast-dl`. This leads to plenty of errors in the logs such as the ones provided below.

```
Failed to get ino for path /podcasts/australia/Better Off Dead/1-20151105-Welcome.jpg.tmp [Error: ENOENT: no such file or directory, stat '/podcasts/australia/Better Off Dead/1-20151105-Welcome.jpg.tmp'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/podcasts/australia/Better Off Dead/1-20151105-Welcome.jpg.tmp'
}
ERROR: [Utils] Failed to get ino for path /podcasts/australia/Better Off Dead/3-20160215-#2 How dare you want to end your life_ Liz's story.mp3.tmp [Error: ENOENT: no such file or directory, stat '/podcasts/australia/Better Off Dead/3-20160215-#2 How dare you want to end your life_ Liz's story.mp3.tmp'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: "/podcasts/australia/Better Off Dead/3-20160215-#2 How dare you want to end your life_ Liz's story.mp3.tmp"
}
```

### Why would this be helpful?

This would remove spurious log entries shown from the console and allow other errors to be visible.

### Future Implementation (Screenshot)

n/a

### Audiobookshelf Server Version

v2.19.4

### Current Implementation (Screenshot)

This is a backend related feature request

nichwall:
Duplicate of https://github.com/advplyr/audiobookshelf/issues/1641

PR #4031, opened 2025-02-22
Refactor ignore file logic

nichwall:
## Brief summary

This PR refactors the "file ignore" for the scanner and watcher.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/2399 and https://github.com/advplyr/audiobookshelf/issues/4029
This helps with https://github.com/advplyr/audiobookshelf/issues/1641, but is not user-configurable like that request is.

## In-depth Description

This adds additional file extensions to the ignore list in the scanner and watcher, and creates a new utility function to ensure both ignore lists stay in sync.

Ignored file extensions are:
- `.part`
- `.tmp`
- `.temp`
- `.crdownload`
- `.download`
- `.bak`
- `.old`
- `.tempfile`
- `.tempfile~`

This function also checks for dotfiles and dotpaths to simplify the calling functions.

This function adds a framework to ignore the file or path whenever a specific string appears anywhere in the relative path. This is to help address vendor-specific paths like `@eaDir` from Synology.

This PR does *not* do anything to the database, it only affects the scanner and watcher.

## How have you tested this?

Created a new library with each of the ignored extensions and using the ignored extension as the "filename" and a valid extension such as `crdownload.mp3` to ensure the filter applies correctly. I also created an `@eaDir` with valid media files to ensure they were ignored correctly.

I then enabled the watcher and performed the same test in a new directory.

## Screenshots

N/A

### DIFF
--- a/server/Watcher.js
+++ b/server/Watcher.js
@@ -5,7 +5,7 @@ const Logger = require('./Logger')
 const Task = require('./objects/Task')
 const TaskManager = require('./managers/TaskManager')
 
-const { filePathToPOSIX, isSameOrSubPath, getFileMTimeMs } = require('./utils/fileUtils')
+const { filePathToPOSIX, isSameOrSubPath, getFileMTimeMs, shouldIgnoreFile } = require('./utils/fileUtils')
 
 /**
  * @typedef PendingFileUpdate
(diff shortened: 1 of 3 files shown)


advplyr:
Thanks! When adding the test file I thought it made sense to put the extname check as the lest step of shouldIgnoreFile.

nichwall:
Ah, yep that makes sense. I didn't have the checks in any particular order. Thanks for updating that!

Issue #4004, opened 2025-02-19
[Bug]: A ; seems to be being added to the custom metadata search


alexander-paterson:
### What happened?

I'm using abs-tract to add in extra metadata providers. The search being sent to abs-tract appears to have a `;` before the word query.

I created a but report in [abs-tract's GitHub](https://github.com/ahobsonsayers/abs-tract/issues/19) and the dev said it appears to be an ABS issue and to create a bug here.

### What did you expect to happen?

After adding a metadata provider through abs-tract I should be able to get search results.

### Steps to reproduce the issue

1. Create a metadata provider pointing to abs-tract
2. Perform a search, no results are returned
3. Check the logs and you can see a `;` before the word query

### Audiobookshelf version

2.19.3

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
2025-02-05T21:34:37.079432255Z INFO Response: 200 OK service: "abs-tract" httpRequest: {url: "http://192.168.8.219:5555/kindle/us/search?mediaType=book&;query=Die+Trying&author=Lee+Child" method: "GET" path: "/kindle/us/search" remoteIP: "172.23.0.1:59900" proto: "HTTP/1.1" requestID: "5c2b925ac873/rZ2rQH0VgA-000001" scheme: "http" header: {accept: "application/json, text/plain, */*" user-agent: "axios/0.27.2" connection: "keep-alive"}} httpResponse: {status: 200 bytes: 3511 elapsed: 3524.220387}
```

advplyr:
I'm not able to reproduce this. Do you get this result when you use other search text?

alexander-paterson:
Okay, this is weird.

I have Goodreads, Kindle US and Kindle UK set up in abs-tract

Looking at the logs, they all have the same `;` in the URL sent to abs-tract

```
[...]

PR #4034, opened 2025-02-22
Add: log custom metadata provider to match other providers

nichwall:
## Brief summary

This PR adds logging for the Custom Metadata Provider URL request to match other providers.

## Which issue is fixed?

Additional logging for https://github.com/advplyr/audiobookshelf/issues/4004

## In-depth Description

Moved URL creation with query string to be before the Axios request and added logging of the generated URL.

## How have you tested this?

Verified not crashing

## Screenshots

N/A


advplyr:
Thanks!

Issue #1321, opened 2022-12-27
[Enhancement]: Podcast episode order (oldest plays first)


raubreywhite:
### Describe the feature/enhancement

In many serial podcasts the listening order is oldest to newest. At the moment, the only playing order is newest to oldest.

It would be good if we could choose on a per-podcast level if we want to play oldest-to-newest, or newest-to-oldest.

<img width="1075" alt="image" src="https://user-images.githubusercontent.com/1532091/209639975-f1dfd5cb-f59f-4b0c-9196-3233d7064cd0.png">

Thank you for audiobookshelf, it is an amazing piece of work!

rudism:
Just want to re-surface [this suggestion](https://github.com/advplyr/audiobookshelf/issues/980#issuecomment-1718432707) by @gbakeman from one of the closed dupe issues. A nice stop-gap here might be for the episode sort drop down to "remember" the last setting for the current user and podcast.

Right now when I want to listen to a podcast in reverse/serial order (on the web or in the app) my process is to go to that podcast's episode listing, click the sort drop down, then reverse the date sort option to be ascending, then start the top episode. If the change to the sort order would "stick" for that podcast it would reduce a couple clicks and make it a bit less cumbersome--even if no other changes were made to the queuing system or "next episode" behavior from other screens or whatever else might make this a more complex issue to address in a more comprehensive way.

hadleyrich:
As above even just the ability to remember last sort would be super useful to me too :)

mathmaniac43:
This may warrant its own issue/discussion, but I would also like a way to download the podcasts in chronological order based on my completion progress. Currently auto-downloads only seem to get new episodes, and I am having trouble even manually targeting older episodes for download.

andranoxi:
I'm catching up on a long running podcast, it would be nice to tap "Play" on the last episode I was listening to on the homepage, and for it to automatically play the next episode.

Or clicking "Play" from the top of a podcast and for it to play the oldest unplayed episode instead of the latest. This should be the default behavior for serialized podcasts.

xFaultx:
Even if a default setting per rss feed would be great. Like setting one to always play oldest first. Honestly there anrent any podcasts I listen to the newest episode first, I always start from the beginning

golgoth85:
+1 for to remembering last sort order, or choose a different default order in ABS settings

advplyr:
[...]

PR #4035, opened 2025-02-23
Play first podcast episode in table

nichwall:
## Brief summary

This PR changes the "Play" button on the podcast page to play the first incomplete episode in the table instead of the "newest published". The "first episode" is the top row, and is dependent on the sorting and filtering options.

## Which issue is fixed?

Related to https://github.com/advplyr/audiobookshelf/issues/1321.

## In-depth Description

This PR adds a new field to the Vuex store to track the podcast episode IDs in the `LazyEpisodesTable`. This takes the current sorting and filtering of the table into account. Then, when pressing the "Play" button, the uppermost incomplete episode is played instead of the most recently released incomplete episode.

This does not fully fix the issue as mentioned in https://github.com/advplyr/audiobookshelf/issues/1321#issuecomment-2492456282 because the sorting is applied to all podcasts instead of being remembered individually. No changes should need to be made for this logic once the sort order is podcast specific.

Commit overview:
1. Add the `sortedEpisodeIds` to the store, which stores the episode IDs of the sorted and filtered table.
2. Change the logic to use the first row of the table.
3. Update comments and remove double reverse of array

**Additional question** Should this just play the very first episode instead of the first incomplete episode? The filter defaults to "incomplete", so if a user has changed to show all episodes, I think it should play whatever is in the first row, regardless of the completion status.

## How have you tested this?

Added several podcasts (some serial and some episodic). Changed sort and filter options, navigating between podcasts to ensure that the state was being updated correctly and the first incomplete episode was correctly played.

## Screenshots

To illustrate what is happening, the episode in the red box is the episode which will be played when pressing the "Play" button above the description.

Screenshot from 2025-02-22 21-59-02

Screenshot from 2025-02-22 21-59-39

### DIFF
--- a/client/components/tables/podcast/LazyEpisodesTable.vue
+++ b/client/components/tables/podcast/LazyEpisodesTable.vue
@@ -89,6 +89,13 @@ export default {
       handler() {
         this.refresh()
       }
+    },
+    episodesList: {
+      handler(newList) {
+        const episodeIds = newList.map((ep) => ep.id)
+        this.$store.commit('setSortedEpisodeIds', episodeIds)
+      },
+      immediate: true
     }
   },
   computed: {
--- a/client/components/tables/podcast/LazyEpisodesTable.vue
+++ b/client/components/tables/podcast/LazyEpisodesTable.vue
@@ -89,13 +89,6 @@ export default {
       handler() {
         this.refresh()
       }
-    },
-    episodesList: {
-      handler(newList) {
-        const episodeIds = newList.map((ep) => ep.id)
-        this.$store.commit('setSortedEpisodeIds', episodeIds)
-      },
-      immediate: true
     }
   },
   computed: {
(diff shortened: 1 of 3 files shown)


advplyr:
I made 2 updates:

1. I updated the episode list to come from the `LazyEpisodesTable` component rather than using the store.
2. I updated the play button on individual episodes to populate the queue using the table episodes instead of the pub date sort.

When making change 2 I fixed an existing bug where if you press play on a finished episode it would play that episode but not add it to the queue.


> **Additional question** Should this just play the very first episode instead of the first incomplete episode? The filter defaults to "incomplete", so if a user has changed to show all episodes, I think it should play whatever is in the first row, regardless of the completion status.

If we made that change would we also want to add every episode to the queue and not check the progress? That would probably make sense also. We can do a separate PR for that if so

Thanks!

nichwall:
Thanks! Yeah, those changes make sense, I spent a while trying to figure out how to reference the elements from the component and the only way I could figure it out was using the store.

Issue #2672, opened 2024-02-27
[Bug]: Users able to add and download duplicate podcast episodes to queue


nichwall:
### Describe the issue

Users can mark the same episode for download multiple times using the "Find Episodes" modal. This adds the episode to the queue multiple times and results in duplicate files being downloaded and added to the podcast. A check against the queue before adding should be sufficient to fix this issue, or ensuring the queue has unique entries.

Download Queue:
![Download queue](https://github.com/advplyr/audiobookshelf/assets/5686638/9b8f6421-a01d-45ad-9208-49a4bc08743a)

Podcast Episodes with library item ID in the filename:
![Podcast episodes](https://github.com/advplyr/audiobookshelf/assets/5686638/be975ffc-4dca-443d-afff-143fdb7b25c8)

Podcast Files
![Podcast files](https://github.com/advplyr/audiobookshelf/assets/5686638/ff853548-fe17-4ded-a761-40b0209156ea)

### Steps to reproduce the issue

1. Select a podcast episode to download using the "Find Episodes" modal
2. Add one or more episodes to the download queue
3. Open the "Find Episodes" modal again and select episodes add to the queue again

### Audiobookshelf version

2.8.0

### How are you running audiobookshelf?

Docker

nichwall:
It is also possible to add episodes to the download queue if they finish downloading while the modal is open because the modal doesn't update on downloads, so adding to queue would also need to check if the episode already exists on the server. This may help with some other duplicate downloads issues that have been reported.

RoyWolfe:
I'm running into a similar issue, probably for the same reason. I'm converting some youtube channels to audio-only to listen to. That conversion is not a fast process and since I have several channels I follow, ABS checks for new episodes again while the same files are still in the queue waiting to be downloaded, causing it to be added multiple times to the download queue.

Issue #4036, opened 2025-02-23
[Bug]: Downloads same episode many times if the episode was not downloaded between scans


samuelvisser:
### What happened?

I had started a huge backlog download job for a podcast. While it was still being downloaded, a new episode dropped. That episode is now added to the download queue 9 times and will be added more presumably as long as the other jobs haven’t finished

![Image](https://github.com/user-attachments/assets/3317bb83-1e40-4911-a25a-b8a6b8b0e051)

### What did you expect to happen?

Audiobookshelf should check, when it found a new episode, whether its already in the download queue

### Steps to reproduce the issue

1. Add a podcast you know will drop a new download today
2. Download the podcast’s entire backlog (or give it any large download job for another podcast)
3. Watch the new episode being added over and over

### Audiobookshelf version

2.19.4

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Windows

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/2672

PR #4041, opened 2025-02-25
Prevent duplicate episodes from being added to queue

nichwall:
## Brief summary

This PR checks if an episode download is already queue to prevent the same episodes being added multiple times.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/2672 and https://github.com/advplyr/audiobookshelf/issues/4036.

## In-depth Description

Before adding a podcast episode to the download queue, the server checks to make sure the URL and Library Item ID does not already exist in the download queue and is not already downloading. I am checking the Library Item ID in case the same podcast has been added to multiple libraries, such as having a dedicated podcast library for individual users who may have added the same podcast.

This does not fully solve the issue where:
1. A user can open the "Find Episodes" modal and add a number of episodes to the download queue
2. Reopen the "Find Episodes" modal before episodes finish downloading
3. Wait for episodes to finish downloading
4. Select the same episodes to download
5. Download the same episodes

This is because the modal does not update the "download state" from the queue. I do not think the download queue should check for existing files because that can get complicated with other files already existing or not being matched, and is more of an edge case.

## How have you tested this?

Tested using the web client to try and add each of the following cases:
- All episodes already in the queue or currently downloading
- Some episodes already in the queue or currently downloading
- No episodes already in the queue or currently downloading

I have not created a dedicated test case for this.

## Screenshots

N/A

### DIFF
--- a/client/components/modals/podcast/EpisodeFeed.vue
+++ b/client/components/modals/podcast/EpisodeFeed.vue
@@ -16,11 +16,12 @@
           v-for="(episode, index) in episodesList"
           :key="index"
           class="relative"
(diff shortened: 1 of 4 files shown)


advplyr:
It works great. I decided to pass the queue data into the modal and compare against it. Working well in my tests.

image

Thanks!

PR #4055, opened 2025-02-27
Fix: flaky 2.15.0 migration test

nichwall:
## Brief summary

This PR fixes the flaky test for the 2.15.0 migration test.

## Which issue is fixed?

Same as https://github.com/advplyr/audiobookshelf/pull/4010

## In-depth Description

Each test now uses a fixed Date object for creation to ensure the test is no longer flaky.

During testing, I found that `LibraryAuthorController` tests are flaky and will usually fail within around 40-70 runs. The longest I saw was 84 runs before failing. Switching to running these tests in isolation (deleted all other tests because I am not sure how to disable them) still had errors at the same approximate frequency, so I think the `LibraryAuthorController` failure is not due to interactions with other tests.

## How have you tested this?

Created a bash script which continuously runs `npm run test` until any test fails. The test ran successfully (without LibraryAuthorController tests) over 300 times in this loop before manually stopping.

## Screenshots

N/A

### DIFF
--- a/test/server/migrations/v2.15.0-series-column-unique.test.js
+++ b/test/server/migrations/v2.15.0-series-column-unique.test.js
@@ -126,9 +126,9 @@ describe('migration-v2.15.0-series-column-unique', () => {
     it('upgrade with duplicate series and no sequence', async () => {
       // Add some entries to the Series table using the UUID for the ids
       await queryInterface.bulkInsert('Series', [
-        { id: series1Id, name: 'Series 1', libraryId: library1Id, createdAt: new Date(), updatedAt: new Date() },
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks! I can reproduce the LibraryItemController test failure but sometimes it takes several hundred runs to fail so it is difficult to debug.

nichwall:
> Thanks! I can reproduce the LibraryItemController test failure but sometimes it takes several hundred runs to fail so it is difficult to debug. 

I removed every tester other than the Logger and LibraryItemController tests to make running the tests repeatedly faster, and the errors still occurred that way so that should make it easier to reproduce.

advplyr:
Yeah I isolated it also. Still taking a long time to get an error

PR #4066, opened 2025-03-01
Added AudiMeta - an extended Audible provider

Vito0912:
## Brief summary

This adds AudiMeta as a provider for Audible which has extended capabilities over Audnexus such as searching for asins in multiple regions and slightly higher rate limits. There are more features, discussed in the Discord channel. 

## Which issue is fixed?

While this PR does not fix it, the provider can be used for https://github.com/advplyr/audiobookshelf/pull/3839

## In-depth Description

I would vote to replace Audnexus (or at least not make it default for Audible) and replace it over the long term with AudiMeta.

Why?

Because Audnexus returns faulty response codes.

Because the dev does not seem to be really interested in continuing the project (Messages in Discord).

Very limited options regarding querying and we have to make many requests.

---

This is a first step to make this transition by making this provider accessible via the dropdown.

When this is implemented for some time I would plan to create a PR to redirect 1% of requests from Audnexus to AudiMeta, and increase this for each release until everything is working.

## How have you tested this?

Matched whole library.

### DIFF
--- a/server/providers/AudiMeta.js
+++ b/server/providers/AudiMeta.js
@@ -1,0 +1,162 @@
+const axios = require('axios').default
+const Logger = require('../Logger')
+const { isValidASIN } = require('../utils/index')
+
+class AudiMeta {
+  #responseTimeout = 30000
+
+  constructor() {
+    this.regionMap = {
+      us: '.com',
+      ca: '.ca',
+      uk: '.co.uk',
+      au: '.com.au',
+      fr: '.fr',
+      de: '.de',
+      jp: '.co.jp',
+      it: '.it',
+      in: '.in',
+      es: '.es'
+    }
+  }
+
+  /**
+   * Audible will sometimes send sequences with "Book 1" or "2, Dramatized Adaptation"
+   * @see https://github.com/advplyr/audiobookshelf/issues/2380
+   * @see https://github.com/advplyr/audiobookshelf/issues/1339
+   *
+   * @param {string} seriesName
+   * @param {string} sequence
+   * @returns {string}
+   */
+  cleanSeriesSequence(seriesName, sequence) {
+    if (!sequence) return ''
+    // match any number with optional decimal (e.g, 1 or 1.5 or .5)
+    let numberFound = sequence.match(/\.\d+|\d+(?:\.\d+)?/)
+    let updatedSequence = numberFound ? numberFound[0] : sequence
+    if (sequence !== updatedSequence) {
+      Logger.debug(`[AudiMeta] Series "${seriesName}" sequence was cleaned from "${sequence}" to "${updatedSequence}"`)
+    }
+    return updatedSequence
+  }
+
+  cleanResult(item) {
+    const { title, subtitle, asin, authors, narrators, publisherName, summary, releaseDate, imageUrl, genres, series, language, lengthMinutes, bookFormat	} = item
+
+    const seriesList = []
+
+    series.forEach((s) => {
+      seriesList.push({
+        series: s.name,
+        sequence: this.cleanSeriesSequence(s.name, (s.position || '').toString())
+      })
+    });
+
+    // Tags and Genres are flipped for AudiMeta
+    const genresFiltered = genres ? genres.filter((g) => g.type == 'Tags').map((g) => g.name) : []
+    const tagsFiltered = genres ? genres.filter((g) => g.type == 'Genres').map((g) => g.name) : []
+
+    return {
+      title,
+      subtitle: subtitle || null,
+      author: authors ? authors.map(({ name }) => name).join(', ') : null,
+      narrator: narrators ? narrators.map(({ name }) => name).join(', ') : null,
+      publisher: publisherName,
+      publishedYear: releaseDate ? releaseDate.split('-')[0] : null,
+      description: summary || null,
+      cover: imageUrl,
+      asin,
+      genres: genresFiltered.length ? genresFiltered : null,
+      tags: tagsFiltered.length ? tagsFiltered.join(', ') : null,
+      series: seriesList.length ? seriesList : null,
+      language: language ? language.charAt(0).toUpperCase() + language.slice(1) : null,
+      duration: lengthMinutes && !isNaN(lengthMinutes) ? Number(lengthMinutes) : 0,
+      region: item.region || null,
+      rating: item.rating || null,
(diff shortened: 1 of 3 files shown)


Vito0912:
@advplyr 
Would you mind looking at this and maybe leaving a comment on what you think about the steps? 
I explained the benefits for the user [mainly that the region doesn't matter when matching ASINs that were already indexed and for multi-region search]. 
I don't like the way 100%, as it spams the selector a bit, but I guess it's a good first start to test it before redirecting some of the traffic, if you agree to it. Logging is made clear for AudiMeta, even if redirecting in the future.

advplyr:
Yeah I'll check it out. I saw some of the conversation and I like the idea but I want to go through it thoroughly when I have time. I've been really tied up the last week

Vito0912:
I only yet received/saw the notification. No problem, take your time.

Vito0912:
Ok, now applied prettier to this PR. I will try to run Visual Studio Code for next PRs now

server/providers/AudiMeta.js
```diff
@@ -0,0 +1,163 @@
...
+   * @param {number} [timeout] response timeout in ms
+   * @returns {Promise<Object>}
+   */
+  asinSearch(asin, region, timeout = this.#responseTimeout) {
+    if (!asin) return null
+    if (!timeout || isNaN(timeout)) timeout = this.#responseTimeout
+
+    asin = encodeURIComponent(asin.toUpperCase())
+    let regionQuery = region ? `?region=${region}` : ''
+    let url = `https://audimeta.de/book/${asin}${regionQuery}`
```
pquerner, line 93: Should the domain be hardcoded here?

server/providers/AudiMeta.js
```diff
@@ -0,0 +1,163 @@
...
+    if (!items.length) {
+      const queryObj = {
+        title: title,
+        region: region,
+        limit: '10'
+      }
+      if (author) queryObj.author = author
+      const queryString = new URLSearchParams(queryObj).toString()
+
+      const url = `https://audimeta.de/search?${queryString}`
```
pquerner, line 145: Hardcoded domain here.

Vito0912:
@pquerner. Built-in providers are always hard-coded because they work differently than custom providers.

See https://www.audiobookshelf.org/guides/custom-metadata-providers/#community-providers for more information on how they work.

You can check Discord to see some conversations about it. AudiMeta is basically the same as Audnex, as it uses the same data sources, except that AudiMeta does not throw errors on a 404, has better searching, and does not care about the region if one is found. (With cons of some few old books missing that are not on Audible anymore)

It also has more endpoints for finding all books in a series, all books by an author, and searching authors directly via Audible (Audnexus, as far as I can tell, only searches their database). In that Discord, there are a few more reasons why I would like to completely replace Audnex in this area—mostly so users not have to switch regions for every book (e.g. if books in German and English audiobooks every time) —but that's up to @advplyr decide at the end if they will be merged/replaced. For stress testing, this is planed implemented as an extra provider currently (when and if this PR is merged), and I plan, depending on whether @advplyr goes with it, to redirect x% of traffic via the AudiMeta API, as the provider `Audible.xxx` is generally named broadly and not after Audnexus anyway (just to be clear, if AudiMeta is ditched, not the Audible names) and scale up

But hardcoding is inteded. You can view the other providers that are built in. The logic for Custom Metadata providers is in the corresponding file in that directory

Vito0912:
> Just looked at audible provider and saw it didnt use hardcoded tld.

Ah, sorry for the confusion. With Audnexus, we currently make a request to the Audible API (using different TLDs) to get the ASINs of the books. Then, for each ASIN, we make a request to Audnexus (hardcoded).

AudiMeta takes the region and search parameters and handles this server-side so that only one request is needed instead of multiple.

binyaminyblatt:
Please merge this I've come across numerous cases where things just don't work properly with Audnexus and I have tried using their API directly sometimes it's just broken and when I used the AudiMeta API it worked so much better

Vito0912:
You can add `https://beta.audimeta.de/<region>` as a custom-metadata provider.

It’s still not as ideal as having it directly implemented, since I don’t get an ASIN when the title already has one. This will make rematches less reliable, but for an initial search, it should work just as well.

While testing the Audible API, I also found a way to get a series description and the author’s description without scraping. Searching for authors should also be much more stable and accurate now as it uses Audible for searching (not like Audnex.us does or how the current main channel of AudiMeta does it).

I don’t have an OpenAPI specification for it yet and will keep it in beta for a week or two, but it is backwards compatible.

tehguitarist:
I was unsure whether to mention this here, or in your github for the project, but a quick test using it as a custom metadata provider, the description is truncated to 91 characters including ellipses, and the language is returned without being capitalised (english instead of English) - but those could also be ABS things. Just worth mentioning in case anyone was going to do a full re-scrape.

Vito0912:
The data comes directly from Audible.  
Audible returns it in lowercase, and the current provider also returns it in lowercase in its data.

Description: I return the short description instead of the long one (publishers set both). This is a mistake I made when I cleaned up the codebase, and "abs" is an extra endpoint I forgot to update. However, it is not an exact length. I will update it today.

The used endpoint in this PR is unaffected and should be correct.

Vito0912:
I fixed the description. As for the language, I think this is a bug in abs, since `german` and `German` should be treated as the same language. 
ABS probably solves this with code, but I have not checked. I do not have access to this as a custom metadata provider, so I cannot change it.

The provider was originally made for ABS, but now is used by other software, so I will not add anything abs specific other that this endpoint. But I will not change data for ABS

I e.g. also don't have any access to the ASIN, which makes rematching very likley to find wrong items. That is why this PR exists. Custom metadata provider are very very restricted.

nichwall:
> I fixed the description. As for the language, I think this is a bug in abs, since `german` and `German` should be treated as the same language. 
> ABS probably solves this with code, but I have not checked. I do not have access to this as a custom metadata provider, so I cannot change it.

ABS just stores the language as a string, it doesn't do any validation or correction on the language (so you have "German", "german", and "ger" at the same time, it is up to the user to standardize).

Vito0912:
I meant more about the capitalization. Ger and German could mean the same, but maybe something else is meant. But German and german are the same so should be handled the same imo. But that was a side node in general. I will probably close this PR soon anyway as Audnexus works for the most part and the added functionality the provider has is currently not usable within ABS

PR #4075, opened 2025-03-04
Fixes search not returning results if description field is not provided by a custom provider

Vito0912:
## Brief summary

Checks if the type is string before passing it to sanitize, because sanitize throws an error if the type is not a string.

Another way would be to return null if the type is not a string instead of the current way to prevent this happening altogether in the sanitizer.

I am not sure if this is possible, but if the type would not be a string, then it would not go through the sanitizer. So if the type is somehow not a string, one could maybe include bad code. I am not that familiar with JavaScript's type system, so I don't know if it's possible that this variable could be something other than a string, null, or a number when numbers would be returned.

### DIFF
--- a/server/providers/CustomProviderAdapter.js
+++ b/server/providers/CustomProviderAdapter.js
@@ -69,25 +69,57 @@ class CustomProviderAdapter {
       throw new Error('Custom provider returned malformed response')
     }
 
+    const toStringOrUndefined = (value) => {
+      if (typeof value === 'string' || typeof value === 'number') return String(value)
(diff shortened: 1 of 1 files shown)


nichwall:
Does this actually cause a crash? The error discussed in https://github.com/advplyr/audiobookshelf/issues/4004#issuecomment-2698283834 is just an error log being printed.

Vito0912:
You are right. I used "crash" in the matter of it's crashing the search. But yeah, it gets catched down the line.
I can change that later

advplyr:
There was no validation of the data being returned from the provider so I added it. I did some testing locally but I didn't test any of your providers.

Thanks!

PR #4088, opened 2025-03-08
Fix empty series delete check

nichwall:
## Brief summary

This PR fixes an issue where the `LibraryItemController` test would fail occasionally.

## Which issue is fixed?

No issue number, referenced in https://github.com/advplyr/audiobookshelf/pull/4055.

## In-depth Description

This PR changes the "author with zero books" check to happen in a Sequelize transaction to ensure race conditions do not happen due to changes in the database between checking the database and deleting the author. This PR also changes the "series with zero books" check to work in the same way.

The issue was specific to the "series with zero books" check. The change to the author function is not strictly necessary, but may as well add the transaction so it does not become an issue in the future.

The previous version of "series with zero books" check would then iterate over each series to do the following actions:
- Check if series is empty
- Delete existing RSS feeds for the series
- Delete the series

This resulted in multiple function calls to delete the series and makes it difficult to revert in case an error occurs during the cleanup, instead of relying on deleting the series to cascade and delete the associated RSS feed.

## How have you tested this?

Verified with existing unit tests, and no failures occurred after running the test over 600 times (previously the test would fail within 100 runs).

## Screenshots

N/A

### DIFF
--- a/server/routers/ApiRouter.js
+++ b/server/routers/ApiRouter.js
@@ -420,40 +420,53 @@ class ApiRouter {
   async checkRemoveAuthorsWithNoBooks(authorIds) {
     if (!authorIds?.length) return
 
-    const bookAuthorsToRemove = (
-      await Database.authorModel.findAll({
-        where: [
-          {
-            id: authorIds,
-            asin: {
-              [sequelize.Op.or]: [null, '']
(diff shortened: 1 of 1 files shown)


advplyr:
Nice catch! I made that update to close the series feeds and did some tests on that.
Thanks!

PR #4089, opened 2025-03-08
Simplify log level determination

nichwall:
## Brief summary

Remove duplicate functions for logging.

## Which issue is fixed?

No issue

## In-depth Description

This PR removes the duplicate `getLogLevelString` function from the scan-related loggers. `Logger` is already included, so this just removes code duplication.

This also increases the total coverage of the `Logger.test.js` with no changes to the test code.

## How have you tested this?

Ran unit tests and a scan to ensure nothing broke.

## Screenshots

N/A

### DIFF
--- a/server/Logger.js
+++ b/server/Logger.js
@@ -21,12 +21,7 @@ class Logger {
   }
 
   get levelString() {
-    for (const key in LogLevel) {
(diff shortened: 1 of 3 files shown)


advplyr:
Thanks!

Issue #2888, opened 2024-04-28
[Enhancement]: (Cosmetic) Picture dimensions should use “×” instead of lowercase x


ZLoth:
### Describe the feature/enhancement

This is purely cosmetic, but currently, the dimensions of a picture are being displayed with an lowercase x. 

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/04981b42-daeb-4a31-b282-8b9971319377)
![image](https://github.com/advplyr/audiobookshelf/assets/6700159/9873244b-9725-47cb-b339-b7d23f9355d9)

It would be nice if it was displayed with the [&times;](https://www.compart.com/en/unicode/U+00D7) symbol instead.

![image](https://github.com/advplyr/audiobookshelf/assets/6700159/ab157d3c-633b-4136-b330-2df2a481b77a)
![image](https://github.com/advplyr/audiobookshelf/assets/6700159/78fc4c43-82c9-42e9-bc7d-5df708ce7e9d)

advplyr:
I updated this for the next release. 

I want to note that for playback speed I was using unicode [⨯](https://www.compart.com/en/unicode/U+2A2F) and it didn't display on a firefox linux browser per https://github.com/advplyr/audiobookshelf/issues/1508

The unicode character you linked [×](https://www.compart.com/en/unicode/U+00D7) should be supported though.

ZLoth:
Thank you for the update. Looking forward to the next release.

criscrafter:
Honestly, I would like this to be a setting, as basically every software and every website uses an lowercase x, and I am used to it and like is way more, because its more compact.

advplyr:
I have no preference at all. Whatever makes the most sense let me know. It's not going to be a setting

ZLoth:
https://en.wikipedia.org/wiki/Multiplication_sign - "[Screen resolution](https://en.wikipedia.org/wiki/Screen_resolution) in pixels, such as 1920 pixels across × 1080 pixels down. Read as "by"."

criscrafter:
Yeah, that is the theory, but in practice nobody does it that way.
Check any Monitor on Amazon and look at the resolution. There will be a x.
The only exception I could find is the small resolution shown in the bottom right corner of Google Images.
My issue is it would make things harder. When you want to copy the resolution into other programs, many of them expect an x.
I am fine with both, but you should consider the added hassle for many users that results from that.

advplyr:
[...]

PR #4100, opened 2025-03-11
Use multiplication symbol instead of letter x

majjejjam:
## Brief summary

Use the proper multiplication symbol (×) for the playback speed.

## Which issue is fixed?

Minor polish.

## In-depth Description

In the playback speed popup, the letter "x" is switched out for the multiplication symbol "×". The order of the number and symbol is reversed so it is read "times 1.0" instead of "1.0 times" (or previously "1.0 x"). Between the symbol and the number, a narrow no-break space (`&#8239;`) is added for proper spacing. The classes of some spans are removed for better alignment.

## How have you tested this?

I have generated the client and ran the server to make sure the changes had no unintended side effects. The playback speed controls still work as usual.

## Screenshots

| Before | After |
| --- | --- |
| bild | Skärmbild från 2025-03-11 12-56-13 |

### DIFF
--- a/client/components/controls/PlaybackSpeedControl.vue
+++ b/client/components/controls/PlaybackSpeedControl.vue
@@ -1,7 +1,7 @@
 <template>
   <div ref="wrapper" class="relative ml-4 sm:ml-8" v-click-outside="clickOutside">
     <div class="flex items-center justify-center text-gray-300 cursor-pointer h-full" @mousedown.prevent @mouseup.prevent @click="setShowMenu(true)">
-      <span class="text-gray-200 text-sm sm:text-base">{{ playbackRateDisplay }}<span class="text-base">x</span></span>
+      <span class="text-gray-200 text-sm sm:text-base">×&#8239;{{ playbackRateDisplay }}</span>
     </div>
     <div v-show="showMenu" class="absolute -top-[5.5rem] z-20 bg-bg border-black-200 border shadow-xl rounded-lg" :style="{ left: menuLeft + 'px' }">
       <div class="absolute -bottom-1.5 right-0 w-full flex justify-center" :style="{ left: arrowLeft + 'px' }">
@@ -11,7 +11,7 @@
         <template v-for="rate in rates">
           <div :key="rate" class="h-full border-black-300 w-11 cursor-pointer border rounded-sm" :class="value === rate ? 'bg-black-100' : 'hover:bg-black hover:bg-opacity-10'" style="min-width: 44px; max-width: 44px" @click="set(rate)">
             <div class="w-full h-full flex justify-center items-center">
-              <p class="text-xs text-center">{{ rate }}<span class="text-sm">x</span></p>
+              <p class="text-xs text-center">×&#8239;{{ rate }}</p>
             </div>
           </div>
         </template>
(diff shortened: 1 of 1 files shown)


Vito0912:
Personally I think before looks cleaner and more intuitive.

majjejjam:
We can definitely keep the order. I do however think it's important to use the proper symbol. The letter "x" has nothing to do with multiplication, besides looking like a cross (but not exactly, most apparent in serif typefaces) and commonly being used instead of it because a cross is harder to type (impossible on a typewriter).

I think the cross together with the spacing is easier to parse. This is especially apparent with more numbers. Compare 1200x500x300mm with 1200 × 500 × 300 mm. When you squint your eyes, the first example is one grey box. The second separates each dimension; it's clear where one ends and another begins. This is of course a different situation, but I think this reasoning still has some merit.

Vito0912:
This is just my personal opinion. In my view, although using the multiplication symbol might be mathematically more correct, the UI is important. All players I know use "x" instead of the multiplication symbol because it fits better (my view and seemingly the ones of the apps).

I don't want to argu or say it is wrong etc, but simply state that good UI is sometimes more important than mathematical correctness. (Even Audible, a profit-driven company, uses "x"). 
again, nothing against your work, but just a opinion for other to decide if they want it.

advplyr:
Early on I used the wrong unicode character for multiplication and it didn't show up on firefox for linux.

https://github.com/advplyr/audiobookshelf/issues/2888#issuecomment-2083790139

I kind of like the x better also but that may be because most other UI use it for playback speed. I could go either way though, if more people prefer the multiplication symbol. Maybe I'll do a discord poll

advplyr:
Re-reading #2888 it seems some users may have stronger feelings about the x vs ×. In that case it was for the resolution.
I really wouldn't want to do a setting for something so minor but now I'm curious how this splits in terms of preference

advplyr:
I posted a poll https://discord.com/channels/942908292873723984/942908295579062335/1349022436347084820

majjejjam:
> good UI is sometimes more important than mathematical correctness

The issue of x vs × is one of typography, not mathematics. In this case, because the UI consists of typography, it's an issue related to UI. Professional typographers recommend using the cross in cases such as these (Matthew Butterick as example: https://practicaltypography.com/math-symbols.html).

The arguments by criscrafter for using the letter x in #2888 are
1. It's the industry standard
2. They're used to it
3. It's more compact
4. It's easier to copy for use in some programs

(1) This is not really valid. We shouldn't blindly copy what the industry does. For example, it's common to see content disappearing behind a white rectangle without shadow or border. I can't find a good example right now, but it basically looks like this (see the text being cut off):
bild

(2) Sure. Being used to something weighs heavily. In this case, I think the other points weigh more though.

(3) I can't see that using the letter x is more compact. And even if it were, white space is important. All good interfaces make use of emptiness in the shape of margins and padding and others.

(4) I don't see users realistically copying either the dimensions of the images in Audiobookshelf or the playback speed for use in external programs.

---

The plus and minus symbols in place now are actual plus and minus symbols:
Skärmbild från 2025-03-11 14-53-07

Here's how it would look with + and -, which you can type on a standard keyboard:
Skärmbild från 2025-03-11 14-53-38

This is arguably worse: the "minus" symbol is smaller and  doesn't align with the horizontal bar of the plus. Now, proper symbols are in place, and if they weren't I'd argue we should use them. It wouldn't matter if it was the industry standard or people were used to them; they make for a worse user interface.

---

Sorry this got a bit long :sweat_smile: I felt I wanted to articulate exactly why I think this is good change. I hope it isn't taken the wrong way.

Also: I just saw the poll now whilst writing this.

nichwall:
> The arguments by criscrafter for using the letter x in #2888 are
> 1. It's the industry standard
> 2. They're used to it
> 3. It's more compact
> 4. It's easier to copy for use in some programs

I would argue points 1 and 2 are the *most* important to be consistent with what users expect. Spot checking a few other apps (SmartAudiobook player, VLC, and YouTube), all look to be using the "1x" instead of the cross with extra space.

But yeah, I agree the copying point doesn't really matter for the playback speed. I'm personally indifferent (and see what happens in the poll), but having the extra spacing in the mobile app would probably look weird because there already isn't a lot of horizontal space

advplyr:
I forgot to update this but the overwhelming consensus on Discord was to keep the "x" so we'll do that

PR #4106, opened 2025-03-11
Fix: Load epub covers via cover-image property

### LINKED ISSUE #4105: [Bug]: Some epubs fail to load the correct cover
### What happened?

Some epubs show different covers when loaded in ABS as when instead using other tools (like for example Calibre). Even changing the cover in Calibre and exporting from there, won't fix the ABS import. 

One example of such book is [here](https://github.com/IDPF/epub3-samples/releases/download/20230704/epub30-spec.epub) (Taken from [Epub samples](https://idpf.github.io/epub3-samples/30/samples.html)). When loaded in ABS it'll look like this: 
Image

When loaded by Calibre:
Image

I expected to have the same cover in ABS as it had in Calibre.

1. Upload the specific epub to ABS
2. Look at wrong cover

### Audiobookshelf version

v2.19.5

Docker

Linux

Firefox

### Logs

```shell

```

### Additional Notes

The issue comes from missing cover metadata entry, as the epub instead uses [cover-image property](https://www.w3.org/TR/epub-33/#sec-cover-image) in manifest to specify the correct image.

Roukanken42:
## Brief summary

Adds ability to load epub covers via [cover-image property](https://www.w3.org/TR/epub-33/#sec-cover-image) in manifest if the cover meta entry is missing.

## Which issue is fixed?

Fixes #4105

## In-depth Description

The implementation respects the [EPUB 3.3 spec](https://www.w3.org/TR/epub-33/#attrdef-item-properties), and loads `property="cover-image maybe-other-props"` tagged manifest items as cover image, in case the cover metadata entry is missing. Fallback to the first image is still present if neither metadata entry, nor property can be found.

## How have you tested this?

I've used this [epub](https://github.com/IDPF/epub3-samples/releases/download/20230704/epub30-spec.epub) as example where this can help. Simply importing this into ABS will have the wrong cover before this change, but the good one after reimporting it with the change. (Same cover as Calibre loads)

## Screenshots

Before: 
image

After:
image

Calibre:
image

### DIFF
--- a/server/utils/parsers/parseEpubMetadata.js
+++ b/server/utils/parsers/parseEpubMetadata.js
@@ -107,7 +107,8 @@ async function parse(ebookFile) {
 
   // Attempt to find filepath to cover image:
   // Metadata may include <meta name="cover" content="id"/> where content is the id of the cover image in the manifest
-  //  Otherwise the first image in the manifest is used as the cover image
+  //  Otherwise find image in the manifest with cover-image property set
+  //  As a fallback the first image in the manifest is used as the cover image
   let packageMetadata = packageJson.package?.metadata
   if (Array.isArray(packageMetadata)) {
     packageMetadata = packageMetadata[0]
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

Issue #4114, opened 2025-03-14
[Bug]: Publish date not save when matched from OpenLibrary


olek-osikowicz:
### What happened?

When I tried to match books with Publish date using OpenLibrary I noticed the year is not saved.
See the video below:

https://github.com/user-attachments/assets/eaeeda8e-f2b6-4902-ac33-d99d4330dfed

### What did you expect to happen?

I would expect the publish year to appear in the form:

![Image](https://github.com/user-attachments/assets/9ffba223-0807-4344-9ef2-89a083bbbf2c)

And then when saved, on the books page as well:

![Image](https://github.com/user-attachments/assets/906d971f-211f-4e0b-a2d1-a86066e455fd)

### Steps to reproduce the issue

See the uploaded video:

1. Edit a book entry
2. Click matching tab.
3. Choose OpenLibrary
4. Click on the matching entry.
5. See the publish year is not submitted.

### Audiobookshelf version

v2.19.5

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Chrome

### Logs

```shell

```

nichwall:
Fixed by https://github.com/advplyr/audiobookshelf/pull/4118

olek-osikowicz:
Thanks for quickly resolving this, I'll close when released!

PR #4118, opened 2025-03-15
Allow number types for payload metadata when updating books.

GabrielGavrilov:
## Brief summary

Allow number types for payload metadata when updating books.

## Which issue is fixed?

[4114](https://github.com/advplyr/audiobookshelf/issues/4114)

## In-depth Description

Allow the use of number types for payload metadata when updating books.

OpenLibrary uses numbers instead of strings to update the published year, so this PR attempts to fix this issue by allowing number types for payload metadata. 

## How have you tested this?

I have replicated issue [4114](https://github.com/advplyr/audiobookshelf/issues/4114) locally and tested the change effects.


advplyr:
I think we should cast it to string if it is a number

GabrielGavrilov:
For sure. Now, every metadata field that is a numeric type gets cast to a string.

advplyr:
Thanks!

PR #4119, opened 2025-03-15
ci: update actions

jfrazx:
## Which issue is fixed?
- update action versions
- don't run docker action if not maintainer repo


advplyr:
Thanks!

PR #4126, opened 2025-03-16
Tailwind 4 migration

mikiher:
## Brief summary

This migrates the web client from Tailwind V3 to V4

## Which issue is fixed?

No issue.

## In-depth Description

V4 has quite a few breaking changes, hence the many changes in the code:
Main changes:
- Ran upgrade tool (which installs tailwindcss v4 and does a first sweep of changes in the code).
- Installed `@tailwindcss/cli` as dev dependency (for cypress)
- Tailwind postcss plugin is now separate from tailwindcss, so installed @tailwindcss/postcss
- Changed nuxt config to use the new plugin
- Had to make manual changes due to:
  - removal of (bg|text|border|ring)-opacity-* utilities
  - changes in tailwind utilities specificity (not fully understood yet).

## How have you tested this?

Not fully tested yet - this is just to share code for now.


mikiher:
_Note_: this still loads the old `tailwind.config.js`. 

I think it can be removed but it needs to be translated into V4 css directives and variables definitions (the upgrade tool didn't do that).

advplyr:
Instead of using the postcss plugin in nuxt.config, I put it in a separate postcss.config.js and it is working now for me. I also removed autoprefixer because I think it is unnecessary now.

I tested by running `npm run dev` and also by `npm run generate` and using the statically built site.

mikiher:
OK, I manually went through all modifications to verify them.
- I reverted the few `!mportnant` additions I made due to the postcss issue you fixed.
- I fixed a typo that I likely introduced with one of my regexp replace expressions
- I fixed a non-class replacement (from `blur' to `blur-sm`) that the upgrade tool has mistakenly made.

I've also done additional testing to find any issues that popped up, and could not find any.

I believe this is ready for review now.

advplyr:
I did a lot of testing and was only able to find a few things. I guess tailwind removed cursor:pointer on buttons by default so I added that in the tailwind @layer base.

Thanks!

Issue #4130, opened 2025-03-17
[Enhancement]: In Audiobook Library → Home Page → Listen Again, option should be available to select between ten most recent books and random display of completed books.


ZLoth:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

This feature request is based upon defect https://github.com/advplyr/audiobookshelf/issues/4129 where it was declared that the behavior was "by design". As currently designed, Audiobook Library → Home Page → Listen Again shows the ten most recent listens. It would be nice to have a toggle so that the Listen Again behaviour can be set to be either ten most recent listens or be similar to the behavior of "Discover" where it brings up ten random completed books.

### Why would this be helpful?

Emulate the behavior of "Discover" so that it isn't a static display.

### Future Implementation (Screenshot)

Not applicable.

### Audiobookshelf Server Version

2.19.5

### Current Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/798e9b99-0ad4-419b-8086-ce1b221448b7)

ZLoth:
Thank you @Vito0912 . This enhancement was made after https://github.com/advplyr/audiobookshelf/issues/4129 was stated to be expected behavior.

PR #4131, opened 2025-03-17
Sort Listen Again randomly (toggable via setting)

### LINKED ISSUE #4129: [Bug]: In Audiobook Library → Home Page → Listen Again, only the ten most recent completed books are displayed
### What happened?

In the "Listen Again" section, it appears that it will list the ten most recent books that are flagged as completed. 

I expected the behavior to be similar to Discover books which is randomized, only for books to be completed. Showing the same ten books gets old after a while. 

When you flag a book as complted, it should show up as the first book in "Listen Again". 

### Audiobookshelf version

2.19.5

Docker

Linux

Chrome

### Logs

```shell
No logs.
```

### Additional Notes

_No response_

Vito0912:
## Brief summary

This resolves #4129.

I think it makes sense to add it the way proposed in that issue, as it would make sense as a contrary part to discover, which is the opposite of listen again.

It makes listen again sort randomly.

---

#4130 does not make sense in my opinion, because it would add more small options and displaying the last 10 listened books does not make that much sense in general. Personally, I never used Listen again, because it only showed the last ten books and it was like "Oh, I listened to that like a week ago."

But maybe others can comment as this is only my view. 

---

For devs, getting the progress i still possible via the API:
`/api/libraries/<id>/items?filter=progress.ZmluaXNoZWQ%3D&sort=progress&desc=1&limit=32`

---

If that is not wanted, just close that PR :)

## Which issue is fixed?

Fixes #4129 

## In-depth Description

replaced sort by progress to random

## How have you tested this?

Own library

### DIFF
--- a/client/components/modals/libraries/LibrarySettings.vue
+++ b/client/components/modals/libraries/LibrarySettings.vue
@@ -75,9 +75,9 @@
       <div class="p-2 w-full md:w-1/2">
         <div class="flex items-center">
           <ui-toggle-switch v-model="listenAgainOrdered" size="sm" @input="formUpdated" />
-          <ui-tooltip :text="$strings.LabelSettingsOnlyShowLaterBooksInContinueSeriesHelp">
+          <ui-tooltip :text="$strings.LabelListenAgainOrderedHelp">
             <p class="pl-4 text-sm">
-              {{ $strings.LabelSettingsOnlyShowLaterBooksInContinueSeries }}
+              {{ $strings.LabelListenAgainOrdered }}
               <span class="material-symbols icon-text text-sm">info</span>
             </p>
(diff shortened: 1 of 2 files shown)


Vito0912:
I added a setting with a default of "random."

One argument on Discord was to see last-played items. I think the purpose of "listen again" is to listen to things again, so random order makes sense as default. For users who want to keep track of what they listened to over the last few days via that feature, they can enable the feature for individual libraries. 

@advplyr

averageyogi:
If it will be selectable, I think the name of the list should change to reflect what's being displayed. "Listen Again" should be a random list of completed titles, "Recently Completed" is what the current implementation should be called, the most recently completed titles.

Issue #3541, opened 2024-10-20
[Enhancement]: Ability to download entire Author and/or Series


pjones112000:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Sometimes when I go on cruises I like taking audiobooks with me to listen offline...it would be nice to be able to download all the audiobooks for a specific or series rather than downloading one at a time....particularly on long cruises or vacations.

### Why would this be helpful?

Currently, I would either need to download one book at a time or copy directly from the filestore to my tablet/phone to listen rather than selecting an entire series or author to download all to my tablet/phone in an automated fashion.

### Future Implementation (Screenshot)

Add option to either the picture of the author or the author details page to download.

### Audiobookshelf Server Version

v2.14.0

nichwall:
Do you mean in the web client or in the mobile app?

If the mobile app, duplicate of https://github.com/advplyr/audiobookshelf-app/issues/188

pjones112000:
Sorry, I meant to mention, the web UI

advplyr:
To further clarify, are you downloading the zip file on the web client and playing this in an audio player on your phone?

pjones112000:
Yes, exactly.   Zip to phone, using an offline music player.  Admittedly,
an offline version of the webclient would be nice but I know that's more
complicated.

pjones112000:
...or even an offline version of the mobile app.

advplyr:
The mobile apps for android and ios support downloading for offline

Chriss78:
Hello just adding/comment this to the Enhancment. A DL option for the WebUI would be helpful for eBook Libraries. Typically a User  DL a  .epub to the PC and copy it to their desired eReader eg Tolino. Especially for series with multiple parts can this be usful and spare a user much clicking each book and afterwards unzipping again etc.
Maybe in combine the funktion with the mark (circle in book cover top left corner) and DL the marked eBooks.

PR #4133, opened 2025-03-17
Adds the option to download selected books

Vito0912:
## Brief summary

Adds a new endpoint that allows bulk downloading multiple library items and adds it to the frontend.

## Which issue is fixed?

None directly, I think. But it comes close to #3541 and I believe this is the better approach instead of implementing an endpoint for Collections, Series, Authors, Playlists, etc.

## In-depth Description

It fetches all the library items and their paths, puts them into one zip file.

Structure is like
```
ZIP
-- file.mp3
-- Book Folder Name/
---- file.mp3
---- file.epub
```

We are limited by GET to download the file as it has a 2000 character limit on most browsers. Since we can't send any request body, we have a limit of 40 IDs we can send to safely prevent any errors. Therefore, the download button is only shown for up to 40 books. But in theory, it would work with more.

## How have you tested this?

Using my own library. Files in root directory, but also in non-root directories.

## Screenshots

I don't have any with public domain works, but just imagine a download button under the three-dot menu when selecting multiple books in the library view.

### DIFF
--- a/client/components/app/Appbar.vue
+++ b/client/components/app/Appbar.vue
@@ -182,7 +182,7 @@ export default {
 
       // The limit of 50 is introduced because of the URL length. Each id has 36 chars, so 36 * 40 = 1440
       // + 40 , separators = 1480 chars + base path 280 chars = 1760 chars. This keeps the URL under 2000 chars even with longer domains
-      if(this.selectedMediaItems.length <= 40) {
+      if (this.selectedMediaItems.length <= 40) {
         options.push({
           text: this.$strings.LabelDownload,
           action: 'download'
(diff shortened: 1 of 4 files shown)


Vito0912:
I am not sure what I am doing wrong. CodeQL hates me ig

advplyr:
Can you set up the recommended VSCode extensions or if you are not using VSCode get something that applies our formatting rules?
For example, semicolons are automatically removed on save as well as a space after function names, etc.

Also when using code from elsewhere if you can update the logs and do tests with both books and podcasts.

Thanks!

Vito0912:
Hey,
I use InteliJ with the prettier plugin enabled. I am not sure why it is not doing it automatically for the Audiobookshelf project, as it does for my other project. I will investigate, maybe some legacy config, from before I installed that plugin.

Regarding podcast I would say we should not support multiple podcasts downloads as they get quite big. I therefore did not test it with podcasts.

Also, I am currently only on mobile, I saw you removed the library where statement. Shouldn't that still be in place. This would allow anyone that has access to at least one library to download any media from the server when knowing the uuid?

(That's why I moved it from LibraryItemsXontroller to Library controller and forgot to change the logs afterwards) to validate access to the books

Vito0912:
I will check my other PR for prettier too. Thanks for pointing that out

Vito0912:
@advplyr I am confused about the prettier config. I have already enabled the format on save option, but nothing happens (the exact same config, when copied to my project, works). Running prettier from the command line in the audiobokshelf directory gives me:

`[warn] Code style issues found in 807 files`., which cant be right.

I will try Visual Studio Code, but you mentioned something about recommended VSCode extensions. I cannot seem to find any resource. Or is there a command I can run, to apply these rules?

advplyr:
I haven't gone deep in understanding all the editor config options so I don't think I'll be much help. The recommended extensions are defined here https://github.com/advplyr/audiobookshelf/blob/master/.vscode/extensions.json.
When running vscode or a vscode based code editor there is a popup that opens up asking if you want to install the recommended extensions.

The `.prettierrc` and `.editorconfig` file should be able to get used by any code editor though I'm just not familiar with others.

PR #4150, opened 2025-03-23
Reduce final docker image size

pinjeff:
## Brief summary
Reduces the final docker image size

## Which issue is fixed?

no issue open

## In-depth Description
this adds a third stage to the build, copying the required files only from the previous stages, this reduces the final image size from 600MB+ down to ~320MB

## How have you tested this?
I tried the image after the build and it seems to be working, is there a test suite to run?


advplyr:
Thanks!

PR #4151, opened 2025-03-23
Fix broken component tests

mikiher:
## Brief summary

This fixes some broken client component tests

## Which issue is fixed?

No issue.

## In-depth Description

Unfortunately, we don't run component tests automatically, so they break from time to time.
This fixes them to run with the current code.

If it's OK with you, I'll add these to the push procedure using a Github workflow (like unit tests).

## How have you tested this?

All existing tests run.

### DIFF
--- a/.github/workflows/component-tests.yml
+++ b/.github/workflows/component-tests.yml
@@ -1,0 +1,48 @@
+name: Run Component Tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Branch/Tag/SHA to test'
+        required: true
+  pull_request:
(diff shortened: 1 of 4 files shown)


advplyr:
Yeah we can add it to the workflow

mikiher:
OK, added the workflow:
- It runs on push and pull_request
- It only runs when client files are changed

advplyr:
I tested this and it looks good, is there anything else you wanted to add here?

mikiher:
No, just forgot to mark it as ready for review.

advplyr:
Thanks!

PR #4162, opened 2025-03-26
Consistent icons in stats page

Toby222:
## Brief summary

For "Items Finished" in stats page, replace SVG icon with equivalent Material Symbol ligature

## Which issue is fixed?

I did not open an issue for this

## In-depth Description

Replaces the inline SVG element for the "Items Finished" stat with the equivalent symbol from the Material Symbol font to be consistent with the other stats.
Noticed this while idly selecting random text on the page and noticing that this icon was not selectable.

## How have you tested this?

Manual edit in browser, did not actually build the program

## Screenshots

Before:
image
After:
image


advplyr:
Thanks!

Issue #4210, opened 2025-04-19
[Bug]: Chapter Lookup Failing


Zerofire03:
### What happened?

When trying to lookup chapters via ASIN, receive an error saying chapters not found.

### What did you expect to happen?

I expect chapters to be looked up and found

### Steps to reproduce the issue

1. Edit book that you want to modify chapters on
2. Click Lookup > Enter ASIN > Search

### Audiobookshelf version

2.20.0

### How are you running audiobookshelf?

Debian/PPA

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Chrome

### Logs

```shell
[...]

PR #4213, opened 2025-04-20
Chapter lookup error interface update

nichwall:
## Brief summary

This PR provides more information when a chapter lookup fails. Instead of closing the modal and showing a toast message, the error is displayed in the modal with some extra information.

## Which issue is fixed?

Addresses https://github.com/advplyr/audiobookshelf/issues/4210 (the issue is already closed, but is representative of a FAQ/common issue)

## In-depth Description

An additional string is added to ensure users are using the ASIN from Audible instead of Amazon because they look the same to a user.

I also localized the "chapter lookup error" from the server by adding a string key to the server response. I left the current error string in case anybody is using it (I doubt it, but just in case).

## How have you tested this?

Tested with valid ASIN, invalid ASIN, and ASIN from Amazon/wrong region.

## Screenshots

I am not in love with the layout, but this was the best I could come up with. The extra space on the top looks weird, but I also prefer the inputs not moving when the messages appear.
Screenshot from 2025-04-19 23-25-25
Screenshot from 2025-04-19 23-25-31
Screenshot from 2025-04-19 23-25-46


advplyr:
Great, thanks!

Issue #4169, opened 2025-03-29
[Enhancement]: Add ability to combine/join M4B files without re-encoding


ZLoth:
### Type of Enhancement

Server Backend

### Describe the Feature/Enhancement

It would be nice if ABS had a Tools function to just combine M4B files without re-encoding. This will save tremendous time. 

### Why would this be helpful?

Audible sells several books in a "Omnibus" format where 2-4 books are inside one file. I tend to break these up into individual books through OpenAudible, but this leaves multiple chapters that need to be combined. There should be a one-touch function to combine these M4B files without re-encoding. 

### Future Implementation (Screenshot)

Can be just one button titled "Join without re-encoding" 

### Audiobookshelf Server Version

2.20

### Current Implementation (Screenshot)

Current workaround, thanks to @Simmias on discord, is to set the following options:

![Image](https://github.com/user-attachments/assets/26ff6a0c-5b05-41fb-989e-8f7ffe6ab194)

I have tested this, and it works successfully. This is much quicker than going through a re-encoding process when it isn't necessary.

PR #4214, opened 2025-04-20
Merge files without encoding

Vito0912:
## Brief summary

If more than one file is available, show an extra option that automatically sets encoding to copy in order to merge the files, thereby disabling advanced options.

Also fixes a visual bug where, if a user cancels a task, the advanced button still appears clickable even though it is not actually clickable. 

---
This time I applied Prettier, but it did nothing (because it's a Vue file). When I used the official formatter for Vue.js, it even changed lines I did not edit. So, I guess this is not desired. If I should apply a formatter, please leave a comment. 

## Which issue is fixed?

#4169

## In-depth Description

N/A. See above

## How have you tested this?

Merging a few books

## Screenshots

grafik

### DIFF
--- a/client/pages/audiobook/_id/manage.vue
+++ b/client/pages/audiobook/_id/manage.vue
@@ -77,9 +77,14 @@
       </div>
       <!-- m4b embed action buttons -->
       <div v-else class="w-full flex items-center mb-4">
-        <button :disabled="processing" class="text-sm uppercase text-gray-200 flex items-center pt-px pl-1 pr-2 hover:bg-white/5 rounded-md" @click="showEncodeOptions = !showEncodeOptions">
-          <span class="material-symbols text-xl">{{ showEncodeOptions || usingCustomEncodeOptions ? 'check_box' : 'check_box_outline_blank' }}</span> <span class="pl-1">{{ $strings.LabelUseAdvancedOptions }}</span>
-        </button>
+        <div class="flex flex-col items-start">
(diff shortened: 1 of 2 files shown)


ZLoth:
Thank you. Looking forward to release,

ABakedWaffle:
Thank you!

advplyr:
One issue is that merging mp3 files to m4b and keeping the mp3 codec isn't a good idea for compatibility. And for most users their multi-file audiobooks are mp3's.

M4b with mp3 codec won't play in iOS as of v18.41.

Users unfamiliar with audio codecs might use that setting because it is faster without realizing they are making audio files that won't play on iOS. Then open a bug report for the iOS app not playing their audio files.

Vito0912:
> M4b with mp3 codec won't play in iOS as of v18.41.

You are right. As of iOS version 18.5, it also does not work. That really surprises me, because as far as I know, Apple proposed the m4b/m4a standard. 
Maybe a disclaimer (for iOS) could be added (since the copy still affects iOS users if shared on platforms like Discord or Reddit). However, I understand if this could cause new issues anyways.

If you want to close this, feel free to do so, as I completely understand.

ZLoth:
> One issue is that merging mp3 files to m4b and keeping the mp3 codec isn't a good idea for compatibility. And for most users their multi-file audiobooks are mp3's.

If there is a single MP3 file detected in the list, then the merge file isn't available.

Vito0912:
Fixed in edge

PR #4223, opened 2025-04-26
Fix server crash when a user requests the RSS feed of an empty collection

Vito0912:
## Brief summary

Fix server crash when a user requests the RSS feed of an empty collection. 

## Which issue is fixed?

None/Discord

## In-depth Description
`reduce` needs at least one item in the array (or intial value). If the array is empty, it throws an error. Since the for loop is not called when `books` is empty, we can just make this null. The RSS feed will still be opened then. An alternative would be to close the RSS feed for any empty collection. I can imagine that this would not be a good solution, because if people remove an item and then instantly add one, it would be unexpected to have a closed RSS feed.

### DIFF
--- a/server/models/FeedEpisode.js
+++ b/server/models/FeedEpisode.js
@@ -215,9 +215,10 @@ class FeedEpisode extends Model {
    * @returns {Promise<FeedEpisode[]>}
    */
   static async createFromBooks(books, feed, slug, transaction) {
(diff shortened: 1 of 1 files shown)


advplyr:
How can I reproduce the crash?

Vito0912:
Create a collection (with a book). Open the RSS feed, delete the last item, then open the RSS feed-url.

Sorry, I forgot about that. Was on Discord.

advplyr:
Thanks!

Issue #4152, opened 2025-03-23
[Enhancement]: Shift Chapter Lookup to account for Audible branding


chriscam85:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Please add a setting to allow for inclusion/exclusion of branding durations when using the chapter lookup.
When I lookup the chapter markers for books I procured from non-Audible sources, I see that the durations are always incorrect, and it gives me pause. I looked at the message being received, and it does appear that the intro/outro durations are included (brandIntroDurationMs and brandOutroDurationMs).
Please allow for us to select whether or not those branding markers should be included in the chapter timings.
Thanks!

### Why would this be helpful?

To ensure accuracy of chapters for audiobooks regardless of where they were sourced, the "Audible" branding should be removed from the calculations.

### Future Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/cf241e55-3466-4b07-b73c-fd65e6bc3a60)

### Audiobookshelf Server Version

v2.20.0

chriscam85:
Alternatively, since the majority of effort is due to the intro branding... Prep-populate the Shift Times with the negative value of brandIntroDurationMs. This way, you do the lookups, apply everything, then do shit times, and Add. The chapters would then be in a "no branding" state.

advplyr:
What's the duration and is it always the same?

chriscam85:
The duration is not always the same. 
As an example, the following three images are from the same series, and the last is another book (different series) which does **not** do the "This is Audible" at the start.

![Image](https://github.com/user-attachments/assets/0a663f28-7b57-4feb-ae8a-81e6264959ac)

![Image](https://github.com/user-attachments/assets/4f5bb4a4-9d69-497a-a100-17de27fdc44c)

![Image](https://github.com/user-attachments/assets/e51c2cf8-e8ff-4b1e-bd32-3e72ec833bc2)

![Image](https://github.com/user-attachments/assets/dbab5594-eba0-469f-aa33-9c654d1ad336)

advplyr:
Have you tested using the shift tool with the brand intro duration amount and it correctly lines up with the book you have?

I don't have any examples in my library to test with and the audible ones I do have with an intro the brandIntroDurationMs doesn't seem to make sense.

chriscam85:
Yes, in all my examples, using the shift tool with the intro duration (as negative) does correct the chapters. 
[...]

PR #4226, opened 2025-04-27
Changes to Chapter-Lookup

Vito0912:
## Brief summary

Adds a new checkbox to remove Audible (Intro and Outro) branding, for example, if the book was not obtained from Audible.
Fixes a bug where trying to subtract more time than the first chapter would not work, even after changing the time in the UI, unless you saved and reopened it.
Makes the error message for timeshifting clearer. 

## Which issue is fixed?

Discord

## In-depth Description

This only happens if the checkmark is checked (which is false by default):
Audible does use the branding times. If the checkbox is enabled, I subtract the branding time from all chapters to move them forward. If a chapter is inside the intro (which I have never seen in a book), it will be set to the next available second.
If a chapter is inside the outro, it is removed (because in that case the outro does not exist).

---

I fixed a bug where, if you tried to add more time than the length of the first chapter and then edited the time of the second chapter to subtract that time, it would still show an error.
The simple fix was to use the second chapter’s start time instead of the first chapter’s end time, since start times are updated on the fly.

I also moved the toast messages to new translation strings. I changed the toast string for the error message, which now better explains how you can subtract more time if you want to. 

## How have you tested this?

Lookup a few books

## Screenshots

grafik


nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/4152

Vito0912:
Thanks! Did not see that

chriscam85:
😍 Thank you!!!

advplyr:
Thanks!

PR #4240, opened 2025-04-29
Improves Year in Review display logic

josh-vin:
## Brief summary

Allows users created in the same year to see some stats on the "Year in Review" page.

## Which issue is fixed?

I could not find an issue regarding this. I could have maybe created one but just decided to fix it. 

## In-depth Description

I just started using AudiobookShelf and noticed that the stats page defaults to the previous year. It looks like there is some logic to allow users to select the current year but that is only if they have more than 1 available year.

I opted to keep that logic and only show previous year stats if those stats exist. In December users with more than one year will start to see their current year recap. New users will see previous year by default as soon as they pass into a new year after their account was created. 

## How have you tested this?

I loaded the page for "Your Stats" with my user createdAt timestamp set with a year in 2025 to verify that it would default to the current year .  (First image)

I then modified the DB for my user to have a createdAt timestamp from 2024. My stats page then pulled up with the default of the previous year with a drop down to select 2025. (Second image).

## Screenshots
image

image

Before the issue was fixed, this was the behavior for users that did not have any data from last year (they were created in the current year):
image

### DIFF
--- a/client/components/stats/YearInReviewBanner.vue
+++ b/client/components/stats/YearInReviewBanner.vue
@@ -164,14 +164,15 @@ export default {
   beforeMount() {
     this.yearInReviewYear = new Date().getFullYear()
 
-    // When not December show previous year
-    if (new Date().getMonth() < 11) {
+    this.availableYears = this.getAvailableYears()
+    const availableYearValues = this.availableYears.map((y) => y.value)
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #4253, opened 2025-05-02
Audiobook tools enhancements

### LINKED ISSUE #1257: [Enhancement]: More fine grained encoding (m4b merge tool)

Currently when using the encoder it automatically encodes for 64kbps and 2 channel but I find this to not always be the preferred encoding options. As an all around those settings are good especially for inputs that for example 128kbps the output is 50% reduced and almost lossless.
But for inputs 32kbps & 1 channel, it actually degrades quality from my testing and also increases file size considerably.

For example

Input: 
32kbps, 2050 Hz, 1 channel 
File Size: 223mb
input

Audiobookshelf output (`-map 0:a -acodec aac -ac 2 -b:a 64k -f mp4 output.m4b`): 
64kpbs, 22050hz, 2 channels. 
File Size: 386mb +60% increase

2-channel

Copy input codec (`-codec copy`):
 8kbps, 22050 Hz, 1 channel
File Size: 240mb +3% increase

codec-copy

From testing the above with a nice pair of headphones the 1 channel `-codec copy` sounded better than audiobookshelf 2 channel encoding. Of course, that's just preference but in any case the 60% increase in file size alone is not worth mapping the audio as 2 channel.

A feature to ffprobe the input and encode based on that I think would be preferred, for example above for CD quality 320kbps, 44100Hz, 2 channel current encoding is great but for 32kbps, 1 channel, just doing a copy I think is preferential.

A potential feature would be to show the user info about the input tracks quality and then offer multiple options to choose from when concatenating into an m4b

### LINKED ISSUE #4169: [Enhancement]: Add ability to combine/join M4B files without re-encoding

advplyr:
## Brief summary

This improves the m4b encoder tool default values and the UI/UX on the audiobook tools page.

## Which issue is fixed?

Fixes #4169 and fixes #1257

## In-depth Description

M4b tool improvements:

- Default to `copy` codec if all audio file codecs are AAC
- Select closest bit rate to one of 32k, 64k, 128k, 192k (instead of always 128k potentially increasing the size)
- Use mono channel if all audio files are mono (instead of always stereo)
- UI improvements
    - Better breakpoints for mobile and small screens
    - Use card with tabs for presets/advanced
    - Show bit rate, channels, and codec in the audio tracks table
    - Show the current bit rate, channels and codec below the presets (bit rate is averaged & channels use the max)

The preset defaults are best explained in the screenshots below.

## Screenshots

When audio tracks are mp3's the default is to encode to AAC
image

When not all audio track codecs are the same the default is to encode to AAC.
Note the bit rate rounds up to the nearest, this can be improved to add more presets
image

When all audio tracks are AAC then default to copy (don't re-encode).
image

When all audio tracks are mono channel then default to mono (if mixed the default is stereo)
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ sw.*
 .DS_STORE
 .idea/*
 tailwind.compiled.css
+tailwind.config.js
--- a/client/components/ui/ToggleBtns.vue
+++ b/client/components/ui/ToggleBtns.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="inline-flex toggle-btn-wrapper shadow-md">
-    <button v-for="item in items" :key="item.value" type="button" class="toggle-btn outline-hidden relative border border-gray-600 px-4 py-1" :class="{ selected: item.value === value }" @click.stop="clickBtn(item.value)">
+    <button v-for="item in items" :key="item.value" type="button" :disabled="disabled" class="toggle-btn outline-hidden relative border border-gray-600 px-4 py-1" :class="{ selected: item.value === value }" @click.stop="clickBtn(item.value)">
       {{ item.text }}
     </button>
   </div>
@@ -9,13 +9,17 @@
 <script>
 export default {
   props: {
-    value: String,
+    value: [String, Number],
     /**
      * [{ "text", "", "value": "" }]
      */
     items: {
       type: Array,
       default: Object
+    },
+    disabled: {
+      type: Boolean,
+      default: false
     }
   },
   data() {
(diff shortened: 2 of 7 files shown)


advplyr:
I added "tailwind.config.js" to the .gitignore because there is a bug with the VSCode tailwind plugin intellisense. For breakpoints like `md:text-lg` the intellisense wasn't working until I added an empty "tailwind.config.js" file and reloaded the extensions.

advplyr:
Screenshot of all UI updates:

image

kuldan5853:
> Use card with tabs for presets/advanced

Does this mean the UI will now be part of the interface like with Edit Medatata etc instead of opening in a fullscreen?

That would already be a nice improvement to the UX flow if true (and should then also be replicated for the metadata embedder). 

Next step then would probably a built-in encoding queue for multi-select "one button encodes"..

advplyr:
> Does this mean the UI will now be part of the interface like with Edit Medatata etc instead of opening in a fullscreen?

No this is still on the same page that you get to from tools. I'm not sure how else to build it that wouldn't be a separate page because there is too much to show in a modal.

I agree on the encoding queue

PR #4261, opened 2025-05-05
Fix .ignore file causing ignores outside the directory

nschum:
## Brief summary

The .ignore file handling based on prefix was incorrect in edge cases.

## Which issue is fixed?

The file "a/.ignore" should only cause the directory "a" to be ignored.
However, it also ignores all files starting with "a".

## In-depth Description

After this fix, it will only ignore paths starting with "a/".

## How have you tested this?

Added unit test.


advplyr:
Great, thanks for updating the unit test

PR #4275, opened 2025-05-10
Add missing series sequence to “Get a Library's Series” response

lcharlick:
## Brief summary

Adds a `sequence` value to each library item returned from the /libraries/:id/series endpoint.

## Which issue is fixed?

N/A

## In-depth Description

Not sure if this is intended or if the docs are just out of date, but the docs do indicate that books within a Book Series object should have a sequence tacked on. This PR adds that.

## How have you tested this?

Fairly minimal, but the change is low impact.

## Screenshots

N/A


advplyr:
The api docs are way out of date. I went ahead and updated it but there will be a dozen more out of date things like that in there.

The series is added to the book object so you can access the sequence there. No point in it being in 2 places

PR #4319, opened 2025-05-21
Audible confidence score

### LINKED ISSUE #4277: [Bug]: Match & Out of  Print Audible Titles
### What happened?

If you match (or worse, quick match) an Audible title that has been reissued or discontinued some "bad things" happen:

If it has been reissued: All information (including reader if it changed) will be updated to reflect the reissue.

If it hasn't been reissued and isn't available any longer if you're *lucky* nothing will happen but if the title matches too much you'll get a *completely* different book wiping out all of the information about that book.

If you're doing a match, you can catch these before the "damage" is done.  With a quick match you will end up doing a lot of fixing.

The software needs to make some better decisions.

If matching and the ASIN doesn't match, do not change any records.  

If the ASIN isn't present in the current audiobookshelf record; double check title & author (and series if applicable).   

This is a bit difficult to reproduce as you would have to own a discontinued or revised titles. 

I will give examples for both cases.

In 2013 Audible released a science-fiction audiobook by Michael McCollum titled "The Clouds of Saturn". (ASIN: B00B89GKI8).    Doing a quick match or match in audiobookshelf has it match with a book called "Saturn's Story" by Jill Logan. (ASIN: B0DYFR8CYC)

In 2017 Audible released "Cold Welcome" by Elizabeth Moon (ASIN: B06Y2H1WCT) with narrator Brittany Pressley.   In 2024 this book was reissued with a new recording (ASIN B0D8517NWS) with narrator Carrie Coello.    No matter what the ASIN and release year are set in audiobookshelf, it will switch the meta information to the reissue.

### Audiobookshelf version

v2.21.0

Docker

None

### Logs

```shell

```

### Additional Notes

Synology host (Debian variant)

mikiher:
## Brief summary

This adds a match confidence score when matching audiobooks using Audible.

## Which issue is fixed?

Fixes #4277 (eventually).

## In-depth Description

When returning match results from Audible, a new match confidence score is calculated based on book duration, title and subtitle, and author.

Details: 
- The duration, title, and author scores are linearly combined
  - duration is weighted 0.7, title 0.2 and author 0.1
- Duration is given 1.0 score if diff is less than 1 minute, then the score drops linearly to 0.6 if diff is 5 mintues, and then more sharply to 0.0 when diff is 10 minutes.
- The title score is calculated by comparing the book title to the actual title query. 
  - If the title query seens to have a subtitle,  we compare it to the book title + subtitle
- The author score is calculated by comparing the book author to the actual author query
  - If the book author has multiple authors (separated by commas), we try to compare against each of the authors and take the maximum score
- if the title is an ASIN, a 1.0 match confidence score is returned (the duration, title, and author scores are not calculated).

The combined score is written to the result record.

I added some code to display it in BookMatchCard.vue (only so you can easily check it - this is not necessarily how we want to display it)

At this point, I haven't yet added the quick match dialog support. This is just for you to comment on the confidence score.

## How have you tested this?

Added extensice unit testing, and ran through many books to see what kinds of score we get. It looks like a threshold of 0.85-0.9 could be used for bypass the intended quick match dialog.

### DIFF
--- a/client/components/cards/BookMatchCard.vue
+++ b/client/components/cards/BookMatchCard.vue
@@ -16,6 +16,7 @@
         <p v-if="book.author" class="text-gray-300 text-xs md:text-sm">{{ $getString('LabelByAuthor', [book.author]) }}</p>
         <p v-if="book.narrator" class="text-gray-400 text-xs">{{ $strings.LabelNarrators }}: {{ book.narrator }}</p>
         <p v-if="book.duration" class="text-gray-400 text-xs">{{ $strings.LabelDuration }}: {{ $elapsedPrettyExtended(bookDuration, false) }} {{ bookDurationComparison }}</p>
+        <p v-if="book.matchConfidence" class="text-gray-400 text-xs">{{ $strings.LabelMatchConfidence }}: {{ book.matchConfidence.toFixed(3) }}</p>
         <div v-if="book.series?.length" class="flex py-1 -mx-1">
           <div v-for="(series, index) in book.series" :key="index" class="bg-white/10 rounded-full px-1 py-0.5 mx-1">
             <p class="leading-3 text-xs text-gray-400">
--- a/client/strings/en-us.json
+++ b/client/strings/en-us.json
@@ -425,6 +425,7 @@
   "LabelLogLevelWarn": "Warn",
   "LabelLookForNewEpisodesAfterDate": "Look for new episodes after this date",
   "LabelLowestPriority": "Lowest Priority",
+  "LabelMatchConfidence": "Match Confidence",
   "LabelMatchExistingUsersBy": "Match existing users by",
   "LabelMatchExistingUsersByDescription": "Used for connecting existing users. Once connected, users will be matched by a unique id from your SSO provider",
   "LabelMaxEpisodesToDownload": "Max # of episodes to download. Use 0 for unlimited.",
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -7,7 +7,7 @@ const FantLab = require('../providers/FantLab')
 const AudiobookCovers = require('../providers/AudiobookCovers')
 const CustomProviderAdapter = require('../providers/CustomProviderAdapter')
 const Logger = require('../Logger')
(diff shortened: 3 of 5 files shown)


mikiher:
Again, at this point please don't merge, this is just to review confidence score calculation.

advplyr:
I'm playing around with this but I don't have good real-world test data.
The biggest benefit I can see from implementing this is preventing quick match from matching with a book with a low confidence score.

mikiher:
Yes, I wanted to get your input on the score itself. I will finish implementing the quick match reject dialog for low confidence scored audiobooks when I get back from vacation next week.

advplyr:
https://github.com/advplyr/audiobookshelf/pull/4383 introduced [fusejs](https://github.com/krisk/Fuse) that I just tested out and it works very well at fuzzy matching. I included the [basic build](https://www.fusejs.io/getting-started/different-builds.html) which is probably a better option than levenshtein

mikiher:
It's a bit hard to assess which matching score is better since the search algorithms don't work the same, but I would hazard to say that, when searching over book provider search results, they likely return the same ordering in most cases except extreme ones. 

The reason is that the search itself (on the provider side) is usually done by more "stupid" algorithms that don't use fuzzy matching, which is why I need to sometime run a few searches to even get any results. The search and sorting over the result set returned by the provider can probably slightly benefit from the fuse.js search, but not much, since those results probably all exactly match the search query (due to the provider's "stupid" search algorithm), and so the fuzzy matching advantages of fuse.js bitap algorithm are probably not realized.

Overall, I'd say that fuse.js is likely more suitable to cases in which the set over which it searches is unfiltered (like the list of podcast episodes, for example).

Also, in the case of Audible results, most of the score is determined by the duration diff, which is something that fuse.js matching doesn't know how to consider.

I need to set parameters and test extensively before I consider switching, and will not do this as part of this PR.

advplyr:
I updated the confidence score to be inside a badge. The badge is green when the score is > 95% and blue otherwise. I chose that number from the tests on my dev library. We can fine tune it down the road.

image

image

advplyr:
Thanks!

Issue #1869, opened 2023-06-25
[Enhancement]: Enable Custom Date-Time Stamp in File Name


Proponent-8247:
### Describe the feature/enhancement

Would it be possible to implement a custom date-time stamp into the file name when downloading files?
Something like how this tool does it? https://github.com/cnovel/PodcastBulkDownloader
This is due to many of the podcasts I listen to don't have proper filename tagging.

erichintime:
This would be great. As stated above, podcast naming is all over the place even within the same feed. 

Being able to prepend a published date to the start of the filename like `YYYY-MM-DD {episode title}.mp3` would be so helpful for large podcast archives and make archives more portable to other players. 

[Podgrab](https://github.com/akhilrex/podgrab) also implements this, though it's not customizable - just on or off in the format I used as an example.

PR #4331, opened 2025-05-24
Feature: Allow to have custom podcast filenames

### LINKED ISSUE #2494: [Enhancement]: Add episode number and date to file name

Adding the episode number and (or - configurable) date to downloaded file name.
After downloading (archiving) my favourite podcast, its folder looks like a huge mess inside without indicating date or number of the episode.

I've seen alternative podcast app, where you can configure to prepend episode number and date to downloaded file name.
Adding this would be a huge improvement. Optionally renaming all previous files (without redownloading from scratch).

Implementing this would greatly improve organization inside podcast folder, which currently can be only alphabetically ordered, and really makes no sense when accessed from outside of audiobookshelf gui (for example windows explorer, when I play in my mp3 player)

Example of such file names:
```
<episode_num>_<YYYY-MM-DD>_<Episode_name>.<ext>
004_2024-01-03_Hello.mp3
003_2024-01-01_World.mp3
002_2023-10-17_Lorem.mp3
001_2022-11-15_Ipsum.mp3
```

My favourite podcast folder currently looks like this:
abs_folder_inside

MagiX13:
## Brief summary

This PR adds the ability to define filename templates for podcasts, as mentioned in #2494 

## Which issue is fixed?

Fixes #2494 and #1869 

## In-depth Description

The filename format template can be defined on a per podcast level, to facilitate distinctions between podcasts with season/episode and those without. The template is a string with some variables that are replaced with the actual content in case the respective variable is available.

The tooltip (see below) explains the options that are available.

## How have you tested this?

I have used this for a few of my podcasts and it works well. I however don't have a podcast that has season/episode format so couldn't test it here

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf-client",
-  "version": "2.23.0",
+  "version": "2.23.1",
   "buildNumber": 1,
   "description": "Self-hosted audiobook and podcast client",
   "main": "index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf",
-  "version": "2.23.0",
+  "version": "2.23.1",
   "buildNumber": 1,
   "description": "Self-hosted audiobook and podcast server",
   "main": "index.js",
--- a/server/models/Podcast.js
+++ b/server/models/Podcast.js
@@ -39,6 +39,8 @@ class Podcast extends Model {
     /** @type {string} */
     this.language
     /** @type {string} */
+    this.podcastFilenameFormat
+    /** @type {string} */
     this.podcastType
     /** @type {boolean} */
     this.explicit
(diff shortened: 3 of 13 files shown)


client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf-client",
-  "version": "2.23.0",
+  "version": "2.23.1",
```
nichwall, line 3: Don't update the version number, that will be taken care of when there is a new release.

Similarly, you should not need to update the package locks.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf",
-  "version": "2.23.0",
+  "version": "2.23.1",
```
nichwall, line 3: Same as previous comment

server/models/Podcast.js
```diff
@@ -131,6 +134,7 @@ class Podcast extends Model {
         itunesId: DataTypes.STRING,
         itunesArtistId: DataTypes.STRING,
         language: DataTypes.STRING,
+	      podcastFilenameFormat: DataTypes.STRING,
```
nichwall, line 137: Do you have Prettier installed? That will auto format the code.

nichwall (changes requested):

client/package.json
MagiX13: This comes with some issues as the database schema needs to be adjusted in order for the features - and audiobookshelf itself to work as some attributes don't exist without migrating.
reverted back to 2.23.0

nichwall: Yes, that is correct. The migration will not occur because it is still the same version, but the version bump is the last commit that happens before a release

MagiX13: Makes sense to do it like that.
However this implies that building it under 2.23.0 with the code merged crashes fatally

server/models/Podcast.js
MagiX13: Done

PR #4358, opened 2025-06-01
feat(docker): Add Healthcheck

### LINKED ISSUE #4352: [Enhancement]: Update either documentation or docker file for audiobookshelf health test
### Type of Enhancement

Documentation

Hi,
there is documentation for audiobookshelf as docker container under
https://github.com/advplyr/audiobookshelf/pkgs/container/audiobookshelf
is insufficient.
There is no valid documentation on how to set up a health check in the compose.yml file to have docker check the audiobookshelf container.
Apparently there is something like http://localhost:13378/healthcheck but it can't be curl-ed since there is no curl in the container and no sufficient other documentation, on how to use it.

Please, elaborate.

Best regards
Heinrich

I want to setup the healthcheck for the docker container.

Something like this should work in the docker compose.yml
    healthcheck:
       test: "curl -f --fail http://localhost:13378/healthcheck || exit 1"
       interval: 30s
       timeout: 3s
       retries: 3

v2.23.0 (docker)

_No response_

Billos:
## Brief summary

This PR adds a simple Healthcheck, querying the healthcheck endpoint

## Which issue is fixed?

Fixes #4352

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,8 +50,7 @@ ARG NUSQLITE3_PATH
 RUN apk add --no-cache --update \
   tzdata \
   ffmpeg \
-  tini \
-  curl
+  tini
 
 WORKDIR /app
 
@@ -70,7 +69,7 @@ ENV SOURCE="docker"
 ENV NUSQLITE3_DIR=${NUSQLITE3_DIR}
 ENV NUSQLITE3_PATH=${NUSQLITE3_PATH}
 
-HEALTHCHECK CMD curl --fail http://localhost:${PORT}/healthcheck || exit 1
+HEALTHCHECK CMD wget --spider -q http://localhost:${PORT}/healthcheck || exit 1
 
 ENTRYPOINT ["tini", "--"]
 CMD ["node", "index.js"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,8 @@ ARG NUSQLITE3_PATH
 RUN apk add --no-cache --update \
   tzdata \
   ffmpeg \
-  tini
+  tini \
+  curl
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


nichwall:
Wget is available in the container, there is not a need to add curl.

https://github.com/advplyr/audiobookshelf/issues/4352#issuecomment-2923585403

Billos:
Woops, changing this !

nichwall:
Generally the server starts up quickly, but we do not know the startup duration of any individual server. Database migrations can also take a while depending on a specific user setup and database/library structure.

I don't think this should be part of the docker file and should be up to individual users to configure.

DieselTech:
What is the point of adding a health check if you're not doing anything with the data it generates? If you're not monitoring the health check and taking corrective actions, then "monitoring" it is useless and just causes more overhead. All this kind of check does is see if the web server is up and responding to web requests. It does not test the data coming from the server or ensure things are actually working.

What your doing by baking this into the `dockerfile` is forcing everyone to use healthchecks even if they don't want them. Then those users need to specifically disable them. Where instead it should be that if you want the health check, you can add it yourself in your docker compose file.

CarnsJalone:
> What is the point of adding a health check if you're not doing anything with the data it generates? If you're not monitoring the health check and taking corrective actions, then "monitoring" it is useless and just causes more overhead. All this kind of check does is see if the web server is up and responding to web requests. It does not test the data coming from the server or ensure things are actually working.
> 
> What your doing by baking this into the `dockerfile` is forcing everyone to use healthchecks even if they don't want them. Then those users need to specifically disable them. Where instead it should be that if you want the health check, you can add it yourself in your docker compose file.

This appears to be a `shallow` health check, which is well defined within the field of container orchestration. However, you are somewhat correct. This health check will only confirm is the process is alive and can respond to requests. In some cases, when a process has become zombified (segfaults), restarting will resolve the underlying issue. In others (configuration, data) this will add additional overhead that will not benefit the process by restarting it. 

Does ABS expose a somewhat deeper health check which can introspect on internals to truly determine the root cause of the health failure so that savvy end-users can resolve their issues? This could be anything from permission issues, corrupted data, incorrect configuration, etc...

There is a great [RFC Draft](https://datatracker.ietf.org/doc/html/draft-inadarei-api-health-check-06) (which has since expired) that denotes guidance for how this `could` be implemented.

nichwall:
> Does ABS expose a somewhat deeper health check which can introspect on internals to truly determine the root cause of the health failure so that savvy end-users can resolve their issues? This could be anything from permission issues, corrupted data, incorrect configuration, etc...

No, the healthceck just returns "okay" if the server is working. The admin will need to look at the crash logs or normal server logs to gain additional information.

PR #4361, opened 2025-06-02
Fix: Audiobook m4b advanced encoder ignore

### LINKED ISSUE #4359: [Bug]: Audiobook m4b advanced encoder ignore
### What happened?

When trying to convert mp3, I specify the parameters in the advanced encoder, but they are ignored. ffmpeg runs with preset parameters

I would like the advanced encoder parameters to be accepted when performing the conversion

1. Open "Make M4B Audiobook File"
2. Tab "Advanced"
3. Set custom options
4. Start encode

### Audiobookshelf version

v.2.24.0

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

Vito0912:
## Brief summary

See detailed

## Which issue is fixed?

Fixes #4359

## In-depth Description

See code :). Just returns the custom values if you are viewing that option tile. It also sets the encoding options after pressing the button to show the information correctly while the task is running.

## How have you tested this?

Local Server

### DIFF
--- a/server/managers/AbMergeManager.js
+++ b/server/managers/AbMergeManager.js
@@ -203,7 +203,15 @@ class AbMergeManager {
     // Move library item tracks to cache
     for (const [index, trackPath] of task.data.originalTrackPaths.entries()) {
       const trackFilename = Path.basename(trackPath)
-      const moveToPath = Path.join(task.data.itemCachePath, trackFilename)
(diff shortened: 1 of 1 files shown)


advplyr:
While testing I came across a separate issue where if re-encoding a single m4b the backup was overwriting the newly encoded m4b because they had the same name. I added a check and appended `.backup` to the filename before the extension.

Thanks!

PR #4370, opened 2025-06-04
Adds ENV for MaxFailedEpisodeChecks

Vito0912:
## Brief summary

Adds ENV (`MAX_FAILED_EPISODE_CHECKS`) for MaxFailedEpisodeChecks. 0 skips the check.

## Which issue is fixed?

No particular, but this would give the ability to higher the amount if someone does hourly checks e.g. (or lower it) or just disable it if the person does not want any checks.

> [!NOTE]
> Currently not tested! Should work, because minor fixes, but will test in the upcoming days (and update with a comment if so)

### DIFF
--- a/server/managers/PodcastManager.js
+++ b/server/managers/PodcastManager.js
@@ -30,7 +30,7 @@ class PodcastManager {
     this.currentDownload = null
 
     this.failedCheckMap = {}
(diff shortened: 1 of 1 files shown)


Vito0912:
@advplyr Now tested and moved to globals as per other envs

advplyr:
Thanks!

Issue #4098, opened 2025-03-10
[Bug]: Podcast cron infinite hang


SuperEllipsis:
### What happened?

Sometimes when the server has been running for a while, it seems to hang on it self, with a simple message of "[CronManager] Podcast cron "0 * * * *" is already executing", it will not download more podcasts from the RSS feeds automatically, until the docker container is restarted. I can still go through each manually and download the episodes from the RSS feed

![Image](https://github.com/user-attachments/assets/ef4efd3c-b793-466d-a6d4-8129229bb52f)

### What did you expect to happen?

I expected it to potentially throw an error/warning in the logs, if it was hanging, and then retry later

### Steps to reproduce the issue

I am unsure of when and how it happens, but usually it is just
1. Setup podcasts with RSS feed to automatically try to pull every hour (standard)
2. Wait a couple of days
3. Automatic download stops working

### Audiobookshelf version

v2.19.5

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

SuperEllipsis:
Good catch, I will do that. And update, if I find something in the debug logs

SuperEllipsis:
I restarted the server, and set the log level to debug. And a couple of days later, it failed again.
Here is the log from the day, where the "Podcast cron \"0 * * * *\" is already executing" starts to appear
[2025-03-13.txt](https://github.com/user-attachments/files/19274923/2025-03-13.txt)

Specifically, here is a short list of the logs around the problem area are (full logs in file above):
```log
{"timestamp":"2025-03-13 02:00:51.069","source":"CronManager.js:208","message":"[CronManager] Finished executing podcast cron 0 * * * * for 22 item(s)","levelName":"DEBUG","level":1}
{"timestamp":"2025-03-13 03:00:00.099","source":"CronManager.js:179","message":"[CronManager] Start executing podcast cron 0 * * * * for 22 item(s)","levelName":"DEBUG","level":1}
[...]

PR #4373, opened 2025-06-04
Potential fix/new knowledge for hangig podcasts

Vito0912:
## Brief summary

From the logs, I strongly suspect that the timeout for the feed request never triggers, which causes the code to hang. As a result, it is never removed from the executing list and is not executed again. To check this (since it is possible that the ssrf filter interferes with the timeout), I added a promise that will throw after the timeout to ensure it works. If this works, we will know that it has something to do with axios and can implement a better fix. This change does not affect any behavior and still allows axios to fail correctly (with an extra second), so it should be safe to merge.

## Which issue is fixed?

Potentially fixe s https://github.com/advplyr/audiobookshelf/issues/4098

## How have you tested this?

Own podcasts with 1 minute cron and custom timeout values


advplyr:
Good idea, thanks!

PR #4374, opened 2025-06-04
Corrects removing of attachments for Trix

Vito0912:
## Brief summary

Currently, files can be added to the editor, which I think is nice and could be brought back in the future. However, because the description is sanitized,(btw. ignore name of branch. Not possible because of sanitization). I noticed that files can be pasted, including images, so maybe we should consider allowing files in the future, but right now, sanitization removes them when the description is saved. To get parity now all attachments are disallowed, also including files

Maybe there's a reason we allowed files that I missed (in that case, please close this), but at the moment, you can add files (such as images or Excel files) that are just lost after saving.

## Which issue is fixed?

None

## In-depth Description

See description

## How have you tested this?

Copy pasting images and files

### DIFF
--- a/client/components/ui/VueTrix.vue
+++ b/client/components/ui/VueTrix.vue
@@ -318,10 +318,8 @@ export default {
       }
     },
     handleAttachmentAdd(event) {
-      // Prevent pasting in images from the browser
-      if (!event.attachment.file) {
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

Issue #4377, opened 2025-06-05
[Enhancement]: Warning when RSS feed fails to load


miltuss:
### Describe the Feature/Enhancement

It is not possible to know if a podcast RSS feed is not working in the current state of ABS. No warning.

### Why would this be helpful?

Add a warning when an RSS feed fails to load. Just like there are warnings when an ebook or audiobook is missing.

mfcar:
Something similar It is being worked on in #2178

PR #2178, opened 2023-10-03
Improve RSS Manager adding Incoming Feed Health Check

mfcar:
This pull request introduces a new feature that improves podcast management by allowing users to check the health status of feed URLs for incoming podcasts:

1. **Feed Health Status**: Users can now easily determine whether a podcast's feed URL is healthy or not. Every time the AudioBookShelf accesses the feed URL, it performs a verification and marks the feed as either healthy or not.

2. **Manager Page**: A new tab has been added on the Confi -> RSS Feeds, where users can view all the podcasts that contain a feed URL. This screen provides the following information:

   - **Podcast Title**
   - **Feed URL**: The Feed URL. It's truncated because sometimes the URL is gigantic
   - **Last Checked**: It's the last date and time that the Feed URL was checked
   - **Last Successful Check Date** (new table column): Displays the date and time of the application able to access the feed URL with success
   - **Feed Healthy** (new table column): Indicates whether the feed URL is marked as healthy or not.
   - **Next Automatic Check**: If the feed has a cron, show the next time that the feed will be checked
   - **Force Recheck**: Allows users to manually trigger a recheck of the feed's health status.

**Reason for the Changes**

This feature was added to provide users with greater transparency and control over the health of incoming podcast feed URLs. It empowers users to make informed decisions about which feeds to subscribe to based on their health status.

**How to Test**

To test this feature, follow these steps:

1. Navigate to the Config --> RSS Feeds.
2. Check the list of podcasts with associated feed URLs.
2. Click the "Force Recheck" button for a podcast to trigger a manual recheck.
3. Verify that the health status is updated accordingly.

**Additional Notes**

- Every time AudioBookShelf accesses the feed, it saves the feed's healthy status (e.g., for downloading new episodes or displaying them on the Podcast Detail page).

**What's missing to finish**

- [ ] The podcast cover arts are not being displayed on the page.

- [ ] I attempted to use the Podcast model, but I am unsure whether I should use the LibraryItem old method or the Podcast Model. As a result, when I execute the following code:

```
await Database.podcastModel.update(podcast, {
 where: { id: podcastId }
 })
```

The update is temporarily saved, but it does not appear to be saved in the database.

-----

The new Page working

https://github.com/advplyr/audiobookshelf/assets/814828/ea3e45af-7ff2-4bcb-a93a-7b947e2a834b

-----

Subscribed Feeds page

image

-----

The RSS Page has been renamed to Opened Feeds and now it's a tab on the RSS Page, that includes a new search box to filter by title.

[...]

PR #4378, opened 2025-06-05
Notifications for failed rss feeds and disabled rss feeds

Vito0912:
## Brief summary

Adds notifications for failed notifications and a separate type that triggers when an RSS feed is disabled because of this action.

## Which issue is fixed?

None, but helps with https://github.com/advplyr/audiobookshelf/issues/4377 and while https://github.com/advplyr/audiobookshelf/pull/2178 is not merged this also gives a way to track disabled podcasts

## In-depth Description

Code ^^

## How have you tested this?

Setting timeout to a ms and using the testpayload


advplyr:
I struggled to find a good way to word this because I wanted to make sure this says it is specifically for scheduled podcast episode download checks. Not to be confused with manual checks by the user or RSS feeds opened in Abs.

Also, with the way that dropdown is built the description had to stay on one line to avoid updating that component.

image

Thanks!

Issue #241, opened 2021-12-14
Socket failed to connect


dietrichmd:
Fresh install using docker-compose and apache2.  I can access site and login, but everything I try to do is greeted with a Socket Failed to Connect error.

Please advise.

Thanks

dietrichmd:
here is proxypass info

   ProxyPreserveHost On 
    ProxyPass /.well-known/ !
ProxyPass / http://192.168.1.11:4040/
ProxyPassReverse /.well-known/ !
ProxyPassReverse / http://192.168.1.11:4040/
Alias / http://192.168.1.11:4040/
RewriteEngine on
    RewriteCond %{HTTP:Upgrade} websocket [NC]
    RewriteCond %{HTTP:Connection} upgrade [NC]
    RewriteRule ^/?(.*) "ws://localhost:4040/$1" [P,L]

dietrichmd:
Solved:  proxy_wstunnel_module was not enabled.  For those looking, its a2enmod proxy_wstunnel

advplyr:
Running a reverse proxy? SSL?

silentArtifact:
Yeah, reverse proxy set up through the DMS os itself, same with SSL. I've got the same thing done with Plex and a bunch of other containers, so I'm not sure what I should do with this.

advplyr:
The difference is that audiobookshelf requires a web socket connection, so the reverse proxy config will probably not be the exact same as other media servers. In NGINX proxy manager there is a toggle for web sockets support.
![image](https://user-images.githubusercontent.com/67830747/153679106-b2a7f5b9-0702-48c6-9740-b26b401986e9.png)

silentArtifact:
Agreed. For synology, I followed these steps:

1. Open Control Panel > Application Portal
2. Change to the Reverse Proxy tab
3. Select the proxy rule for which you want to enable Websockets and click on Edit
4. Change to the "Custom Header" tab
5. Click Create > WebSocket
6. Click Save

Now it works!

advplyr:
Awesome, thanks for following up with your steps. It is still a bit messy, but I added your steps for synology and the screenshot for nginx proxy manager in the readme.
https://github.com/advplyr/audiobookshelf/blob/master/readme.md#reverse-proxy-set-up

silentArtifact:
Hey, I should be thanking _you_ for all the work you've done on this! It's exactly what I've been looking for, and I seriously appreciate all your effort!

advplyr:
[...]

PR #4381, opened 2025-06-06
fix(Dockerfile): Wire ROUTER_BASE_PATH to build arg

treyturner:
## Brief summary

Create a clear interface for building an image with an empty or alternate `ROUTER_BASE_PATH` without needing to modify source, shell into containers, or rebuild at all.

## Which issue is fixed?

Not sure if I'd consider this a fix per-se, but it may ease tension on sentiments along the lines of #3874 

## In-depth Description
The best I could make of the [saga of this change](https://github.com/advplyr/audiobookshelf/issues/3874#issuecomment-2614154863) given my available time is that if one desires a containerized instance of audiobookshelf with an empty `ROUTER_BASE_PATH`, they're responsible for building it themselves.

I haven't tested extensively yet as I'm still getting familiar with some of the features, but if I've understood the process recommended for doing so at present, it should work to set `ROUTER_BASE_PATH` empty before building the apps. I was surprised then to see then this variable wasn't wired up to a Docker build arg where it can be easily set from outside the build or containers without modifying the source:

```sh
docker build --build-arg ROUTER_BASE_PATH= . -t audiobookshelf:no_router_base_path
```

This PR was created both to ask the experts whether this is a complete and valid approach, and if so, offered as a small-diff pressure-release valve for people like myself who were initially flustered about the opinionated nature of this change.

## How have you tested this?
I tested local and remote access via browser, the official Android app, and RSS feeds via PodcastAddict.

## Screenshots
abs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,13 @@
 ARG NUSQLITE3_DIR="/usr/local/lib/nusqlite3"
 ARG NUSQLITE3_PATH="${NUSQLITE3_DIR}/libnusqlite3.so"
+ARG ROUTER_BASE_PATH="/audiobookshelf"
 
 ### STAGE 0: Build client ###
 FROM node:20-alpine AS build-client
+
+ARG ROUTER_BASE_PATH
+
+ENV ROUTER_BASE_PATH=${ROUTER_BASE_PATH}
 
 WORKDIR /client
 COPY /client /client
@@ -14,8 +19,10 @@ FROM node:20-alpine AS build-server
 
 ARG NUSQLITE3_DIR
 ARG TARGETPLATFORM
+ARG ROUTER_BASE_PATH
 
 ENV NODE_ENV=production
+ENV ROUTER_BASE_PATH=${ROUTER_BASE_PATH}
 
 RUN apk add --no-cache --update \
   curl \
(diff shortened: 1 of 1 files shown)


treyturner:
Thought I might have had some problems loading content after this change, but it ended up being [`okhttp` caught up in a recently implemented bot-blocking ACL on my reverse proxy](https://github.com/advplyr/audiobookshelf/issues/241#issuecomment-2603227637).

treyturner:
Ok, I finally got a chance to test this again and it does work. Does it seem like a reasonable approach to improving configurability?

fmaass:
Confirming this works. I hit the `/audiobookshelf`-baked-into-the-client problem (#3874, and the OIDC facet in #5287) serving ABS at a bare domain behind Traefik.

Built v2.35.1 with exactly this diff and `docker build --build-arg ROUTER_BASE_PATH=`:

- Client bakes `<base href="/">`, assets at `/_nuxt/*` (no `/audiobookshelf` prefix).
- Behind a reverse proxy routing the domain root → container, it serves cleanly at root.
- OIDC (Authelia) now works end-to-end from the bare domain: `authOpenIDAutoLaunch` redirects straight to the IdP on first visit (with the stock image, auto-launch does **not** fire at the bare root — you get the manual SSO button every time), `redirect_uri` is `https://host/auth/openid/callback` which matches the IdP registration, and the same-origin callback validation (#5287) accepts root callbacks.
- Verified in a real browser: bare domain → SSO → logged in, clean URL throughout.

It's a minimal, opt-in build-arg (defaults to `/audiobookshelf`, so existing builds are unaffected) — the cleanest way to get a true root deployment without patching source. Would be great to get this reviewed/merged.

PR #4383, opened 2025-06-06
Use fuse.js for podcast episode search

### LINKED ISSUE #4382: [Enhancement]: Improve podcast episode matching
### Type of Enhancement

Server Backend

When manuallu matching an audio file to a podcast episode in Audiobookshelf, users can search for the correct episode by typing part of its title. However, the current search behavior seems to rely on Levenshtein distance, which doesn't work well with partial or numeric queries. For example, searching for "212" to find an episode titled "Super Podcast - Episode 212 - How to start writing book" returns no results, likely because the distance score is too high. I would be great to use fuzzy search instead.

A suggested improvement would be use fuzzy search algorithm. This would help users find episodes even if they only remember a number or a keyword from the title.

In example episode name "Super Podcast - Episode 212 - How to start writing book" should be found by searching "212" or "How to stat writig"

v2.24.0

_No response_

JKubovy:
Replace levenshtein distance with fuse.js fuzzy searching library. Search in episode's title and subtitle

## Brief summary

This PR improves the manual podcast episode search by replacing the Levenshtein distance algorithm with [Fuse.js](https://github.com/krisk/fuse), a lightweight fuzzy search library. The result is a more intuitive and forgiving search experience when linking audio files to episodes.

## Which issue is fixed?

Fixes #4382 

## In-depth Description

Previously, episode search relied on Levenshtein distance, which often failed to return results when users searched using partial titles, numbers, or keywords — e.g., searching "212" for "Super Podcast - Episode 212 - How to start writing book".

This PR replaces that logic with Fuse.js, a fuzzy matching library. Fuse allows more flexible search by ranking matches based on relevance instead of string similarity distance. With this change, users no longer need to copy full or exact episode titles to find a match.

Fuse.js is added as a new dependency. It's licensed under Apache 2.0, which is compatible with the project’s GPLv3 license.

While this update is currently applied only to episode search, Fuse.js could potentially be reused in other areas of the app where improved search UX is desired.

## How have you tested this?

- Manually add podcast and download one episode
- In edit dialog (Match tab) a try to search multiple episode names
- Verified that relevant results are returned even with partial input


advplyr:
The fuse library is nice. We are only using the [basic build](https://www.fusejs.io/getting-started/different-builds.html) and so by pulling out the basic min js file we are only adding ~18KB instead of ~500KB.

My one issue was with quick matching where we use the first result that gets returned from that function. I did some testing and I thought it made sense to lower the threshold to `0.1` for quick match.

Thanks!

PR #4386, opened 2025-06-07
Add clickable seek-to-time links for timestamps in descriptions

AemonCao:
## Brief summary

This PR introduces a new feature that automatically detects timestamps (e.g., `59:23` or `01:03:55`) in an item's description and converts them into clickable links. When a user clicks on one of these links, the player will seek to that specific time in the audio.

This enhances the user experience by allowing creators to easily reference parts of the audio in the description, and listeners to jump directly to those points of interest.

## In-depth Description

1.  A new computed property, `processedDescription`, uses a regular expression to find all timestamp patterns in the item's HTML description.
2.  Each found timestamp is wrapped in an `<a>` tag with a `data-time-seconds` attribute containing the time converted to total seconds.
3.  An event delegation pattern is used on the description container (`@click="handleDescriptionClick"`). This avoids attaching numerous event listeners and efficiently captures clicks on the dynamically generated links.
4.  When a timestamp link is clicked, the handler extracts the time in seconds from the `data` attribute and calls the existing `timestampClick` method to seek the player.

### Before

`39:14-53:42 My favorite part`

image

### After

`<a href="#">39:14</a>-<a href="#">53:42</a> My favorite part` (Clicking "39:14" seeks the audio to 39 minutes and 14 seconds).

image

https://github.com/user-attachments/assets/d2b034f7-327d-410f-ac09-d65fe13caf98

### DIFF
--- a/client/components/modals/podcast/ViewEpisode.vue
+++ b/client/components/modals/podcast/ViewEpisode.vue
@@ -16,7 +16,7 @@
         </div>
       </div>
       <p dir="auto" class="text-lg font-semibold mb-6">{{ title }}</p>
-      <div v-if="description" dir="auto" class="default-style less-spacing" v-html="description" />
+      <div v-if="description" dir="auto" class="default-style less-spacing" @click="handleDescriptionClick" v-html="processedDescription" />
       <p v-else class="mb-2">{{ $strings.MessageNoDescription }}</p>
 
       <div class="w-full h-px bg-white/5 my-4" />
@@ -70,6 +70,9 @@ export default {
     description() {
       return this.episode.description || ''
     },
+    processedDescription() {
+      return this.linkifyTimestamps(this.description)
+    },
     media() {
       return this.libraryItem?.media || {}
     },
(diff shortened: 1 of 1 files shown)


Vito0912:
~~Like the idea of interactive descriptions, but isn't that what chapters are for, which is directly below the description and has also jumpable markers?~~
~~Especially since this will convert things that are not meant to be timestamps~~

It's for podcasts. Nvm, good idea imo

AemonCao:
Possible mismatching scenarios:

1. Bible Verses
  `Genesis 1:10`

2. Ratios or Scores
`The final score was 2:10.`
`Adjust the aspect ratio to 16:09.`

AemonCao:
> Possible mismatching scenarios:
> 
> 1. Bible Verses
>    `Genesis 1:10`
> 2. Ratios or Scores
>    `The final score was 2:10.`
>    `Adjust the aspect ratio to 16:09.`

I added some conditionals in [`07432cc` (#4386)](https://github.com/advplyr/audiobookshelf/pull/4386/commits/07432cc26c260d592744d1079ddb270a07d5bc05), which can reduce the possibility of false matches.

advplyr:
This was implemented in the mobile app here https://github.com/advplyr/audiobookshelf-app/pull/1208

I tested yours and it has the same issue with Syntax [that I mentioned](https://github.com/advplyr/audiobookshelf-app/pull/1208#issuecomment-2119358941)

I think we should use the regex and method of parsing that the app is using since it has been in place for a year now without any issues.

I prefer your method of attaching a click listener to the outer div instead of adding a listener to each timestamp. We should update the mobile app to do that as well.

advplyr:
I implemented this using the regex from the mobile app here https://github.com/advplyr/audiobookshelf/pull/4419

Thanks!

PR #4393, opened 2025-06-10
Fix filesystem pathexists path join

advplyr:
## Brief summary

This is a follow up to #4342 to fix an edge case

## Which issue is fixed?

No issue

## In-depth Description

When using `Path.posix.join` with a posix file path and a Windows file path the resulting join isn't normalized.

For example:
```js
Path.posix.join("C:/Users/Test/audiobooks", "..\\test")
```

returns `C:/Users/Test/audiobooks/..\test`

`Path.join` returns `C:/Users/Test/test` as expected

This is an issue because `fs.pathExists` is normalizing the posix and non-posix path

## How have you tested this?

Making API calls. The web ui doesn't use relative paths.

### DIFF
--- a/server/controllers/FileSystemController.js
+++ b/server/controllers/FileSystemController.js
@@ -89,7 +89,6 @@ class FileSystemController {
     }
 
     const { directory, folderPath } = req.body
-
     if (!directory?.length || typeof directory !== 'string' || !folderPath?.length || typeof folderPath !== 'string') {
       Logger.error(`[FileSystemController] Invalid request body: ${JSON.stringify(req.body)}`)
       return res.status(400).json({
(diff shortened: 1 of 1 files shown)


Vito0912:
In my development build, I cannot get it to work because the file path is normalized, but the libraryFolder.path is not. I think this is a Windows-specific error due to the different direction of slashes.
I made some changes in a PR I will do tomorow

Vito0912:
https://github.com/Vito0912/audiobookshelf/commit/d2d558754f290b56cf1e7a73c765337cb0ddba5d (Line 131)

advplyr:
Ah yeah I forgot to put the path back to posix after the join. I just did a quick fix here https://github.com/advplyr/audiobookshelf/commit/22f6e86a12bd4a52f5cbddd76ea043cae3e3f3d0

Vito0912:
Thx

PR #4394, opened 2025-06-11
Added the ISBN for Audible providers (returned data)

Vito0912:
## Brief summary

Added the ISBN for Audible providers

## Which issue is fixed?

N/A

## How have you tested this?

Matched a few books with ISBN and ASIN


advplyr:
Thanks!

PR #4399, opened 2025-06-11
Adds more options to pathexists and allows uploading media files/book files if not currently for item

Vito0912:
# TBA


Vito0912:
Never came to really draft the detailed description and iirc someone had issues with this PR. The branch will exists further
If someone wants to reopen this PR and or wants to do changes, go for it.

Issue #2114, opened 2023-06-26
[Enhancement] Add library filter for explicit content & not explicit content


ikrEErki:
If a small overlay/banner could automatically be added to the cover of all explicit books would be great. Then you can just scroll through your library and easily find books you have missed labeling for explicit

slyfoxgh:
+1 For adding a new filter to quickly view Explicit books in your library. Similar to Tags or Genres

iconoclasthero:
+1
I'm writing a script to use the API endpoints to change over whole libraries and it would be nice if I could only change the ones needing changing.

iconoclasthero:
https://github.com/iconoclasthero/abc/blob/master/abs.setexplicit
requires
https://github.com/iconoclasthero/abc/blob/master/abs.getlibraries

This worked for me.  I may continue to refine it, but for now it will get the job done.

It's pretty clear that doing this does not solve the problem of having empty libraries with suggestive names like...
'I store all my smut in this library but it's empty'

Issue #2833, opened 2024-04-08
[Enhancement]: Show Abriged Icon in Shelf


mutschler:
### Describe the feature/enhancement

Like for Explicit Books: The icon in shown in the List views. However the Abridged Icon is only shown on the Book Detail view.
Would be great have the Icon in the list view as well so it's easier to spot if there's an Abridged and normal version of the book.

advplyr:
I'm not sure how to show NOT abridged in the UI

nichwall:
I think the "not abridged" is for the filter, like how there is "Has ebook" and "Has no ebook", but there is currently only an "Abridged" filter in the web app. This is separate to adding the "Abridged" indicator under to book next to the "Explicit" E icon.

advplyr:
Yeah for the filter I was looking at where that would be shown in the UI. I guess it would have to be a submenu like ebooks but with only 2 options.

Demian98:
@advplyr in my opinion, the abridged information is quite important for the listener. It would be great if it is not just display as a small icon. It should also be in the details list:
![grafik](https://github.com/advplyr/audiobookshelf/assets/77210091/825ea532-5f25-41e8-ad01-6bab6300d738)

What do you think about this?

advplyr:
I don't like putting it there because I don't think boolean fits in that detailed list view.
What would it look like?
Abridged: Yes/No?

I believe Audible or one of the book providers puts (Abridged) in the title of the book. You could also do it that way.

PR #4425, opened 2025-06-21
Add explicit filter

Vito0912:
## Brief summary

Add explicit filter for audiobooks.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/2114. The inversion would still be handled by https://github.com/advplyr/audiobookshelf/issues/2833 if we decide to rename the issue.

## How have you tested this?

Added books to the explicit list and applied a filter for it.


advplyr:
I added a check for if the user has permission to access explicit content before adding it to the library filters.

Thanks!

PR #4428, opened 2025-06-23
Add ipaddress to playbacksession

peter1490:
## Brief summary

Store IP address per listening session instead of per device.

## Which issue is fixed?

No issue.

## In-depth Description

Previously, the application only stored the last known IP address for a device. This meant that if a user connected from different networks, the IP address history for their listening sessions was lost, and the UI would only show the most recent IP for the device, not the IP used for that specific session.

This change modifies the PlaybackSession object and its corresponding database model to include an ipAddress field. A database migration has been added to support this new column. The PlaybackSessionManager has been updated to capture the IP address from the current request when a session is created, ensuring the session-specific IP is saved. The frontend modal for viewing session details has also been updated to display the IP address from the session data, providing an accurate history of the IP used for each listening session.

## How have you tested this?

I ran the full server-side test suite using the npm test command. All 295 tests passed, confirming that the modifications did not introduce any regressions.

To test the feature, I started a new listening session and verified that the IP address was displayed correctly. Next, I initiated another listening session (on a different IP address) and confirmed that the IP addresses shown for each listening session were indeed different.

### DIFF
--- a/server/migrations/v2.20.1-add-ipaddress-to-playbacksession.js
+++ b/server/migrations/v2.20.1-add-ipaddress-to-playbacksession.js
@@ -1,0 +1,67 @@
+/**
+ * @typedef MigrationContext
+ * @property {import('sequelize').QueryInterface} queryInterface - a Sequelize QueryInterface object.
+ * @property {import('../Logger')} logger - a Logger object.
+ *
+ * @typedef MigrationOptions
+ * @property {MigrationContext} context - an object containing the migration context.
+ */
+
+const migrationVersion = '2.20.1'
+const migrationName = `${migrationVersion}-add-ipaddress-to-playbacksession`
+const loggerPrefix = `[${migrationVersion} migration]`
+
+/**
+ * This migration script adds the ipAddress column to the playbackSessions table.
+ *
+ * @param {MigrationOptions} options - an object containing the migration context.
+ * @returns {Promise<void>} - A promise that resolves when the migration is complete.
+ */
+async function up({ context: { queryInterface, logger } }) {
+  logger.info(`${loggerPrefix} UPGRADE BEGIN: ${migrationName}`)
+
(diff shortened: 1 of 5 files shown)


server/migrations/v2.20.1-add-ipaddress-to-playbacksession.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @typedef MigrationContext
+ * @property {import('sequelize').QueryInterface} queryInterface - a Sequelize QueryInterface object.
+ * @property {import('../Logger')} logger - a Logger object.
+ *
+ * @typedef MigrationOptions
+ * @property {MigrationContext} context - an object containing the migration context.
+ */
+
+const migrationVersion = '2.20.1'
```
Vito0912: Looks like you work on an outdated branch? 
(Current version is 2.25.1, not 2.20.0)

peter1490: It seems I've miss read the Migrations README.
I am indeed working with the latest master branch, it just appears that I've set an incorrect migration name. Sorry...

As this is my first PR, do you have any other feedback ?
Thank you.

Vito0912: > As this is my first PR, do you any other feedback ?

I do not work on migrations, so I can't give you more information. That was just something that caught my attention right away. I do like the idea, though. I never noticed that it would do for the ips.
One tip: Use Prettier before committing (if not already done). I often forget to do this because my IDE does not run it for ABS automatically, for some reason.

peter1490: Yeah the feature is simple but I think useful for history tracking.

Thanks for the tip, I've enabled prettier now 👍

PR #4441, opened 2025-06-28
POC: Add LastSeenManager to batch user activity updates

Vito0912:
## Brief summary

This pull request introduces a new `LastSeenManager` to handle user activity tracking and periodic updates to the `lastSeen` timestamp in the database without just relying on the socket. The changes integrate this manager into the server's lifecycle.

## Which issue is fixed?

N/A

## In-depth Description

Collects users into a Set, flushes every 5 minutes, and sets the timestamp. This is not very accurate (it can be up to 5 minutes off), but it should be the most efficient solution. It might be okay to add the real timestamp, but I think it is not necessary. When authentication happens, the time is updated accurately. So we just add a estimate and do not change the current behaviour

## How have you tested this?

My own server

## Screenshots

N/A

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -37,6 +37,7 @@ const CronManager = require('./managers/CronManager')
 const ApiCacheManager = require('./managers/ApiCacheManager')
 const BinaryManager = require('./managers/BinaryManager')
 const ShareManager = require('./managers/ShareManager')
+const LastSeenManager = require('./managers/LastSeenManager')
 const LibraryScanner = require('./scanner/LibraryScanner')
 
 //Import the main Passport and Express-Session library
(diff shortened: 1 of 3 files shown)


nichwall:
Does adding to a set keep or replace the original item? If it replaces the original item, it would be trivial to use the most recent seen time. I agree that not being accurate to the minute is not very important for seeing if someone is online during maintenance or general user management.

Vito0912:
A set just contains "unique" values. 
I thought that it shows a green dot when online?
This ofc only works with websockets again, but I would assume that 5 minutes for maintenance should be sufficient (or one can look at the active listening sessions). 
But at the end it wouldn't really make a difference to use a map. 
We talk about microsecond at most.

PR #4451, opened 2025-07-01
WIP [Help wanted]: Koreader sync

Vito0912:
## Brief summary

Does not work yet.

- [x] endpoints  
- [ ] push (partially working already)  
- [ ] pull (should work, but it does not)  
- [ ] auth (we probably need an "app" password or something similar)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf",
-  "version": "2.25.1",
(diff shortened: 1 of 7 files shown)


Vito0912:
If someone wants to build on top of that, go for it
There are currently multiple projects adding support via a bridge. No need anymore

Issue #4375, opened 2025-06-04
[Enhancement]: Spice Rating


bunkabee:
### Describe the Feature/Enhancement

I'm sharing my audiobookshelf with a few coworkers and I'd like to be able to integrate the spice rating of the books on a scale of 1 to 5 (chili peppers, flames, stars, etc.). This is usually identified as the chili pepper emojis on booktok, but any sort of spice rating feature in the book details would be nice. 

### Why would this be helpful?

It would be helpful to have spice ratings in the book details to help users determine what is and isn't appropriate to listen to for their kids or themselves. Most people have different ideas of what is considered explicit content, and I've recently learned that my definition is not the same as some people who share my server. Having a "spice" rating would be beneficial especially to help filter by certain "spice" levels depending on reading mood. 

### Future Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/827cbaf8-b0ee-4f00-9140-0c6194d6e9b8)

### Audiobookshelf App Version

iOS App - 0.9.81

### Current Implementation (Screenshot)

N/A

nichwall:
There are multiple ratings and review feature requests in the server, such as https://github.com/advplyr/audiobookshelf/issues/1153

The simplest solution for you to implement using current functionality is to use tags, or separate books by library.

astoryarc:
I would love to see this as an extension of the ratings system once it's implemented

Issue #1184, opened 2022-11-20
[Enhancement]: Parse Audible rating when using Match


DarrowLykos:
### Describe the feature/enhancement

I'd like to be able to save the Audible rating when books are matched or manually add my own so users can sort my library by rating.

Dr-Blank:
There have been a couple of issues created with the intention of implementing somewhat of a rating system.

Like #701 to fetch rating from goodreads and implement a rating system for users, and #398 to parse from ID3 tags.

I want to second this as it would definitely be helpful in maintaining a record and selecting new books to read. This would also help in curating personalized recommendations to users based on the ratings (an idea to consider for #1118).

---

I am thinking it would be an integer associated with every media item (maybe from 0 to 10), which then can be represented in either binary form (like or dislike) or a star rating system, as per the user's preference.

advplyr:
Also related #1153

advplyr:
Also related #901

advplyr:
I keep looking into implementing this then fall short on how this should be added.

If anyone can mock up a UI/UX then that would be helpful. As is the case for most features the code is the easy part, the UI/UX is the difficult part.

DarrowLykos:
> I keep looking into implementing this then fall short on how this should be added.
> 
> If anyone can mock up a UI/UX then that would be helpful. As is the case for most features the code is the easy part, the UI/UX is the difficult part.

I was thinking it would be a simple 5 star rating system that appears in the group [Narrators, Publish Year, Genres, etc] working in the same way GoodReads rating system does. It would show the parsed rating from GoodReads/Audible/iTunes and allow the user to submit their own local rating or, if possible (but not necessary), connect to their own account on GoodReads etc and update the rating on their actual account with GoodReads etc.

advplyr:
At the moment we won't support any integration with Goodreads since they closed their public API.

In this case we wouldn't be submitting user ratings anywhere so we would only be able to have it for your server. This is the main issue with this feature because we are talking about 2 different things.

1. Ratings from external sources
2. Internal ratings that you make inside Abs

These would need to be shown separately

Issue #901, opened 2022-08-10
[Enhancement]: Ratings for books and authors


plague-doctor:
### Describe the feature/enhancement

It would be great if we could add a rating both for author and books.

Preferably in `.opf` file maybe something like:
```
<dc:rating>4.5</dc:rating>
```
range 0-10? Anything would be useful.

Issue #701, opened 2022-06-07
[Enhancement]: Book rating system / connect with Goodreads 


scuppasteve:
### Describe the feature/enhancement

As Audiobookshelf is on the road to supplanting other audiobook players and as a library. I would like to see the option to rate books as they are completed. Either as part of a book finishing playing, where it prompts a rating on the app and saves it to the server. Or you can just go in at any time and rate any books same as you can mark books completed.

A second request would be to have Goodreads rating score listed in the book info, as well as syncing your rating in Audiobookshelf with your Goodreads "completed" books and saving the ratings there.

cyturralde:
I like this but if you wanted to take it a step farther (I don't know if this is possible or not), it would be great to have an implementation that when you are done reading on here it marks it as read on Goodreads as well.

scuppasteve:
> marks it as read on Goodreads as well

Yep, that is what i was hoping as well. I have a collection setup for books i want to read, but if that could be sync'd from Goodreads as well that, would be awesome. Ultimately in the short term i would just like an easy way of recording my rating of a book and i can manually enter it on Goodreads later, i just want to record my main impressions immediately.

Jdiesel87:
Goodreads no longer has a public API

https://github.com/advplyr/audiobookshelf/issues/281

kymahi:
> > Goodreads no longer has a public API
> > #281
> 
> How does Readarr sync back and for books in collection / want to read / etc?

It is still supporting the existing API for those who have keys for it, but at no longer distributing new keys, and there’s talks of what their next steps in regards to a public API is.

https://help.goodreads.com/s/article/Does-Goodreads-support-the-use-of-APIs

raviwarrier:
If not sync with goodreads, can we have an internal rating system. This way, users of the same library can at least see what others think of the book they are considering reading.

advplyr:
Goodreads won't be supported see https://github.com/advplyr/audiobookshelf/issues/281#issuecomment-1159572840

Storing a rating with library items is something to consider with enough support. Another relevant request about ratings is #398

Otsira:
Hi guys. I'm thinking of tackling this proposal as so: Adding a tag with the rate and review in the metadata and the ability to export the reviews as a csv to be imported in [goodreads](https://help.goodreads.com/s/article/How-do-I-import-or-export-my-books-1553870934590). 

Is there anything else I should add?

raviwarrier:
[...]

PR #4452, opened 2025-07-01
Add book ratings

peter1490:
## Brief summary

This PR introduces a comprehensive book rating system to Audiobookshelf. Users can now rate books on a 1-5 star scale, give an explicit rating and view community average ratings for both rating types. The system also supports fetching ratings from external providers like Audible during metadata matching.

## Which issue is fixed?

Here are the enhancements requests:
*   #4375
*   #1184
*   #901
*   #701

## In-depth Description

This feature adds a full-stack implementation for book ratings.

**Backend:**
*   **Database:** New tables `userBookRatings` and `userBookExplicitRatings` have been added to store user-specific ratings. The `books` table has been extended to store ratings from external metadata providers.
*   **API:** New endpoints (`/rate` and `/rate-explicit`) have been added to the API to submit and update ratings. The `/item/:id` endpoint now returns rating information.
*   **Controllers:** The `LibraryItemController` has been updated to handle the logic for saving user ratings, calculating community averages, and fetching rating data.
*   **Settings:** New server settings have been added to allow administrators to enable or disable the simple rating system, explicit rating and community ratings.
*   **Scanner:** The metadata scanner has been updated to pull in ratings from matched providers.

**Frontend:**
*   **Rating Component:** A new reusable `RatingInput.vue` component has been created for capturing star (or flame) ratings.
*   **Item Page:** The book details page (`/item/:id`) now displays the user's rating, the community rating, and an explicit rating. Users can add or update their rating directly from this page.
*   **Settings Page:** The admin's settings page has been updated with toggles to manage the new rating features.
*   **Edit Modals:** The metadata editing modals have been updated to include and preserve provider's rating information.
*   **Developer Experience:** The client-side development setup has been migrated from Webpack to Vite for a faster development experience, and the necessary Cypress configuration has been updated.

This solution provides a robust and configurable way to handle book ratings, enhancing the user experience by providing more metadata and interaction possibilities with their library items.

## How have you tested this?

The changes have been tested both manually and with automated tests.

**Automated tests:**
*   Unit tests have been added for the new backend logic in `LibraryItemController.test.js`.
*   Cypress component tests have been created for the new `RatingInput.vue` component to ensure it functions correctly.

**Manual testing steps:**
1.  Navigated to `Settings > General` and enabled "Enable Ratings", "Enable Community Ratings", and "Enable Explicit Ratings".
2.  Opened a book details page and verified the new rating section is visible.
3.  Clicked on the stars to add a personal rating and confirmed it was saved and displayed correctly after a page refresh.
4.  Clicked the flames to add a personal explicit rating and confirmed it was saved and displayed correctly after a page refresh.
5.  Observed the community rating values (from rating with another user).
6.  Used the "Match" feature on a book and checked if a rating was populated from the metadata provider.
7.  Edited a book's details and ensured the rating information was not lost.

## Screenshots

**Item Page:**

With all settings enabled (rating + community rating + explicit rating): 
image

With just rating enabled + explicit rating enabled: 
image

With just the simple rating system enabled:
image

With just the explicit rating enabled:
image

**Matchin Tool Tab:**

When the provider has a rating value:
image

**Edit Details Tab:**

image

**Settings Page:**

By default just the simple rating system is enabled:
image


client/store/libraries.js
```diff
@@ -12,7 +13,8 @@ export const state = () => ({
   numUserPlaylists: 0,
   collections: [],
   userPlaylists: [],
-  ereaderDevices: []
+  ereaderDevices: [],
+  libraryItemsCache: {}
```
nichwall, line 17: Why is this cache here versus in the normal global cache?

client/plugins/init.client.js
```diff
@@ -61,44 +61,36 @@ Vue.prototype.$sanitizeFilename = (filename, colonReplacement = ' - ') => {
   const MAX_FILENAME_BYTES = 255
 
   const replacement = ''
-  const illegalRe = /[\/\?<>\\:\*\|"]/g
+  const illegalRe = /[\\\?<>\\:\*\|\"]/g
```
nichwall, line 64: Why is this no longer checking for the forward slash?

Vito0912:
Note, I am just a contributor, not a maintainer. So nothing that I say has any value other than an opinion.

It looks like this is a cool new feature, especially since you can turn it off.  
I quickly looked through the code, and I noticed you left many development changes in it.  
You also removed some things, like `toOldJSON`, and replaced them with a different prototype method. Is there a reason? I use TypeScript more, but that does not seem like the right way to do it in JavaScript either.

Also, I think we should not hardcode Audible or include an Audible logo, because it makes it harder to scale and leaves out other providers who could give ratings that are also already in ABS.
Also custom metadata providers should be supported.

You also made some refactoring changes to some files without a clear pattern that also not used in your PR any than the refactor.


Also some strings where changed without any obvious reason. e.g. `LabelSettingsEnableWatcher`, did you use AI to refactor?

Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM node:20-alpine AS build-client
 
 WORKDIR /client
 COPY /client /client
-RUN npm ci && npm cache clean --force
+RUN npm install && npm cache clean --force
```
nichwall, line 9: Why the change to install from ci?

client/components/cards/LazyBookCard.vue
```diff
@@ -78,7 +78,7 @@
         </div>
 
         <!-- Error widget -->
-        <ui-tooltip cy-id="ErrorTooltip" v-if="showError" :text="errorText" class="absolute bottom-4e left-0 z-10">
+        <ui-tooltip cy-id="errorTooltip" v-if="showError" :text="errorText" class="absolute bottom-4e left-0 z-10">
```
nichwall, line 81: Why change the capitalization here?

nichwall:
I agree with Vito. This looks cool, but there are a *lot* of changes that seem a bit weird and are not really explained. I left a few comments on a few lines, but have not read through most of it because there are so many lines.

client/plugins/init.client.js
peter1490: This is obviously a mistake, thanks for catching it. I will revert this change.

Dockerfile
peter1490: This is a left over from development, reverting this... Thank you.

client/components/cards/LazyBookCard.vue
peter1490: I couldn't make the frontend testing to work correctly, it had many bugs that needed to be fixed and one of them was this cy-id value.

client/store/libraries.js
peter1490: This keeps all library item state management in one place, making it easier to maintain consistency and debug issues. Altough I am not a developper so If that doesn't seem sound to you, I can remove this part.

peter1490:
Thank you for your comment, these are all valid concerns.

Please keep in mind that I am not a developper, my field is cybersecurity, so I do know how to code but I am by no means a developper. I am only here to try and help push this project forward.


> I quickly looked through the code, and I noticed you left many development changes in it. 

Apologies, I will make sure to remove unnecessary dev changes.

> You also removed some things, like `toOldJSON`, and replaced them with a different prototype method. Is there a reason? I use TypeScript more, but that does not seem like the right way to do it in JavaScript either.

`toOldJSON` does not return the ratings, same for `toOldJSONExpanded`, so I wasn't sure if I should modify them or just create `_getExpandedItemWithRatings`. Apologies if this seems trivial, I do appreciate your suggestions (if you have any).

> Also, I think we should not hardcode Audible or include an Audible logo, because it makes it harder to scale and leaves out other providers who could give ratings that are also already in ABS. Also custom metadata providers should be supported.

That's a very good point, I again was unsure if I should just support currently available Providers. Currently only Audible API supports a rating, this is why I went with Audible, but making it fully dynamic would be a better approach.
 
> You also made some refactoring changes to some files without a clear pattern that also not used in your PR any than the refactor.

You are right, there are many changes that are not explained at all. Those changes might have been made either by personal choice when trying to debug or by actual code logic needs.

Is leaving a comment on every little change (that is not self explanatory) the way to go about it ? or should I just avoid doing this when the changes do not relate directly to the PR ?

> Also some strings where changed without any obvious reason. e.g. `LabelSettingsEnableWatcher`, did you use AI to refactor?

I do use AI, and when it added the new strings, it also modified the LabelSettingsEnableWatcher string, I made sure to not modify anything more but this one slipped through.


Hope this clears somethings up, again I appreciate any other suggestion or comment you might have. I am here to learn :)

client/store/libraries.js
Vito0912: I am confused. If you are 
> not a develop[...]er

how did this PR was created then?

I already had the supicion that some AI was used because of the usual styles and mistakes they often make, but it seems like this was mostly done by AI from your comment above (please correct me if wrong). This can still be okay, but there should be a note or warning that this code might not be correct. I will add more comments about things I find a bit strange.

client/components/modals/item/EditModal.vue
```diff
@@ -245,8 +245,22 @@ export default {
         this.processing = false
       }
     },
-    libraryItemUpdated(expandedLibraryItem) {
-      this.libraryItem = expandedLibraryItem
+    libraryItemUpdated(updatedLibraryItem) {
+      if (this.libraryItem && this.libraryItem.id === updatedLibraryItem.id) {
+        // The updated item from the server doesn't contain rating data, so we preserve it from the client-side model.
```
Vito0912, line 250: But why doesn't it. Shouldn't it be returned by the server. Was this written before the backend. I am not sure here, but if it's not returned by the backend, it probably should.

client/components/modals/item/tabs/Match.vue
```diff
@@ -534,6 +551,9 @@ export default {
         if (match.narrator && !Array.isArray(match.narrator)) {
           match.narrator = match.narrator.split(',').map((g) => g.trim())
         }
+        if (match.rating) {
+          match.rating = parseFloat(match.rating)
```
Vito0912, line 555: This could produce NaN - Should be converted at Provider level

server/models/UserBookRating.js
```diff
@@ -0,0 +1,53 @@
+const { DataTypes, Model } = require('sequelize')
+
+class UserBookRating extends Model {
```
Vito0912, line 3: Wouldn't one model with a type be better? (General question)

server/scanner/Scanner.js
```diff
@@ -236,6 +236,12 @@ class Scanner {
       }
     }
 
+    if (matchData.rating && (!libraryItem.media.providerRating || options.overrideDetails)) {
+      updatePayload.providerRating = matchData.rating
+      updatePayload.provider = 'audible'
```
Vito0912, line 241: Why only Audible?
OpenLibrary also supports ratings

And also custom metadata provider

server/models/UserBookRating.js
nichwall: This almost feels like it would fit more as a new column on the `progress` for a user, since it is already joined to the response. I haven't dug through a lot of the code.

Vito0912: I would suggest not adding any more columns to the media progress unless the backend is rewritten. Adding a new column means larger payloads, and since ABS sends all user progress, this will add several kilobytes (or even a megabyte for some users) of data to each call. Also, at least two third-party apps call the user endpoint every few seconds.
But in general I would agree with you, altough the queries could become more complex then?

nichwall:
> To answer your questions:
> > Is leaving a comment on every little change (that is not self explanatory) the way to go about it ? or should I just avoid doing this when the changes do not relate directly to the PR ?
> 
> latter. You do not want to mix up multiple things in one PR that has nothing to do with your actual PR/ changes made in that PR. Especially non for example or build files.

This is one of the most important parts to me. Seeing a PR with just shy of 20k lines changed and the first few changes I read being unrelated changes makes it difficult to review. Comments are helpful, especially if the change is to fix a specific bug that is not obvious, but you don't need comments on every line change.

> > You also removed some things, like toOldJSON, and replaced them with a different prototype method. Is there a reason? I use TypeScript more, but that does not seem like the right way to do it in JavaScript either.

> toOldJSON does not return the ratings, same for toOldJSONExpanded, so I wasn't sure if I should modify them or just create _getExpandedItemWithRatings. Apologies if this seems trivial, I do appreciate your suggestions (if you have any).

The `toOldJson` is so the API works with existing clients. The data model has been updated but the API is largely unchanged, so translation is still necessary for the mobile apps and 3rd party clients.

server/models/UserBookRating.js
nichwall: I agree with you on the larger data payload, but queries should be able to remain unchanged other than adding a column to the table join because anywhere the rating would be used will also be showing the user progress so the row is already fetched from the database.

Then, whenever the backend changes to include the progress with the item fetch instead of being included with the `user` object, both could be updated at the same time.

client/store/libraries.js
peter1490: Yes, most of the code is AI, and even though I tried to check every change, things got a bit out of hand.

I am currently trying to properly reread and revert every unnecessary and incoherent change, but I feel like I'm wasting your guys' time at this point. Should I just close this PR?

Thank you for your advice.

nichwall:
Due to an increasing number of AI-generated PRs and limited time, I am not planning to review any more AI-generated PRs. I think this PR can be closed due to similar sentiments expressed by other contributors.

Additionally, this PR has not received any updates to address most of the concerns in this conversation. Thanks for expressing interest in Audiobookshelf, contributions are welcome but we ask that they be something more manageable to review.

danieljrich:
Semi-related, I just created a PR for adding Ratings fetched from Audible.com on match. Only 200 lines 🙃 
Would love anyone's two cents on it. Specifically on retroactively fetching the rating for existing items.
PR: https://github.com/advplyr/audiobookshelf/pull/4795

PR #4498, opened 2025-07-16
pkg bug. add @yao-pkg/pkg devDependencies

huyunan:
pkg bug


jaxley:
Thank you! Took me a while to figure out the missing packages on my Mac to get this working today. This will help other contributors. 

I think only other thing i had to on Mac is fakeroot from homebrew.

PR #4263, opened 2025-05-05
Add new api endpoint for direct playing audio files using session id

advplyr:
## Brief summary

When direct playing, Audio track URLs include the user token. As outlined in #4259 this is a security concern when casting because you aren't making the request to your server.

This PR creates a separate API endpoint at `/public/session/:sessionId/track/:index` that is only accessible while the session is open. This endpoint debug logs the session id, username and track index.

## Which issue is fixed?

Fixes #4259

## In-depth Description

Open playback sessions are automatically closed after 36 hours and a new session id (UUIDv4) is created each time a session is opened.

There is a remaining update to be made for HLS streams because they still include the user token in the URL.
This PR fully addresses #4259 because HLS isn't supported for casting, only direct play.

PR #4520, opened 2025-07-22
Fix podcast episode track index null in playback session

advplyr:
## Brief summary

Make sure podcast episode audio track index is always 1 in playback sessions & support 1 or 0 track index in the direct play endpoint.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf-app/issues/1635

## In-depth Description

In [v2.21.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.21.0) a bug was fixed where podcast episodes scanned in from the file system had an audio track with an index of `null`. https://github.com/advplyr/audiobookshelf/commit/7c114a051a08d401b5b6dded1620a6f1f92f05c0

In [v2.22.0](https://github.com/advplyr/audiobookshelf/pull/4263) new API endpoints for direct playing was added that use the audio track index in the url.

When playing an episode with an index value of `null` in the web client it falls back to transcoding because the track isn't found.

In mobile app v0.10.0-beta the new direct play endpoints are used. 
When playing the `null` index episode on iOS it falls back to transcoding. When playing on Android it crashes the server due to a bug where Android is still using the direct play endpoint when transcoding. That separate bug is how this bug was found.

The fix is to two-fold:
1. Updated the `/public/session/:id/track/:index` endpoint to support index `0` or `1` being passed in for podcast episodes.
2. Update the podcast episode audio track in the playback session to always use index 1

## How have you tested this?

Tested on web, iOS and Android with null audio track index

### DIFF
--- a/server/controllers/SessionController.js
+++ b/server/controllers/SessionController.js
@@ -288,7 +288,12 @@ class SessionController {
       return res.sendStatus(404)
     }
 
-    const audioTrack = playbackSession.audioTracks.find((t) => t.index === audioTrackIndex)
+    let audioTrack = playbackSession.audioTracks.find((t) => toNumber(t.index, 1) === audioTrackIndex)
+
+    // Support clients passing 0 or 1 for podcast episode audio track index (handles old episodes pre-v2.21.0 having null index)
+    if (!audioTrack && playbackSession.mediaType === 'podcast' && audioTrackIndex === 0) {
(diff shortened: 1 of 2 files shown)


advplyr:
The Android transcoding bug fix https://github.com/advplyr/audiobookshelf-app/pull/1636

PR #4528, opened 2025-07-24
Update SideRail.vue

FelixSche:
Changed cursor at version to pointer

## Brief summary

It´s just a little CSS fix to change the cursor.

## Which issue is fixed?

On the webclient the version, which acts as a button to open the changelog menu had the wrong cursor.

## In-depth Description

Its just a quick little CSS update. It´s not a big issue but it still affects everyone who uses the webclient. It helps keeping the app more consistent.

## How have you tested this?

I set the service up locally and also reran all cypress tests to check if I somehow broke anything else.


FelixSche:
I know this is such a tiny change, but it bothered my inner peace that the cursor was wrong 😅

advplyr:
Ah yeah I've seen that a few times but always in the middle of something else then forget to go back and fix it.
Thanks!

PR #4552, opened 2025-08-01
Replace some SVG icons with material-symbols

Toby222:
## Brief summary

Follow-up to #4162 :⁾
Replaces some embedded SVG icons with their Material Symbols equivalents

## Which issue is fixed?

Inconsistency in icons

## In-depth Description

Did a search for `svg` in the codebase and picked at the low-hanging fruit of SVGs right next to material icons, then replaced the svgs with icon equivalents.

## How have you tested this?

Ran with `npm run prod` and compared with non-dev server.

## Screenshots

Before:
side bar before
explicit indicator before
stats page before

After:
side bar after
explicit indicator after
stats page after


advplyr:
Nice, thanks!

PR #4557, opened 2025-08-03
Allow a whitelist of CORS origins

Vito0912:
## Brief summary

As more tools become available for ABS, they depend on the browser to make requests. 
Because of CORS, this is not possible. Right now, there is no simple way to enable CORS without using environment variables, and there is no option to allow it for specific origins, which would be better. 
Because of this, some people use proxy methods, which put users at high risk of abuse and stolen credentials. This PR adds an option to let certain origins make requests with CORS.

## How have you tested this?

My own online toolbox

## Screenshots

grafik


Vito0912:
Hey @advplyr,

it would be awesome if this could be reviewed soon/ included in the next release as it would make things more secure and easier for "external plugins" to be created. Normally I wouldn't ask for this, but it would be nice if this can be reviewed (and if needed changed) and then merged asap.

advplyr:
Hey good idea. I'll check this out later today

advplyr:
Instead of automatically saving I think this should have a save button like the prefixes setting.

image

This is because validation doesn't work well with this component. If you enter an invalid URL it just says there is an invalid URL but it stays in the select input.

Vito0912:
Yeah, I also noticed that, but I thought a save button inside the settings page could look a bit weird and the message would be a good trade-off.
I will look into that on how to make it more responsive for the user.

Vito0912:
I updated the code to it is more responsive

advplyr:
I increased the width of the inputs and separated that setting out into a new "Security" section since it didn't belong with Display

image


Thanks!

Issue #4555, opened 2025-08-02
[Bug]: Container stops when trying to play new book for the first time


bjhiltbrand:
### What happened?

I have a couple new books. When I try to play them for the first time, the container crashes with a fatal exception:

```[Server]` Unhandled rejection: TypeError: Cannot read properties of null (reading 'path')```

It seems that I can simply restart the container, and start playing the book again, and everything runs as it should. I am using the Android client when this happens, if that's pertinent. Any thoughts?

### What did you expect to happen?

I would expect the book to play as normal.

### Steps to reproduce the issue

1. Put a new book on the server
2. Attempt to play book.
3. An playback error pops up in the client (I'm using Android when this happens)
4. Check the container to see that it stopped with the fatal error in logs
5. Restart container
6. Play book again without issue.

### Audiobookshelf version

2.26.3 Docker

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Other (list in "Additional Notes" box)

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
audiobookshelf
date,stream,content
2025/08/02 12:06:28,stdout,[2025-08-02 12:06:28.269] INFO: [SocketAuthority] Socket Connected to /socket.io hZOHIR9NQGoyGAlKAAAB

2025/08/02 12:06:25,stdout,[2025-08-02 12:06:25.534] INFO: [Watcher] "Main" Ready

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.481] INFO: Listening on port :80

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.455] INFO: [Watcher] Initializing watcher for "Main".

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.413] INFO: [BackupManager] Auto Backups are disabled

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.411] INFO: [BackupManager] 0 Backups Found

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.388] INFO: [LogManager] Init current daily log filename: 2025-08-02.txt

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.384] INFO: [Database] ANALYZE completed

2025/08/02 12:06:22,stdout,[2025-08-02 12:06:22.352] INFO: [Database] running ANALYZE

[...]

PR #4562, opened 2025-08-06
Fixes hls crash on Android

Vito0912:
## Brief summary

The new getTrack is made for real files instead of hls streams. Since we use the metadata object, which does not exist on the hls audioTrack object, the server crashes. This PR adds a redirect to the hls source. I have NOT checked if it works with the android app and if it can follow redirects, but this at least stops the server from crashing. The hls stream should be reached using the correct URL anyway (shouldn't it?).

I am not up to date on the mobile repo, but it would need to be changed (in case it does not follow redirects) so it requests hls streams directly

## Which issue is fixed?

fixes https://github.com/advplyr/audiobookshelf/issues/4555

## In-depth Description

Look at the code xD. Short description should give enough hints.
SessionController.js
PlaybackSessionManager.js
AudioTrack.js

## How have you tested this?

local dev server

### DIFF
--- a/server/controllers/SessionController.js
+++ b/server/controllers/SessionController.js
@@ -302,6 +302,16 @@ class SessionController {
     const user = await Database.userModel.getUserById(playbackSession.userId)
     Logger.debug(`[SessionController] Serving audio track ${audioTrack.index} for session "${req.params.id}" belonging to user "${user.username}"`)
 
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks, redirecting works. I implemented it here https://github.com/advplyr/audiobookshelf/pull/4565

Issue #4513, opened 2025-07-20
[Enhancement]: Customizable left and right indents in book reader


Pathfinder9021:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

As a user, I would like to be able to customize the left and right indents or set the maximum width which text might occupy when reading books.

Optionaly, it would be great having the top and bottom indents customizable as well.

### Why would this be helpful?

This can improve the ease of reading books on landscape-oriented displays. 

### Future Implementation (Screenshot)

**Approach 1**
A new setting can be added to the reader settings that sets the maximum width that text can occupy when reading books. This can be a percentage of the screen width.
<img width="711" height="396" alt="Image" src="https://github.com/user-attachments/assets/01a6151d-7e21-42c8-9822-789a335d5175" />

**Approach 2**
A new settings can be added to the reader settings that sets the indents on the left and right when reading books. This can be a percentage of the screen width.
<img width="710" height="431" alt="Image" src="https://github.com/user-attachments/assets/0d47f534-3c2b-4c0e-b45f-ef847a274f97" />

Changing the maximum width will force the eye to focus on the middle of the screen.
<img width="2872" height="1508" alt="Image" src="https://github.com/user-attachments/assets/2ee7a5e7-d8b7-4720-90fa-e5a831bf3ae5" />

### Audiobookshelf Server Version

v2.26.1

### Current Implementation (Screenshot)

Currently, text takes up the entire width of the tablet PC or laptop screen, which is inconvenient if the display diagonal exceeds 12 inches. 

<img width="2862" height="1546" alt="Image" src="https://github.com/user-attachments/assets/35bbca94-9654-48e6-a544-236178e548a4" />
<img width="2878" height="1620" alt="Image" src="https://github.com/user-attachments/assets/f8eacfda-473a-4208-abf7-edeef522e72c" />

Issue #4489, opened 2025-07-14
[Enhancement]: Vertical page scrolling for books


tearsofdeath:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Continuous vertical scrolling when reading epub books

### Why would this be helpful?

It's a personal preference thing for sure, but for most people continuous vertical scrolling is just plain better than horizontal paged scrolling.

I can *tolerate* paged scrolling, but it's just annoying to me, plus the pages sometimes get stuck in the transition between pages when using the touchscreen rather than the volume buttons 

### Future Implementation (Screenshot)

You know what this looks like 

### Audiobookshelf Server Version

2.26.0

nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/3641

PR #4568, opened 2025-08-08
Epub reader vertical continuous scroll and customizable page width

DavevanIersel:
## Brief summary

Added a setting for the epub viewer to continuously scroll vertically, and a setting which allows setting the viewing width of the page (works for paginated too)

## Which issue is fixed?

- Adds customizable page width #4513 
- Adds vertical page scrolling #4489 

## In-depth Description

I added two settings: page width and page flow. Page width will allow the user to set the maximum width of the reader. The navigation arrows will stay aligned to the edge of the screen but the document will become smaller, thus allowing a more focused reading experience. Page flow has two options: Paginated or Scrolled. Paginated is the same horizontally paginated experience as before. Scrolled becomes a vertical, infinite scroll. 

In scrolled mode, half a screenheight is added to the final page to allow for some more room to scroll up. Without this you'd get to the end of the final page and the last word would be 'glued' to the bottom of your screen.

## How have you tested this?

Tested on v2.27.0 on Chrome and Firefox. I made sure to test different mobile screensizes using devtools to make sure no issues occur with the width and height changes. 

## Screenshots

image

Page width:
By default stays the same:
image

But it can now be set to any width the user wants:
image

Scrolled view:
image


Keydrain:
This is also a feature I'd love to have. Let me know how I can assist.

PR #4594, opened 2025-08-16
Testing/Feedback wanted: Add podcast support to metadata embedding tools

### LINKED ISSUE #4304: [Enhancement]: "Embed Metadata"-tool for podcasts (similar to audiobooks)
### Type of Enhancement

Web Interface/Frontend

While the Audiobook File Management Tools include the tool "Embed Metadata", there is no equivalent tool for podcasts. This enhancement would allow users to have ABS write the current metadata from ABS to the underlying files' id3-tags.

If you download the files or send them to a friend, the (additional) metadata stored in ABS is currently stuck there and can't be easily shared with the files. This enhancement would allow the transfer of up-to-date metadata for podcasts.

An "Embed Metadata"-tool like the one used for audiobooks:
<p align="center">
  
</p>

v2.22.0

<p align="center">
  
</p>

Vito0912:
## Brief summary

This adds support for embeding metadata, changed in ABS (or generally pulled from the RSS feed) to be embeded back to the files. 
I tested it on some podcasts and it got results that are working. Sadly it's a bit of a hacky solution, so maybe a complete rewrite is the better solution. But I wanted to make a start so maybe someone want to jump of from it and makes this PR way better

## Which issue is fixed?

Fixes #4304 and Discord

## In-depth Description

See brief summary, but it basically checks if it's a podcast and then creates the metadata object for every new episode to embed into the files. Since this was not intended, we return the "podcasts" metadata at the API level instead. The files still update correctly.

## How have you tested this?

Podcasts from itunes

## Screenshots
grafik

### DIFF
--- a/server/utils/ffmpegHelpers.js
+++ b/server/utils/ffmpegHelpers.js
@@ -152,29 +152,7 @@ module.exports.downloadPodcastEpisode = (podcastEpisodeDownload) => {
     const podcastEpisode = podcastEpisodeDownload.rssPodcastEpisode
     const finalSizeInBytes = Number(podcastEpisode.enclosure?.length || 0)
 
-    const taggings = {
-      album: podcast.title,
-      'album-sort': podcast.title,
-      artist: podcast.author,
-      'artist-sort': podcast.author,
-      comment: podcastEpisode.description,
-      subtitle: podcastEpisode.subtitle,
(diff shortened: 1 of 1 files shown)


Vito0912:
Do not plan to support this PR anymore due to other PRs being open never got merged nor commented.

PR #4595, opened 2025-08-16
Fix listening url log with ipv6 host

renesat:
## Brief summary

Fix incorrect listening url in log when use ipv6 host

## In-depth Description

Run with ipv6 host return incorrect url in log:
```
> node index.js --host ::
...
[2025-08-16 17:55:56.554] INFO: Listening on http://:::12378
...
```

After change:
```
> node index.js --host ::
...
[2025-08-16 17:55:56.554] INFO: Listening on http://[::]:12378
...
```

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -11,6 +11,7 @@ const axios = require('axios')
 const { version } = require('../package.json')
 
 // Utils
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #4596, opened 2025-08-16
Fix freeze on some audio sources

renesat:
## Brief summary

Fixed freeze when downloading some episodes.

## In-depth Description

On one of the podcasts (https://podster.fm/rss.xml?pid=48291), a situation arose where the download did not complete when downloading an episode. Even timeouts did not work. Through trial and error, I found out that the request would complete with a 200 status code, but would return an empty stream, causing ffmpeg to wait indefinitely for the input stream.

I found out that the podcast did not like the value passed in Accept. By default, `Accept: application/json, text/plain, */*` is passed. But if you pass `Accept: */*`, everything starts working and produces the necessary redirects.

## How have you tested this?

I checked all my podcasts for functionality on the master version with this change.

### DIFF
--- a/server/utils/ffmpegHelpers.js
+++ b/server/utils/ffmpegHelpers.js
@@ -118,6 +118,7 @@ module.exports.downloadPodcastEpisode = (podcastEpisodeDownload) => {
           method: 'GET',
           responseType: 'stream',
           headers: {
+            'Accept': '*/*',
             'User-Agent': userAgent
           },
           timeout: global.PodcastDownloadTimeout


Vito0912:
Apart from that being the solution, you should probably also notify the podcast provider.  
json doesn't make sense here, that is true, but still `*/*` is included (if you say that's the default behavior, not sure for axios), because it should just read it as a priority list:

> The HTTP Accept [request](https://developer.mozilla.org/en-US/docs/Glossary/Request_header) and [response header](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) indicates which content types, expressed as [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types), the sender is able to understand. In requests, the server uses [content negotiation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Content_negotiation) to select one of the proposals and informs the client of the choice with the [Content-Type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Type) response header. In responses, it provides information about which content types the server can understand in messages to the requested resource, so that the content type can be used in subsequent requests to the resource.

So it still should have worked

renesat:
Yes, I understand that everything should work without this change. I've basically written to the podcast provider, but I haven't received a response yet.

I just thought that this change shouldn't break anything, but it fixes this problem.

advplyr:
There have been many small tweaks like this we've had to make to accommodate podcast providers. For the user agent string I had to set up a fallback because there was no single user agent that worked for all podcasts.

Thanks!

PR #4608, opened 2025-08-19
Remove token from hls url

laxandrea:
## Brief summary

PR #4266 removed authorization for hls streams so there is no need to add the token to the url of the audio chucks

## Which issue is fixed?

Following PR #4263

## In-depth Description

Showing the token in the URL could be unsafe in some situations

## How have you tested this?

Browser with custom client

## Screenshots

Screenshot 2025-08-19 at 15 44 06

Screenshot 2025-08-19 at 15 46 02


advplyr:
Thanks!

PR #4621, opened 2025-08-24
Inode bug fix

### LINKED ISSUE #1447: [Bug]: Books being "overwritten" during scans due to mistaken inode identity
### Describe the issue

Docker-compose running ABS 2.2.12 on Ubuntu Linux 22.04; all folders mapped to the same 8TB HDD (not SDD). Prefer OPF data, and ODMs as well. Only one Library configured (I've reported this in the past when I had two libraries, but I wasn't able to capture any real information at the time).

I've noticed this many times, but always too late to get useful logs: I'd remember loading a book, but when I search for it, it isn't there. So I re-download. Later, I'd find a book that has the same cover as the one I re-downloaded but some other book's title. When I look at its folder and files, I find that the title is correct, but the cover is from the newly missing book.

Interestingly, it turns out the "missing" book is right there on the drive where I'd put it, and absolutely nothing is ever suspicious about either the original book or the one that replaced it leaving only its cover - nothing's mixed up on the disk. So this is not a data loss problem, only metadata confusion.

So I've finally noticed this happening in time to actually capture logs -- the problem was that this is apparently random, so it usually gets books I don't notice because I loaded them a long time ago, or for which my logs are long gone.

The smoking gun is a line saying something like `[2023-01-27 03:12:49] DEBUG: [Scanner] scanFolderUpdates: Library item found by inode value "olderbook => newerbook" (Scanner.js:584)`. (Those two are the names of the folders directly containing the books.) I can see this line is intended to handle hard links, but every time it appears in my repo it's mistaken, there's no hard link there (if you want I'll deep-scan for duplicate inode numbers, but I've already confirmed none of the three books in this example share any common inode numbers for any of their files).

My first example involves two "=>" lines that I actually WATCHED swap the same cover's title twice - it was random that I was looking in the right place at the right time. The original book is "The Skull of the World"; it had been loaded and Matched on the 22nd, and today the 26th I happened to be looking at newly loaded books, and I saw the cover for Skull of the World appear with the title "American Resistance", which I'd just gotten. Within a few second, the title changed again to "Unfree Speech" (still keeping the same "Skull" cover). Needless to say, both of those books had just been moved into the repo - but now I only have one book with the wrong cover and a jamble of metadata left over because the nonfiction books didn't have a subtitle or series so the resulting book had both of those from the original fiction book.

How's that for an intro? Any questions? Do I need to provide logs? Can I trim them a little? I've got them saved in full.

1. Reproducing the issue seems unlikely. This is triggered by a combination of loading a new book, plus the existence of an older book, plus unknown and possibly random factors.

### Audiobookshelf version

2.2.12

Docker

jaxley:
## Which issue is fixed?

Fixes #1447 
## In-depth Description

This PR changes the database to store an additional foreign key, deviceId, that combined with inode (ino) uniquely identifies a file across linux filesystems. This resolves problems caused by duplicate inodes across different filesystems where the metadata and file associations get corrupted (see above issue).

I added a new column to the libraryItems table, deviceId. I also added an index to ensure the combination of ino + deviceId is distinct.

I changed the queries that just used inode to query also deviceId and I changed the fsUtils to return deviceId when getting file stats.

This was the simplest change that kept inode and deviceId distinct. I had considered concatenating inode and deviceId and storing in the existing ino field to achieve a single unique id, but that seems yucky.

### Other changes

Along the way, I corrected various data type annotations that were incomplete or incorrect and added others to help identify problems and better understand the code and function arguments. e.g. several should have returned Promise<foo> but just returned foo.

I upgraded a couple unit testing libraries that were outdated. I have not upgraded others as that should be done via a separate PR - especially if any nuxt updates are breaking changes.

## How have you tested this?

I created extensive unit tests for all changes and modified the code to be unit-testable to achieve the right amount of unit testability.

I also ran the server locally and uploaded several ebook files while debugging the flow and monitoring the logs. Scanned the library folder as well to verify that the files scanned properly. 

## Screenshots
N/A


server/migrations/changelog.md
```diff
@@ -16,3 +16,4 @@ Please add a record of every database migration that you create to this file. Th
 | v2.19.1        | v2.19.1-copy-title-to-library-items          | Copies title and titleIgnorePrefix to the libraryItems table, creates update triggers and indices             |
 | v2.19.4        | v2.19.4-improve-podcast-queries              | Adds numEpisodes to podcasts, adds podcastId to mediaProgresses, copies podcast title to libraryItems         |
 | v2.20.0        | v2.20.0-improve-author-sort-queries          | Adds AuthorNames(FirstLast\|LastFirst) to libraryItems to improve author sort queries                         |
+| v2.29.0        | v2.29.0-add-deviceId          | Adds deviceId to libraryItems table to uniquely identify files in a filesystem                        |
```
nichwall: This should be v2.30.0, because the migration is for v2.30.0

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf",
-  "version": "2.29.0",
+  "version": "2.30.0",
```
nichwall: Don't update the version number of the server in the PR. This will be done as part of the release process.

server/scanner/LibraryItemScanner.js
```diff
@@ -139,26 +139,15 @@ class LibraryItemScanner {
...
-      path: libraryItemData.path,
-      relPath: libraryItemData.relPath,
-      isFile: isSingleMediaItem,
-      mediaMetadata: libraryItemData.mediaMetadata || null,
-      libraryFiles
-    })
+    return await buildLibraryItemScanData(libraryItemData, folder, library, isSingleMediaItem, libraryFiles)
   }
 
+  async setDataFromPath(path) {}
```
nichwall: Is this dangling from something else you were doing? This is not called by anything and matches another function name with two parameters.

nichwall (commented):
I started looking through the code but haven't tried running it yet and didn't have time to actually go through the migration/tests, but the basic query functionality seems to make sense.
I saw you added several comments that said "BUGBUG", can you elaborate? I assume you mean these are other bugs you found.

package.json
jaxley: Ahh, makes sense.

server/scanner/LibraryItemScanner.js
jaxley: Yes, dangling. Decided against taking on extensive refactoring in this PR.

jaxley:
> I started looking through the code but haven't tried running it yet and didn't have time to actually go through the migration/tests, but the basic query functionality seems to make sense. I saw you added several comments that said "BUGBUG", can you elaborate? I assume you mean these are other bugs you found.

Yes, these are places where there appears to be a bug in the code. For example, there are places where LibraryFiles are added as objects to LibraryItem.libraryFiles instead of a JSON version of the object. Those all need to be changed to be consistent.

I also found a problem with how symlinked library directories are handled. On a mac, /tmp is a symlink to /private/tmp. If you create a library and added the filder /tmp/somefolder to the library, it will break when uploading new content. This is because somewhere the symlink /tmp/somefolder is replaced with the "realpath" /private/tmp/somefolder. Later, some code that was added to deal with some bug with relPathToReplace incorrectly replaces part of the fully-qualified file path using the correct library folder path of '/tmp/somefolder' but the WRONG fully-qualified file path that has been resolved to remove the symlink (so starts with /private/tmp/somefolder)... This results in corrupting the file path and leaving the word "private" in it.  A workaround is to add /private/tmp/somefolder instead of the symlinked path /tmp/somefolder as a library folder.  I started looking into fixing this but there are so many places paths are resolved, it wasn't obvious where the code should be changed.  IMO, if someone gives a symlinked path, they DO NOT WANT that to be resolved to the actual current path. Why? Because if/when they move that somewhere else, the resolved paths will all break -- but they wouldn't if it kept the symlinked path names...

jaxley (commented):
> I started looking through the code but haven't tried running it yet and didn't have time to actually go through the migration/tests, but the basic query functionality seems to make sense.

I built the debian package and installed on a clone of my original audiobookshelf server. It's working great. I've been able to add files without any corruptions or errors logged. And run full library scans without corrupting the files.

jaxley:
I've been running this in production for the past 2 months with no issues.

Any idea when this can be merged into mainline?

moonblade:
Are there an updates here on when it can be merged? facing the same inode issue on nfs.

jaxley:
It's almost the year anniversary of this PR. Can the maintainers please identify a roadmap timeframe when it can get merged in? 

Or, what else can I do to get this over the line?

nichwall:
I can try looking at this again in the next few weeks (in the middle of moving apartments), so feel free to ping me again in a few weeks if I forget.

I don't really have a good answer other than it's a few thousand line change and will take time to review. It does help that you've been running it locally for reliability questions. :)

nichwall:
I started going through the PR code changes again, but thought of an edge case this might not handle. This change uses the inode and device ID (for multiple file systems/disks) to ensure inodes do not collide, but does this work for a single file system and disk with a small inode pool?

I am specifically thinking about an increasing number of reports from Docker on Mac users reporting a similar issue related to inodes, but I don't think this would fix the issue if they have the same device ID for everything. I don't have a way to test this myself, but maybe we could instead use the inode and file size in bytes for "uniqueness"? Just a thought. I would need to hunt down the specific GH issues and Discord conversations for the Mac reports.

I also realize this review is incredibly delayed, but it might be easier to move the JSDoc and comment updates to a second PR, especially if something needs to change from my thought above.

FerriteGiant:
Any chance for progress on this in the near future? My setup using unraid is pretty much unusable now. I could probably help with some testing.

Issue #4634, opened 2025-08-30
[Bug]: Book detail UI stops working after going from one specific book to another


Phantomwise:
### What happened?

Hi, here's a weird one:
 
- I added books 1-8 of the Harry Fox series by Andrew Clawson, in order.
- In the home view, I can open the details of book 1, then click the left arrow to go to book 2, and onwards until book 8. I can also open the details of book 8, then click the right arrow to go to book 7 then 6 then 5 then 4. I can do the same to go from book 3 to 2 to 1.
- However, when I try to go from book 4 to book 3, it kinda freezes the UI. I still see the details of book 4, I can scroll up and down in it, but none of the buttons work. I can't even exit the book details window. There's no "socket disconnected" popup. It doesn't crash the container.
- I can open the details of book 3 and 4 directly without issue, the problem only happens when I specifically go from book 4 to book 3. Book 3 to 4 is completely fine.
- If I rescan book 3 and book 4 it doesn't help.
- Every time it happens I need to manually refresh the page to get it working again. Then I try again and the same thing happens.
- If I go to the book series view, the same thing happens: I can go from one book to another except from book 4 to book 3.

### What did you expect to happen?

Being able to go from book 4 to book 3.

### Steps to reproduce the issue

¯\_(ツ)_/¯

### Audiobookshelf version

v2.29.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Other (list in "Additional Notes" box)

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Logs

```shell
# Too long for this field so I added the file as an attachment. Relevant logs start at 2025-08-30 14:13

# Here are the last 100 lines or so:
{"timestamp":"2025-08-30 14:19:42.060","source":"SocketAuthority.js:198","message":"[SocketAuthority] User Offline phantomwise","levelName":"DEBUG","level":1}
{"timestamp":"2025-08-30 14:19:42.061","source":"SocketAuthority.js:202","message":"[SocketAuthority] Socket 5FDISAzJfOdhAU56AAAh disconnected from client \"phantomwise\" after 75633ms (Reason: transport close)","levelName":"INFO","level":2}
{"timestamp":"2025-08-30 14:19:43.128","source":"ApiCacheManager.js:48","message":"[ApiCacheManager] count: 0 size: 0","levelName":"DEBUG","level":1}
[...]

PR #4636, opened 2025-08-30
Fix: MultiSelect causes web client to become unresponsive if duplicate keys exist

nichwall:
## Brief summary

This PR adds a unique identifier to the Vue for loop to ensure all keys have a unique identifier.

## Which issue is fixed?

Related to https://github.com/advplyr/audiobookshelf/issues/4634

## In-depth Description

If the same key exists multiple times in a MultiSelect component, the unique key requirement for `v-for` is violated. By adding an index to the key, we ensure that every key is unique, but do not change the values displayed in the front-end.

This can only happen if the database incorrectly includes the same value in an array, such as the same tag applied to the same book multiple times. Instead of filtering out the duplicate tags, these should be displayed to ensure consistent behavior when comparing against the API or another client and allow the user to fix the duplicate tag.

This does not fix the root of the problem where the same tag can be added to the same book, but fixes the symptom.

## How have you tested this?

Used database with the same tag on a book multiple times and ensured:
- Correctly removes selected tag when no duplicates exists (unchanged)
- Correctly adds new tags (unchanged)
- Correctly removes selected tag when duplicate tags exist
Screenshot from 2025-08-30 10-07-57
Screenshot from 2025-08-30 10-08-01

- Correctly removes all duplicate tags when unselected using the dropdown

Screenshot from 2025-08-30 10-07-27
Screenshot from 2025-08-30 10-07-33

## Screenshots

See above to match with test case.

### DIFF
--- a/client/components/ui/MultiSelect.vue
+++ b/client/components/ui/MultiSelect.vue
@@ -4,10 +4,11 @@
     <div ref="wrapper" class="relative">
       <form @submit.prevent="submitForm">
         <div ref="inputWrapper" role="list" style="min-height: 36px" class="flex-wrap relative w-full shadow-xs flex items-center border border-gray-600 rounded-sm px-2 py-1" :class="wrapperClass" @click.stop.prevent="clickWrapper" @mouseup.stop.prevent @mousedown.prevent>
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #4639, opened 2025-08-31
Normalize tags and genres in BookFinder results

Vito0912:
## Brief summary

Mentioned in https://github.com/advplyr/audiobookshelf/issues/4634#issuecomment-3239425190 Audible returns the same tag twice. This caused #4636. To avoid more issues, this PR adds validation to convert the strings/array into a Set first to make them unique. 

On a side note, I am not sure why we return genres as an array and tags as a string (at least for the Audible Provider), and why both also work the other way around. This code normalizes it to always return an array. In my tests, this did not break anything, especially since ", " could interfere with actual tag comma separations.

## Which issue is fixed?

Related to #4634 

## In-depth Description

See short description

## How have you tested this?

Matching the mentioned book and a few other ones.

### DIFF
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -595,6 +595,15 @@ class BookFinder {
         book.description = htmlSanitizer.sanitize(book.description)
         book.descriptionPlain = htmlSanitizer.stripAllTags(book.description)
       }
+      if (book.tags) {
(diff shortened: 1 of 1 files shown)


advplyr:
I thought it would be better to fix this in the providers instead so I did that here https://github.com/advplyr/audiobookshelf/pull/4927

PR #4647, opened 2025-09-03
Add armv7 support to docker builds

### LINKED ISSUE #197: [Feature Request] ARMv7 support
Hi, it would be awesome if this could be installed on my raspberry pi since that's what I use to host my Plex and other libraries.

Cheers. Stay safe.

Artox:
Update package.json and Dockerfile to support the armv7 platform.

Similar to arm64 it is supported by docker's buildkit and uses
an architecture-specific nusqlite3 binary.

This binary is not yet available at mikiher/nunicode-sqlite, PR is open: mikiher/nunicode-binaries#1

Further update Dockerfile to build web client on the build host without emulation as a performance optimization.

The server build step still relies on emulation for compiling code native to the target platform.

Closes #197

CI workflow is not updated yet, before proper testing.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -71,7 +71,7 @@ jobs:
           tags: ${{ github.event.inputs.tags || steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           context: .
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: true
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max


nichwall:
armv7 support was removed in server version 2.8.0
https://github.com/advplyr/audiobookshelf/releases/tag/v2.8.0

Artox:
> armv7 support was removed in server version 2.8.0 https://github.com/advplyr/audiobookshelf/releases/tag/v2.8.0

I tested current version and it is working well, I resolved all the small issues.
It is a pity to drop armv7, I am hoping to show that keeping it is little effort.

I previously used the amd64 .deb with qemu and it was not quite usable. The native version however works flawlessly.

https://hub.docker.com/repository/docker/notjosua/audiobookshelf-unstable-arm-fork/general


EDIT: If binaries are to be dropped, at least it will be nice to support users rebuilding locally for armv7.

nichwall:
Thanks for taking the time to put the PR together, but after more discussion, we will not be adding support for armv7 back in.

Support was originally dropped for a few reasons:
- Binary support didn't exist at the time (as you mentioned)
- We don't have hardware set up to verify it is working for every release (limited automated testing exists, so we can't rely on automated tests)
- The number of users on armv7 is a very small percentage, and as you mentioned users can still run from source or build locally
- A lot of other projects have also dropped support for 32-bit installs. A nice write-up from the LSIO group is here and words it better than me https://www.linuxserver.io/blog/a-farewell-to-arm-hf

When we originally dropped support a year and a half ago, some people were unable to update to ABS 2.8.0, but nearly all of them were just using a 32-bit OS on 64-bit hardware and could then upgrade. Otherwise, I am finding less than 10 conversations between GitHub and Discord about running on armv7 since support was dropped.

I am currently working on rewriting all of the documentation, so I can add an FAQ about 32-bit armv7 support and that users can still build/run the app locally.

PR #4662, opened 2025-09-10
Added badge collection changes

rashmi0114:


nichwall:
Is there a reason you're putting all of this as a plugin instead of making it persistent with the database?

Vito0912:
Badges/Achievements were decided to be out of scope for ABS. I am not sure if this opinion has changed thouhg.

Still, this looks very AI-generated, which is not necessarily bad, but the PR has no description. Some files changed are not directly related to your PR, and you also left a binary file in it.  
Additionally, some things are intercepted that could work without intercepting every request and reading from the DB.

Just some things I noticed on overlooking the PR

advplyr:
Definitely AI generated

Issue #4007, opened 2025-02-19
[Enhancement]: Autoplay books in series


RA341:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Audiobooks in a series should automatically autoplay to the next book once a book ends if it has a next book.

This can be toggled accordingly.

### Why would this be helpful?

It would be nice to start a series and continue playing until I stop, while pretty niche 

it's annoying to switch to the next book when doing chores or not having access to the device from where the ABS is playing. 

### Future Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/ae1855b1-eeef-4ee5-8583-bab106a35813)

### Audiobookshelf Server Version

v2.19.3

colvdv:
I would add that a custom queuing feature would be nice in addition to autoplaying books within a series.

averageyogi:
> I would add that a custom queuing feature would be nice in addition to autoplaying books within a series.

In the web player Playlists and Collections should continue as a queue and auto-play the next at the end of each book, they don't queue in the android app.

colvdv:
> In the web player Playlists and Collections should continue as a queue and auto-play the next at the end of each book, they don't queue in the android app.

Ah yes, that is a feature on web. Thank you for pointing that out.

NadCat1:
I second this. Having autoplay would be especially helpful while driving

DarkAngelGH:
Playlists and Collections should queue in the Android app; without that, the Collections and Playlists aren't really useful (within the Android app), as they don't provide any additional functionality.  I'm really hoping this is implemented soon.

nrevolt:
> Playlists and Collections should queue in the Android app; without that, the Collections and Playlists aren't really useful (within the Android app), as they don't provide any additional functionality. I'm really hoping this is implemented soon.

Agreed, such a basic feature should not be missing from any iterations of the player,

averageyogi:
> Playlists and Collections should queue in the Android app; without that, the Collections and Playlists aren't really useful (within the Android app), as they don't provide any additional functionality. I'm really hoping this is implemented soon.

This is the issue on queuing in the app repo: https://github.com/advplyr/audiobookshelf-app/issues/416. It's one of the most requested features.

PR #4698, opened 2025-09-25
Add autoplay for next book in series after queue ends

Vito0912:
## Brief summary

This PR adds automatic playing a series if the queue ends. It intentionally does NOT add all the items to the queue, a. to save bandwith and b. to give users the possibility to add items to the queue.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/4007

## In-depth Description

This adds a new route to get the next book in a series. This could be done in the frontend, but I decided not to because a series can have hundreds of books and fetching the full series at once would not be good. This endpoint can also be used by other apps to display the following books and is a small addition. I hope this makes sense.  

The next book in a series only plays if the queue is empty and there is a next book available. If auto play in the queue is disabled, it will not play. I am not sure if that is the best way (since people might want to use the queue and not automatically listen to series books, but I think adding a new control just for this is unnecessary).

## How have you tested this?

Own server

## Screenshots

None

### DIFF
--- a/client/components/app/MediaPlayerContainer.vue
+++ b/client/components/app/MediaPlayerContainer.vue
@@ -183,29 +183,75 @@ export default {
   methods: {
     mediaFinished(libraryItemId, episodeId) {
       // Play next item in queue
-      if (!this.playerQueueItems.length || !this.$store.state.playerQueueAutoPlay) {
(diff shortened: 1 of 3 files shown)


Vito0912:
Is mostly frontend, therefore closing

PR #4716, opened 2025-10-02
Async Cover Search

mikiher:
## Brief summary

This replaces the existing synchronous and sequential cover search with an asynchrounous and parallel search.

## Which issue is fixed?

These's no current issue, although currently cover search with `All` option selected always times out and returns no results due to a faulty AudiobooksCovers provider (backend always times out). This resolves the issue thoroughly, by replacing the existing synchronous and sequential provider search with an asynchronous (using WebSockets) and parallel search.

### Server

1. New CoverSearchManager (server/managers/CoverSearchManager.js)
    - Handles parallel provider searches instead of sequential
    - Each provider has a 10 second timeout
    - Uses AbortController for proper cancellation
    - Tracks active searches by requestId
    - Streams results as they arrive via callbacks
2. SocketAuthority Updates (server/SocketAuthority.js)
    - Added WebSocket event handlers:
      - search_covers - Starts a new search
      - cancel_cover_search - Cancels active search
    - Emits events back to client:
      - cover_search_result - Streaming results from each provider
      - cover_search_complete - All providers finished
      - cover_search_error - Fatal errors
      - cover_search_provider_error - Individual provider failures
      - cover_search_cancelled - Cancellation confirmed
    - Auto-cleanup on socket disconnect
3. Reduced provider request timeout from 30 seconds to 10.
4. Shortened lengthy axios error messages

### Client

Updated Cover.vue:
- Covers are shown as they arrive.
- Improved UX with Cancel button
- Generates unique request IDs: cover-search-{timestamp}-{random}
- Sets up/removes WebSocket listeners on mount/unmount
- Handles streaming results incrementally
- Cancels searches automatically when:
  - Component is unmounted (tab switch)
  - Modal is closed
  - User presses Cancel
  - WebSocket disconnects

## How have you tested this?

- Tested on many book covers
- Since audiobookcovers providers is currently faulty 100% of the time, this allowed testing timeout and provider failure

### DIFF
--- a/client/components/modals/item/tabs/Cover.vue
+++ b/client/components/modals/item/tabs/Cover.vue
@@ -51,21 +51,21 @@
     <form @submit.prevent="submitSearchForm">
       <div class="flex flex-wrap sm:flex-nowrap items-center justify-start -mx-1">
         <div class="w-48 grow p-1">
-          <ui-dropdown v-model="provider" :items="providers" :label="$strings.LabelProvider" small />
+          <ui-dropdown v-model="provider" :items="providers" :disabled="searchInProgress" :label="$strings.LabelProvider" small />
         </div>
         <div class="w-72 grow p-1">
-          <ui-text-input-with-label v-model="searchTitle" :label="searchTitleLabel" :placeholder="$strings.PlaceholderSearch" />
+          <ui-text-input-with-label v-model="searchTitle" :disabled="searchInProgress" :label="searchTitleLabel" :placeholder="$strings.PlaceholderSearch" />
         </div>
         <div v-show="provider != 'itunes' && provider != 'audiobookcovers'" class="w-72 grow p-1">
-          <ui-text-input-with-label v-model="searchAuthor" :label="$strings.LabelAuthor" />
+          <ui-text-input-with-label v-model="searchAuthor" :disabled="searchInProgress" :label="$strings.LabelAuthor" />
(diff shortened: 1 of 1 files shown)


Vito0912:
@mikiher Is there a reason for removing audiobookcovers? I didn't use them myself, but I assume some do. 
They had some issues for a month, but the API should be working again.

mikiher:
It was not working up until this morning and was timing out. 
I can bring it back, but generally I think we should not rely on flaky services. 

One way to resolve this is to have another Providers option that is a subset of all existing ones, that we think are relatively stable and valuable (calling it "Best" or something similar), and make it the default. 

Right now some of the providers aren't stable (audiobookcovers), some take too much time (openlibrary), some return many irrelevant results (itunes), or duplicates (audible regions) - we could remove all of these from "Best", and users will still have the option to select "All", or individual providers. What do you think?

Vito0912:
I am not sure about that, but I think that's a great idea.

I just wanted to leave a message to tell that audiobookscover works again and was confused on why removing it.

mikiher:
OK, this is ready for review now.

- Added a Best option that searches only audible.com, google, and fantlab.
- Reverted removal of AudiobookCovers.com
- Tidied up client logging and error toasts for cover search

advplyr:
Much better, working great. Thanks!

PR #4700, opened 2025-09-25
feat(docker): Add PUID/PGID support for volume permissions

Vylyne:
## Brief summary

Introduces an entrypoint script and 'gosu' to allow users to specify Process User ID (PUID) and Process Group ID (PGID) via environment variables.

## Which issue is fixed?

Fixes #4471

## In-depth Description

Introduces an entrypoint script and 'gosu' to allow users to specify Process User ID (PUID) and Process Group ID (PGID) via environment variables.

This resolves permission errors encountered when binding host volumes into the container, as the entrypoint script now runs 'chown' on /config and /metadata as root before dropping privileges and running the application.

- Adds 'gosu' dependency to the runtime stage.
- Implements 'docker-entrypoint.sh' to handle user switching and permission fixing.
- Updates ENTRYPOINT to use the wrapper script, passing CMD as arguments.

## How have you tested this?

I built the container and I'm running it now with a non root user it had no issues working with the files when previously it did when I used the user: directive in my docker compose.

Build Results
```shell
docker build -t audiobookshelf/server:puid-fix .
[+] Building 0.4s (23/23) FINISHED
 => [internal] load build definition from Dockerfile
 => => transferring dockerfile: 2.10kB
 => [internal] load metadata for docker.io/library/node:20-alpine
 => [internal] load .dockerignore
 => => transferring context: 198B
 => [internal] load build context
 => => transferring context: 58.91kB
 => [build-server 1/7] FROM docker.io/library/node:20-alpine@sha256:eabac870db94f7342d6c33560d6613f188bbcf4bbe1f4eb47d5e2a08e1a37722
 => CACHED [stage-2 2/8] RUN apk add --no-cache --update   tzdata   ffmpeg   tini   gosu
 => CACHED [stage-2 3/8] WORKDIR /app
 => CACHED [build-client 2/5] WORKDIR /client
 => CACHED [build-client 3/5] COPY /client /client
 => CACHED [build-client 4/5] RUN npm ci && npm cache clean --force
 => CACHED [build-client 5/5] RUN npm run generate
 => CACHED [stage-2 4/8] COPY --from=build-client /client/dist /app/client/dist
 => CACHED [build-server 2/7] RUN apk add --no-cache --update   curl   make   python3   g++   unzip
 => CACHED [build-server 3/7] WORKDIR /server
 => CACHED [build-server 4/7] COPY index.js package* /server
 => CACHED [build-server 5/7] COPY /server /server/server
[...]

Issue #4471, opened 2025-07-10
[Bug]: Unable to start container with non-root user


mill1000:
### What happened?

When attempting to run audiobookshelf via Docker as a non-root user the container is unable to start due to EACCES: permission denied errors.

### What did you expect to happen?

The container to start normally as a non-root user.

### Steps to reproduce the issue

1. Setup minimal docker compose example.
```yaml
services:
  audiobookshelf:
    image: ghcr.io/advplyr/audiobookshelf:latest
    container_name: audiobookshelf
    restart: unless-stopped
    user: 1000:1000
    environment:
      - TZ=America/Denver
      - PORT=13378
    ports:
      - 13378:13378
    volumes:
      - ./audiobooks:/audiobooks
      - audiobookshelf-config:/config
      - audiobookshelf-metadata:/metadata

volumes:
  audiobookshelf-config:
  audiobookshelf-metadata:
```
2. Bring container up: `docker compose up`
3. Watch it fail

### Audiobookshelf version

v2.25.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
Attaching to audiobookshelf
audiobookshelf  | Running in production mode.
audiobookshelf  | Options: CONFIG_PATH=/config, METADATA_PATH=/metadata, PORT=13378, HOST=undefined, SOURCE=docker, ROUTER_BASE_PATH=/audiobookshelf
audiobookshelf  | [2025-07-09 21:31:10.777] INFO: === Starting Server ===
audiobookshelf  | [2025-07-09 21:31:10.780] INFO: [Server] Init v2.25.1
audiobookshelf  | [2025-07-09 21:31:10.780] INFO: [Server] Node.js Version: v20.19.2
audiobookshelf  | [2025-07-09 21:31:10.780] INFO: [Server] Platform: linux
audiobookshelf  | [2025-07-09 21:31:10.781] INFO: [Server] Arch: x64
audiobookshelf  | [2025-07-09 21:31:10.793] FATAL: [Server] Unhandled rejection: [Error: EACCES: permission denied, mkdir '/metadata/streams'] {
audiobookshelf  |   errno: -13,
audiobookshelf  |   code: 'EACCES',
audiobookshelf  |   syscall: 'mkdir',
audiobookshelf  |   path: '/metadata/streams'
audiobookshelf  | } 
[...]

PR #4717, opened 2025-10-02
Move to a non-root model.

Vylyne:
## Brief summary

Moved to non-root user model.

## Which issue is fixed?

Related to questions in #4471
Superseeds #4700 

## In-depth Description

- Added buildargs to specify the User ID and Group IDs.
- entrypoint now defaults to running as 1000:1000 rather then 0:0.
I noticed myself I had some issues initially running with the user directive. 

## How have you tested this?

I'm running on a build with this change incorporated and VOLUME Instruction  added to docker file.
ghcr.io/vylyne/audiobookshelf:latest

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,9 @@
+# globally defining Arguments and Defaults
 ARG NUSQLITE3_DIR="/usr/local/lib/nusqlite3"
 ARG NUSQLITE3_PATH="${NUSQLITE3_DIR}/libnusqlite3.so"
+# default Process user id and group id
+ARG PUID=1000
+ARG PGID=1000
 
 ### STAGE 0: Build client ###
 FROM node:20-alpine AS build-client
@@ -25,7 +29,7 @@ RUN apk add --no-cache --update \
   unzip
 
 WORKDIR /server
-COPY index.js package* /server
+COPY index.js package* /server/
 COPY /server /server/server
 
 RUN case "$TARGETPLATFORM" in \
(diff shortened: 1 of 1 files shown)


Ionshard:
Does this not force the non-root user to always be 1000:1000, since the ARG is baked into the image. I try and run my docker containers as their own user/group. This wouldn't allow that.

e.g. I have a `audiobookshelf` user with UID `983`. I do this because the base user of the server is UID `1000` and I don't want the docker to have access to the base user's files.

Vylyne:
not exactly, It uses the build time arguments to set what user the container will run the entrypoint as if no user directive is added. When the User directive is used, then the entry point is always run as what ever UID:GID is provided by the user directive.|

so if you use the docker run ... --user 983:983 ... or use the user: 983:983 in docker compose the container will be running as that user no matter what is specified in the docker file.

If you're building your own image you can edit the args, or add the build argument values that you want to the build command and you won't need to the --user or user: directive. So that's the "better" way to do it but the easy way is using the image and just forcing what you want it to be at runtime, it works just as well and you should be able to just change it to anything you want.

Only other change On this I'm thinking could be added is to see if there's something in tini or in index.js that we could/should add to prevent a user for user --user 0:0, like just have it exit out and throw an error that it should not be run as root.

voruti:
> Only other change On this I'm thinking could be added is to see if there's something in tini or in index.js that we could/should add to prevent a user for user --user 0:0, like just have it exit out and throw an error that it should not be run as root.

I think this shouldn't be forced. The default should be non-root, as that's best practice, but if someone explicitly wants to use root, they should be allowed to do so.

Edit: maybe add a warning when running as root?

Vylyne:
> I think this shouldn't be forced. The default should be non-root, as that's best practice, but if someone explicitly wants to use root, they should be allowed to do so.

Yeah that was my thought too

> Edit: maybe add a warning when running as root?
If we did that I'd suggest it should be in the Interface and the logs. Not just the logs cause someone who does it by accident may not see the logs. But yeah I think adding any protections or warnings around running as root would be a different change anyway. 

So I'll leave this as is, just doing the move away from running as root to a default non-root users.

I did want to pick a different default user ID but with most major linux distro's starting users at 1000 I went with that because it's the lowest barrier for the least experienced user to use. Since that's likely the UID they're using when they're logging in to the host, (if not root).

lectavison:
is it possible to set the default to 1001:1001 since Libation uses that one for its exports, use an environment variable for those that use easypannel?

Vylyne:
I've noticed  that there are a number of pulls/downloads on my dev image. that I don't think are entirely just me.

I've added a fork sync workflow that runs daily at 2am to pull the merge the changes from advplyr:master into my main, so my latest image should pick up any changes to official upstream release while/until the pr is approved, made obsolete, rejected, or whatever the maintainers choose to do. I'd still recommend using the official release, really this message is meant for people who know what they're doing and have already started playing with my dev image, I'm less likely to nuke the main or make changes that can break it then I am to my dev repo. Anything I do that could affect the functionality of the image I'll do in a dev branch first before pushing into my main. so while less table and less recommended then Official Release it should be more stable than a dev build.

> is it possible to set the default to 1001:1001 since Libation uses that one for its exports, use an environment variable for those that use easypannel?

With this changes I've proposed in this PR it would be possible to set the "user" flag when using docker run commands or the "user" property/directive on the service in docker compose. I do not know how or if you can do that with easypanel. If you want to test this It also appears there should be an option to change the "app service image" for the template when you're deploying it or using it or something in easypanel, (if i'm looking at the same easypanel docs). If you can figure that out you could test with my image and see if that meets you're ask. If it does then let us know give us a +1 to request PR approval.

I personally would not recommend going the path of Environment Variables to change the what user the container runs as because this would require the container to run as root then change to that user.

Vylyne:
Cleaned up branch to only have 1 commit and only the changes to the docker file, removed the changes to the build workflow.

PR #4721, opened 2025-10-05
Update LocalAudioPlayer and chapters view for improved functionality

solar141:
## Brief summary

This PR fixes two issues related to audio playback in the Audiobookshelf client: preventing multiple audio streams from playing simultaneously and ensuring chapter previews stop at the end of the chapter without auto-advancing.

## Which issue is fixed?

Chapter Preview Doesn’t Stop at Chapter End
Multiple Audio Streams Playing Simultaneously

## In-depth Description

Multiple Audio Streams Playing Simultaneously:

The issue occurs when multiple audio streams (e.g., from different chapters or previews) start playing at the same time, leading to overlapping sounds.
Solution: Added event listeners to detect when a new audio stream begins and automatically pause any currently playing audio. This ensures only one stream is active at a time.
Updated files: client/pages/audiobook/_id/chapters.vue and client/players/LocalAudioPlayer.js.
This is a common edge case for users previewing multiple chapters quickly, improving the overall user experience by preventing audio chaos without affecting core playback logic.

Chapter Preview Doesn’t Stop at Chapter End:

The problem is that when previewing a single chapter, the player auto-advances to the next chapter instead of stopping at the end.
Solution: Removed the auto-advance logic specifically for single-chapter preview mode, allowing the playback to end naturally at the chapter's conclusion.
Updated file: client/pages/audiobook/_id/chapters.vue.
This fix targets preview functionality, which is useful for users scanning books, and doesn't impact full audiobook playback.

## How have you tested this?

Open an audiobook page in a local development environment (run npm run dev).
Start playing a chapter preview.
Quickly start another chapter preview.
Verify that the first audio pauses immediately when the second starts, with no overlap.
Test with multiple browsers/tabs to simulate concurrent plays.
Ensured no regression in standard full-book playback.

### DIFF
--- a/client/pages/audiobook/_id/chapters.vue
+++ b/client/pages/audiobook/_id/chapters.vue
@@ -595,6 +595,12 @@ export default {
       this.hasChanges = hasChanges
     },
     playChapter(chapter) {
+      // Pause any existing audiobook playback to prevent multiple streams
+      if (this.$store.state.streamLibraryItem) {
+        this.$eventBus.$emit('pause-item')
+      }
+
+      console.log('$store.state.streamLibraryItem', this.$store.state)
       console.log('Play Chapter', chapter.id)
       if (this.selectedChapterId === chapter.id) {
         console.log('Chapter already playing', this.isLoadingChapter, this.isPlayingChapter)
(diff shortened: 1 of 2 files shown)


client/pages/audiobook/_id/chapters.vue
```diff
@@ -641,15 +647,8 @@ export default {
...
-        if (nextTrack) {
-          console.log('Playing next track', nextTrack.index)
-          this.currentTrackIndex = nextTrack.index
-          this.playTrackAtTime(nextTrack, 0)
-        } else {
-          console.log('No next track')
-          this.destroyAudioEl()
-        }
+        // stop at end of chapter
+        this.destroyAudioEl()
```
nichwall, line 651: This does not work if a chapter ends in the middle of a track.

client/players/LocalAudioPlayer.js
```diff
@@ -231,6 +231,11 @@ export default class LocalAudioPlayer extends EventEmitter {
   }
 
   play() {
+    // Emit pause-chapter event to stop any other chapter playing
+    if (this.ctx.$eventBus) {
```
nichwall, line 235: This does not stop chapters from playing when I tested.

PR #4734, opened 2025-10-09
Pdf zoom in zoom out

farhanisrakalvy:
Description:
I have added a zoom control feature that allows users to zoom in and out using the keyboard. Press “+” to zoom in and “-” to zoom out. The maximum zoom level is 200%, while the minimum is 100%. Additionally, a display box shows the current zoom level. Users can also choose between two options: "Fit Width" and "Fit Page."

File Modified:
audiobookshelf\client\components\readers\ ComicReader.vue

 
Screenshot 2025-10-08 204055


Vito0912:
Without reviewing this, you still have the pdf/epub files in the history. 
You need to force push to remove them from your commit history (meaning to remove all commits with those files and force push one correct commit without the files). 
Currently, the files are still part of the tree. 
Also, you should not add the files to the .gitignore, as they should not be part of /docs. - Although it probably won't hurt. Just a comment

nichwall:
This PR looks like it still has the media files in the git history as Vito mentioned. This PR also touches many more files than just the ComicReader, most of which are not related to the actual described change.

Due to an increasing number of AI-generated PRs and limited time, I am not planning to review any more AI-generated PRs. I think this PR can be closed due to similar sentiments by other contributors.

Issue #3765, opened 2024-12-31
[Bug]: Removing a book from the filesystem doesn't remove it from audiobookshelf


dessalines:
### What happened?

I removed a book from my filesystem, but for some reason it's still showing up as missing on audiobookshelf.

What's worse is that there's no way to remove it now, either from within the android app, or the web interface.

### What did you expect to happen?

A rescan should remove books that are missing.

### Steps to reproduce the issue

1. Delete a book
2. It still shows up and is marked as missing in audiobookshelf
3. There's no way to delete it.

### Audiobookshelf version

v2.7.16

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Chrome

nichwall:
ABS does not delete books if they are removed from the file system and are marked as missing so you are aware that ABS thinks files are missing. Can you explain what you mean by "no way to delete it"?

You can edit the book and press the "delete" button, or go to the Issues tab in the navigation bar in the web client and delete missing books. This tab may not show up in a mobile browser depending on the screen dimensions.

dessalines:
I can't see the issue tabs on either android, or the web app.

Also why is it showing missing books? I don't know of any other media server that does that behavior. Not jellyfin, navidrome, or any other app. If you remove the file, it may keep the metadata stored, but it doesn't display missing items in the user interface. 

I hit this because I sometimes download a few different versions of audiobooks to see which one is the best quality, and then remove the ones that don't. I did this, and now it shows versions of books I deleted on my main page.

nichwall:
Can you share a screenshot of the issues tab not showing up?

The metadata is no longer tracked when it is deleted, so it shows up to let the user know that there is metadata without a corresponding file. This metadata is not deleted in case the file being inaccessible is temporary (drive was not mounted correctly, file permissions changed, etc)

dessalines:
Android app: no issues tab, incorrectly shows the books I deleted.

![Screenshot_2024-12-31-17-20-37-75_e9782661401ed6f222682e1e02b74bd4](https://github.com/user-attachments/assets/7b2e6ee6-19b5-4827-a05a-9a0ad83ee7df)

Web app: No issues tab, shows missing metadata on the front page for books I deleted.

[...]

PR #4745, opened 2025-10-12
[Enhancement] Clairfy when user has insufficent permissions to delete items from issues tab

LiamWilkinson:
## Brief summary

Add tool tip to disabled delete button when user does not have permissions to delete items on issues tab.

## Which issue is fixed?

https://github.com/advplyr/audiobookshelf/issues/3765

## In-depth Description

It's unclear that the permissions to delete an item that has issues require permissions a user may not have. When you visit the issues tab logged in as a user that does not have permissions there is no indication that the user has no permission to delete it just doesn't show the button. It would be helpful to know in that case that the user needs to either log into an account with permissions to delete or contact someone that can delete that item from the library.

Example attached:

Screenshot (115)

## How have you tested this?

Logged in via a user that does not have permission to delete and see the disabled button, hovering over the disabled button the tool tip displays a message explaining that the user does not have permission to delete:

Screenshot (113)

Regression tested with a user that has permissions to delete:

Screenshot (114)

## Screenshots

Attached under testing instructions and PR description.

### DIFF
--- a/client/components/app/BookShelfToolbar.vue
+++ b/client/components/app/BookShelfToolbar.vue
@@ -74,6 +74,9 @@
         <!-- issues page remove all button -->
         <ui-btn v-if="isIssuesFilter && userCanDelete && !isBatchSelecting" :loading="processingIssues" color="bg-error" small class="ml-4" @click="removeAllIssues">{{ $strings.ButtonRemoveAll }} {{ $formatNumber(numShowing) }} {{ entityName }}</ui-btn>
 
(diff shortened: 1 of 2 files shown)


advplyr:
Hmm this isn't consistent though with everywhere else that we do not show the delete option when the user doesn't have permission to delete

nichwall:
I agree this can probably be closed, because it does not match the Delete options being hidden everywhere else when a user does not have Delete permissions.

PR #4748, opened 2025-10-14
Add MusicBrainz provider

### LINKED ISSUE #4125: [Enhancement]: MusicBrainz Metadata
### Type of Enhancement

Server Backend

Allow matching books to and pulling data down from MusicBrainz.

I understand this request may sound odd, but there is some logic behind me wanting this. I listen to Big Finish's audio drama's, and I recently set up audio bookshelf because Jellyfin's support for audio books is lacking. It works really well, however, sadly most of the metadata providers are pretty lacking in regards to Big Finish's catalogue. They all seem to have some of the early monthly adventures, however that's a pretty small amount of the whole Big Finish catalogue. MusicBrainz however, does include audio books, and has a pretty complete collection of the [Big Finish library](https://musicbrainz.org/label/dd10cb9b-f1cd-4d9c-8b6c-1a5082fe6d43).

It would be quite helpful for me at the least if I could use MusicBrainz to get metadata for these dramas. I do appreciate this may be a niche feature though!

I can make a rough mock up if needed, but I'm not sure it is, I think it's implementation would be similar to other providers like Google Books.

v2.19.5

The matching page:
Image

jwillikers:
## Brief summary

Adds a [MusicBrainz](https://musicbrainz.org/) provider as a source for audiobook metadata and cover art.

## Which issue is fixed?

Fixes #4125.

## In-depth Description

This PR makes it possible to retrieve metadata and cover art for audiobooks from MusicBrainz releases.
The metadata on MusicBrainz is contributed and maintained by its community.
The MusicBrainz API is documented [here](https://musicbrainz.org/doc/MusicBrainz_API) and rate-limiting is documented [here](https://musicbrainz.org/doc/MusicBrainz_API/Rate_Limiting).
For example data, the [Wind and Truth Release Group](https://musicbrainz.org/release-group/200278c4-e2bf-4a99-8758-8253b282e226) is probably a good place to start.
This PR adds a proper MusicBrainz provider implementation with support in the Match and Cover Art dialogues.
Cover Art is provided through the [Cover Art Archive](https://coverartarchive.org/), which is maintained by the Internet Archive.

Right now, I've opted to use the `score` returned from the MusicBrainz search as the confidence rating, although it should be noted that the MusicBrainz score does not incorporate duration. Not sure if that's the best choice or not.

The `genres`, `tags`, `abridged`, and `explicit` fields are all determined via MusicBrainz `tags`. Genres in MusicBrainz are kind of different than genres for audiobooks, and in MusicBrainz, genres are just a specific set of tags anyways. Any entity in MusicBrainz can have tags, and sometimes they make sense in certain places over others. For instance, tags related to audiobook genres often apply to the written work, so they make the most sense on the work and work series entities. However, `unabridged` and `abridged` tags make more sense at the release group level. I basically implemented genres as everything not beloging to a set of special purpose tags, and the presence or abscence of some of those special purpose tags is used to define the values for the `abridged` and `explicit` fields.

For language, I directly use the the [ISO 639-3](https://en.wikipedia.org/wiki/ISO_639-3) three-letter language code used by MusicBrainz. Maybe I should convert this to the full name of the language to match what the Audible provider returns?

### Remaining Work

I could use help with the following, assuming they should even be implemented.

- [ ] Display the disambiguation comment in the Match list, since similar releases are common, i.e. digital and CD releases.
- [ ] Reduce MusicBrainz API usage. Right now, I gather pretty much all of the information for each match listed in the search results. Gathering the necessary metadata used in the search results requires a separate API call for each release to get details including total duration, authors, and narrators. MusicBrainz allows selectively requesting the information that should be included in the response. Not all of the collected metadata is displayed for the matches in the search results listing, like genres, tags, and publishers. Getting all of the tags and genres requires additional API calls for related works and series entities. Those extra API calls result in rate-limiting. Ideally I could provide only the minimum metadata necessary in the results list and upon selection of a match, retrieve the remaining metadata. I'm not sure how to go about that. Another method to avoid rate limiting would be to cache some of the API responses, particularly those for work and series entities which are often the same among similar releases.

### What's Not Implemented in this PR

- [ ] Chapters lookup in the chapters editor. The current chapters editor is made for ASIN's and not set up to allow other providers. It will need to be updated to allow using a MusicBrainz Release ID to obtain the chapters from a MusicBrainz Release's track list or annotation. Since chapters are not incorporated directly from the Match dialogue, returning chapters there isn't useful. I've already implemented functions to parse chapters from a MusicBrainz track list as part of this PR, but figured a follow up PR would be better suited to implementing support for multiple chapters providers.
- [ ] Using AcoustID fingerprinting to lookup a MusicBrainz release. This is significantly more complex and is limited in its utility until the AcoustID server supports tracks with lengths greater than ~9 hours. This would also most likely provide a method for submitting missing AcoustID fingerprints to the AcoustID server.
- [ ] Support for podcasts. I'm not familiar with how podcasts are handled on MusicBrainz, but I know that they are. I'm not planning on adding this myself, but it should be possible.

## How have you tested this?

So far, I've tested its usage to retrieve metadata and cover art with [Skyward](https://musicbrainz.org/release-group/1e97eb8b-9d20-4cb2-be3f-d498a06f70c1).

### DIFF
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -9,11 +9,13 @@ const CustomProviderAdapter = require('../providers/CustomProviderAdapter')
 const Logger = require('../Logger')
 const { levenshteinDistance, levenshteinSimilarity, escapeRegExp, isValidASIN } = require('../utils/index')
 const htmlSanitizer = require('../utils/htmlSanitizer')
+const MusicBrainz = require('../providers/MusicBrainz')
 
 class BookFinder {
   #providerResponseTimeout = 10000
 
   constructor() {
+    this.musicbrainz = new MusicBrainz()
     this.openLibrary = new OpenLibrary()
     this.googleBooks = new GoogleBooks()
     this.audible = new Audible()
@@ -23,11 +25,12 @@ class BookFinder {
     this.audiobookCovers = new AudiobookCovers()
     this.customProviderAdapter = new CustomProviderAdapter()
 
-    this.providers = ['google', 'itunes', 'openlibrary', 'fantlab', 'audiobookcovers', 'audible', 'audible.ca', 'audible.uk', 'audible.au', 'audible.fr', 'audible.de', 'audible.jp', 'audible.it', 'audible.in', 'audible.es']
+    this.providers = ['google', 'itunes', 'musicbrainz', 'openlibrary', 'fantlab', 'audiobookcovers', 'audible', 'audible.ca', 'audible.uk', 'audible.au', 'audible.fr', 'audible.de', 'audible.jp', 'audible.it', 'audible.in', 'audible.es']
 
     this.verbose = false
   }
 
+  // todo Not sure what this is for, but MusicBrainz can also be used to perform ISBN lookups.
   async findByISBN(isbn) {
     var book = await this.openLibrary.isbnLookup(isbn)
     if (book.errorCode) {
@@ -190,6 +193,22 @@ class BookFinder {
     const region = provider.includes('.') ? provider.split('.').pop() : ''
     const books = await this.audible.search(title, author, asin, region, this.#providerResponseTimeout)
     if (this.verbose) Logger.debug(`Audible Book Search Results: ${books.length || 0}`)
+    if (!books) return []
+    return books
+  }
+
+  /**
+   *
+   * @param {string} title
+   * @param {string} author
+   * @param {string} asin
+   * @param {string} isbn
+   * @returns {Promise<Object[]>}
+   */
+  async getMusicBrainzResults(title, author, asin, isbn) {
+    const books = await this.musicbrainz.search(title, author, asin, isbn)
+    // todo Sort by closest score and then duration?
+    if (this.verbose) Logger.debug(`MusicBrainz Book Search Results: ${books.length || 0}`)
     if (!books) return []
     return books
   }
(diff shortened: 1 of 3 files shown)


server/finders/BookFinder.js
```diff
@@ -609,8 +644,8 @@ class BookFinder {
         searchResults.push(...providerResults)
       }
     } else if (provider === 'best') {
-      // Best providers: google, fantlab, and audible.com
-      const bestProviders = ['google', 'fantlab', 'audible']
+      // Best providers: musicbrainz, google, fantlab, and audible.com
+      const bestProviders = ['musicbrainz', 'google', 'fantlab', 'audible']
```
Vito0912: I would not add the provider directly to the best providers yet. This was newly introduced to increase speed with the cover search and are provider that have a wide range of audiobooks or/and books

server/finders/BookFinder.js
Vito0912, line 466: This is a big chunk of duplicate logic (that can be easily deduplicated). This is not directly related to your PR, but imho it would be better to try using an unified design, so this can expanded further to more providers.

(Tbh I don't fully understand why this is not applied for all providers. Providers that have more data can then use a fine-tuned logic)

Vito0912:
Btw. on an unrelated note: 
Iirc you mentioned having some scripts to contribute to MusicBrainz? 
I have quite a collection of Audible audiobooks in a structured format that we might be able to integrate with MusicBrainz, or do you think this is out of scope of MusicBrainz (no clue, for me it's just a music/track platform)? If so, you can reach out to me on DC or via email in my GitHub.

server/finders/BookFinder.js
jwillikers: Done, I dropped it.

server/finders/BookFinder.js
jwillikers: I've combined the sort sections for both providers now to be the same. It should be trivial to allow it for more, if not all providers now.

PR #4750, opened 2025-10-15
Add metadata providers API and use them on web client

### LINKED ISSUE #3820: [Enhancement]: Metadata Provider per Library allow Custom Metadata Providers to be Selected
### Type of Enhancement

Web Interface/Frontend

Currently when you add a Custom Metadata provider you can manually search using this, but you cannot set the Metadata Provider in the Library view to the custom provider. 

image

Allow you to have a custom provider for a particular library for example Big Finish on a library that is exclusive with their books. 

Dropdown should be amended to include any custom metadata providers configured under Settings - Custom Metadata Providers

v2.17.7

Settings - Library - Edit - Details

mikiher:
## Brief summary

Up until now, the client maintained its own list of providers, and that was always in danger of getting out of sync with the server.
This PR adds new API endpoints to get the different providers.

## Which issue is fixed?

This didn't have an associated bug, but on the way it fixes #3820 

### Changes in the server

These API endpoints were added:
```js
    this.router.get('/search/providers/books', SearchController.getBookProviders.bind(this))
    this.router.get('/search/providers/books/covers', SearchController.getBookCoverProviders.bind(this))
    this.router.get('/search/providers/podcasts', SearchController.getPodcastProviders.bind(this))
    this.router.get('/search/providers/podcasts/covers', SearchController.getPodcastCoverProviders.bind(this))
    this.router.get('/search/providers/authors', SearchController.getAuthorProviders.bind(this))
    this.router.get('/search/providers/chapters', SearchController.getChapterProviders.bind(this)
```
Notes:
- Book Cover providers include cover-only providers, and All and Best options
- All providers include relevant custom metadata providers.
- Also took the chance to validate query parameters for existing /search API in SearchController.

### Changes in the client

- Prefetched all used lists in layout init (`layouts/default.vue`)
- Updated relevant lists when custom providers is added/removed
- Changed `store/scanners.js` state, getters, actions, and mutations to use the new api
- Made required changes in users of `scanners.js`
- Removed unnecessary fetches of lists - all fetches are now initiated from `layouts/default.vue`

Notes:
- The author and chapter providers are not used currently in the client, as there's only one provider - we can add them in the future if needed.

## How have you tested this?

- Checked that Match and Cover tabs contain all providers as before (for books and podcasts)
- Checked that LibraryEdit shows all available providers including custom ones.
- Checked that BatchQuickMatchModal displays available providers.

### DIFF
--- a/server/controllers/SearchController.js
+++ b/server/controllers/SearchController.js
@@ -42,7 +42,7 @@ class SearchController {
    * Fetches a library item by ID
    * @param {string} id - Library item ID
    * @param {string} methodName - Name of the calling method for logging
-   * @returns {Promise<LibraryItem>}
+   * @returns {Promise<import('../models/LibraryItem').LibraryItemExpanded>}
    */
   static async fetchLibraryItem(id) {
     const libraryItem = await Database.libraryItemModel.getExpandedById(id)
--- a/server/finders/PodcastFinder.js
+++ b/server/finders/PodcastFinder.js
@@ -7,9 +7,9 @@ class PodcastFinder {
   }
 
   /**
-   * 
-   * @param {string} term 
-   * @param {{country:string}} options 
+   *
+   * @param {string} term
+   * @param {{country:string}} options
    * @returns {Promise<import('../providers/iTunes').iTunesPodcastSearchResult[]>}
    */
   async search(term, options = {}) {
@@ -20,12 +20,16 @@ class PodcastFinder {
     return results
   }
 
+  /**
+   * @param {string} term
+   * @returns {Promise<string[]>}
+   */
   async findCovers(term) {
     if (!term) return null
     Logger.debug(`[iTunes] Searching for podcast covers with term "${term}"`)
-    var results = await this.iTunesApi.searchPodcasts(term)
+    const results = await this.iTunesApi.searchPodcasts(term)
     if (!results) return []
-    return results.map(r => r.cover).filter(r => r)
+    return results.map((r) => r.cover).filter((r) => r)
   }
 }
 module.exports = new PodcastFinder()


server/finders/BookFinder.js
```diff
@@ -402,7 +402,8 @@
       let authorCandidates = new BookFinder.AuthorCandidates(cleanAuthor, this.audnexus)
 
       // Remove underscores and parentheses with their contents, and replace with a separator
-      const cleanTitle = title.replace(/\[.*?\]|\(.*?\)|{.*?}|_/g, ' - ')
+      // Use negated character classes to prevent ReDoS vulnerability
+      const cleanTitle = title.replace(/\[[^\]]*\]|\([^)]*\)|{[^}]*}|_/g, ' - ')
```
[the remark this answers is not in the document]
mikiher: Fixed.

server/finders/BookFinder.js
```diff
@@ -668,7 +669,9 @@
   let stripped = keepSubtitle ? title : stripSubtitle(title)
 
   // Remove text in paranthesis (i.e. "Ender's Game (Ender's Saga)" becomes "Ender's Game")
-  let cleaned = stripped.replace(/ *\([^)]*\) */g, '')
+  // Use a safe two-pass approach to prevent ReDoS vulnerability
+  let cleaned = stripped.replace(/\([^)]*\)/g, '') // Remove parenthetical content
```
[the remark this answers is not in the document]
mikiher: Fixed.

mikiher:
I now also fixed all CodeQL security complaints.

Most importantly, I think it's best if we use the new `utils` function `getQueryParamAsString`  in all controllers. 
It does two things:
- always returns string | null (which just reading it from req.query doesn't guarantee)
- caps the length at 1000.

mikiher:
Also added a small unrelated fix to a bug in new cover search, which accidentally broke itunes cover search.

server/routers/ApiRouter.js
```diff
@@ -283,6 +283,12 @@ class ApiRouter {
     this.router.get('/search/podcast', SearchController.findPodcasts.bind(this))
     this.router.get('/search/authors', SearchController.findAuthor.bind(this))
     this.router.get('/search/chapters', SearchController.findChapters.bind(this))
+    this.router.get('/search/providers/books', SearchController.getBookProviders.bind(this))
+    this.router.get('/search/providers/books/covers', SearchController.getBookCoverProviders.bind(this))
+    this.router.get('/search/providers/podcasts', SearchController.getPodcastProviders.bind(this))
+    this.router.get('/search/providers/podcasts/covers', SearchController.getPodcastCoverProviders.bind(this))
+    this.router.get('/search/providers/authors', SearchController.getAuthorProviders.bind(this))
+    this.router.get('/search/providers/chapters', SearchController.getChapterProviders.bind(this))
```
advplyr: I'm thinking that these endpoints are unnecessary.

Right now we are returning the custom metadata providers in "/api/libraries/:id" when "?include=filterdata" query param is on the request.
https://github.com/advplyr/audiobookshelf/blob/a87ea327151fcb7ec67ce981bb6a44878a9ab343/server/controllers/LibraryController.js#L222-L232

Custom metadata providers were added on the "filterdata" library request so it isn't a great implementation, but I do like limiting the number of API requests by returning the relevant data for that library when fetching the library.

With this implementation we would be making 4 additional API requests on every page load when they may be unnecessary if the user isn't doing any matching. Also, they may not have a book or a podcast library which would be 2 unnecessary requests either way. Podcast covers will always be an unnecessary request since we don't have any of those and I doubt we ever will since covers come from podcast RSS feeds.

I think we should either:
a) Make this a single "providers" API endpoint that is only requested if necessary
b) Return the providers for the specific library when the library is fetched. This could be via the "filterdata" query param like we are doing now or we could come up with a more fitting API endpoint to fetch the library + data.

mikiher: A few comments:
1. I concede that having numerous endopoints is unneccessary.
2. Providers are not library specific (you can argue that they are library-type specific, but that's not the same thing). 
   - I think it's also a poor choice to put custom providers in `/libraries/:id` response - they're also not library specific
   - providers are also needed when adding a new library - there's no /libraries/:id payload at that point.
3. The current implementation doesn't make additional calls on every page load. It only fetches them when the whole app is reloaded (Ctrl+R).
   - This is why I put the fetch in default.vue
   - I only fetch them again when a custom provider changes.

With that in mind, I'm going to make the following changes:
1. Use a single API point `/search/providers` to get all providers, including custom providers.
   - I'll minimize the payload so it doesn't have repetitions, and build the state variables on the client.
3. Remove the custom providers payload from /libraries/:id?include=filterdata
4. Only call `/search/providers` when a list is needed.
5. Continue to only call `/search/providers` once (unless custom providers are added/removed) during the app's lifetime.

Does this seem OK?

Also regarding your comment:

> Podcast covers will always be an unnecessary request since we don't have any of those and I doubt we ever will since covers come from podcast RSS feeds.

I disagree. Covers from itunes are sometimes slightly different from the ones in the podcast's RSS feed. I like the choice. Also, I understand custom providers for podcasts are not supported right now. Won'r they be supported in the future?

advplyr: Those changes look good.

> I disagree. Covers from itunes are sometimes slightly different from the ones in the podcast's RSS feed. I like the choice. Also, I understand custom providers for podcasts are not supported right now. Won'r they be supported in the future?

Yes, I think we should support custom podcast providers in the future. What I mean is that I don't see there ever being a case where we would have a podcast provider that is for covers only. That comment can be disregarded though since we won't be doing additional db reads or requests to fetch cover providers.

mikiher:
OK, PTAL again.

server/controllers/SearchController.js
```diff
@@ -16,25 +38,121 @@ const { isValidASIN } = require('../utils')
...
+          valid: false,
+          error: {
+            status: 400,
+            message: 'Invalid request query params'
+          }
+        }
+      }
+    }
+    return { valid: true }
+  }
```
advplyr: This check appears to be unnecessary now that the query params are being validated with `getQueryParamAsString`.

mikiher: Right now it's not redundant, because getQueryParamAsString can return null when the input is illegal (either an Array, or too long) - but then it's true that `getQueryParamAsString` could have thrown an error instead of returning a null value. I'll modify it to do so.

mikiher: OK, Done.
- Removed `validateStringParams` and `validateRequiredString`
- Modified `getQueryParamAsString` to throw validation errors and handle required params

server/controllers/SearchController.js
```diff
@@ -16,26 +38,70 @@
...
+      const libraryItem = await SearchController.fetchLibraryItem(id)
 
-    const results = await BookFinder.search(libraryItem, provider, title, author)
-    res.json(results)
+      const results = await BookFinder.search(libraryItem, provider, title, author)
+      res.json(results)
+    } catch (error) {
+      Logger.error(`[SearchController] findBooks: ${error.message}`)
+      if (error instanceof ValidationError || error instanceof NotFoundError) {
+        return res.status(error.status).send(error.message)
```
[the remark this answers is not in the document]
mikiher: Fixed.

server/controllers/SearchController.js
```diff
@@ -45,20 +111,24 @@
...
-      Logger.error(`[SearchController] findCovers: Invalid title sent in query`)
-      return res.sendStatus(400)
+      let results = null
+      if (podcast) results = await PodcastFinder.findCovers(title)
+      else results = await BookFinder.findCovers(provider, title, author)
+      res.json({ results })
+    } catch (error) {
+      Logger.error(`[SearchController] findCovers: ${error.message}`)
+      if (error instanceof ValidationError) {
+        return res.status(error.status).send(error.message)
```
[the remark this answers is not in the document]
mikiher: Fixed.

server/controllers/SearchController.js
mikiher: Also Fixed the security warnings. The remaining codeCL failure seems either bogus or unrelated.

mikiher:
I addressed your latest comment (the code looks much better now). PTAL again.

mikiher:
Also upgraded codeql-actions to v3, as v2 is retired.

advplyr:
This is great, thanks!

PR #4756, opened 2025-10-19
Change token expiry

Vito0912:
## Brief summary

This increases the default tokenExpiry for the refresh and access token. 

## Which issue is fixed?

/

## In-depth Description

It was discussed some time ago that the default duration might eventually be increased /decreased. 
Since there have been no reports of authentication issues related to them, imho I think it's fair to say that the time has come where it should be safe to increase the refresh token and decrease the access token.
Especially for people switching between apps, 7 days can easily be reached.

## How have you tested this?

/

## Screenshots

/

### DIFF
--- a/server/auth/TokenManager.js
+++ b/server/auth/TokenManager.js
@@ -12,9 +12,9 @@ class TokenManager {
 
   constructor() {
     /** @type {number} Refresh token expiry in seconds */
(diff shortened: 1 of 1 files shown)


advplyr:
Yep, this works. Thanks!

PR #4770, opened 2025-10-23
Don't require listening today for 'Days in a row' stat

jamerst:
## Brief summary

Adjust logic for calculating 'Days in a row' stat to not require listening on the current day.

I usually listen at the end of the day, so this allows viewing the stat before you've listened for the day.

## Which issue is fixed?

N/A - no associated issue currently

## In-depth Description

Changes day counting logic to start counting from yesterday instead of today, so that there doesn't need to be any activity for the current day.

If there is activity for the current day this is counted at the end once the last day with no activity has been found.

## How have you tested this?

Set some temporary test data in the `ApiRouter.getUserListeningStatsHelpers` method to mock appropriate activity on dates:
image

## Screenshots

No visual changes, only change is to the value shown for "Days in a row".

With no activity on the current day:
image

With activity on the current day:
image

### DIFF
--- a/client/components/stats/DailyListeningChart.vue
+++ b/client/components/stats/DailyListeningChart.vue
@@ -186,14 +186,14 @@ export default {
     daysInARow() {
       var count = 0
       while (true) {
-        var _date = this.$addDaysToToday((count * -1) - 1)
-        var datestr = this.$formatJsDate(_date, 'yyyy-MM-dd')
+        const _date = this.$addDaysToToday(count * -1 - 1)
+        const datestr = this.$formatJsDate(_date, 'yyyy-MM-dd')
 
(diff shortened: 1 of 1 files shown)


nichwall (approved):

advplyr:
Thanks!

jamerst:
Thanks for merging!

Guessing this change also needs making to https://github.com/advplyr/audiobookshelf-app? Doesn't look like they share code

nichwall:
> Guessing this change also needs making to https://github.com/advplyr/audiobookshelf-app? Doesn't look like they share code

That is correct, they are two separate repositories.

advplyr:
fyi https://github.com/advplyr/audiobookshelf/issues/4865
not sure how to handle

Issue #4772, opened 2025-10-24
[Enhancement]: Placement of Websocket Connection Message


timpoppleton:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

My connection to the ABS websocket connection occasionally fails when on my work network, this is just due to the company proxy.

This brought up a little bit of a design issue, the placement of the "socket connected/disconnected" popup.

Currently it's located at the bottom centre of the web interface, this covers up the web player controls.

I was wondering if this might be simply moved to the top right corner perhaps?

### Why would this be helpful?

When the message is up, it's still possible to play the audio and it'd be useful to use the controls to pause/play/skip when this message comes up.

### Future Implementation (Screenshot)

See below and just place it under the username in the top right corner.

### Audiobookshelf Server Version

v2.30.0

### Current Implementation (Screenshot)

<img width="900" height="197" alt="Image" src="https://github.com/user-attachments/assets/0f52264a-1a5b-4380-baae-2f1cc20664db" />

rcarteraz:
I had this same thing happen recently and agree, it's frustrating. Would be best if moved.

PR #4782, opened 2025-10-29
move socket toast to top of the page

brandonartner:
## Brief summary

Minor change to move the websocket message/error toast to the top of the screen on the web client.

## Which issue is fixed?

Enhancment https://github.com/advplyr/audiobookshelf/issues/4772

## In-depth Description

Play/Pause button would become inaccessible when the websocket message or error was displayed.

Screenshot from 2025-10-29 14-55-31

Moving the toast message to the top of the screen is a minimal change to prevent this.

Screenshot from 2025-10-29 15-06-44

## How have you tested this?

Opened and played an audiobook.
Turned off dev server to produce  websocket disconnect error.
Toast now shows at top of page rather than covering the pause button at the bottom.

### DIFF
--- a/client/assets/app.css
+++ b/client/assets/app.css
@@ -250,3 +250,12 @@ Bookshelf Label
 .abs-btn:disabled::before {
   background-color: rgba(0, 0, 0, 0.2);
 }
+
+/* Padding for toastification toasts on the bottom to not cover media player*/
+.media-player .Vue-Toastification__toast.toast-mb-40 {
+  margin-bottom: calc(var(--spacing) * 40);
+}
+
+.media-player .Vue-Toastification__toast.toast-mb-48 {
+  margin-bottom: calc(var(--spacing) * 48);
+}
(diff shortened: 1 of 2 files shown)


Vito0912:
I think the idea is good, but can't we just add extra space at the bottom if the player is open?

The implemented position blocks navigation on the PWA (/ devices that have smaller screens), and because of how the OS works, the socket will disconnect and reconnect when opening the app, preventing the user from switching.

brandonartner:
> I think the idea is good, but can't we just add extra space at the bottom if the player is open?
> 
> The implemented position blocks navigation on the PWA (/ devices that have smaller screens), and because of how the OS works, the socket will disconnect and reconnect when opening the app, preventing the user from switching.

Updated the changes. Moved the toast back to the bottom and added a watcher based on when the media player is opened or closed to add/remove new css classes that add extra space below toast. CSS classes are based on the responsive height of the media player.

Let me know if it's preferred not to add watchers and would rather just have the toast be above where the media player would be regardless of if it is rendered.

Screenshot from 2025-11-01 11-50-47 Screenshot from 2025-11-01 11-50-38

Screenshot from 2025-11-01 11-51-06 Screenshot from 2025-11-01 11-51-18

PR #4788, opened 2025-10-31
Increasing the timeout for BookFinder - some metadata providers need more time

TN-SKYC:
## Brief summary
Some custom metadata providers are slow due to the providers throttling the responses. In v2.30 the timeout went down to 10s (from 30s). Even the 30s sometimes was not enough.

## Which issue is fixed?
Timeout increased for waiting for data from metadata providers.

## In-depth Description
Changing the timeout from 10s to 120s to let custom metadata providers function properly - we have no control over providers decisions to throttle the responses.

## How have you tested this?
Yes, it's a simple timeout increase. When timeout hits, logs show:
```
2025-10-28 20:52:36.963 ERROR [CustomMetadataProvider] Search error [AxiosError: timeout of 30000ms exceeded] { code: 'ECONNABORTED', config: { transitional: { silentJSONParsing: true, forcedJSONParsing: true, clarifyTimeoutError: false }, adapter: [Function: httpAdapter], transformRequest: [ [Function: transformRequest] ], transformResponse: [ [Function: transformResponse] ], timeout: 30000, xsrfCookieName: 'XSRF-TOKEN',
```

### DIFF
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -11,7 +11,7 @@ const { levenshteinDistance, levenshteinSimilarity, escapeRegExp, isValidASIN } 
 const htmlSanitizer = require('../utils/htmlSanitizer')
 
 class BookFinder {
-  #providerResponseTimeout = 10000
+  #providerResponseTimeout = 120000
 
   constructor() {
     this.openLibrary = new OpenLibrary()


Vito0912:
Sorry to be the bummer again, but if a request takes 2 minutes to complete, then you either a) DOS the provider with your metadata implementation or b) have a very poor implementation.  
I am not sure if this is already an ENV, but in my opinion, it should be an ENV instead of being set from 30s to 120s. 30s should be the maximum for default set. 
Even the slower OpenLibrary often responds in under 10 seconds.

TN-SKYC:
I agree, it should be an ENV. The change is from 10s. 30s was back in 2.29.

Vito0912:
Yeah ik. But 10s to 120s sounds a bit more dramatic and so I opted for 30s because of this (and I think "reverting" it to 30s in sensible)

PR #4795, opened 2025-11-03
[Enhancement] Simple ratings from audible.com & custom providers

danieljrich:
## Brief summary

Adds star rating support for audiobooks, displaying ratings from Audible.com when available and allowing manual rating entry/editing.

**NOTE: this will only affect new items. In order to get rating for existing items, quick-match will need to be run server-wide**

## Which issue is fixed?

N/A - New feature

## In-depth Description

This PR adds star rating functionality for audiobooks:

- **Database**: Adds `rating` column to `books` table (FLOAT, nullable)
- **Display**: Shows 5-star ratings with half-star support on:
  - Item details page (below author name)
  - Match card in match modal
  - Sort by rating option in library dropdown
  - Subtitle display when sorting by rating
- **Editing**: Rating can be added/edited in the Details tab and Match modal
- **Removal**: Setting rating to 0 removes it (saves as null)
- **Sorting**: New "Rating" option in sort dropdown with proper null handling

Rating values come from Audible metadata when matching, but can also be manually entered for any book regardless of ASIN.

## How have you tested this?

1. Matched a book from Audible.com and verified rating displays with stars
2. Manually added/edited ratings in Details tab
3. Set rating to 0 to verify it gets removed
4. Sorted library by rating and verified display (asc, and dec)
5. Verified ratings show in match modal and can be edited there
6. Quick-Match on existing matched item successfully added rating.

## Screenshots
rating on item
image
sort by rating
image

set to 0 to remove
image

no stars if set to 0
image

rating on match view
image


Vito0912:
This looks great. 

I did not test it (and maybe the few lines already took care of it). Do you respect quick matches or only matches made through the web interface?

 This is probably answered by my first question automatically, but do you also allow other providers (especially custom metadata providers) to add this information?

danieljrich:
@Vito0912 great call outs. Thank you!

1. I missed quick match. Oversight on my part. But adding that does make it very easy to retroactively go back through and just quick match existing media to fetch the rating. So thanks for that! I just tested it and have commited the change

2.  From what I saw in the code; rating is already mentioned only in the audible.js providers file. So right now it's just Audible. I commited some changes so it's supported for custom providers.

client/components/cards/LazyBookCard.vue
```diff
@@ -349,6 +349,13 @@ export default {
         if (this.mediaMetadata.publishedYear) return this.$getString('LabelPublishedDate', [this.mediaMetadata.publishedYear])
         return '\u00A0'
       }
+      if (this.orderBy === 'media.metadata.rating') {
+        if (this.mediaMetadata.rating) {
+          const ratingValue = typeof this.mediaMetadata.rating === 'object' && this.mediaMetadata.rating.average ? this.mediaMetadata.rating.average : Number(this.mediaMetadata.rating)
```
Vito0912: Why is it unsure which type it is?

client/components/modals/item/tabs/Match.vue
```diff
@@ -174,6 +174,16 @@
           </div>
         </div>
 
+        <div v-if="selectedMatchOrig.rating" class="flex items-center py-2">
+          <ui-checkbox v-model="selectedMatchUsage.rating" checkbox-bg="bg" @input="checkboxToggled" />
+          <div class="grow ml-4">
+            <ui-text-input-with-label v-model="selectedMatch.rating" type="number" step="0.1" min="0" max="5" :disabled="!selectedMatchUsage.rating" label="Rating" />
+            <p v-if="mediaMetadata.rating" class="text-xs ml-1 text-white/60">
+              {{ $strings.LabelCurrently }} <a :title="$strings.LabelClickToUseCurrentValue" class="cursor-pointer hover:underline" @click.stop="setMatchFieldValue('rating', typeof mediaMetadata.rating === 'object' && mediaMetadata.rating.average ? mediaMetadata.rating.average : Number(mediaMetadata.rating))">{{ typeof mediaMetadata.rating === 'object' && mediaMetadata.rating.average ? mediaMetadata.rating.average.toFixed(1) : Number(mediaMetadata.rating).toFixed(1) }}</a>
```
Vito0912: Same as above

client/components/modals/item/tabs/Match.vue
```diff
@@ -569,8 +585,19 @@ export default {
...
       updatePayload.metadata = {}
 
+      // Always include rating if it exists in the match
+      if (this.selectedMatch?.rating !== undefined && this.selectedMatch?.rating !== null) {
+        const ratingValue = Number(this.selectedMatch.rating)
+        if (!isNaN(ratingValue) && ratingValue > 0) {
+          updatePayload.metadata.rating = ratingValue
+        } else if (ratingValue === 0 || isNaN(ratingValue)) {
+          // Set to null to remove rating
+          updatePayload.metadata.rating = null
```
Vito0912: Does undefined would make more sense here. To match the other behaviour (line 600)

server/providers/CustomProviderAdapter.js
```diff
@@ -109,7 +109,8 @@ class CustomProviderAdapter {
         tags: toStringOrUndefined(tags),
         series: validateSeriesArray(series),
         language: toStringOrUndefined(language),
-        duration: !isNaN(duration) && duration !== null ? Number(duration) : undefined
+        duration: !isNaN(duration) && duration !== null ? Number(duration) : undefined,
+        rating: rating !== undefined && rating !== null ? (!isNaN(Number(rating)) && Number(rating) > 0 ? Number(rating) : undefined) : undefined
```
Vito0912: Here it's undefined now?

Vito0912 (commented):
I am just a contributor so my comments do not mean anything nor do you have to change these. Just left a few comments

client/components/cards/LazyBookCard.vue
danieljrich: good catch. This is a result of my earlier iterations I missed when changing.

client/components/modals/item/tabs/Match.vue
danieljrich: fixed.

client/components/modals/item/tabs/Match.vue
danieljrich: agreed! Fixed

server/providers/CustomProviderAdapter.js
danieljrich: fixed

Lxstr:
If I may, this seems like a cool feature, however for many like myself it's additional visual activity and I would prefer this feature completely off by default.

danieljrich:
> If I may, this seems like a cool feature, however for many like myself it's additional visual activity and I would prefer this feature completely off by default.

@Lxstr Totally support that. Good call out, and I appreciate the feedback.

I just pushed up a change that adds a toggle switch to Settings, and sets it to "off" by default.

image

palenshus:
Any chance this will get checked in soon?

PR #4797, opened 2025-11-04
Add a 'Match after scan' option to the schedule scan config

indiefan:
## Brief summary

Adds an optional configuration toggle to the "Schedule" tab in edit library to also run "Match Books" after the manually scheduled scans take place. Includes a slider for minimum confidence level threshold to accept the match.

## Which issue is fixed?

I'm not aware of an open issue for this feature. Allows for manually scheduling matching to be performed at a regular cadence.

## In-depth Description

Adds a boolean to the library settings model and confidence level threshold, then uses these in the CronManager to execute the match after the scan completes. Refactors ApiRouter a bit to avoid a circular dependency with the CronManager.

## How have you tested this?

Manually triggered in my dev build and ran npm test in both the root and client folders (are there other tests I should run?).

## Screenshots

image

image

### DIFF
--- a/client/components/modals/libraries/EditModal.vue
+++ b/client/components/modals/libraries/EditModal.vue
@@ -126,6 +126,7 @@ export default {
           skipMatchingMediaWithIsbn: false,
           autoScanCronExpression: null,
           matchAfterScan: false,
+          matchMinConfidence: 0,
           hideSingleBookSeries: false,
           onlyShowLaterBooksInContinueSeries: false,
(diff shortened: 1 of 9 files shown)


JonasFocke01:
Is the match taking place only on timed full-library matches, or is it going to match after the "watch filesystem" event happens?

I looked over 250 issues and 100 PRs to find this fix, and im VERY happy someone had the same thought as myself. Thank you for your work!

PR #4807, opened 2025-11-06
Add ability to hide and unhide series

### LINKED ISSUE #870: [Enhancement]: Nested Series or add a whole Series to a collection

So for books that have series or multiple series in it would be great to have a overall group then when you go there it shows you the Seasons then when clicking the Season it shows the book.  Example the Dr. who books,

### LINKED ISSUE #2092: [Enhancement]: Hide sub series in the general library view

I have an idea how to solve the problem with the subseries in a simple way.

Currently, I do it in such a way that I also enter the main series in the "Series" field for each sub-series. This has the effect that all sub-series are displayed within the main series when the "collapse series" option is activated.

The problem with this solution is that in this case the subseries are shown twice in the library view. Once within the main series and once as independent series in the general library view.

One solution to this problem would be to mark the sub-series. To do this, one would have to enter a kind of code word, e.g. "sub-series", in the series input field where one enters the series name(s) for an audio book. This firmly defined code word causes Abs not to create a new series, but to remove the series marked in this way from the general library view but not from the main series. The series marked in this way would then only be visible within the main series and would no longer appear twice in the database.

DoctorDalek1963:
## Brief summary

Adds the ability to hide or unhide a series. A hidden series won't show up in the series tab, but is still visible as a collapsed sub-series. Series are hidden on a per-user basis.

## Which issue is fixed?

Fixes #870. Fixes #2092.

## In-depth Description

Adds an option in the toolbar menu when viewing a series to hide/unhide. Hiding a series adds it to a list `user.extraData.hiddenSeries`. This is also where we store which series are hidden from "continue series" on the home page.

When the client asks the server for all the series, the SQL query now filters out the series which are hidden for that user. Currently this is not respected on the home page, so the "recent series" section will show hidden series. I'm not sure if they should be hidden here also.

Hidden series will still be suggested in the dropdown when adding a book to a series, and they will still appear on the page for a book, under the name like normal. This means hidden series are still easily accessible. Hidden series will also be considered for the "continue series" section of the home page. For use cases involving nested series, it is easy to remove just the sub-series from the "continue series" section to avoid clutter.

## How have you tested this?

I've tested this on my own server with 179 Big Finish audio dramas, 22 series, and now over 10 hidden series.

## Screenshots

New series menu with option to hide series


server/controllers/MeController.js
```diff
@@ -381,6 +381,44 @@ class MeController {
...
+      return res.sendStatus(404)
+    }
+
+    const hasUpdated = await req.user.unhideSeries(req.params.id)
+    if (hasUpdated) {
+      SocketAuthority.clientEmitter(req.user.id, 'user_updated', req.user.toOldJSONForBrowser())
+    }
+    res.json(req.user.toOldJSONForBrowser())
+  }
+
```
Vito0912, line 444: Should probably merged into one, and POST instead of GET. We change something.

Vito0912 (changes requested):

server/controllers/MeController.js
DoctorDalek1963: I was just copying the example of `removeSeriesFromContinueListening` and `readdSeriesFromContinueListening` from directly above. I can combine the hide/unhide endpoints into one POST, but these two should probably be combined in a different PR as well.

Vito0912 (approved):
LGTM, but I am just a contributor

DoctorDalek1963:
@advplyr Could you review this please?

jadero:
That would actually be a sufficient remedy for [this](https://github.com/advplyr/audiobookshelf/issues/5069) feature request. Looking forward for this feature to be merged.

thelistenersfury:
Can't wait for this to be merged in, it's exactly what I wanted

Vito0912:
I am not a maintainer and can't say for sure, but as this includes frontend changes and the frontend gets rewritten, I don't think this will be merged at all, as recently no real frontend changing PRs got merged and on newer ones it got commented they won't be merged

mynery:
While I support this change, I kinda don't think it fixes the mentioned issues, because it repeats the work for all users.
It's fine for a single user instance, but I personally would love to have a button where every user could decide whether they are interested in being shown subseries or not. As far as I understand, this change would practically mean that every new user who joins will have to start off with removing a bunch of series if they want to have a view intended by the owner of the library.

Issue #4798, opened 2025-11-04
[Bug]: EAC3 playback fails on browser and some android devices.


kpmgeek:
### What happened?

Dolby Atmos E-AC3 encoded MP4 files fail to play in the browser on Firefox on linux, and also fail to play on some android phones using both the official app and Lissen due to those phones not included said codec for MediaCodec.  Playback reports progress but no audio is ever produced. On iOS and Android devices supporting EAC3 decoding internally, this works perfectly.

### What did you expect to happen?

I expected playback to work, presumably by server-side transcoding. Bundling an ffmpeg-based decoder on the client side would also work.

### Steps to reproduce the issue

1. 

### Audiobookshelf version

2.30.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Logs

```shell
General
Complete name                            : /home/kpmgeek/Downloads/Harry Potter and the Philosopher’s Stone (Full-Cast Edition).m4b
Format                                   : MPEG-4
Format profile                           : Base Media / Version 2
Codec ID                                 : mp42 (dby1/isom/iso2/mp41)
File size                                : 2.80 GiB
Duration                                 : 8 h 41 min
Overall bit rate mode                    : Constant
Overall bit rate                         : 769 kb/s
Title                                    : Harry Potter and the Philosopher’s Stone (Full-Cast Edition)
[...]

PR #4808, opened 2025-11-07
Support eac3 and ac3 in transcodes

Vito0912:
## Brief summary
Basically, HLS allows AC-3. The problem is that if the browser does not support AC-3, it fails again. Since the MIME type is MP4, we have to force the codec by codec. 
This has the downside that browsers that fail and transcode an AC-3 file will only get an AAC file. But I think this should never happen because it would then perform a direct play. So this should fix EAC-3/AC-3 not playable

## Which issue is fixed?

#4798 

## In-depth Description

See the conversation in the attached issue and my short description

## How have you tested this?

Tried to play an EAC-3 file with Firefox which does not support it. Before it failed. Now it worked. It should not affect any other codecs. So nothing that was not broken should be able to be broken now.

Thanks to @TheStaticTurtle

### DIFF
--- a/server/objects/Stream.js
+++ b/server/objects/Stream.js
@@ -275,7 +275,7 @@ class Stream extends EventEmitter {
 
     const codecOptions = [`-loglevel ${logLevel}`, '-map 0:a']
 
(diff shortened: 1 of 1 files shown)


Vito0912:
@advplyr I know you are busy with the frontend, but this might be a PR to look into in the near future as eac-3 seems to be the format the new Harry Potter books come with, which are very popular

advplyr:
Thanks!

PR #4815, opened 2025-11-10
Update wording for library watcher settings and scan note

### LINKED ISSUE #4095: [Enhancement]: Rephrase some settings
### Type of Enhancement

Web Interface/Frontend

(I have submitted an identical issue for the mobile app.)

Some of the binary settings include the word "Enable" or "Disable" in the string. "Enable" is redundant, and "Disable" is confusing: to use such a setting, you enable the disabling of a functionality. I propose rephrasing these strings so that they are easier to parse. Here are the ones I found, and my suggestion (the extra information would need updating as well):

## Root account > Settings > Scanner
Enable Watcher → Automatically scan library for changes

## Backups
Enable automatic backups → Automatic backups

I think this would make the settings easier to understand.

Before:
Image

After:
Image

v.2.19.5

_No response_

mikiher:
## Brief summary

This provides a slightly better solution for #4095, by changing the wording a bit on watcher settings and the note appearing on the library scan schedule settings.

## Which issue is fixed?

Fixes #4095 

## In-depth Description

The wording goes back to use the word "watch" instead of "scan" in the watcher settings, and also updates the note on the Library Scan Schedule that refers to those settings.


advplyr:
Works for me, thanks!

Issue #4584, opened 2025-08-12
[Enhancement]: Add sort options for narrators in Narrators page


Phantomwise:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Adding sort options for narrators in the "Narrators" page:
- sort by first name
- sort by last name
- sort by number of books
- and maybe allow to reverse the order?

### Why would this be helpful?

In the "Narrators" page, there are two columns: "Name" and "Books", but there are no options to change the default sorting order, which is by narrator first name and that isn't the most helpful.
- Sorting by last name would be helpful to find a specific narrator.
- Sorting by number of books would be very convenient to show one's favorite narrators, that have the most books, first.
- I'm not sure how much useful it would be to be able to reverse the sort order though.

### Future Implementation (Screenshot)

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/f344751e-09b4-4418-86ff-91f51939e685" />

Gext and arrow in white for the current sorting order, and in grey for the orders that are not currently used. That's not easy to distinguish at first glance though.

Alternatively, maybe there's no need for an option to reverse the order, in which case an arrow only under the current sorting column would make it easier to tell which order is currently used?

### Audiobookshelf Server Version

v2.28.0

### Current Implementation (Screenshot)

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/674476f3-b65c-4f22-8e0b-b4d0f9e5839e" />

Narrators are sorted by first name, and I need to scroll a while to find my favorite narrators since they're not on the first page.

PR #4828, opened 2025-11-13
#4584 sort options for narrators

S-Reinhard:
## Brief summary

Allow users to sort narrators by name and book count.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/4584

## In-Depth Description

**How the narrator sorting works:**

Users can sort narrators by clicking on the corresponding table header cells.  
When hovering over these cells, the cursor changes to a pointer. Because there are no lines separating the tabel cells, the text in the corresponding cell is underlined.  
Clicking the same header again toggles the sorting direction (ascending/descending).  
An arrow icon visually indicates the current sorting direction and only appears in the active column to show which column is being sorted.

To ensure the visual indicators and sorting logic remain in sync, both are driven by shared state properties like the currently selected attribute and sort order.

Those parameters are used as query params when fetching the narrators.
On the server side the query params from the request are replaced with known strings to ensure working default and fallback sorting options, before sorting the list of narrators using naturalSort().

**Why this solution is effective:**

It follows the design principle of *external consistency*—the behavior and appearance mimic commonly used sortable tables, making the feature intuitive and familiar for users.

Server side sorting will work in the future even when paging is added.

**Areas for Improvement:**

The current implementation uses a Unicode arrow character returned from a function. This wasn't my first choice; I would have preferred using an icon to support `aria-labels`. However, I was unable to insert a standalone icon without also introducing a button. If icon-based implementation is preferred, I would appreciate guidance on the icon library and documentation used in this project, along with relevant code examples.

Fetching a new list of narrators every time the order changes is a bit slow and creates unnecessary traffic, this could be improved by caching.

**Who is affected:**

While the change is small, it affects many users because it adds function to the web client. 

## How Have You Tested This?

I ran the project inside a dev container.
Then I added sample audiobooks as test data and navigated to the narrator page in a web browser. I verified the behavior when hovering over, clicking, and repeatedly interacting with the header cells of sortable columns.

Additionally, I confirmed that the edit and delete buttons next to each narrator still function as expected.

## Screenshots

Before:
image
After:
image

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "audiobookshelf-client",
-  "version": "2.32.1",
+  "version": "2.33.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "audiobookshelf-client",
-      "version": "2.32.1",
+      "version": "2.33.0",
       "license": "ISC",
       "dependencies": {
         "@nuxtjs/axios": "^5.13.6",
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "audiobookshelf-client",
-  "version": "2.32.1",
+  "version": "2.33.0",
   "buildNumber": 1,
   "description": "Self-hosted audiobook and podcast client",
   "main": "index.js",
--- a/client/pages/config/sessions.vue
+++ b/client/pages/config/sessions.vue
@@ -66,7 +66,11 @@
                 <p class="text-xs">{{ getPlayMethodName(session.playMethod) }}</p>
               </td>
               <td class="hidden sm:table-cell max-w-32 min-w-32">
-                <p class="text-xs truncate" v-html="getDeviceInfoString(session.deviceInfo)" />
+                <p class="text-xs truncate">
+                  <template v-for="(line, index) in getDeviceInfoLines(session.deviceInfo)">
+                    <br v-if="index > 0" :key="'br-' + index" />{{ line }}
+                  </template>
+                </p>
               </td>
               <td class="text-center w-24 min-w-24 sm:w-32 sm:min-w-32">
                 <p class="text-xs font-mono">{{ $elapsedPrettyLocalized(session.timeListening) }}</p>
@@ -130,7 +134,11 @@
               <p class="text-xs">{{ getPlayMethodName(session.playMethod) }}</p>
             </td>
             <td class="hidden sm:table-cell max-w-32 min-w-32">
-              <p class="text-xs truncate" v-html="getDeviceInfoString(session.deviceInfo)" />
+              <p class="text-xs truncate">
+                <template v-for="(line, index) in getDeviceInfoLines(session.deviceInfo)">
+                  <br v-if="index > 0" :key="'br-' + index" />{{ line }}
+                </template>
+              </p>
             </td>
             <td class="text-center">
               <p class="text-xs font-mono">{{ $elapsedPretty(session.timeListening) }}</p>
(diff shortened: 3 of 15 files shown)


Vito0912:
Disclaimer: I am just a contributor

Imho, this should be handled on the server side and not on the client side (even though there is no pagination).

Also, just a note as there has been an increasing number of AI-generated PRs. I don’t think yours is one, but when the description is written by AI like here, it might delay even checking it (please note that PRs currently have a longer review time than usual anyway though).

S-Reinhard:
@Vito0912 Why should this be done on the server side? In my opinion, this would generate unnecessary traffic, as sorting does not require information from the backend, nor does it impose a high load on the client. Have I overlooked something?

Vito0912:
What I mean is, this just works because there is no pagination, which imho is bad. And this would not work if there was pagination
So if in the future pagination would be added that would break. If it's done server side (like most of the others) it would work

S-Reinhard:
@Vito0912 it's been a while but I had time to implement server side sort but that is really slow compared to client side.

Vito0912:
Server-side sorting only makes sense when used with pagination to **improve** speed. But I do not see how it would be slower than sorting on the client, except for a small delay from the database each time.

Do not get me wrong, but this is also the least efficient way to implement server-side sorting. It needs to be done inside the SQL query. In its current state, it gives no benefit at all, that is true. [It was the same before - but it might be related to SQLite limits, because I remember there were problems with natural sorting or something similar.]

> [!NOTE]
> Anyway, I am just a contributor, but I am also quite sure that this will not be merged. The frontend is currently being rewritten, which means that any change to the Vue frontend (/client folder) will be fully removed/obsolete. I do not know why this is not shared/communicated clearly or at all to be said (only a few new PRs mention it), while around 100 PRs are still open, even though at least half of them cannot be merged and will not be merged because of the UI rewrite.
> I know the team works hard, but this is sadly the worst communication I have seen for anyone who wants to contribute

From one of the newer PRs:
> Just an FYI, but web client front-end changes are not actively being merged or reviewed right now because we are in the middle of migrating away from Vue to a new framework. You're welcome to leave this open in the meantime (either as inspiration or if someone else wants to use it in the meantime).

S-Reinhard:
@Vito0912  I am aware that an SQL query would have been faster, but the way that the response data is aquired would have required major changes compared to what this feature does. And the sort function I used in the end was already implemented before, I only had to make it possible to change the sorting direction and attribute. The sort function comes from a library optimized for speed and so I tested if it would be fast enough. And the sort itself was fast enough but the communication between frontend and backend was to slow, but to be honest my laptop is really struggeling to run this dev container so it might not be a problem for other people. I also tried caching, but that caused a bug, when loading from cache after changing sort directions, the narrator entries were in editing mode and their names were not visible. So I didn't commit the caching.

However thank you for telling me that frontend changes won't be merged. I wish that stood in the "How to Contribute" part of the readme,.
Do you know another small Issue, that I could do instead?

Vito0912:
Yeah. I by no means want to say it's bad. Just wanted to explain that that implementation has no benefit over client side. (Personally, but that's personal xD, I think Server-Side is always a bit better)

But anyway to your actual question: 
I can search a bit and check some. There are some issues I noted down, but stopped big parts of ABS contributions, so still have a few simpler on hand.
If it's just about having your name in that repo you can also translate a string. It will be added as a commit if GitHub is connected with Weblate

S-Reinhard:
@Vito0912 I want to contribute something real. Could you show me some of the issues you noted down. Most things I found were client fixes.

Vito0912:
I looked up the list and sadly every one would need some change at the frontend :( or is somewhat complex

PR #4857, opened 2025-11-25
Feat/metadata id matching

Cutch:
## Brief summary
Add an api to sync metadata files. And add the abs id to the metadata.json to support when inode detection fails.

## Which issue is fixed?

NA

## In-depth Description
This PR does two things, first it adds a tool api the syncs all metadata files when toggling "Store metadata with item info" metadata files are not updated until a change is made. This has led to adding and remove tags on all files to update all library items. The api POST /tools/batch/update-metadata-files can be used to scan and update all items in the library.

The second change is adding the abs item id to the metadata.json. inode are not always supported on different systems/copying changing files makes a new inode. Storing the abs id in the metadata helps limit issues when files are moved and change.

I have noticed with windows inodes do not always work. Renames are detected in Watcher, but Scanner treats items like a new item. Ideally Watcher should pass this info directly into Scanner instead of relying on inodes. However, even that may not work on all systems, the metadata.json id should allow a much wider range of compatibility.
Protection when copying is added to prevent a copy from changing the original library item.
If an item cannot be detected or a copy is detected the new item will still get a new library entry/id.

## How have you tested this?

Locally tested for the metadata changes, copying, moving, deleting, creating. Api tested via postman.


server/controllers/ToolsController.js
```diff
@@ -150,6 +150,29 @@ class ToolsController {
...
+   */
+  async updateAllItemMetadata(req, res) {
+    if (!req.user.isAdminOrUp) {
+      Logger.warn(`Non-admin user "${req.user.username}" other than admin attempted to batch scan library items`)
+      return res.sendStatus(403)
+    }
+
+    const libraryItems = await Database.libraryItemModel.findAll()
+    for (const libraryItem of libraryItems) {
+      await libraryItem.saveMetadataFile()
```
Vito0912, line 170: Note: I am just a contributer 

But if we do that, we probably should also add the covers here and rename the endpoint. Metadata at least can be forced. For covers there currently is no "easy" way.
But that's just my opinion 

Ik it's not directly related, but such endpoint only would make sense if it also forces the covers (if the setting is enabled), because the creation of metadata already is a process that can be done very easily (although not very beautiful)

server/models/LibraryItem.js
```diff
@@ -574,6 +574,7 @@ class LibraryItem extends Model {
     let jsonObject = {}
     if (this.mediaType === 'book') {
       jsonObject = {
+        absId: this.id,
```
Vito0912: It's an ABS file, wouldn't id make more sense?

Vito0912 (commented):
Awesome idea imho

server/models/LibraryItem.js
Cutch: You're right, originally i had thought this may have been merged elsewhere, and it was to prevent conflict. But it looks metadata is always its own object, i'll make that update

server/controllers/ToolsController.js
Cutch: I agree that it's useful however i think having a separate endpoint would be better. Since there is two toggles, this would allow for the same flexibility, in case there is a user that only switches one or the other

Vito0912: But it would be only generated in the new folder when the switch is generated anyways. So it does not change anything and in opposite could potentially fix issues when users deleted the cover folder as there is no way to regenerate it - when they have it saved with the items in the past. But againy just my opinion as a contributor and not maintainer.
Two endpoints for the same "action" seems a bit useless (especially since metadata can be regenerated already, covers not)

Cutch: The goal of this feature is that the API to update the metadata could be used when the toggle is triggered. Ideally we could actually start the task or call the api when the toggle is triggered for the first time. 
This works because the metadata is duplicated in the database and can be written out to the json file.

From looking into it, cover images/local images are not stored elsewhere, only the paths are stored in the database so if an image is lost it cannot be recreated. When triggering the toggle ideally a job should be run that moves the images from the internal directory to the book directory for vice-versa.

I think for your use case, where the image is deleted, there is nothing that can recover the file. The best you could do is re-scan the file which should recognize the missing coverPath and attempt to re-extract it.

Vito0912: > The goal of this feature is that the API to update the metadata could be used when the toggle is triggered.

But I mean it's the same case for the cover?

The cover is not stored somewhere else, yes. But it can be embedded in the file or also be a file carry to the file, that was not generated by ABS. Also that is just one use case. As for your metadata it is exactly the same thing that would move it to the books. There is no way to do it otherwise.
But that's my last comment on this . I also understand if you are another opinion. This endpoint just is very useless imho

## Edit

I think I now understand what you mean by "updating", because you want it to populate your new id? If so that should not need an API call but happen automatically after the migration. But I did not look into the code any further, so this might be the case already. But this would make sense what you mean by your comment


Yeah, I think that is what you mean
> is triggered for the first time

which would make it even more useless imho. I thought you at least wanted to introduce this endpoint to be able to move the metadata.json files between the metadata folder and side carry 

But again, last comment on this from me as I am just a contributer. Maybe advplyr thinks it's a proper way

Cutch: Just a note yes i agree this should have a migration, i'll have to look into if there is a process for it.
Yes it could be used for migration purposes for this specific commit.
However, the goal here was to create an api that could be called when "Store metadata with item" is toggled.
I have not hooked it up to the toggle as it can take some time to run, so instead its a tool that can be used instead of changing all of the tags and setting them back in order to cause a metadata save.

Vinnecent:
This would be a major game changer for me.
I use a vps that has a different nextcloud server mounted via rclone (the reason for this is that it is extremely cheap).
The disadvantage is that I have to scan the library manually if I add new audiobooks and this takes around 3 hours (I have around 1800 audiobooks).

PR #4870, opened 2025-11-27
Fix region parameter extraction in findChapters

Vito0912:
## Brief summary
Forgotten migration

## Which issue is fixed?

Discord

## In-depth Description

Forgotten migration

## How have you tested this?

Not tested. But should work. Please test xD


advplyr:
Thanks!

PR #4873, opened 2025-11-29
Fix Upload Lookup

Vito0912:
## Brief summary

When uploading a book through the interface and pressing the "Automatically fetch..." button, it fails because a recent refactor in the finder logic now requires a libraryId to be provided. Since any item not yet uploaded has no id, this caused the error.  
This PR fixes the issue and makes the button work again.  
It also disables the heuristic for scoring and sorting since it is not available anyway if there is no id available. Currently, the API returns a low confidence score when not providing an id, which is incorrect.

## Which issue is fixed?

DC

## In-depth Description

See above

## How have you tested this?

Uploaded and matched book


advplyr:
Thanks!

PR #4881, opened 2025-12-01
Wrap ensureDir in try-catch blocks in manager classes

mikiher:
## Brief summary

This wraps `ensureDir` calls in try catch blocks on some of the manager classes.
It's done to try to debug issues experienced by some users reported in [this abs-windows bug](https://github.com/mikiher/audiobookshelf-windows/issues/56) 

## Which issue is fixed?

Currently no issue. If I find the root cause the abs-windows bug, I will likely open a corresponding server issue.

## In-depth Description

Since the current failure unfortunately does not contain a stack trace and is not caught by any try-catch block, I put a try-catch block around suspected `ensureDir` calls and log an error before re-throwing.

## How have you tested this?

I just brought up a server and made sure everythiing still works. 
This will not change any existing functionality - it just logs errors in the location they happened so they can be identified.

### DIFF
--- a/server/managers/BackupManager.js
+++ b/server/managers/BackupManager.js
@@ -48,13 +48,6 @@ class BackupManager {
   }
 
   async init() {
-    // Validate backupPath before attempting to create it
-    if (!this.backupPath || typeof this.backupPath !== 'string' || this.backupPath.length < 2) {
-      Logger.error(`[BackupManager] Invalid backup path configured: "${this.backupPath}". Falling back to default.`)
-      // Reset to default backup path
-      global.ServerSettings.backupPath = Path.join(global.MetadataPath, 'backups')
-    }
-
     try {
       const backupsDirExists = await fs.pathExists(this.backupPath)
       if (!backupsDirExists) {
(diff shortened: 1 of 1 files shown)


server/managers/BackupManager.js
```diff
@@ -48,9 +48,34 @@ class BackupManager {
...
+        Logger.info(`[BackupManager] Attempting to use default backup path: "${defaultBackupPath}"`)
+        global.ServerSettings.backupPath = defaultBackupPath
+        try {
+          await fs.ensureDir(defaultBackupPath)
+        } catch (fallbackError) {
+          Logger.error(`[BackupManager] Failed to create default backup directory: ${fallbackError.message}`)
+          throw new Error(`[BackupManager] Failed to create default backup directory at "${defaultBackupPath}"`, { cause: fallbackError })
+        }
+      } else {
+        throw new Error(`[BackupManager] Failed to ensure backup directory at "${this.backupPath}"`, { cause: error })
```
advplyr: If the backupPath is getting changed we need to save the server settings. Otherwise if the server settings get updated this change will be overwritten. https://github.com/advplyr/audiobookshelf/blob/329e3c71795398d93a5e12fe7045f7ca4d0c098e/server/Database.js#L435

mikiher: OK, I removed the backup fallback logic and just wrapped with a try-catch block like the other places.

advplyr:
Thanks!

Issue #3680, opened 2024-12-05
[Bug]: Error on getting Podcast Feed when trying to import "Lage der Nation"


maulenski:
### What happened?

Want to import the following podcast: https://podcasts.apple.com/de/podcast/lage-der-nation-der-politik-podcast-aus-berlin/id1092957894
Was found through internal search with "Lage der Nation"
Got following error: "Fehler beim Abrufen des Podcast Feeds" / "Error on getting podcast feed"

### What did you expect to happen?

Podcast is added to available podcasts

### Steps to reproduce the issue

1. go to podcasts
2. add
3. search for "Lage der Nation"
4. click on found Podcast to add to library

### Audiobookshelf version

v2.17.3

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Additional Notes

also tried different links/feeds from Spotify or directly from the Podcasts Website.

maulenski:
yes it seems to be the same/similar issue. My Container is running in TrueNAS Scale. Adding the mentioned environmental variable does not work (net.ipv6.conf.all.disable_ipv6=1). Am i doing something wrong?

maulenski:
i also tried now inserting this line into sysctl configuration in Truenas Scale UI, not working

maulenski:
`2024-12-05 14:12:56.750155+00:00[2024-12-05 15:12:56.747] ERROR: [podcastUtils] getPodcastFeed Error AxiosError: getaddrinfo EAI_AGAIN feeds.lagedernation.org
2024-12-05 14:12:56.750291+00:00at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26) {
2024-12-05 14:12:56.750336+00:00hostname: 'feeds.lagedernation.org',
2024-12-05 14:12:56.750476+00:00syscall: 'getaddrinfo',
2024-12-05 14:12:56.750520+00:00code: 'EAI_AGAIN',
2024-12-05 14:12:56.750560+00:00errno: -3001,
2024-12-05 14:12:56.750600+00:00config: {
2024-12-05 14:12:56.750636+00:00transitional: {
2024-12-05 14:12:56.750675+00:00silentJSONParsing: true,
2024-12-05 14:12:56.750789+00:00forcedJSONParsing: true,
2024-12-05 14:12:56.750880+00:00clarifyTimeoutError: false
[...]

Issue #3579, opened 2024-11-02
[Bug]: Audiobookshelf is not able to subscribe to German Podacast "Lage der nation" 


femanuel:
### What happened?

Audiobookshelf in Podcast mode is not able to subscribe to German Podacast "Lage der Nation" (https://feeds.lagedernation.org/feeds/ldn-mp3.xml), although the podcast can be found via the built-in search function. Several months ago this was still working.

### What did you expect to happen?

ABS should be able to subscribe to the podact, but it issues a red error message "Fehler beim Abrufen des Podcast" (German)

### Steps to reproduce the issue

1.  Try to subscribe to podcast "Lage der Nation" which can be found via the search function or use the URL 'https://feeds.lagedernation.org/feeds/ldn-mp3.xml' to do this manually.

### Audiobookshelf version

2.16.2 and earlier

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Logs

```shell
2024-11-02 12:59:55.205

DEBUG

[iTunes] Searching for podcast with term "lage der nation"

2024-11-02 12:59:55.761

DEBUG

[iTunes] Podcast search for "lage der nation" returned 1 results

2024-11-02 12:59:57.770

DEBUG

[podcastUtils] getPodcastFeed for "https://feeds.lagedernation.org/feeds/ldn-mp3.xml"

2024-11-02 13:00:05.375

ERROR

[...]

Issue #3346, opened 2024-08-30
[Bug]: [podcastUtils] getPodcastFeed Error Error: getaddrinfo EAI_AGAIN feeds.lagedernation.org


ego1105:
### What happened?

The podcast "Lage der Nation" stopped working in abs. I deleted it and tried to re-add it without success. Either by searching for "lage der nation" or by adding "https://feeds.lagedernation.org/feeds/ldn-mp3.xml". 

The logs say:
ERROR [podcastUtils] getPodcastFeed Error Error: getaddrinfo EAI_AGAIN feeds.lagedernation.org 

### What did you expect to happen?

This podcast can be added and episodes are downloaded.

### Steps to reproduce the issue

1. Try to add "lage der nation" podcast
2. "Failed to get podcast feed" message appears
3. log messageg appears.

### Audiobookshelf version

2.12.3

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
{"timestamp":"2024-08-30 09:09:35.347","source":"podcastUtils.js:229","message":"[podcastUtils] getPodcastFeed for \"https://feeds.lagedernation.org/feeds/ldn-mp3.xml\"","levelName":"DEBUG","level":1}

{"timestamp":"2024-08-30 09:09:40.352","source":"podcastUtils.js:269","message":"[podcastUtils] getPodcastFeed Error Error: getaddrinfo EAI_AGAIN feeds.lagedernation.org","levelName":"ERROR","level":4}
```

### Additional Notes

Delete and re-add of podcast.
docker compose down and up.

advplyr:
I tested that podcast and had no issue adding it and downloading episode. So far the `EAI_AGAIN` issue has been brought up many times and every time it is an issue with the users network.
We can re-open this is if it is shown to be a bug with Abs. Users in the Discord may be able help troubleshoot a network issue.

BorisWilhelms:
I have the same error on the same feed. DNS seems to work. I am running ABS in a docker container. I am able to curl the feed inside the container, but ABS seems not be able to download the feed.

BorisWilhelms:
I did some more testing. I am using the offical docker container running on Raspberry PI. I can resolve the hostname on the PI directly and also inside the container. A curl inside the container also works.

[...]

PR #4885, opened 2025-12-02
Implement experimental DNS pre-resolution, enabling Podcast downloads even with partial DNS resolution failures

PassionateBytes:
## Brief summary

This PR introduces an optional configuration parameter `EXP_DNS_RESOLUTION`. When enabled, the server attaches a custom `lookup` function to the http/https agents used for outgoing requests (axios). The lookup tries the system resolver (`getaddrinfo`) first, and only when that fails with `EAI_AGAIN` falls back to explicit `A`/`AAAA` queries via Node's `dns.resolve4`/`dns.resolve6` (c-ares), which tolerate one address family failing while the other resolves.

This enables successful web requests in situations that would otherwise fail due to partial DNS resolution failures, i.e. when a DNS query resolves an `A` record of a hostname while rejecting requests for `AAAA` records (`SERVFAIL` or `REFUSED`), or vice versa. This failure mode is specific to the Alpine (musl) container environment; see root cause below.

## Root cause

Node's `dns.lookup` (used by axios for name resolution) calls libc `getaddrinfo`. For an unspecified address family, musl sends the `A` and `AAAA` queries in parallel and fails the **entire** lookup with `EAI_AGAIN` if **either** response carries rcode `SERVFAIL`, even when the other family returned valid addresses. This is deliberate musl behavior and still present in current musl master, in `name_from_dns` ([`src/network/lookup_name.c`](https://git.musl-libc.org/cgit/musl/tree/src/network/lookup_name.c)):

```c
for (i=0; i<nq; i++) {
    if (alens[i] < 4 || (abuf[i][3] & 15) == 2) return EAI_AGAIN;
```

glibc returns the successfully resolved family instead, which is why this only reproduces in the Alpine-based Docker image and not when running the server on glibc systems. Upgrading Alpine/musl does not fix it.

Reproduction with a stub DNS server that answers `A` normally and `SERVFAIL` for `AAAA`:

| Environment | Call | Result |
|---|---|---|
| `node:20-alpine` (musl) | `dns.lookup` | `EAI_AGAIN`, request fails |
| `node:20-slim` (glibc) | `dns.lookup` | returns the A record |
| `node:20-alpine` | `dns.resolve4` (c-ares) | returns the A record |

## Implementation notes

- `server/utils/resilientDns.js` exports `resilientLookup` (drop-in replacement for `dns.lookup`, usable as the `lookup` option of `http.Agent`/`https.Agent`) and `getAgentsForUrl` (builds the agent pair for a request, composing the SSRF request filter with the custom lookup via `ssrf-req-filter`'s exported `requestFilterHandler`).
- With `EXP_DNS_RESOLUTION=1`, `Server.js` sets `axios.defaults.httpAgent`/`httpsAgent` with the custom lookup; the SSRF-filtered call sites (`podcastUtils.getPodcastFeed`, `fileUtils.downloadFile`) get it through `getAgentsForUrl`.
- Because request URLs are never rewritten, TLS/SNI and certificate validation follow the normal path, and axios' native redirect handling (`follow-redirects`) stays active. Redirect hops resolve through the same agents, so no manual redirect or loop handling is needed.
- `getaddrinfo` remains the primary resolver, so `/etc/hosts`, nsswitch, and container DNS behave exactly as before; the fallback only engages on `EAI_AGAIN`. With healthy DNS the flag is a no-op.
- `PREFER_IPV6=1` switches the fallback order to `AAAA` before `A`.
- Unit tests in `test/server/utils/resilientDns.test.js`; verified end-to-end in the Alpine image against a stub DNS server (default lookup fails with `EAI_AGAIN`, custom lookup succeeds, SSRF filter still blocks private-range targets with the custom lookup active).

## Which issue is fixed?

- https://github.com/advplyr/audiobookshelf/issues/4887
- https://github.com/advplyr/audiobookshelf/issues/4011
- https://github.com/advplyr/audiobookshelf/issues/3876
- https://github.com/advplyr/audiobookshelf/issues/3680
- https://github.com/advplyr/audiobookshelf/issues/3579
- https://github.com/advplyr/audiobookshelf/issues/3346

Additionally, a discussion of the issue server side can be found here:
- https://talk.lagedernation.org/t/lage-cdn-dns-issues/27017/18?u=passionatebytes

## Example of the failure mode

The German news podcast 'Lage der Nation' hosted their feeds on a DNS setup that consistently answered `SERVFAIL` for `AAAA` queries while resolving `A` records fine (since fixed on their side, but the same misconfiguration exists elsewhere):

```bash
$ nslookup feeds.lagedernation.org
Server:		127.0.0.53
Address:	127.0.0.53#53

Non-authoritative answer:
Name:	feeds.lagedernation.org
Address: 139.162.176.53
;; Got SERVFAIL reply from 127.0.0.53
** server can't find feeds.lagedernation.org: SERVFAIL
```

In the Alpine container this made every request to that host fail with:

```
ERROR: [podcastUtils] getPodcastFeed Error AxiosError: getaddrinfo EAI_AGAIN feeds.lagedernation.org
```

For those wanting to try an up-to-date build including this patch: https://hub.docker.com/r/passionatebytessolutions/audiobookshelf

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -2,13 +2,12 @@ const Path = require('path')
 const Sequelize = require('sequelize')
 const express = require('express')
 const http = require('http')
+const https = require('https')
 const util = require('util')
 const fs = require('./libs/fsExtra')
 const fileUpload = require('./libs/expressFileupload')
 const cookieParser = require('cookie-parser')
 const axios = require('axios')
-const { AxiosError } = require('axios')
-const dns = require('dns').promises
 
 const { version } = require('../package.json')
 
@@ -17,6 +16,7 @@ const is = require('./libs/requestIp/isJs')
 const fileUtils = require('./utils/fileUtils')
 const { toNumber } = require('./utils/index')
 const { getRequestOrigin } = require('./utils/requestUtils')
+const { resilientLookup } = require('./utils/resilientDns')
 const Logger = require('./Logger')
 
 const Auth = require('./Auth')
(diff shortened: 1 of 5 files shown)


Vito0912 (approved):

server/Server.js
```diff
@@ -86,6 +87,66 @@ class Server {
...
+        async (error) => {
+          if (error.response && [301, 302, 303, 307, 308].includes(error.response.status) && error.response.headers.location) {
+            const redirectUrl = error.response.headers.location
+            Logger.debug(`[Server] Following ${error.response.status} redirect to ${redirectUrl}`)
+            return axios({
+              ...error.config,
+              url: redirectUrl
+            })
+          }
+          return Promise.reject(error)
```
Vito0912: This looks a bit scary as this would create an infinite redirect loop, or am I missing something.
Didn't meant to approve btw. I am read-only anyway

PassionateBytes: Good catch, and fair point! I'll implement a counter so we won't recurse infinitely.

PassionateBytes: I have made another commit which addresses this. This keeps track of the number of redirects, as well as the visited URLs. This allows us to stop at the `maxRedirects` count, while catching circular redirects early.

PassionateBytes:
For those wanting to try an up-to-date build of ABS, including this patch - I have published a build here and aim to keep publishing updated versions for every release until this PR is merged: https://hub.docker.com/r/passionatebytessolutions/audiobookshelf

PassionateBytes:
@advplyr would love to get some feedback on this one. Theres a small community of people waiting for this fix to get into a release.
Please let me know your thoughts on this or any blockers

binyaminyblatt:
Would it be possible to add an additional config that would allow for DNS over https and I could provide the server I want to use for it

PR #4740, opened 2025-10-10
fix #1617 (useradd: user 'audiobookshelf' already exists)

Yetangitu:
This change fixes the problem of failing upgrades on dpkg-based systems (see #1617) by reworking the check for whether the `audiobookshelf` user/group already exists.

PR #4892, opened 2025-12-03
Update Debian preinst for [Bug]: Debian ugprade fails with exit 9 #4888

### LINKED ISSUE #4888: [Bug]: Debian ugprade fails with exit 9
### What happened?

got an error while upgrading debian package in debian trixie amd64:
> Vorbereitung zum Entpacken von .../03-audiobookshelf_2.31.0_amd64.deb ...
> Removed '/etc/systemd/system/multi-user.target.wants/audiobookshelf.service'.
> Creating system user: audiobookshelf in audiobookshelf with audiobookshelf user-daemon and shell /bin/false
> useradd: Benutzer »audiobookshelf« existiert bereits
> dpkg: Fehler beim Bearbeiten des Archivs /tmp/apt-dpkg-install-36NNH5/03-audiobookshelf_2.31.0_amd64.deb (--> unpack):
>  »neues pre-installation-Skript des Paketes audiobookshelf«-Unterprozess gab den Fehlerwert 9 zurück
> Created symlink '/etc/systemd/system/multi-user.target.wants/audiobookshelf.service' → '/usr/lib/systemd/system/audiobookshelf.service'.

> root@fileserver:/var/lib/dpkg/info# ./audiobookshelf.preinst 
> Creating system user: audiobookshelf in audiobookshelf with audiobookshelf user-daemon and shell /bin/false
> useradd: Benutzer »audiobookshelf« existiert bereits
> root@fileserver:/var/lib/dpkg/info# echo $?
> 9 

problem is pipefail in combination with grep -q.
plz replace the line
> if ! getent passwd | grep -q "^$user:"; then
with
> if ! (getent passwd ; true ) | grep -q "^$user:"; then
to get rid of the error.

pls patch the preinstall script

1. apt-get update
2. apt-get upgrade

### Audiobookshelf version

2.31.0

Debian/PPA

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

Galen007:
## Brief summary

grep -q stops the getent execution because it finds the searched expression, this causes an non zero exitcode which stops the script because of the "set -o pipefail". This is only if the user already exists.

## Which issue is fixed?
Fixes #4888

## In-depth Description
add a true to the getent, makes sure the if condition above uses the exitcode of the grep instead of the getent.

## How have you tested this?
manually patched current package and installation worked fine.

### DIFF
--- a/build/debian/DEBIAN/preinst
+++ b/build/debian/DEBIAN/preinst
@@ -22,7 +22,7 @@ add_user() {
   declare -r descr="${4:-No description}"
   declare -r shell="${5:-/bin/false}"
 
-  if ! getent passwd | grep -q "^$user:"; then
+  if ! (getent passwd ; true) | grep -q "^$user:"; then
     echo "Creating system user: $user in $group with $descr and shell $shell"
     useradd $uid_flags --gid $group --no-create-home --system --shell $shell -c "$descr" $user
   fi
@@ -39,7 +39,7 @@ add_group() {
     declare -r gid_flags="--gid $gid"
   fi
 
-  if ! getent group | grep -q "^$group:" ; then
+  if ! (getent group; true) | grep -q "^$group:" ; then
     echo "Creating system group: $group"
     groupadd $gid_flags --system $group
   fi


Vito0912:
Probably duplicate of https://github.com/advplyr/audiobookshelf/pull/4740

PR #4906, opened 2025-12-12
When adding items to playlist, sort the playlist sections alphabetically.

sir-wilhelm:
## Brief summary

When adding items to playlist I noticed they are not sorted alphabetically like they are on the playlist page. I have one named `_Up Next`, named that way so it always shows up on top, but when adding items to playlist it's not sorted alphabetically.

This PR sorts the playlist.

## Which issue is fixed?

None

## In-depth Description

I just modified `sortedPlaylists()` on the playlist modal to ensure playlist that are selected (`isItemsIncluded`) come before those that are not, and if the items are in the same group to sort by name.
This makes each section sorted alphabetically (playlist the item is added to and the rest of the playlist)

## How have you tested this?

I ended up using github's codespaces, uploaded some small .mp3s and used them as "audiobooks" to test the playlist modal.

## Screenshots
Current:
image
After changes:
image

### DIFF
--- a/client/components/modals/playlists/AddCreateModal.vue
+++ b/client/components/modals/playlists/AddCreateModal.vue
@@ -97,7 +97,10 @@ export default {
             ...playlist
           }
         })
-        .sort((a, b) => (a.isItemIncluded ? -1 : 1))
+        .sort((a, b) => {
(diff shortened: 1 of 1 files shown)


advplyr:
Makes sense, thanks!

Issue #4765, opened 2025-10-22
[Enhancement]: UX: Add library item to top as well as bottom of both Public Collections and Private Playlist


ZLoth:
### Type of Enhancement

None

### Describe the Feature/Enhancement

In the dialogue boxes for adding to Collections and Playlists, there should be a button to add a library item to the top or bottom of a list. 

### Why would this be helpful?

This is some usability enhancements as part of previously submitted requests for collections and playlists:

* https://github.com/advplyr/audiobookshelf/issues/3364
* https://github.com/advplyr/audiobookshelf/issues/3357

The current functionality of ABS is to add new items to the bottom of a collection or playlist. In my use case, I maintain a private playlist of titles that I want to listen to. If I have a new library item and I want to prioritize it for upcoming listening, then I have to go into the actual playlist and drag it upward. This gets fustrating if one has a long playlist like this:

<img width="1312" height="1054" alt="Image" src="https://github.com/user-attachments/assets/d6128d35-edef-4f6e-8151-24cc7d83a432" />

Imagine dragging a new titlke upward of 244 steps. 

### Future Implementation (Screenshot)

Please note that the ideal symbol would be an up or down arrow pointing to a straight line, I didn't see such symbols on the character picker, so I tried some substitutes.

<img width="502" height="179" alt="Image" src="https://github.com/user-attachments/assets/876de508-b2e3-4cc9-9e96-04a6e7f59d21" />
<img width="501" height="206" alt="Image" src="https://github.com/user-attachments/assets/56719b55-2af6-4f25-8a4a-4291d49dbee1" />

### Audiobookshelf Server Version

2.30

### Current Implementation (Screenshot)

The current images show the plus sign, and add to the bottom of the list.

<img width="502" height="179" alt="Image" src="https://github.com/user-attachments/assets/939b637c-c704-4a8c-bacd-bd60b5e06640" />
<img width="501" height="206" alt="Image" src="https://github.com/user-attachments/assets/78fe314b-c699-4dc4-af16-1f7bc57ff79b" />

PR #4907, opened 2025-12-12
Added buttons to add items to the top or bottom of a playlist.

sir-wilhelm:
## Brief summary

Updated the add to Playlist modal to support adding items to the top or bottom of playlist.

## Which issue is fixed?

Fixes part of #4765. I only addressed playlist, since I don't use Collections.
If you want the same treatment for Collections let me know.

## In-depth Description

I don't know if this is the best solution, but I thought it makes more sense to update the [batch add items to a playlist](https://api.audiobookshelf.org/#batch-add-items-to-a-playlist) endpoint to check for an optional `addToTop` property and then adjust where the item is added. That seemed better/easier than adding it at the bottom, then making more request to move it after the fact.

If you have a preferred implementation I'd be happy to update the PR after feedback. 

## How have you tested this?

I used github's codespaces via VSCode to make and test the server/UI changes.

## Screenshots
image
image

### DIFF
--- a/client/components/modals/playlists/AddCreateModal.vue
+++ b/client/components/modals/playlists/AddCreateModal.vue
@@ -15,7 +15,7 @@
         <div class="w-full overflow-y-auto overflow-x-hidden max-h-96">
           <transition-group name="list-complete" tag="div">
             <template v-for="playlist in sortedPlaylists">
-              <modals-playlists-user-playlist-item :key="playlist.id" :playlist="playlist" class="list-complete-item" @add="addToPlaylist" @remove="removeFromPlaylist" @close="show = false" />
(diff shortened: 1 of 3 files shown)


nichwall (approved):
This is working for me.

I ran into a bug during initial testing where I added to a playlist and the interface did not update (in a playlist but could not add or remove it until I closed and opened the modal again). However, I was unable to recreate this after multiple additional tests and restarting the server a few times.

sir-wilhelm:
I checked this branch out again to test. I build/ran it with `cd client; npm run generate; cd ..; npm run dev` and was unable to reproduce the bug you saw.

nichwall:
Thanks for checking. I could not reproduce it again, so it may have been something in my browser that got cached from me switching between a bunch of branches for reviewing.

sir-wilhelm:
I'll re-open if this is still an issue/desired feature on the new client (once it's out).

PR #4916, opened 2025-12-15
Add HOST environment variable to Dockerfile

Minituff:
## Brief summary

This PR fixes the `HOST` being `undefined` on Docker

## Which issue is fixed?

This fixes the container's web UI not being accessible 

## How have you tested this?

The workaround it is set `HOST=0.0.0.0` as a Docker environment variable.

## Screenshots

This is the default output in the console (without the proposed changed)
Screenshot 2025-12-15 at 10-18-41 Portainer local

This is the console output with the proposed changed

Screenshot 2025-12-15 at 10-20-38 Portainer local

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,7 @@ COPY --from=build-server ${NUSQLITE3_PATH} ${NUSQLITE3_PATH}
 EXPOSE 80
 
 ENV PORT=80
+ENV HOST=0.0.0.0
 ENV NODE_ENV=production
 ENV CONFIG_PATH="/config"
 ENV METADATA_PATH="/metadata"


advplyr:
Wouldn't this break users using ipv6? I think undefined is the preferred default in this case

Minituff:
Probably yea, but I would guess there are more ipv4 users than ipv6. 

Right now out of the box the container doesn't work unless I add that environment variable.

We could always add the equivalent ipv6 value in the docs for those users. 

Ideally, we can infer whether or not they use IPv4 or IPv6 and set it appropriately.

advplyr:
What do you mean it doesn't work? When passing in undefined to express.listen it is already checking whether ipv6 or ipv4 and setting appropriately.

There are well over 100k users running that dockerfile without issue. If we push an update that breaks ipv6 we will be flooded with reports within the hour

Minituff:
Interesting, maybe it has something to do with how I deployed it? 

I was completely unable to reach the container via it's IP/port unless I added that environment variable. 

Let me test it again, because if it is working for everyone else then it must be a "me" problem 😃

advplyr:
Someone in the Discord might know. There are a lot of skilled Docker devs in there

Minituff:
Roger that. Okay, then let's close this PR. I definitely don't want to break everything 😕

PR #4935, opened 2025-12-25
Add natural volume scaling option for web client

### LINKED ISSUE #4889: [Bug]: Non-logarithmic volume slider
### What happened?

When using the volume slider, the top portion of the volume slider has very little effect, compared to a slither at the bottom.

The volume scale is logarithmic, matching perception.

1. Play an audio track
2. Manipulate the volume control

### Audiobookshelf version

2.31.0

Docker

Linux

Firefox

### Logs

```shell

```

### Additional Notes

I can't *confirm* that the slider doesn't use the correct scale, it just certainly doesn’t feel like it

lerndmina:
## Brief summary

Adds a "Natural volume scaling" toggle in Player Settings that applies a quadratic curve to volume output, making the volume slider feel more responsive and natural across its full range.

## Which issue is fixed?

Fixes #4889

## In-depth Description

The current volume slider uses a linear scale, but human hearing perceives loudness logarithmically. This causes the top portion of the slider to feel unresponsive while small adjustments at the bottom cause large perceived volume changes.

This PR applies a quadratic curve (`volume²`) when the setting is enabled:
- 50% slider → 25% actual volume
- 70% slider → 49% actual volume
- 100% slider → 100% actual volume

This gives users finer control in the quiet range where small changes are more noticeable, while moving through the loud range faster where differences are harder to perceive.

The setting is disabled by default. Only affects web client playback.

## How have you tested this?

1. Started the dev server and client with hot-reload
2. Played an audiobook in the web client
3. Opened Player Settings and toggled "Natural volume scaling" on/off
4. Adjusted volume slider while audio was playing
5. Confirmed the slider feels more responsive with the setting enabled
6. Verified setting persists after page refresh

## Screenshots

image


nichwall:
This works in my testing.

Is there a reason this is a setting instead of just changing the calculation? We aren't changing the max/min volume so at worst it would be a one-time change for people who are not listening at 100% volume in the web client.

lerndmina:
> This works in my testing.
> 
> Is there a reason this is a setting instead of just changing the calculation? We aren't changing the max/min volume so at worst it would be a one-time change for people who are not listening at 100% volume in the web client.

Not really a specific reason, just thought people might want to toggle it on / off. The setting can be removed easily though.

nichwall:

https://github.com/audiobookshelf/audiobookshelf-client-react/pull/204

PR #4952, opened 2026-01-01
Display localized/styled text for selected filter.

sir-wilhelm:
## Brief summary

Currently the selected filter drop down displays the id used to do the filtering by the server. This change aims to use the friendly / localized value instead of the id.

## Which issue is fixed?

None, discussed in discord.

## In-depth Description

The drop down is populated by `selectedText()`, and it handles quite a few edge cases. I just added 4 more filter types to the list. If a filter is selected it'll look up the friendly name from the drop down's value list and select it based on the selected id.

## How have you tested this?

Ran it on my dev instance.

## Screenshots

Original uses ids in the drop down:

After the change it'll use the friendly name:
image

### DIFF
--- a/client/components/controls/LibraryFilterSelect.vue
+++ b/client/components/controls/LibraryFilterSelect.vue
@@ -338,6 +338,18 @@ export default {
             const series = this.series.find((se) => se.id == decoded)
             if (series) filterValue = series.name
           }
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

Issue #4937, opened 2025-12-26
[Enhancement] Add “Audible Series ID/ASIN” field to Series metadata


H2OKing89:
### Type of Enhancement

Server Backend

## Feature request: Store Audible Series ASIN (Series URL / ID) on Series

**Problem / Motivation**

Audiobookshelf lets us rename a Series to whatever we want (which is great), but once we do that there’s no “hard link” back to the original Audible Series page.

For **books**, we already commonly rely on ASINs as stable identifiers for matching and automation. But **Series** are currently just a name + description conceptually, which means:

* Two different Audible series can share the same (or very similar) name.
* Users may rename series (ex: add “(Unabridged)” or reorder text) and lose the ability to map back to the canonical series on Audible.
* External tooling and metadata workflows can’t reliably “round-trip” between ABS ↔ Audible for series-level logic.

Right now, the series object is essentially:

* `id`
* `name`
* `description`
* timestamps (`addedAt`, `updatedAt`) [Audiobookshelf API](https://api.audiobookshelf.org/)

…and `PATCH /api/series/<id>` only accepts `name` and `description` [Audiobookshelf API](https://api.audiobookshelf.org/), so there’s no place to store a provider ID.

---

### Concrete example (why I want this)

**Audiobook ASIN URL** (for a specific volume):

```text
Mushoku Tensei vol_01
https://www.audible.com/pd/B0CJWTXLPJ
```

**Audible Series URL** (for the whole series page):

```text
Mushoku Tensei: Jobless Reincarnation
https://www.audible.com/series/B0CJYVSQFB
```

In ABS I might name the series as:

* `Mushoku Tensei`
* `Mushoku Tensei (English)`
* `Mushoku Tensei: Jobless Reincarnation (Unabridged)`
* etc.

But I’d still love to preserve **the canonical Audible series ID** (`B0CJYVSQFB`) so the series can always be mapped back to Audible even if I rename it.

---

### Proposed solution (minimal + future-proof)

Add an optional field to **Series** for Audible mapping, something like:

* `audibleSeriesAsin: string | null`

  * Example: `"B0CJYVSQFB"`
* optionally (nice-to-have):

  * `audibleSeriesUrl: string | null` (derived from the ASIN + region/domain)
  * OR a more generic structure like:

    * `externalIds: { [provider: string]: string }`
    * example: `externalIds: { "audible.com": "B0CJYVSQFB" }`

#### Why I like the generic `externalIds` idea

[...]

PR #4960, opened 2026-01-04
feat: Add Audible series ASIN field to Series entity

H2OKing89:
## Brief summary

Adds first-class support for **Audible Series ASINs** by introducing a new `audibleSeriesAsin` field on the **Series** entity. This enables stronger integration with Audible metadata and allows series to be reliably linked to official Audible catalog entries.

## Which issue is fixed?

Fixes #4937

### What this enables

- Store a **10-character Audible Series ASIN** directly on a Series (`audibleSeriesAsin`)
- Accept either:
  - Raw ASIN (e.g. `B08G9PRS1K`)
  - Full Audible series URL (ASIN auto-extracted)
- Prevent accidental ASIN loss during common edit flows (blank updates won’t clear an existing ASIN)
- Search library by **series name OR series ASIN**

### ✅ Normalization + validation
- Extracts ASINs from pasted Audible URLs
- Validates format: `^[A-Z0-9]{10}$` (uppercase alphanumeric, length 10)

### ✅ Data integrity safeguards (no accidental nuking)
- **Frontend:** real-time validation + visual feedback
- **Model layer:** validation + normalization
- **Sequelize hook:** `beforeValidate` ensures validation runs for all update paths
- **Controller safeguard:** blank/empty updates won’t overwrite an existing ASIN

### Why this approach?

- **Low-friction:** ASINs are captured automatically from Audible metadata when available, but can still be entered manually.
- **URL-friendly:** Pasting a full Audible series URL "just works" via ASIN extraction.
- **Safe by default:** Safeguards prevent accidental clearing/overwriting of an existing ASIN during normal edit flows.

### Database & model layer
- Migration `v2.33.0` adds `audibleSeriesAsin` + index
- Series model:
  - Normalization helper (`normalizeAudibleSeriesAsin()`)
  - Validation rules for the 10-char format
- `beforeValidate` hook guarantees consistent validation across CREATE / PATCH / PUT
- Backend guard prevents empty updates from clearing existing ASIN data

### UI
- New `AsinInput.vue` component:
  - URL detection + ASIN extraction on paste
  - Real-time validation state + feedback
- Integrated into Series edit flow
- Protects against clearing when selecting an existing series with a blank ASIN field

### Metadata + scanning
- Audible provider returns series ASINs in results when available
- Quick match + interactive match capture and persist series ASINs automatically

### Search
- Library search now supports lookup via **series ASIN** in addition to the name

### ASIN flow architecture

```text
[...]

PR #4962, opened 2026-01-04
feat: Enable series name editing with duplicate validation

H2OKing89:
## Brief summary

Enables editing of series names directly from the book/series edit modal with proper duplicate name validation. Previously, the series name field was locked (disabled) when editing existing series, forcing users to use workarounds or third-party tools.

## Which issue is fixed?

Addresses user feedback from #4937:
- @iconoclasthero: "I just want to be able to edit the series name... I have yet to fathom why that's a locked field."
- @Vito0912: Noted users need external tools for renaming series

Also resolves the long-standing TODO in `SeriesController.js`:
```javascript
// TODO: Currently unused in the client, should check for duplicate name
```

### Why was it locked?

The series name field was disabled in `EditSeriesInputInnerModal.vue` when editing existing series, likely to prevent accidental renames that affect multiple books. However, this forced users to:
- Manually delete and recreate series
- Use third-party tools like [abstoolbox](https://abstoolbox.vito0912.de/tools)
- Work around the limitation

### What this PR does

**Backend (`SeriesController.js`):**
- ✅ Implements duplicate name validation (resolves TODO)
- ✅ Checks if another series with the same name exists in the same library before renaming
- ✅ Updates `nameIgnorePrefix` when series name changes (important for sorting)
- ✅ Returns descriptive error messages for validation failures

**Frontend:**
- ✅ Removes `:disabled` lock on series name input
- ✅ Adds `originalSeriesName` tracking to detect changes
- ✅ Validates duplicate names before submitting
- ✅ Calls `PATCH /api/series/:id` immediately when renaming existing series
- ✅ Shows backend error messages in toast notifications
- ✅ Prevents accidental overwrites with proper checks

### How it works

1. **User edits series** → Opens edit modal, name field is now editable
2. **User changes name** → Frontend tracks the change
3. **Frontend validation** → Checks if new name conflicts with existing series in library
4. **Submit** → Sends PATCH request to backend
5. **Backend validation** → Double-checks for duplicates, updates `nameIgnorePrefix`
6. **Success** → Series renamed, all books in series automatically see the new name (linked by UUID, not name)
7. **Socket event** → `series_updated` notifies all clients

### Why this is safe

- **Books link to series by UUID**, not by name - renaming doesn't break relationships
- **Duplicate validation** on both frontend and backend prevents name collisions
- **Same library only** - allows same series names across different libraries
- **Immediate PATCH** - changes are persisted to database immediately, not deferred

### Manual Testing
1. ✅ **Rename existing series** - Changed "Harry Potter" to "Harry Potter Series"
   - All books updated correctly
   - Series list shows new name
   - Sorting still works (nameIgnorePrefix updated)

2. ✅ **Duplicate name validation** - Tried renaming to an existing series name
   - Frontend: Shows toast error "A series with that name already exists in this library"
   - Backend: Returns 400 with descriptive message
   - Original name preserved

3. ✅ **Empty name validation** - Tried submitting empty name
   - Shows "Must enter a series" error
   - Form doesn't submit

4. ✅ **New series creation** - Still works as before
   - Can type new series name
   - If matches existing series, uses that series ID

5. ✅ **Cross-library series** - Same series name in different libraries
   - Works correctly, only checks within same library

### Automated Testing
- ✅ All 315 existing tests pass
- ✅ No regressions introduced

### Edge Cases Tested
- Renaming to same name (no-op, no error)
- Renaming with different capitalization (allowed)
- Multiple books in renamed series (all updated)
- Socket updates propagate to other clients

## Screenshots
<details>
<summary>Screenshots (click to expand)</summary>

Before (series name disabled):
image

After (series name editable):
image
</details>

**Validation in action:**
When attempting to rename to existing series name:
- Toast appears: "A series with that name already exists in this library"
- Form doesn't submit
- Original name preserved

### DIFF
--- a/client/components/modals/EditSeriesInputInnerModal.vue
+++ b/client/components/modals/EditSeriesInputInnerModal.vue
@@ -8,7 +8,7 @@
         <div class="bg-bg rounded-lg px-2 py-6 sm:p-6 md:p-8" @click.stop>
           <div class="flex">
             <div class="grow p-1 min-w-48 sm:min-w-64 md:min-w-80">
-              <ui-input-dropdown ref="newSeriesSelect" v-model="selectedSeries.name" :items="existingSeriesNames" :disabled="!isNewSeries" :label="$strings.LabelSeriesName" @input="seriesNameInputHandler" />
+              <ui-input-dropdown ref="newSeriesSelect" v-model="selectedSeries.name" :items="existingSeriesNames" :label="$strings.LabelSeriesName" @input="seriesNameInputHandler" />
             </div>
             <div class="w-24 sm:w-28 md:w-40 p-1">
               <ui-text-input-with-label ref="sequenceInput" v-model="selectedSeries.sequence" :label="$strings.LabelSequence" />
--- a/client/components/widgets/SeriesInputWidget.vue
+++ b/client/components/widgets/SeriesInputWidget.vue
@@ -19,6 +19,7 @@ export default {
     return {
       selectedSeries: null,
       originalSeriesSequence: null,
+      originalSeriesName: null,
       showSeriesForm: false
     }
   },
@@ -61,6 +62,7 @@ export default {
       }
 
       this.originalSeriesSequence = _series.sequence
+      this.originalSeriesName = _series.name
       this.showSeriesForm = true
     },
     addNewSeries() {
@@ -71,6 +73,7 @@ export default {
       }
 
       this.originalSeriesSequence = null
+      this.originalSeriesName = null
       this.showSeriesForm = true
     },
     submitSeriesForm() {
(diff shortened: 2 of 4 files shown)


Vito0912:
As I said in my other comment, I do not think it is a good idea to make this changeable so easily. In my opinion, this should be added to the metadata utils, where other such actions are already possible, although I am just a contributor, so whatever. But that is not why I am commenting.

It is cool that you seem to have found AI that works for you. But you help nobody if you open AI-generated PRs. ABS already had a flood of AI PRs and you contributed 5 new ones.

It adds review work for things you did not write and likely do not fully understand, though I do not want to assume this about you. Even your answers are copied straight from an AI. The following was, in no world, written by a human: https://github.com/advplyr/audiobookshelf/pull/4960#issuecomment-3708235195
Just wanted to state this.

AI is a nice assistance. Submitting a AI only PR and even answering trough AI is no assistance.

H2OKing89:
Yeah the endpoint already existed. It just wasn’t being used because the UI prevented renames and the server-side handler still had a TODO about duplicate-name checking.

This PR adds the missing duplicate-name guard (and updates nameIgnorePrefix) and then wires the UI to call the existing PATCH when the name changes. just enabling the existing single-series edit path.

iconoclasthero:
I just think it should be up to the individual library maintainers to be able to manually edit series names without having to delete the series and re-**type** it into a new name (since I can't even copy the old one) just to be able to make _any_ corrections—which I still can't edit later.  To me, it feels like unnecessarily restrictive UI.  I made this point two years ago and, while I run into it infrequently because I do not read _that many_ books that are elements of a series, it is frustrating when I do run into it.
image

As an e.g., if I want to change this title in **my** library to "MurderBot Diaries," without deleting the series and retyping it in a new one and reentering the element number, I'm SOL. 

I'm not sure what "I do not think it is a good idea to make this changeable so easily"—which is the same as saying "I think this should be hard/harder to do"—is supposed to be protecting...me from balling up my own library?\*  I don't know what "other metadata utils" is referring to, but the field is right there, I don't want to have to go poking around anywhere else just to change e.g., "Murderbot" to "MurderBot."

> For the reason why it is the case: Probably because this would affect many books in a book dialogue. Similar how you cannot rename a genre or tag, but you can via the setting. But dunno

If this speculation is correct, I am ok with it having the same back-end effect as the present method of correcting series names, i.e., I delete the series and then recreate a new one on exactly one book record leaving all other records to be manually edited as well.  I don't use batch editing, but it would certainly be an obvious place to edit the series name.  

This is a point of appreciable _"frustration"_ for me every single time I run into it and would love to see it change...and as such I should probably run this response through ChatGPT to tone it down for me!  (I.e., I need the opposite of K&P's Luther, the anger translator.)

\*FWIW - If this _is_ just designed to protect users from themselves, then a longer-range solution could be tiered settings levels e.g., in Kodi.

Vito0912:
Just for the record. I do not mean to delete and add the series again. This would remove the set position.

But like for generes and tags, you have the option to rename them at a global place in the UI, also without removing and adding and is done in two clicks.
This is probably done because the same tags and genres can be used across libraries. 
So after that fact maybe making the textfield editable is really the better option, because series are not across libraries. But my comment has no effect on mergability anyway.

I still think it would be cleaner in the settings, because then it would be all at one place and the benefit, you would be able to delete a series, but again, not will change if this PR is merged, just some comment

(And because that PR here is 100% AI generated anyway I also didn't do much thinking here, because that was not the main point I was making)

PR #4970, opened 2026-01-08
Make the download icon easier to see on share page for books with light covers

sir-wilhelm:
## Brief summary

This change improves the accessibility of the download icon on the share page.

## Which issue is fixed?

n/a

## In-depth Description

The current page always displays a light icon, but it's can be quite invisible for books with light covers. This just takes advantage of a property that's already being set (`this.coverBgIsLight`) to use a dark icon if the cover/background is light.

## How have you tested this?

Tested on dev instance with light cover, dark cover, and no cover.

## Screenshots

pre change:
image
image

post change:
light cover:
image
image

dark cover:
image
image

no cover (uses the dark cover color):
image

### DIFF
--- a/client/pages/share/_slug.vue
+++ b/client/pages/share/_slug.vue
@@ -14,7 +14,7 @@
         </div>
 
         <ui-tooltip v-if="mediaItemShare.isDownloadable" direction="bottom" :text="$strings.LabelDownload" class="absolute top-0 left-0 m-4">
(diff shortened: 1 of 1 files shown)


nichwall (approved):
Looks good to me, works in my testing on a dev server.

sir-wilhelm:
I'll re-open if this is still an issue on the new client (once it's out).

PR #4973, opened 2026-01-09
Fix screen reader compatability issue for the "jump backwards" button for media controls

KiwiHour:
## Brief summary

Fixes screen reader compatability issue for the "jump backwards" button for media controls

## Which issue is fixed?

Pointed out by jonathan859 on discord

## In-depth Description

Very simple solution, follows the same format as the forward button. 

## How have you tested this?

The program runs fine, I do not have a screen reader to test but considering how small the change is it should work perfectly

## Screenshots
n/a


advplyr:
Thanks!

Issue #4934, opened 2025-12-24
[Bug]: Series field is case-sensitive and reverts to original casing, making correction difficult


JamesDean3:
### What happened?

When an audiobook is imported and the Series field is automatically populated, the series name is treated as case-sensitive. This allows the same series to exist multiple times with different capitalization (for example, Example Series and example series).

If I try to correct the casing by removing the series and re-adding it with the desired capitalization, the series name reverts back to the original (undesired) casing after saving.

If I instead try to add the series again with the correct casing while the incorrect one exists, audiobookshelf reports that the series already exists for that book (so essentially it's fault checking is case insensitive, and sees that it's already set with that name, so it will neither let you add both or recognize there's a change that needs to be saved when removing the undesired one).

The only option is to remove the series All together, save, then open the book again and add it with the correct capitalization.

### What did you expect to happen?

Series names should be treated as case-insensitive, or editing a series name’s casing should persist after saving, making it possible to correct capitalization without workarounds. 

### Steps to reproduce the issue

1. Import an audiobook that automatically populates the Series field.

2. Ensure the imported series name has incorrect or inconsistent capitalization.

3. Open the book up to be edited in audiobookshelf and remove the series entry.

4. Re-add the same series name with corrected capitalization.

5. Save the changes.

6. Reopen the book’s edit page and observe that the series name has reverted to the original capitalization.

### Audiobookshelf version

2.31.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

### Additional Notes

I realize I'm one version behind, but I didn't see this in the change log, and as far as I can tell, there isn't currently a bug report mentioning this.

nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/4255 and https://github.com/advplyr/audiobookshelf/issues/4033

Issue #4033, opened 2025-02-22
[Bug]: Different case sensitivity behavior for some tasks.


karmazinondrej:
### What happened?

Hi, 
I had different font case for "Serie", which resulted into showing it as two separate series. 

But when I tried to correct it by deleting Serie and rewriting it (just changing the cases), it did not allow to save, as there was no change. I had to delete it, save and then rewrite it.

### What did you expect to happen?

I would expect to take it as case-insensitive for Series and other "groupping" types.

If not, then I would expect to behave the same for editing, but there it already is case-insensitive.

I expect consistent behavior at least.

### Steps to reproduce the issue

Name the series with different cases and try to get it into one serie.

### Audiobookshelf version

 v2.19.3 

### How are you running audiobookshelf?

Other (list in "Additional Notes" box)

### What OS is your Audiobookshelf server hosted from?

Other (list in "Additional Notes" box)

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Edge

### Logs

```shell

```

### Additional Notes

lxc container on proxmox

PR #4976, opened 2026-01-09
Fix case-insensitive lookup for series and authors during import

### LINKED ISSUE #4255: [Bug]:Series grouping is case-sensitive — identical names with different letter case are treated as separate series.
### What happened?

Audiobooks with identical series names that differ only in letter casing (e.g., Harry Potter vs harry potter) are not grouped together. This results in duplicate series entries in the library UI.

Books with series names that differ only by case should be grouped together under one series entry. Comparison should be case-insensitive. For example, Harry Potter and harry potter should be treated as the same series.

1.	Import two audiobooks with identical series names that differ only by case (e.g., Harry Potter and harry potter).
2.	Open the library and view the series list.
3.	Observe that the books are not grouped under the same series.

### Audiobookshelf version

2.21.0

Docker

Linux

Chrome

### Logs

```shell

```

### Additional Notes

This causes confusion and disorganization in larger libraries where metadata inconsistencies are common. Making the grouping logic case-insensitive would greatly improve usability.

rbennion:
## Brief summary

Prevent duplicate series and authors from being created when names differ only in casing (e.g., "Harry Potter" vs "harry potter") during library scanning/import.

## Which issue is fixed?

Fixes #4255

Related to #4033 and #4934

## In-depth Description

When importing books, the code has two paths for looking up existing series/authors:

1. **Database fallback** (`getByNameAndLibrary`) - uses `fn('lower', col('name'))` for case-insensitive matching ✅
2. **Filter data cache** - was using `se.name === seriesName` for case-sensitive matching ❌

This inconsistency meant:
- First scan (no filter data loaded): Case-insensitive lookup → correct
- Subsequent scans (filter data cached): Case-sensitive lookup → creates duplicates!

**The fix:** Changed both `getSeriesIdByName` and `getAuthorIdByName` in `Database.js` to use case-insensitive comparison when using the filter data cache:

```javascript
// Before (buggy):
return this.libraryFilterData[libraryId].series.find((se) => se.name === seriesName)?.id || null

// After (fixed):
const seriesNameLower = seriesName.toLowerCase()
return this.libraryFilterData[libraryId].series.find((se) => se.name.toLowerCase() === seriesNameLower)?.id || null
```

This affects both series and authors, preventing duplicates for both during import.

## How have you tested this?

1. All 315 existing unit tests pass
2. Manual testing steps:
   - Import a book with series "Harry Potter"
   - Import another book with series "harry potter" 
   - Both books now correctly group under the same series (using the first casing encountered)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/server/Database.js
+++ b/server/Database.js
@@ -624,7 +624,9 @@ class Database {
     if (!this.libraryFilterData[libraryId]) {
       return (await this.authorModel.getByNameAndLibrary(authorName, libraryId))?.id || null
     }
-    return this.libraryFilterData[libraryId].authors.find((au) => au.name === authorName)?.id || null
+    // Case-insensitive comparison to match getByNameAndLibrary behavior
+    const authorNameLower = authorName.toLowerCase()
+    return this.libraryFilterData[libraryId].authors.find((au) => au.name.toLowerCase() === authorNameLower)?.id || null
   }
 
   /**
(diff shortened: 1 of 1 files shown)


nichwall:
Hey, just as an FYI we are not reviewing AI generated PRs due to an increasing number of them and limited development and review time. Similar PRs were also created a few days ago by someone else.

rbennion:
No problem. I am a huge ABS user and also a big user of Claude code. I thought I would go ahead and see if I could fix those two. I did review them myself though. ;). Cheers!

PR #5004, opened 2026-01-25
Access token refresh grace period

### LINKED ISSUE #4630: [Bug]: Need to re-login very often
### What happened?

After the update to authentication, it's automatically logging out multiple times per day. It usually is slow for a bit, then goes back to the login screen. This is happening on both the web UI and the Android app.

I'm not sure but the login session should last longer.

1. Just use the app until the account is logged out.

### Audiobookshelf version

v2.29.0

Docker

Chrome

### Logs

```shell

```

### Additional Notes

Unraid OS

nichwall:
## Brief summary

This pull request adds a brief 1 minute grace period when generating new access tokens to address a race condition when rotating tokens.

## Which issue is fixed?

Attempted fix https://github.com/advplyr/audiobookshelf/issues/4630

## In-depth Description

A race condition can occur in the token refresh logic when the Access Token has expired, because the `rotateTokensForSession` saves the new session regardless of whether it has already been updated. This allows a client to send two refresh requests to the server, where each generates a new refresh token. The first refresh token will be returned to the client, but the second one overwrites the first in the database, causing the user to be unable to generate a new access token an hour later when it expires.

To fix this, I added two new columns to the session table, `lastRefreshToken` and `lastRefreshTokenExpiresAt`, to give a short grace period to the second request. When checking if the access token is valid, we check if the current refresh token matches and has not expired, or if the last refresh token has not expired during its 1 minute grace period (we can make this shorter, but the race condition is more likely to occur with larger databases so erred on the side of caution).

Then, when rotating the tokens, we only store the new tokens if the `refreshToken` matches what is now the `lastRefreshToken` to prevent the second request from overwriting the first request. If this write to the database does not occur, the rotation function instead loads the refresh token from the database and returns this token instead of generating a new one.

## How have you tested this?

I was unable to recreate the original issue with a short lived access token (3 seconds), but tested by navigating between pages on a web browser (desktop) and Android app at the same time, with brief pauses to verify each token was refreshing correctly.

## Screenshots

No front end changes.


server/auth/TokenManager.js
```diff
@@ -183,20 +183,44 @@ class TokenManager {
...
+    }
 
     // Update the session with the new refresh token and expiration
     session.refreshToken = newRefreshToken
-    session.expiresAt = newExpiresAt
-    await session.save()
+    session.expiresAt = new Date(Date.now() + this.RefreshTokenExpiry * 1000)
+
+    // Only update the session if the refresh token hasn't changed since we originally read it
+    const [numUpdated] = await Database.sessionModel.update(session, {
```
kevinhaugen:
```suggestion
    const [numUpdated] = await Database.sessionModel.update({
        refreshToken: session.refreshToken,
        expiresAt: session.expiresAt,
        lastRefreshToken: session.lastRefreshToken,
        lastRefreshTokenExpiresAt: session.lastRefreshTokenExpiresAt
      }, {
```

Sequelize’s Model.update() expects a plain object. Passing a model instance may appear to work, but leads to unexpected behavior. I put together a [little gist](https://gist.github.com/kevinhaugen/0f55384ce5b1174bbb3261e43b5d44b6), and you can see the same refresh token keeps getting returned on the rotation request. Swapping to the plain object resolves it. You can also run the script to mock the original issue. 

Original
```
running 10 concurrent refresh requests...
refresh responses: ok=2 bad=8
bad by status: { '401': 8 }
```

Grace
```
running 10 concurrent refresh requests...
refresh responses: ok=10 bad=0
```

I ran into this issue while using the Plappa and Prologue Apple Watch apps. The token on the watch gets invalidated and you can't auth again until you're back by your paired phone. I came to open a similar PR!

nichwall: Hey, thanks for catching that! I'm not able to view your gist, it just redirects to this PR.

kevinhaugen: Oh shoot! I fixed the link.

nichwall: Thanks! That is a much more robust way to test it than what I was doing. :)
I can sit down and look at this again in the next few days.

advplyr:
thanks, this is working well and makes sense. will include in the next release

PR #5006, opened 2026-01-26
Fix applying chapters not fully replacing existing chapter list

### LINKED ISSUE #4939: [Bug]: Applying chapters does not fully replace existing chapter list
### What happened?

Image
Image

I’m attempting to update the chapters for a book. The original audiobook contains 61 chapters, but only 36 chapters are returned during a lookup search.

After I select **Apply Chapters**, the chapter list partially updates: the first 36 chapters are replaced with the new titles and timestamps, but the remaining original chapters are still present.

Applying chapters should completely replace the existing chapter list with the chapters returned from the lookup search.

Open a book with an existing chapter list (61 chapters).
Perform a chapter lookup.
Observe that only 36 chapters are returned.
Click Apply Chapters.
The first 36 chapters update, but the remaining original chapters are not removed.

### Audiobookshelf version

v2.32.1

Docker

None

### Logs

```shell

```

### Additional Notes

How are you running audiobookshelf?
Docker (Synology NAS)

What OS is your Audiobookshelf server hosted from?
Synology DSM (NAS)

CutSnake01:
## Brief summary

Fix bug where applying Audible chapters didn't fully replace the existing chapter list when the new list had fewer chapters.

## Which issue is fixed?

Fixes #4939

## In-depth Description

When applying chapters from Audible lookup, if the new chapter list had fewer chapters than the existing one, the old extra chapters were being retained. For example, going from 61 to 36 chapters would result in a hybrid list with 36 updated chapters plus 25 leftover old ones.

The bug was in the `applyChapterData()` function which incorrectly fell back to old chapters when the Audible chapter array was exhausted. This fallback should only happen for explicitly locked chapters, not for all remaining old chapters.

The fix removes the fallback to old non-locked chapters, ensuring the chapter list is fully replaced rather than merged with leftover old chapters.

This affects any user who:
- Has a book with more chapters than Audible returns
- Uses the chapter lookup feature

## How have you tested this?

1. Open a book with 61+ chapters in the chapter editor
2. Perform Audible chapter lookup (returns ~36 chapters)
3. Click "Apply Chapters"
4. Verify only 36 chapters remain (not 61)
5. Verify locked chapters are preserved if any were set

## Screenshots

N/A - No UI changes, logic fix only

### DIFF
--- a/client/pages/audiobook/_id/chapters.vue
+++ b/client/pages/audiobook/_id/chapters.vue
@@ -753,8 +753,6 @@ export default {
         } else if (audibleChapters[audibleIdx]) {
           merged.push({ ...audibleChapters[audibleIdx], id: i })
           audibleIdx++
-        } else if (this.newChapters[i]) {
-          merged.push({ ...this.newChapters[i], id: i })
         }
       }
       this.newChapters = merged


nichwall (approved):
This is working in my testing.

PR #5015, opened 2026-01-29
Fix log purging

sir-wilhelm:
## Brief summary

Scanner logs are never purged (despite there being a server setting for it).
Daily logs purge on startup (w/ a potential off by one error if you start the server on a new day), and the continuous purging does not work, since new logs were never tracked in `dailyLogFiles`.

## Which issue is fixed?

No issue, this was brought up in discord.
https://discord.com/channels/942908292873723984/942914154254176257/1465893426049323103

## In-depth Description

Two commits, one for adding Scanner log purging and another for fixing the daily log purging.

LibraryScan/scanner changes:
The Scanner makes sure the `scanLogDir` is created, but so does the `LogManager`. They both don't _need_ to, but I did not want to decide where that logic should reside, so I left it in both places.
I added a purge scanner logs function/call after each scan is complete.

**Note**: the default is 2, and scanner logs have never been purged, so there could be quite a bit of clean up once this hits prod instances.

LogManager/daily log changes:
extracted a common function to do the purging/apply the log retention.
I call the function after daily logs are created and on startup.
`dailyLogFiles` now tracks when new files are created (so they can be purged).

## How have you tested this?

Dev instance. I tested having extra logs on start up and rolling over to a new day to test daily purging. Running scans to make sure those are purging was also tested.

## Screenshots

n/a


server/managers/LogManager.js
```diff
@@ -36,6 +36,12 @@ class LogManager {
     return global.ServerSettings.loggerDailyLogsToKeep || 7
   }
 
+  async enforceDailyLogRetention() {
```
nichwall, line 43: Can we enforce a minimum of 1 daily log to keep?

I don't think any edge cases would arise by deleting the log file immediately after logging every time and then recreating it the next time we log a line, but I definitely see that having some performance problems.
It also defeats the purpose of all logging to immediately delete the file if it is 0.

server/scanner/LibraryScan.js
```diff
@@ -137,6 +139,40 @@ class LibraryScan {
...
+    await this.purgeOldScanLogs(scanLogDir)
+  }
+
+  /**
+   * Keep the most recent N scan logs in metadata/logs/scans.
+   * Where N is the server setting `loggerScannerLogsToKeep`.
+   *
+   * @param {string} scanLogDir
+   */
+  async purgeOldScanLogs(scanLogDir) {
```
nichwall: I think this makes more sense to have in `LogManager`. All of the scanning and deletion of log file functionality already exists there (would need some updates to handle both arrays easily).

Otherwise I don't have a problem with this being in this file for now.

nichwall (commented):
Looks good to me.

High level, I think it makes sense for the `LogManager` to own all log limit enforcement, including scan logs.

server/managers/LogManager.js
sir-wilhelm: It's currently using `loggerDailyLogsToKeep()`, it'll use whatever is in the database settings and default to 7.

server/scanner/LibraryScan.js
sir-wilhelm: Ya, it makes sense/would be easy to have a `purgeOldScanLogs()` in `LogManager` since it already knows the log path and would not need it to be passed in.

I think it might still have it's own method for cleaning up the files, since the current `enforceDailyLogRetention()` utilizes the `removeLogFile(filename)` method that handles file deletion/`dailyLogFIles` array clean up.

server/managers/LogManager.js
nichwall: That makes sense. I mean what happens if the database has a zero? I realize this is not currently exposed in the web interface, just thinking about for in the future.

While writing this response, I realized the more important thing is that we probably want to say "0" is "unlimited", or keep all log files (similar to backup size and podcast episodes to keep).

server/scanner/LibraryScan.js
nichwall: Yeah, those functions would need to be rewritten to account for two different sets of logs to track.

Maybe it makes sense to just entirely get rid of the daily log file array from the `LogManager`, because it doesn't really do anything other than say if we have files to delete. I would need to think about that more, but either way that sort of change would be in its own PR and we could address "reusing functions" at that point.

sir-wilhelm: Ya, I could also see having a new PR for moving the logic of `saveLog` in `LibraryScan` to the `LogManager` or `Logger` object. Seems odd it's doing the path/file writing instead of just passing the `logLines` off to some other class to deal w/ it.

For this PR should I move `purgeOldScanLogs()` into `LogManager` and have the `LibraryScan` call it off the logger (ex: `await Logger.logManager?.purgeOldScanLogs()`)?

If so I think I'd also change:
```
const scanLogDir = Path.join(global.MetadataPath, 'logs', 'scans')

await fs.ensureDir(scanLogDir)
```
to just be:
```
const scanLogDir = Logger.logManager.ScanLogPath
```
Since the `LogManager` does the `ensureDir` call, and already has the logic of building the path.

server/managers/LogManager.js
sir-wilhelm: That makes sense. I'll add `if LogsToKeep <= 0 return` (for both daily/scan).

sir-wilhelm:
Moved the scanner log purge into `LogManager` and added support for keeping all logs by setting the value to 0 or a negative number.

Tested changes on my dev instance and they still purge scanner logs correctly.

PR #5036, opened 2026-02-06
Improved subtitle parsing to account for bare colon in title

kctdfh:
## Brief summary

Improved subtitle parsing so that only `: ` (colon followed by a space) is treated as a title/subtitle separator. Previously, a bare `:` was matched, which incorrectly split titles like [10:04](https://www.goodreads.com/book/show/20613582-10) into title "10" and subtitle "04".

## Which issue is fixed?

This is an edge case that I encountered in my own server and decided to just fix it. I haven't seen any reports of this issue from other users.

## In-depth Description

The subtitle parsing logic previously used a bare colon (`:`) to detect and split title/subtitle boundaries. This works for the vast majority of books (e.g. "Everything Is Miscellaneous: The Power of the New Digital Disorder"). However, this splitting logic fails for titles that contain a colon as part of the name itself (e.g. [10:04](https://www.goodreads.com/book/show/20613582-10) by Ben Lerner, or something like "Making the Mission:Impossible Movies", which isn't a real book but serves as a good example).

This PR changes all subtitle-splitting logic to require a space after the colon (`": "`) before treating it as a separator. This is consistent with standard conventions for title/subtitle formatting and would also not trigger incorrect splits for titles with bare colons. Note that the subtitle-splitting logic of `getSubtitle(folder)` in `scandir.js` is not changed because there we depend on " - " only.

Lastly, I've kept the focus of this PR narrow to just accounting for the space after the colon in subtitle parsing, but there are some related improvements that could be made in the future. For example, accounting for the rare case where a combined book title + subtitle might have more than one colon. See [this book](https://www.goodreads.com/book/show/6654264-code-name) which could be received as "Mission: Impossible Code Name: Judas" or [this book](https://www.goodreads.com/book/show/27879233-humanoid-encounters) which is actually officially called "Humanoid Encounters: 1995-1999: The Others amongst Us". These are really narrow edge cases but there could be other cases of multiple colons in title + subtitle out there because apparently Kindle Direct Publishing doesn't ensure that the book title doesn't include a colon (see [this discussion](https://www.kdpcommunity.com/s/question/0D52T00005FHSzMSAX/book-title-has-2-colons-before-subtitle-on-amazon-sales-page?language=en_US)). I propose no fix for these cases and don't really have an opinion either. I think the current heuristic (i.e. everything after ": " is the subtitle) is good enough.

## How have you tested this?

- All existing + new unit tests pass.
- Verified that standard colon-space subtitles (e.g., "Anna Karenina: subtitle", "The Great Gatsby: A Novel") are still correctly split.
- Verified that bare-colon titles (e.g., "10:04", "Making the Mission:Impossible Movies") are no longer incorrectly split.
- Verified the changes in the web client by passing the Title in the nfo file and setting Title and Subtitle to null in the other active metadata priorities/sources) so that they come from the nfo file parser.

## Screenshots

Before
1004_before

After
1004_after


nichwall (approved):
Looks good to me and runs locally.

advplyr:
Thanks!

PR #5042, opened 2026-02-08
Fix OpenAPI spec description

openam:
## Brief summary

I tried to implement a Custom Metadata Provider, and when I tried to use it my times were way too long with parsing the duration into seconds. The results came out correctly when I set the duration as minutes.

## Which issue is fixed?

I didn't create an issue. I searched quickly, and didn't see anything.

## Screenshots

With parsing duration into seconds
image

With parsing duration into minutes
image


advplyr:
Thanks!

PR #5045, opened 2026-02-09
fix: support uploading files to existing library item folders

### LINKED ISSUE #1894: [Bug]: Cannot Upload to existing folder (via api)
### Describe the issue

I would like to upload a new file to an existing folder to extend a podcast.
I get an 500. From the code in the upload controller this is intentional. I cannot imagine why, so I thought this must be untentional.

If it is intentional, how can I upload a new file to an existing podcast then?

1. use the api to upload to an existing folder
2. 500 -> Folder already exists

### Audiobookshelf version

v2.2.23

Docker

ScottamDendar:
## Brief summary

This PR attempts to address enhancement #1894, Attempting to upload a file/book to existing library items. I simply changed the `pathexists` api call to accept the filepath of a file within an item. The UI was also updated to call the `pathexists` endpoint per file within an item, as a result.

## Which issue is fixed?

Closes #1894 

## In-depth Description

Someone tried to compliment my audiobook with an ebook supplimant, and ran into this issue. I thought it would be a good into the project.

The update is rather straightforward, but I decided to allow partial uploads. this means if at least 1 file _doesnt_ exist (meaning the rest are dupes), the upload will proceed with the new file. I could change it though that all the files must not exist before uploading.

I understand there is a UI update in the works, but considering its months away, and this update is very tiny, I was hoping it would be acceptable.

## How have you tested this?
I ran the `npm run test` command, but found the controller I updated had no unit tests. If desired, i could start some, though itd certainly take me a little longer.

1. create library
2. upload file 'foo.m4b' with author 'anyguy'
3. attempt to upload file 'foo.m4b' with author 'anyguy' again
  - it should fail, as it used to
4. upload file 'foo.pdf' with author 'anyguy'
5. Notice Library has book 'Foo' from 'anyguy' with both uploads being successful.

## Screenshots

Attached a video of the change working, and also blocking a re-upload (im used to work where the video would embed, I could post it privately to youtube if we still want it embedded lol).
https://github.com/user-attachments/assets/54e35a66-681e-484c-87df-84bf40de9fa8

Without my changes, the red error happens each time.

### DIFF
--- a/client/pages/upload/index.vue
+++ b/client/pages/upload/index.vue
@@ -377,26 +377,33 @@ export default {
 
       const itemsToUpload = []
 
-      // Check if path already exists before starting upload
-      //  uploading fails if path already exists
+      // Check if path already exists for each file before starting upload
+      //  uploading fails if all paths already exist
       for (const item of items) {
-        const exists = await this.$axios
-          .$post(`/api/filesystem/pathexists`, { directory: item.directory, folderPath: this.selectedFolder.fullPath })
-          .then((data) => {
-            if (data.exists) {
-              if (data.libraryItemTitle) {
-                this.$toast.error(this.$getString('ToastUploaderItemExistsInSubdirectoryError', [data.libraryItemTitle]))
-              } else {
-                this.$toast.error(this.$getString('ToastUploaderFilepathExistsError', [Path.join(this.selectedFolder.fullPath, item.directory)]))
+        let filesToUpload = []
+        for (const file of item.files) {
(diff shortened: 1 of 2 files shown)


ScottamDendar:
~~Update commit message to "feat: etc."~~ Done

ScottamDendar:
Closing issue: The new react client doesn't use the same api, so the updates made to the server and the vue client are not needed. Also, the new client does not have the same issue.

PR #5052, opened 2026-02-12
Add placholder books

### LINKED ISSUE #3221: [Enhancement]:
### Type of Enhancement

Web Interface/Frontend

When purchasing audiobooks, the goal is often to complete a specific series. To better keep track of the books you already own and those you still need to buy, it would be helpful to have the option to add placeholder audiobooks to a series. These placeholders could consist of a cover and matching metadata (both also editable like already possible), but would not be playable (perhaps indicated with an alpha/saturation filter on the cover or an extra icon like a little shopping cart?). 

The feature would allow users to better manage their audiobooks (purchase planning, organization/tracking of books). Also its just more asthetic to look at the series in that way when all the books are displayed.

Saturation to black and white, added little shopping cart in the corner, also alpha set to 35%
Example

v2.11.0

Example2
Here two books are missing for this series. But would I also remember it in a few months without looking it up again?

JensBickel:
## Brief summary
I found the placeholder issue/feature request on GitHub and I found the placeholder feature request on GitHub and implemented a first version of it.
See: https://github.com/advplyr/audiobookshelf/issues/3221

This PR adds manually creatable placeholders for existing book series and at the author level.

## Which issue is fixed?
closes #3221 
## In-depth Description
This implementation allows users with upload permission to add placeholders at both the author and series levels.

Placeholders are tracked in the database and can be selected as upload targets, which allows in-place promotion to a real library item during upload.
This makes it possible to prepare and organize incomplete series entries before media files are available.
When a matching upload is completed, the placeholder is reused instead of creating a duplicate item, keeping metadata continuity.

Placeholders are visually distinct from regular books.

The implementation is backed by the file system to closely match the behavior of real books.

## How have you tested this?

I added unit tests for the newly introduced code and manually tested the behavior in two browsers.
I verified that placeholders can be created from both author and series views and appear with distinct placeholder styling in the UI.
I also validated the upload flow by targeting an existing placeholder and confirming it is promoted in place to a regular library item without creating duplicates.

## Screenshots
Before
After
Details

### DIFF
--- a/server/controllers/AuthorController.js
+++ b/server/controllers/AuthorController.js
@@ -353,8 +353,11 @@ class AuthorController {
       return res.status(400).send('Library item already exists at that path')
     }
 
+    let createdPlaceholderDirectory = false
     try {
+      const placeholderDirectoryExisted = await fs.pathExists(outputDirectory)
       await fs.ensureDir(outputDirectory)
+      createdPlaceholderDirectory = !placeholderDirectoryExisted
     } catch (error) {
       Logger.error(`[AuthorController] Failed to create placeholder directory "${outputDirectory}"`, error)
       return res.status(500).send('Failed to create placeholder directory')
@@ -430,6 +433,11 @@ class AuthorController {
     } catch (error) {
       Logger.error('[AuthorController] Failed to create author placeholder', error)
       await transaction.rollback()
+      if (createdPlaceholderDirectory) {
+        await fs.remove(outputDirectory).catch((cleanupError) => {
+          Logger.error(`[AuthorController] Failed to remove placeholder directory after rollback "${outputDirectory}"`, cleanupError)
+        })
+      }
       return res.status(500).send('Failed to create placeholder')
     }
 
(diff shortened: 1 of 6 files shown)


JensBickel:
I have chosen a Draft as I am currently looking into improving the test coverage further.

advplyr:
This isn't something we're going to support right now. This is also far too big of a change for a first time contributor. It takes as long or longer to review something like this than it does to write it

JensBickel:
@advplyr  
Thanks for taking the time to look at this and for the honest feedback. I completely understand your concerns about the size of the PR and the review effort involved, especially for a first-time contributor.

I came across issue #3221 while looking for ways to organize my planned purchases and thought I’d give it a try. If this feature isn’t something you’re planning to support, I totally respect that.

That said, if there’s interest in the idea in principle, I’d be happy to explore a smaller, more incremental approach or work within any constraints you’d want such a feature to meet. I just wanted to better understand whether it’s out of scope entirely or simply not the right time.

Thanks again for maintaining this awesome project that already brought me countless hours of fun listening to my audio-book library.

advplyr:
It's completely out of scope right now so I don't think you should spend time exploring it for the main project. A lot of users have built external tools and apps for things out of scope

PR #5063, opened 2026-02-16
IDOR fixes

### LINKED ISSUE #5062: [Bug]: IDOR in Media Progress Controller
### What happened?

I was looking through the `server` code and noticed that the `server/controllers/MeController.js` does not validate the user when deleting the `mediaProgresses` data. This allows an attacker to delete the progress of a victim user, if they can identify the GUID of the `mediaProgresses` record (which is admittedly fairly difficult unless there's a leak somewhere).

I'm unsure how an attacker would get the `mediaProgressId` for an attacker organically, and due to it being a GUID they can't really brute-force it. Because of this, it would likely lower the severity of this finding fairly drastically. However, it *is* still a vulnerability in software that I use myself, so I'm hoping to make it better, even if only slightly.

The server code should validate that the `mediaProgresses` record belongs to the user making the request (or potentially an admin?)

1. To test this, I ran audiobookshelf in a docker container using the default `docker-compose.yml` file,  added a couple audio books to the `audiobooks` folder, and created a library for that folder. I then setup 2 users (`user1` and `user2`). I logged in as `user1`, listened to the book for a bit, and saw that a record was created in `mediaProgresses` for my user (validated in sqlite with `select * from mediaProgresses mp INNER JOIN users u on mp.userId = u.Id where u.username = 'user1';`. I then logged out, and in a new browser session logged in as `user2`. Using the bearer token for `user2`, I crafted a request using BurpSuite to request `DELETE /api/me/progress/<victim-media-progress-id>`. 

Image

After the request, the `mediaProgresses` record was missing from the database. 

### Audiobookshelf version

v2.32.1

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

mandreko:
## Brief summary

This code should add authorization checks in the `MeController.js` to prevent an obscure IDOR vulnerability.

## Which issue is fixed?

Fixes #5062 

## In-depth Description

This PR fixes 3 IDOR (Insecure Direct Object Reference) vulnerabilities in the MeController that allowed authenticated users to access and manipulate resources belonging to other users. 

1. IDOR in Media Progress Deletion
2. Missing Authorization in Bookmark Operations
3. IDOR in Listening Sessions

These issues would allow:
- Users could delete any media progress record by ID
- Users could create/modify/delete bookmarks for restricted content
- Users could view listening sessions for content they shouldn't access

I attempted to follow code conventions where I saw. But feel free to make modifications as needed ;)

## How have you tested this?

I added supporting unit tests (lots of copy/paste since I'm not a huge js developer)

### DIFF
--- a/test/server/controllers/MeController.test.js
+++ b/test/server/controllers/MeController.test.js
@@ -547,9 +547,7 @@ describe('MeController - IDOR Security Tests', () => {
 
       // Create mock context with getUserItemListeningSessionsHelper
       const mockContext = {
-        getUserItemListeningSessionsHelper: sinon.stub().resolves([
-          { id: 'session1', timeListening: 300, startedAt: Date.now() }
-        ])
+        getUserItemListeningSessionsHelper: sinon.stub().resolves([{ id: 'session1', timeListening: 300, startedAt: Date.now() }])
       }
 
       const fakeReq = {


advplyr:
Thanks!

PR #5068, opened 2026-02-17
Describe nginx proxy manager fix for frequent logouts

### LINKED ISSUE #4630: [Bug]: Need to re-login very often

jneidel:
See #4630 and https://github.com/advplyr/audiobookshelf/issues/4630#issuecomment-3842888850

## Brief summary

Like I mentioned on the issue, I was logged out frequently and this configuration change on my nginx proxy manager fixed this for me and for @tburnsed (who intially proposed it).

## Which issue is fixed?

Partially fixes #4630 (partial since many proxy manager seem to have this issue, but this documentation only applies to nginx proxy manager)

## How have you tested this?

For two weeks on my homeserver setup (dockerized audiobookshelf, dockerized nginx proxy manager). Connections from web and android via admin account.

## Screenshots

Image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -93,6 +93,22 @@ Toggle websockets support.
 
 <img alt="NGINX Web socket" src="https://user-images.githubusercontent.com/67830747/153679106-b2a7f5b9-0702-48c6-9740-b26b401986e9.png" />
 
+Add this Custom Nginx Configuration (under "Edit Proxy Host" > settings cog):
+```nginx
+proxy_connect_timeout 86400s;
+proxy_send_timeout 86400s;
+proxy_read_timeout 86400s;
+send_timeout 86400s;
+
+proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+proxy_set_header X-Forwarded-Proto $scheme;
+proxy_set_header X-Real-IP $remote_addr;
+
+proxy_set_header Upgrade $http_upgrade;
+proxy_set_header Connection "upgrade";
+proxy_set_header Host $host;
+```
+
 ### NGINX Reverse Proxy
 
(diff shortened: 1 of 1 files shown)


nichwall:
See comments in https://github.com/advplyr/audiobookshelf/issues/4630#issuecomment-3914700615

I don't think this actually fixes the issue.

jneidel:
> I don't think this actually fixes the issue.

It has fixed it for me (I did not change anything else.)
I've disabled the configuration for the moment. Now the issue should return. I'll report back.

nichwall:
I'm glad it's working for you as is, but I think this is fixing a symptom rather than actually fixing the issue.

jneidel:
If the issue only arises with a proxy in the middle, but not with the raw app, that would suggest that it is not an application issue, but a proxy configuration issue.
Application issues can't be fixed by changing proxy configuration.
Proxy configuration issues could however be fixed by improving the application though (e.g. by aligning with what the proxies expect or allow for with their default behavior.)

I'm not trying to dissuade your PR. If that provides a fix to everyone across all proxies that would be great of course.
I was not aware of your PR at the time of opening and was just trying to provide more visibility for, what was to me, the only fix currently available.

jneidel:
I removed the nginx config and still did not get logged out. So something else must have fixed it. Not sure what I could have been.

PR #5073, opened 2026-02-20
Improve personalized/discover query performance and cache invalidation behavior

kevingatera:
## Brief summary

This PR improves API read performance (especially personalized/discover), reduces cache churn from high-frequency writes, and adds targeted DB indexes for the discover query path.

## Which issue is fixed?

N/A (no linked issue yet)

## In-depth Description

This started from investigating slow home screen loads tied to personalized/discover.

What changed:

- Cache invalidation is now more targeted for high-churn models (`session`, `mediaProgress`, `playbackSession`, `device`) so frequent progress/session/device writes do not flush all API cache entries.
- Personalized shelf loading was parallelized where shelf queries are independent.
- Discover candidate queries were simplified to avoid expensive join-heavy filtering during candidate count/selection.
- DB indexes were added for hot discover filters:
  - `mediaProgresses(userId, mediaItemId, isFinished, currentTime)`
  - `bookSeries(seriesId, bookId)`

Search payload compatibility note:

- I reverted the search payload-minification change to preserve existing API behavior for third-party clients.
- With compatibility preserved, search responses remain large in expected cases (around 2.4 MB in my probe for `q=galaxy`).

One note: I was unsure whether to include a package version bump only to force migration execution. I kept the migration changes in this patchset, but if maintainers prefer no version bump in this type of PR, I can adjust to match the project release/versioning flow.

## How have you tested this?

I tested on a SQLite-backed deployment with repeated cold/warm probes and server-log timing checks, and then daily-drove the build for a couple of days.

Before vs after (representative)

- Cold personalized (`/api/libraries/:id/personalized?minified=1&include=rssfeed,numEpisodesIncomplete`)
  - Before targeted discover/query improvements:
    - personalized: 12.38s worst observed
    - discover shelf: 10.35s worst observed
  - After query-path + index work:
    - cold cycles observed: 1.05s, 2.73s, 0.22s
    - discover shelf in those cycles: 0.90s, 2.35s, 0.19s

- Additional direct API behavior after fixes
  - personalized cache-miss cases now often fall in sub-second to low single-digit seconds instead of prior double-digit spikes
  - warm personalized calls remain very fast

Correctness checks

- Verified key endpoints still return 200 with expected response structure:
  - `/api/libraries/:id/personalized`
  - `/api/libraries/:id/items`
  - `/api/libraries/:id/series`
  - `/api/libraries/:id/search`
- Verified search response keeps expanded shape (`libraryFiles` and `media` present on `libraryItem`) and observed ~2.4 MB payload in compatibility mode.
- Verified migration execution and index creation using `PRAGMA index_list/index_info`.
- Verified query planner usage with `EXPLAIN QUERY PLAN` (including use of the new mediaProgress index).

## Screenshots

N/A - backend/server/query performance changes only.

### DIFF
--- a/server/migrations/v2.32.2-add-discover-query-indexes.js
+++ b/server/migrations/v2.32.2-add-discover-query-indexes.js
@@ -14,7 +14,7 @@ const loggerPrefix = `[${migrationVersion} migration]`
 const indexes = [
   {
     table: 'mediaProgresses',
-    name: 'media_progress_user_item_finished_time',
+    name: 'media_progresses_user_item_finished_time',
     fields: ['userId', 'mediaItemId', 'isFinished', 'currentTime']
   },
   {
--- a/server/models/BookSeries.js
+++ b/server/models/BookSeries.js
@@ -48,6 +48,10 @@ class BookSeries extends Model {
           {
             name: 'bookSeries_seriesId',
             fields: ['seriesId']
+          },
+          {
+            name: 'book_series_series_book',
+            fields: ['seriesId', 'bookId']
           }
         ]
       }
--- a/server/models/MediaProgress.js
+++ b/server/models/MediaProgress.js
@@ -80,6 +80,10 @@ class MediaProgress extends Model {
         indexes: [
           {
             fields: ['updatedAt']
+          },
+          {
+            name: 'media_progresses_user_item_finished_time',
+            fields: ['userId', 'mediaItemId', 'isFinished', 'currentTime']
           }
         ]
       }


nichwall:
I have only glanced through the code, but initial things are:
- I don't think we want to change the API behavior of searches. The API is not fully documented and while this might work with the web client and first party beta apps, 3rd party apps may still expect the full payload. If this is not part of the home page optimization, it should probably be in its own PR.
- If possible we want to avoid adding explicit SQL syntax so we can eventually support other database systems (why we are using an ORM in the first place). We do it in other places, but want to move those instances to Sequelize instead of using Literals.

Yes, no version bump. We can update the script version number to prevent merge conflicts.

kevingatera:
Hi, good points on both items:
- I agree changing search payload shape can impact 3rd-party clients, so I can remove that from this PR and keep this one focused on home page/perf work.
- I also agree on avoiding explicit SQL where possible. And here too I can revise the current changes to reduce/avoid new raw literals and keep to Sequelize patterns where practical.

kevingatera:
@nichwall I've now made the changes, noting that going back to full payload in my case went back to a full 2.4MBs of data crossing the wire

advplyr:
I've been testing these changes and they are a great improvement. Since this includes a migration file I won't merge this until the day of release but it will be included

advplyr:
Thanks!

Issue #4628, opened 2025-08-27
[Bug]: Incorrect books identification and count after merging authors


Phantomwise:
### What happened?

There are kinda 2 issues, one with merging authors, and a related issue with deleting books from merged authors.

- In an old librrary, I had books from the same author for which one book used a different way to write the author's name than the other books ("Daphne Du Maurier" instead of "Daphne du Maurier", "S.A. Chakraborty" instead of "Shannon Chakraborty") in the embedded metadata
- On the "Authors" page, I had two entries for those authors.
- On the "Authors" page, I renamed the odd author entry into the same name than the other books used, so it was merged (renamed "Daphne Du Maurier" into "Daphne du Maurier", and "S.A. Chakraborty" into "Shannon Chakraborty").
- I didn't check if after the merge the remaining author listed all the books (tested later, it doesn't, see Additional Notes)
- Today I deleted the books in question (full delete, files included)
- The author page of those authors doesn't show any books, since they've been deleted:

<img width="642" height="542" alt="Image" src="https://github.com/user-attachments/assets/700726c9-afee-4fa8-8168-8e1f56caeaf5" />
<img width="660" height="539" alt="Image" src="https://github.com/user-attachments/assets/193c5bae-3d4c-436d-a778-ef32e1d28531" />

- But in the "Authors" view, the entry for those authors still displays "1 Books":

<img width="1366" height="768" alt="Image" src="https://github.com/user-attachments/assets/2a0c8e51-18af-4211-aec6-ea82926cdfbf" />

- I re-downloaded those books and added them to my new library
- Same issue with the author names
- I think I merged the authors (not certain so I did another test later, in case I misremember and had actually edited the author field manually for one instead, see Additional Notes)
- The remaining "Daphne du Maurier" author doesn't list the books from the other variation, but "Shannon Chakraborty" does.

### What did you expect to happen?

- When merging authors, the remaining author's page and book count includes books of the entry that was deleted during the merge
- The number of books by an author is the same whether it's in the author's page or in the general "Authors" view.
- Deleted books don't get counted

### Steps to reproduce the issue

1. Add two books with different authors
2. Go to the "Authors" view
3. Rename one of the authors into the exact name of the second author
4. Rescan manually if you want, it doesn't change anything
5. Check the entry for the remaining author in the "Authors" view

### Audiobookshelf version

v2.29.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

[...]

PR #5076, opened 2026-02-21
AuthorController: Update author name in match if changed

### LINKED ISSUE #5075: [Bug]: When matching author by ASIN, author name is not updated
### What happened?

1. I quick matched an author by name
2. I changed the name of the author and saved
3. I quick matched again
4. The author's name was not updated and "No updates necessary" toast was shown

I expected the name of the author to revert to the original name, since the match was by ASIN and the matched author had a different name from the saved one.

1. 

### Audiobookshelf version

v2.32.1

Built from source

Windows

Chrome

### Logs

```shell

```

### Additional Notes

_No response_

mikiher:
## Brief summary

When quick matching an author, a check is added if the matched author name is different from the saved author.

## Which issue is fixed?

Fixes #5075

The author name check was missing and therefore the author name was not updated as expected.
I added a diff check similar to other fields, and if there's a diff, it updates the author name and sets `hasUpdates` appropriately.

## How have you tested this?

Ran the repro steps in the bug. Now the expected update happens.

### DIFF
--- a/server/controllers/AuthorController.js
+++ b/server/controllers/AuthorController.js
@@ -368,6 +368,11 @@ class AuthorController {
       hasUpdates = true
     }
 
+    if (authorData.name && req.author.name !== authorData.name) {
+      req.author.name = authorData.name
+      hasUpdates = true
+    }
+
     if (hasUpdates) {
       await req.author.save()
 


advplyr:
I realize this is more complex actually. When updating the author name we also need to update `lastFirst`.

Mainly though we have to merge the author if another author is found with the same name. When looking at the code that does this in the update function I realized there is a bug because author names should only merge if they are in the same library.
I'm going to fix that bug now

advplyr:
I'm not sure how that bug has been able to survive for years. I just fixed it and found one issues referencing it: https://github.com/advplyr/audiobookshelf/issues/4628

I'm sure there are many related issues about this because it will mess up the displayed number of books count and all types of things.

mikiher:
> Mainly though we have to merge the author if another author is found with the same name. When looking at the code that does this in the update function I realized there is a bug because author names should only merge if they are in the same library.

Sorry, but I disagree. We can definitely suggest a merge, but why should we force it? There's also no uniqueness constraint in the DB. I am quite sure there are authors that share a name. I think this is a forced complication that should not be there.

I am aware that this likely requires some decisions to be made when editing authors, but these need to be resolved smartly, not by forcing a single author.

pwinnski:
> Sorry, but I disagree. We can definitely suggest a merge, but why should we force it?

That ship sailed long ago. Authors are combined based on name already. The bug was related to the same author name existing in different libraries. 

"We can definitely suggest a merge" implies a profound conceptual shift and an explosion of code and functionality.

mikiher:
Yeah, I discussed this with @advplyr, and he doesn't want client changes in the current code. We need to come up with a design and a plan to change the auto-merge behavior in the future.

PR #5078, opened 2026-02-21
fix: server-side comic page extraction with caching

### LINKED ISSUE #3505: [Bug]: Big comics are basically unusable at the moment
### What happened?

Since the server was fixed to be better at importing cbr/cbz comic books (no crashes anymore, thanks!), I have been trying to implement a comic library. 

The built-in reader works okay for smaller books (20, 30mb filesize), but as soon as files get over in the multiple hundred MB range, they either load infinitely (never opening), or if they open (I tested this on a 200mb file), it can take up to 30-60 seconds for a single page turn, while the whole system basically becomes unresponsive. 

This makes using the comic book feature basically unusable at the moment. 

This was the same across the App and the Web client, so I assume it is a backend issue. 

Smoother performance / working at all also for larger cb files

1. Add a big comic book (300+ mb) to your library and try to open it. 

When it (eventually) opens, try to scroll through more than a few pages. 

### Audiobookshelf version

v2.14.0

Windows Tray App

Windows

None

### Logs

_No response_

### Additional Notes

_No response_

mattmanning:
## Brief summary

Server-side comic page extraction with caching to fix slow page loads for large CBR/CBZ files.

## Which issue is fixed?

Fixes #3505

## In-depth Description

Large comic books (300+ MB) were causing 30-60 second page turns because the web client downloaded the entire archive before extracting pages client-side.

**Solution:**
- New `ComicCacheManager` extracts pages on-demand and caches to disk
- New endpoint `GET /api/items/:id/comic-pages` returns page list metadata
- New endpoint `GET /api/items/:id/comic-page/:page` returns individual pages
- Updated `ComicReader.vue` to fetch pages individually
- Browser preloads 2 pages ahead, 1 behind

**Performance:**
| Metric | Before | After |
|--------|--------|-------|
| Initial load | Entire file (300MB+) | Metadata (~1KB) |
| Page turn | Re-extract from memory | Single image (~2MB) |
| Browser memory | Entire archive | Current page only |

## How have you tested this?

1. Built Docker image, deployed to local Audiobookshelf (v2.32.1)
2. Tested with Chainsaw Man CBZ (189 pages)
3. Browser automation verified: pages load, navigation works, preloading works

**Environment:** Docker/Linux, Chrome

## Screenshots

Pages load instantly instead of 30-60s wait.

---
**Note:** Companion mobile app PR: advplyr/audiobookshelf-app#1809

### DIFF
--- a/server/Auth.js
+++ b/server/Auth.js
@@ -18,7 +18,11 @@ const { escapeRegExp } = require('./utils')
 class Auth {
   constructor() {
     const escapedRouterBasePath = escapeRegExp(global.RouterBasePath)
-    this.ignorePatterns = [new RegExp(`^(${escapedRouterBasePath}/api)?/items/[^/]+/cover$`), new RegExp(`^(${escapedRouterBasePath}/api)?/authors/[^/]+/image$`)]
+    this.ignorePatterns = [
+      new RegExp(`^(${escapedRouterBasePath}/api)?/items/[^/]+/cover$`),
+      new RegExp(`^(${escapedRouterBasePath}/api)?/authors/[^/]+/image$`),
+      new RegExp(`^(${escapedRouterBasePath}/api)?/items/[^/]+/comic-page/[0-9]+`)
+    ]
 
     /** @type {import('express-rate-limit').RateLimitRequestHandler} */
     this.authRateLimiter = RateLimiterFactory.getAuthRateLimiter()


server/controllers/LibraryItemController.js
```diff
@@ -1157,6 +1157,145 @@ class LibraryItemController {
...
+    } catch (error) {
+      Logger.error(`[LibraryItemController] Failed to get comic pages: ${error.message}`)
+      res.status(500).send('Failed to read comic file')
+    }
+  }
+
+  /**
+   * GET api/items/:id/comic-page/:page/:fileid?
+   * Get a single comic page (extracted and cached on server)
+   * Public endpoint (no auth required, like covers)
```
nichwall, line 1220: This should have authentication. Unauthenticated requests for cover images do leak information, but do not give access to any actual items or library files like this would.

nichwall:
Also as an FYI, the entire web front end is currently being rewritten and migrated to React, so front end changes will likely not be fully reviewed or merged until after that point. You're welcome to leave this open in the meantime.

This generally looks good from quickly glancing over it.

Issue #5079, opened 2026-02-23
[Enhancement]: Lock screen reflects "Use chapter track" setting


n8jadams:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

When the user in the web application selects the "Use chapter track" feature while listening to an .m4b file with chapter markers, the UI shows the listening progress in the context of the current chapter and not the entire audio file.

<img width="953" height="1153" alt="Image" src="https://github.com/user-attachments/assets/ca54c5bf-0e9d-4101-9aa3-2af68a53e355" />

However, in the lock screen the UI still shows the listening progress + tracker in the context of the entire audio file.

![Image](https://github.com/user-attachments/assets/e34d2cd2-3448-4bae-8dfb-a9eac0def78c)

### Why would this be helpful?

It brings consistency to the UX. Furthermore, it is too easy to accidentally scrub the listening position too far and totally lose your spot.

### Future Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/59e76ff5-f969-47e0-8b1c-e8a8d0d7d497)

### Audiobookshelf Server Version

v2.32.1

### Current Implementation (Screenshot)

![Image](https://github.com/user-attachments/assets/e34d2cd2-3448-4bae-8dfb-a9eac0def78c)

(Correctly tracking the listening in the context of the single chapter, not the full audio file.)

PR #5080, opened 2026-02-23
feat: MediaSession respects "Use chapter track" setting

n8jadams:
## Brief summary

- When "Use chapter track" is enabled in player settings, the Media Session API now reports the current chapter's duration/position instead of the full audiobook
- OS lock screen and notification center scrubbers now span only the current chapter
- Seek requests from OS controls are mapped back to the correct absolute position

## Which issue is fixed?

Implements https://github.com/advplyr/audiobookshelf/issues/5079

### The Problem

When playing an M4B audiobook with embedded chapters (e.g., a 12-hour book with 30 chapters), the OS media controls (lock screen, notification center, Control Center) show a scrubber spanning the entire 12 hours! This makes precise navigation nearly impossible - a tiny thumb movement might skip 30 minutes.

### The Solution

When the user enables `Use chapter track` in player settings, we intercept the Media Session API to report values relative to the chapter instead of to the full audio file.

### Technical Implementation

1. Position State Reporting (`updateMediaSessionPositionState`)

Instead of:
```javascript
setPositionState({ duration: 43200, position: 7200 })  // 12hr book, 2hr in
```

We calculate:
```javascript
chapterStart = 7000      // Chapter 5 starts at 1:56:40
chapterEnd = 8440        // Chapter 5 ends at 2:20:40
chapterDuration = 1440   // 24 minutes
chapterPosition = 200    // 3:20 into chapter

setPositionState({ duration: 1440, position: 200 })
```

2. Seek Mapping (`mediaSessionSeekTo`)

When the user drags the OS scrubber, the OS sends a seek request relative to what we reported. We map it back:

```javascript
// OS requests: "seek to 600 seconds" (10 min into the 24-min chapter)
// We translate: chapterStart + 600 = 7000 + 600 = 7600
audio.currentTime = 7600  // Correct absolute position
```

3. Metadata Updates (`updateMediaSessionForChapter`)

When chapters change, we update the displayed title to show the chapter name instead of the book title.

### Why This Approach?

This was the minimum code change necessary to get the consistent UX between the web app interface and the OS media controls. This also gets the UI/UX in consistency with other apps, such as Audible.

## How have you tested this?

1. Tested on an iPad using the iOS Safari.
2. Tested for regressions. During playback and in the lock screen, the full audio progress shows if the "Use chapter track" is disabled, or if the audio file doesn't have tracker markers.
3. Tested new functionality, enabled "Use chapter track", locked the screen, now only the current chapter appears. Scrubbing and skipping ahead/behind 10s works as expected. Chapters transition at the right time.

## Screenshots

Before while listening to chapter 1 of Fourth Wing:
audiobookshelf-screen-2

After listening to chapter 1 of Fourth Wing:
audiobookshelf-screen-3

### DIFF
--- a/client/components/app/MediaPlayerContainer.vue
+++ b/client/components/app/MediaPlayerContainer.vue
@@ -148,6 +148,10 @@ export default {
     currentChapter() {
       return this.chapters.find((chapter) => chapter.start <= this.currentTime && this.currentTime < chapter.end)
     },
+    useChapterTrack() {
+      const _useChapterTrack = this.$store.getters['user/getUserSetting']('useChapterTrack') || false
+      return this.chapters.length ? _useChapterTrack : false
+    },
     title() {
       if (this.playerHandler.displayTitle) return this.playerHandler.displayTitle
       return this.mediaMetadata.title || 'No Title'
@@ -291,6 +295,8 @@ export default {
     setPlaybackRate(playbackRate) {
       this.currentPlaybackRate = playbackRate
       this.playerHandler.setPlaybackRate(playbackRate)
+      // Update position state with new playback rate
+      this.updateMediaSessionPositionState()
     },
     seek(time) {
       this.playerHandler.seek(time)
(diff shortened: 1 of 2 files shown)


nichwall:
Just an FYI, but web client front-end changes are not actively being merged or reviewed right now because we are in the middle of migrating away from Vue to a new framework. You're welcome to leave this open in the meantime (either as inspiration or if someone else wants to use it in the meantime).

n8jadams:
I'll revisit this again once the React version is released. Thanks!

PR #5089, opened 2026-02-26
feat: add autocomplete attributes for password manager support

meek2100:
### Brief summary

Adds standard HTML `autocomplete` attributes to the login and initial server setup forms to improve support for browser-native and third-party password managers (e.g., Apple Keychain, 1Password, Bitwarden).

### Which issue is fixed?

Addresses the need for improved credential autofill support.

### In-depth Description

This PR updates the `ui-text-input` and `ui-text-input-with-label` components to accept an `autocomplete` prop, which is then bound to the underlying HTML `<input>` element.

The `login.vue` page has been updated to use these new props:

* **Initial Server Setup**: Sets `autocomplete="username"` and `autocomplete="new-password"` to help password managers suggest and save root credentials during setup.
* **Login**: Sets `autocomplete="username"` and `autocomplete="current-password"` to ensure saved credentials are correctly mapped and autofilled by the OS/browser.

This is a non-breaking change that follows standard Web accessibility and security practices for credential handling.

### How have you tested this?

1. **Manual Web Test**: Verified in a browser (Chrome/Firefox) that the "Save Password" prompt appears correctly after a successful login.
2. **Autofill Verification**: Confirmed that saved credentials appear as suggestions in the username field.
3. **Setup Test**: Verified that the password manager suggests a "Strong Password" for the `new-password` fields during the server initialization screen.


advplyr:
thanks!

Issue #5018, opened 2026-01-31
[Enhancement]: Postgres support


twsouthwick:
### Type of Enhancement

Server Backend

### Describe the Feature/Enhancement

Add PostgreSQL as an alternative database backend while keeping SQLite as the default.

#2046 already exists, but this a proposal that I wanted to discuss how to actually accomplish this. 

### Why would this be helpful?

In some hosting setups (such as kubernetes) it's easier to separate out the data concerns from the app so there's more control over the storage setup.

## Approach

This breaks the addition of PostgreSQL support to two steps: prepatory work and enablign postgres.

The existing codebase has some existing sqlite-specific code. To prepare for postgres, there are two approaches that could be taken to address this:

### Phase 1A: Dialect Abstraction (no behavior change)

- [ ] Create dialect helper module for SQL generation
- [ ] Refactor raw queries to use helpers instead of inline SQLite syntax
- [ ] Make migrations dialect-aware (conditionally skip SQLite-only operations)

**Example dialect specific changes:**

| Current (SQLite) | PostgreSQL equivalent |
|------------------|----------------------|
| `json_each(column)` | `jsonb_array_elements_text(column)` |
| `GROUP_CONCAT(col, ', ')` | `STRING_AGG(col, ', ')` |
| `COLLATE NOCASE` | `LOWER(column)` |
| `IFNULL(a, b)` | `COALESCE(a, b)` |

### Phase 1B: Schema Normalization (alternate approach)

Convert JSON blob columns to proper relational tables, eliminating dialect-specific JSON functions entirely.

- [ ] Create lookup tables for shared entities
- [ ] Create junction tables for many-to-many relationships
- [ ] Write migration to extract JSON data into new tables
- [ ] Update Sequelize models with proper associations
- [ ] Refactor queries to use JOINs instead of JSON functions

**JSON columns actively queried (require normalization):**

| Model | Column | New Tables |
|-------|--------|------------|
| `books` | `narrators` | `narrators` + `book_narrators` |
| `books` | `tags` | `tags` + `book_tags` |
| `books` | `genres` | `genres` + `book_genres` |
| `podcasts` | `tags` | `tags` + `podcast_tags` |
| `podcasts` | `genres` | `genres` + `podcast_genres` |

**JSON columns NOT queried (keep as-is):**

These columns store metadata blobs read/written as a whole and don't need SQL-level filtering:

| Model | Columns |
|-------|---------|
[...]

Issue #2046, opened 2023-08-31
[Enhancement]: PostgreSQL support


adocampo:
### Describe the feature/enhancement

It would be great to have PGSQL to speed up things and be more consistent. SQLite is fine for development, but it has many issues for production: it's slow and prone to corruption easily.

advplyr:
There have been a few discussions around this on Discord and #1584 #1202

Those were back when Abs was still using JSON files as a db. Basically we need an embedded database so that users aren't required to setup their own db. Once the new ORM (sequelize) is fully in place and the old data models get phased out we can work on supporting external databases like postgres.

Currently the migration to SQlite & sequelize is in-progress

advplyr:
Make backups. Docker is not the only install method. I don't think requiring an external db is a good UX

adocampo:
Certainly I do backups always I have to work with a sqlite DB, but I don't even trust them, I faced in the past the backups weren't working either, even doing them with the official method.... anyway, my suggestion is NOT to force the user to use any DB, but let them select (like nextcloud or home-assistant does, for example, which by default uses SQLite, so the user can just jump that step, or configure the external database if they want).
Speaking of UX, using a proper DB engine will boost the UX :P

FreedomBen:
I manage Postgres databases and huge distributed web apps for my day job, and I think SQLite is a much, much better UX for the vast number of users who aren't going to scale vertically.  Having the option for postgres would be cool, but SQLite should stay the default.  Just my opinion of course.  Even for people needing to serve a lot of users, you can vertically scale incredibibly far with a SQLite database, and backups are as simple as an rsync.

adocampo:
I can agree with all you say. Many people will find SQLite more than enough and probably won't use another engine and will be annoyed if they are forced to configure anything additional.

But I have a question regarding when you say UX. That means "User eXperience", and I can't see why a fastest SQL engine can degrade the user experience. Because PGSQL provides excellent performance thanks to its advanced indexing, query optimization/planning, parallelization, caching mechanisms, and multi-core scaling capabilities.

[...]

Issue #1584, opened 2023-03-05
[Enhancement]: Are you open to a PR for MongoDB?


RXWatcher:
### Describe the feature/enhancement

We love ABS but it's having issues with performance on over 100K audio books. Our goal is 300K and there is no way ABS will handle that quantity in its current form. We feel that the DB used is likely a large bottleneck. If we were to supply code to move everything to mongodb, would you be open to accepting the PR?

We can fork and go out on our own but that 1) Doesnt help the community 2) We'd have to continue to merge your chances into our repo. I dont think either is ideal so I'd rather the mongodb code go into the public repos.

I feel this is a win-win for both parties.

Please let me know your thoughts

advplyr:
The DB is something we have been talking about upgrading for a while. The hesitation with MongoDB would be its ability to be embedded/portable. Did you have a method in mind for implementing MongoDB?

RXWatcher:
I'll have to think about that. Mongodb came to mind because of it's no-sql structure and being fairly straight forward when moving from json files. Let me give it some thought but I appreciate you not slamming the door on the idea.

advplyr:
I like MongoDB and it would be great if this was an embeddable option because it would save a lot of time migrating.

The bottleneck of the current db is likely because we are loading most of the data into memory on server startup. If we switch to a standard db solution then we would fetch that data when necessary instead of filling up memory.
For small/medium sized libraries the current db will be faster then other standard dbs because it is just reading/writing to a file, no indexing. There is a bigger problem with the current db in that if a file write gets interrupted then it can result in data loss.

I tested another no-sql db that was built to be embedded ([LiteDB](https://github.com/mbdavid/LiteDB)) and the performance was not good. My last search for embedded no-sql dbs was about a year ago so maybe something has come along.

I realized we will probably need to setup a relational model and use sqlite like everyone else. With that we could eventually offer users the ability to use their own external db instead of the embedded one.
I'm still open to anything here though.

rasmuslos:
I would suggest using  [knex](https://knexjs.org/) to handle data. This way the end user could decide for themselves what database they want to use. ABS could come preconfigured with SQLite but with an easy way to setup Postgres for big libraries like this one. Knex supports PostgreSQL, MariaDB, SQLite, and some more.

RXWatcher:
Yeah, ideally Id want postgres but thought the leap from the current database to a sql database was too much of an ask.

advplyr:
[...]

PR #5098, opened 2026-03-03
Add optional PostgreSQL backend support with migration safeguards and regression coverage

kevingatera:
## Why

Audiobookshelf currently defaults to SQLite, which is simple for most installs, but some deployments need an external DB backend for reliability and infrastructure fit.

This PR adds PostgreSQL as an optional backend while preserving SQLite as the default and maintaining backward compatibility.
It also includes targeted safety fixes for real-world PostgreSQL edge cases discovered during migration and testing.

### 1) Optional PostgreSQL backend support (SQLite remains default)
- Add dialect and config selection:
  - `DB_DIALECT=sqlite|postgres` (optional)
  - `DATABASE_URL` for PostgreSQL
- Keep SQLite behavior unchanged when PostgreSQL is not configured.
- Add PostgreSQL dependencies (`pg`, `pg-hstore`).

### 2) Dialect-aware query compatibility
- Introduce `server/utils/sqlDialectHelpers.js` for cross-dialect SQL generation.
- Refactor key query paths to avoid SQLite-specific syntax in PostgreSQL, including:
  - case-insensitive sort behavior
  - JSON array and value operations
  - numeric casting behavior for text-backed fields
- Guard sequence and year sorting with safe casts to avoid runtime failures from non-numeric legacy values.

### 3) Migration and schema-safety hardening
- Add `server/scripts/migrateSqliteToPostgres.js` for SQLite to PostgreSQL data migration.
- Add preflight checks to prevent data loss:
  - detect overlong varchar source values
  - detect non-integer source values for integer target columns
- Harden JSON normalization for malformed and double-encoded payloads.
- Improve migration metadata handling and case-insensitive table and column mapping for PostgreSQL.

### 4) PostgreSQL runtime bug fixes found during validation
- Fix legacy user lookup behavior for PostgreSQL (`oldUserId` path).
- Fix media progress `updatedAt` sync path to avoid raw SQL table-name mismatches.
- Fix collapse-series edge cases:
  - empty include list (`IN ()`) generation
  - aliasing issues in PostgreSQL SQL
  - display title fallback alias compatibility

### 5) Focused regression tests (high signal)
Added tests for the exact risky paths that caused failures:
- `test/server/utils/queries/libraryItemsBookFilters.test.js`
  - collapse-series empty ID set
  - safe ID quoting
  - PostgreSQL join alias safety
  - failure-path query diagnostics
- `test/server/models/MediaProgress.test.js`
  - valid and invalid `lastUpdate` handling
- `test/server/models/User.test.js`
  - UUID and legacy old-id lookup behavior on PostgreSQL and SQLite
- `test/server/scripts/migrateSqliteToPostgres.test.js`
  - JSON normalization and integer and varchar preflight checks
- `test/server/utils/sqlDialectHelpers.test.js`
  - dialect helper correctness

## Backward compatibility

- SQLite remains the default behavior.
- Existing SQLite users are unaffected unless PostgreSQL is explicitly configured.
- PostgreSQL code paths are opt-in and covered by targeted regression tests.

## Validation

Local:
- `npm test` passed (`356 passing`)
- targeted compatibility suite passed (`69 passing`)

CI (branch):
- Unit tests passed
- Integration tests passed

Observed runtime impact during migration validation (same host, from app logs):
- Discover load time improved from about `5.23s` (SQLite period average) to about `0.08s` (PostgreSQL period average).
- Personalized shelves load time improved from about `7.35s` to about `0.12s`.

## Notes

- This PR intentionally focuses on safe, minimal compatibility and migration behavior.
- It does not force external DB usage. PostgreSQL is optional.

Beta image tag for testing:
- `ghcr.io/kevingatera/audiobookshelf:homelab-postgres-5871b01f6dd9f1cd690b9a70bdc227a2d5db2643`

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -20,9 +20,13 @@ on:
       - index.js
       - package.json
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   build:
-    if: ${{ !contains(github.event.head_commit.message, 'skip ci') && github.repository == 'advplyr/audiobookshelf' }}
+    if: ${{ !contains(github.event.head_commit.message, 'skip ci') && (github.repository == 'advplyr/audiobookshelf' || github.repository == 'kevingatera/audiobookshelf') }}
     runs-on: ubuntu-24.04
 
     steps:
@@ -53,6 +57,7 @@ jobs:
             ${{ runner.os }}-buildx-
 
       - name: Login to Dockerhub
+        if: ${{ github.repository == 'advplyr/audiobookshelf' }}
         uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
@@ -63,7 +68,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
-          password: ${{ secrets.GHCR_PASSWORD }}
+          password: ${{ github.token }}
 
       - name: Build image
         uses: docker/build-push-action@v6
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,7 @@ ARG NUSQLITE3_PATH
 RUN apk add --no-cache --update \
   tzdata \
   ffmpeg \
+  postgresql-client \
   tini
 
 WORKDIR /app
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,8 @@
         "p-throttle": "^4.1.1",
         "passport": "^0.6.0",
         "passport-jwt": "^4.0.1",
+        "pg": "^8.19.0",
+        "pg-hstore": "^2.3.4",
         "semver": "^7.6.3",
         "sequelize": "^6.35.2",
         "socket.io": "^4.5.4",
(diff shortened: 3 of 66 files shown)


kevingatera:
Related context for reviewers:

- PostgreSQL support request tracking: https://github.com/advplyr/audiobookshelf/issues/2046
- Newer duplicate and request thread: https://github.com/advplyr/audiobookshelf/issues/5018
- Earlier external database backend discussion: https://github.com/advplyr/audiobookshelf/issues/1584
- Prior compatibility-related PR (ID handling groundwork): https://github.com/advplyr/audiobookshelf/pull/3600

This PR keeps SQLite as the default, adds optional PostgreSQL support, and includes regression coverage for the edge cases hit during validation (JSON normalization, numeric cast safety, aliasing, legacy oldUserId lookup, and migration safety preflight checks).

kevingatera:
I pushed a tightening pass focused on reviewer-risk areas and sqlite/postgres parity.

Changes in this pass:
- Added dialect/default and table-detection coverage in `test/server/Database.test.js`.
- Added podcast and series query parity coverage in:
  - `test/server/utils/queries/libraryItemsPodcastFilters.test.js`
  - `test/server/utils/queries/seriesFilters.test.js`
- Expanded postgres-specific migration manager coverage in `test/server/managers/MigrationManager.test.js`.
- Added uppercase UUID regression coverage in `test/server/models/User.test.js`.
- Reduced collapse-series fallback log noise (`warn` to `debug`) in `server/utils/queries/libraryItemsBookFilters.js`.

kevingatera:
This thread is also relevant to the design choices here: https://github.com/advplyr/audiobookshelf/discussions/5070

I share the concern about keeping migration/model complexity under control if both backends are supported.

That is why this PR tries to keep the blast radius small:
- SQLite stays the default path unless Postgres is explicitly enabled.
- Postgres is opt-in only (`DB_DIALECT` and `DATABASE_URL`).
- Dialect differences are kept in helper/query layers with targeted regression tests, instead of splitting core model behavior.
- Migration execution is guarded with preflight checks and dry-run-first behavior.

If this is still too broad for a first step, I can split follow-up work into smaller PRs (core dialect/query parity first, migration tooling after).

Chrtol:
Any update on this?

kevingatera:
I've actually been running this for months to see how things go annd what breaks. There will be an update soon

kevingatera:
Sorry for the thrashing in here. This has been an open PR for a while and I've had to rebase it onto master a few times to keep it mergeable, which is why the commit list kept shifting. It should be settled now: rebased onto latest master (through v2.36.0), and I fixed a batch of things I ran into while daily driving this on postgres.

The main addition is backups. Postgres mode now uses `pg_dump` custom format, stored as `absdatabase.postgres.dump` inside the backup zip, and restore picks the right path based on what's in the archive. The SQLite path is untouched. The db password goes through `PGPASSWORD` instead of argv, the dump/restore commands have a timeout so a hung connection can't stall the backup scheduler, and the backup listing now skips a corrupt archive instead of throwing on the whole page.

I also hit two upstream migrations (v2.33.0 and v2.35.0) that compare index/column names case-sensitively when they introspect the schema. That silently no-ops or outright crashes on postgres, since unquoted identifiers fold to lowercase. Both are patched, but fair warning that any future migration reading `information_schema` will need the same treatment.

Smaller fixes: postgres triggers now maintain the denormalized library item columns (SQLite gets this from the app layer), user cache lookups no longer miss on mixed-case usernames/emails, and a shelf query fallback is scoped to postgres only so the SQLite code path stays exactly as it was.

For reviewers: I tried to put the why behind each change in the commit messages, especially where the causation isn't obvious from the diff (the case-folding bugs, for example). Not every commit has that level of detail, but the non-obvious ones should.

Still running this branch against postgres 14 in production, same as the last few months. Redeployed the rebased head there to be sure: migrations applied clean, triggers got created, and I checked the pg_dump output actually restores with `pg_restore`.

Known limits are the same as before: the migrator needs a superuser role, and it loads some tables fully into memory.

binyaminyblatt:
Please consider merging this!
For anyone managing a large library, Audiobookshelf is known to suffer from slow database queries and high latency on library/home views. Adding native PostgreSQL support would provide a massive performance boost, while solving critical infrastructure needs:

- **Query Performance & Throughput**: Eliminates SQLite query bottlenecks and locks during library scans, search, and multi-user streaming.

- **High Availability**: Decouples state from the application host—if the machine running the server crashes, the container can fail over immediately without data corruption.

- **Replication & Resilient Backups**: Enables native streaming replication and continuous off-site backups without needing to lock or take down the app.

Really appreciate all the work put into this PR!

binyaminyblatt:
I wanted to try this so I setup the docker container I couldn't figure out how to migrate so I logged into the docker container with a shell to run the script that needs to be a little easier and then when I ran it I got
```
client             index.js           node_modules       package-lock.json  package.json       server
/app # ./server/
Auth.js             Server.js           auth/               libs/               models/             routers/            utils/
Database.js         SocketAuthority.js  controllers/        managers/           objects/            scanner/
Logger.js           Watcher.js          finders/            migrations/         providers/          scripts/
/app # ./server/
Auth.js             Server.js           auth/               libs/               models/             routers/            utils/
Database.js         SocketAuthority.js  controllers/        managers/           objects/            scanner/
Logger.js           Watcher.js          finders/            migrations/         providers/          scripts/
/app # ./server/scripts/migrateSqliteToPostgres.js 
ash: ./server/scripts/migrateSqliteToPostgres.js: Permission denied
/app # node ./server/scripts/migrateSqliteToPostgres.js 
[migrate] sqlite source: /config/absdatabase.sqlite
[migrate] postgres target schema: public
[migrate] dry run: false
[migrate] tables to migrate: migrationsMeta, settings, users, apiKeys, sessions, libraries, libraryFolders, authors, series, books, podcasts, podcastEpisodes, libraryItems, bookAuthors, bookSeries, collections, collectionBooks, playlists, playlistMediaItems, mediaProgresses, devices, playbackSessions, feeds, feedEpisodes, mediaItemShares, customMetadataProviders
[migrate] overlong source values detected for varchar columns:
[migrate]   podcasts.imageurl -> podcasts.imageurl (max=255, actualMax=283, overRows=2)
[migrate]   podcastEpisodes.subtitle -> podcastepisodes.subtitle (max=1000, actualMax=2548, overRows=189)
[migrate]   podcastEpisodes.enclosureurl -> podcastepisodes.enclosureurl (max=255, actualMax=358, overRows=47)
[migrate] failed: Error: Migration aborted to prevent truncation/data loss. Widen target column types first.
    at main (/app/server/scripts/migrateSqliteToPostgres.js:362:13)
```
so I changed ran 
```sql
ALTER TABLE podcasts ALTER COLUMN imageurl TYPE TEXT;
ALTER TABLE podcastepisodes ALTER COLUMN subtitle TYPE TEXT;
ALTER TABLE podcastepisodes ALTER COLUMN enclosureurl TYPE TEXT;
```
then I got 
```
/app # node ./server/scripts/migrateSqliteToPostgres.js 
[migrate] sqlite source: /config/absdatabase.sqlite
[migrate] postgres target schema: public
[migrate] dry run: false
[migrate] tables to migrate: migrationsMeta, settings, users, apiKeys, sessions, libraries, libraryFolders, authors, series, books, podcasts, podcastEpisodes, libraryItems, bookAuthors, bookSeries, collections, collectionBooks, playlists, playlistMediaItems, mediaProgresses, devices, playbackSessions, feeds, feedEpisodes, mediaItemShares, customMetadataProviders
[migrate] failed: Error: Migration writes are destructive. Set ALLOW_DESTRUCTIVE_TARGET=true after confirming the target database can be truncated.
    at main (/app/server/scripts/migrateSqliteToPostgres.js:378:15)
```
then when I ran it with `ALLOW_DESTRUCTIVE_TARGET=true` I got 

```
/app # ALLOW_DESTRUCTIVE_TARGET=true node ./server/scripts/migrateSqliteToPostgres.js 
[migrate] sqlite source: /config/absdatabase.sqlite
[migrate] postgres target schema: public
[migrate] dry run: false
[migrate] tables to migrate: migrationsMeta, settings, users, apiKeys, sessions, libraries, libraryFolders, authors, series, books, podcasts, podcastEpisodes, libraryItems, bookAuthors, bookSeries, collections, collectionBooks, playlists, playlistMediaItems, mediaProgresses, devices, playbackSessions, feeds, feedEpisodes, mediaItemShares, customMetadataProviders
[migrate] failed: error: permission denied to set parameter "session_replication_role"
    at /app/node_modules/pg/lib/client.js:652:17
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (/app/server/scripts/migrateSqliteToPostgres.js:382:7) {
  length: 126,
  severity: 'ERROR',
  code: '42501',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'guc.c',
  line: '3604',
  routine: 'set_config_with_handle'
}
```
I had to make the user a superuser to make it work this is a problem just because I could make the user a superuser doesn't mean everyone can

nichwall:
I don't think we're ready to merge this yet. We have a large number of open PRs that we need to get a handle on and merging this would impact a lot of other PRs that include database migrations, along with additional migrations I'm already planning to improve performance for a number of queries.

Merging this would require testing database migrations for two different database backends, which adds to the complexity of addressing these changes.

binyaminyblatt:
I started testing this and I don't think its as ready as it needs to be to be merged see my earlier comment about what I found I look forward to when this is ready but until then I recommend making this a draft

PR #5104, opened 2026-03-05
Per-book playback speed

### LINKED ISSUE #1173: [Enhancement]: Save other player configurations along with progress.

I usually listen to 2 or more titles at a time, and the speed for every title differs. Coming from maBook, I miss this feature where it would save the player preferences for each book along with the progress statistic. This could be a server-side feature or just for client, that is up to your discretion. I think this would be more sensible if it were to be client-side, since every client has a different use case, one might listen to on phone with different speed/loudness than may their tv or something.

For now, only the listening speed is the player configuration that could be stored for every title, but in future if you add more configuration, for example, **loudness** or **equalizer**, that could also benefit from this design choice as those settings would also be unique to every book.

octopotato:
## Brief summary

Saves playback speed per book/episode so each title remembers its own narration speed. Also fixes keyboard shortcuts (`Shift+Arrow`) not persisting playback rate at all.

## Which issue is fixed?

- Fixes #1173

## In-depth Description

Stores playback rate per item in `mediaProgress.extraData.playbackRate`. When a book/episode is opened, it checks for a saved per-item rate and falls back to the global user setting if there isn't one.

No migration needed — `playbackRate` is just a new optional key in the existing `extraData` JSON column. Nothing changes for existing users until they start setting per-item speeds.

The keyboard shortcut fix is separate but related — `increasePlaybackRate`/`decreasePlaybackRate` bypassed `playbackRateChanged`, so they never persisted at all (pre-existing bug).

> [!NOTE]
> This does not add a dedicated setting in the player settings for changing the "global default speed." The last speed change on any item is treated as the new global default for items without a per-item rate. A dedicated global rate dropdown in the player settings modal would be a nice follow-up but I kept it out of this PR to keep the change contained.

## How have you tested this?

- Set different speeds on two books, switched between them — each keeps its own rate
- Hard refreshed — rates persist from the database
- Played a book with no saved rate — falls back to global setting correctly
- Tested Shift+ArrowUp/Down — rate now saves
- Tested with podcast episodes (Planet Money) — per-episode speed persists correctly
- Full test suite passes (324 tests including 9 new MediaProgress model tests)

## Screenshots

https://github.com/user-attachments/assets/c1087a79-e24d-40d9-9e6f-e7ba7c4fcd0b

### DIFF
--- a/test/server/models/MediaProgress.test.js
+++ b/test/server/models/MediaProgress.test.js
@@ -1,0 +1,102 @@
+const { expect } = require('chai')
+const sinon = require('sinon')
+
+const { Sequelize } = require('sequelize')
+const Database = require('../../../server/Database')
+const Logger = require('../../../server/Logger')
+
+describe('MediaProgress', () => {
+  let mediaProgress
+
+  beforeEach(async () => {
+    global.ServerSettings = {}
+    Database.sequelize = new Sequelize({ dialect: 'sqlite', storage: ':memory:', logging: false })
+    Database.sequelize.uppercaseFirst = (str) => (str ? `${str[0].toUpperCase()}${str.substr(1)}` : '')
+    await Database.buildModels()
+
+    const user = await Database.userModel.create({ username: 'testuser', type: 'root' })
+    const library = await Database.libraryModel.create({ name: 'Test Library', mediaType: 'book' })
+    const libraryFolder = await Database.libraryFolderModel.create({ path: '/test', libraryId: library.id })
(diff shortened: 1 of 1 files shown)


nichwall:
Thanks for opening a PR.
For your awareness, the frontend is currently being fully rewritten and migrated to React, so we are not reviewing PRs that modify the existing web client at this time. You are welcome to leave this PR open in the meantime, and we may use it as inspiration after the new web client is ready or you can update this PR at that time.

octopotato:
Ah I wasn't aware of that, my bad! I'll split out the server side changes since that's _really_ what I'm after since I was trying to implement this in a 3rd party app that way at least the API will have the support and the web client can be implemented once the migration to React is done.

PR #5106, opened 2026-03-05
Add playbackRate to MediaProgress extraData

### LINKED ISSUE #1173: [Enhancement]: Save other player configurations along with progress.

octopotato:
## Brief summary

I added `playbackRate` to `mediaProgress.extraData` so each book/episode can store its own playback speed server-side. Any client (web, mobile, third-party) can use it.

## Which issue is fixed?

- Partially fixes #1173. Just the server changes here, the web client will need to wait until the React migration is done (see #5104)

## In-depth Description

I'm storing playback rate per item in `mediaProgress.extraData.playbackRate`. The existing `PATCH /api/me/progress/:libraryItemId/:episodeId?` endpoint accepts `playbackRate` in the payload, and `GET /api/me/progress` returns it in the response.

I'm already using this on the iOS side in [AudioBooth](https://github.com/AudioBooth/AudioBooth/pull/167) for per-book narration speed. The web client implementation is in #5104, kept separate since the frontend is being rewritten to React.

## How have you tested this?

- Verified `playbackRate` persists in the database via `PATCH /api/me/progress/:id`
- Verified it's returned in `getOldMediaProgress()` serialization
- unit tests

## Screenshots

```bash
# 1. Check current playbackRate (null, no per-book rate set)
$ curl -s "/api/me/progress/$ITEM" | jq .playbackRate
null

# 2. Set playbackRate to 1.5x
$ curl -X PATCH "/api/me/progress/$ITEM" -d '{"playbackRate": 1.5}'
HTTP 200

# 3. Read it back
$ curl -s "/api/me/progress/$ITEM" | jq .playbackRate
1.5

# 4. Update to 2.0x
$ curl -X PATCH "/api/me/progress/$ITEM" -d '{"playbackRate": 2.0}'
HTTP 200

# 5. Confirm update
$ curl -s "/api/me/progress/$ITEM" | jq .playbackRate
2

# 6. Verify other progress fields are unaffected
$ curl -s "/api/me/progress/$ITEM" | jq "{playbackRate, currentTime, progress, isFinished}"
{
  "playbackRate": 2,
  "currentTime": 107.001282,
  "progress": 0.015263561750603215,
  "isFinished": false
}
```

### DIFF
--- a/server/models/MediaProgress.js
+++ b/server/models/MediaProgress.js
@@ -169,6 +169,7 @@ class MediaProgress extends Model {
       hideFromContinueListening: !!this.hideFromContinueListening,
       ebookLocation: this.ebookLocation,
       ebookProgress: this.ebookProgress,
+      playbackRate: this.extraData?.playbackRate || null,
       lastUpdate: this.updatedAt.valueOf(),
       startedAt: this.createdAt.valueOf(),
       finishedAt: this.finishedAt?.valueOf() || null
@@ -207,6 +208,12 @@ class MediaProgress extends Model {
       // Old model stored progress on object
       this.extraData.progress = Math.min(1, Math.max(0, progressPayload.progress))
       this.changed('extraData', true)
+    }
+
+    if (progressPayload.playbackRate !== undefined) {
+      this.extraData.playbackRate = progressPayload.playbackRate
+      this.changed('extraData', true)
+      delete progressPayload.playbackRate
     }
 
     this.set(progressPayload)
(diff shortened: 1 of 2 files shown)


octopotato:
Converting to draft. Got some feedback that it would make more sense to wait until the data model is finished so that it has a dedicated API

PR #5112, opened 2026-03-09
Add progress stall for ffmpeg podcast downloads

### LINKED ISSUE #4116: [Bug]: Failed/hanged podcast episode download blocking download queue forever
### What happened?

When a download stuck for some reason (I suppose due to network issue or host-side issue), the "downloading episode x" status will hang forever and thus block all following download task, until a restart of abs. 

Solutions I can think of: 
1. A timeout mechanism for downloading tasks, retry/abort the task when the download progress does not change for a certain time. Maybe move the retry task to the bottom of the download queue to prevent dead loop, or set a max retry count. 
2. An option to manually retry/abort download tasks. 

Happened quite randomly in my case, could be any episode download task. More likely when I try to download a bunch of past episodes of one podcast at once, maybe some host have rate limit.

### Audiobookshelf version

v2.19.5

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

Vito0912:
## Brief summary

See Discord below. Just a quick fix, nothing major

## Which issue is fixed?

It *should* fix: https://discord.com/channels/942908292873723984/954760207131615264/1480565704670314516
fixes #4116 

## In-depth Description

After Axios receives a chunk, the timeout is ignored. ffmpeg waits forever. So if the stream breaks, it will stall forever. Because `.on('progress')` is a constant that should be called at least once a second if active data is being streamed, if there is no progress for 60 seconds, it sees the podcast as stalled.

## How have you tested this?

No. This is only theoretical. I made sure it will not break downloads (I hope).

### DIFF
--- a/server/utils/fileUtils.js
+++ b/server/utils/fileUtils.js
@@ -297,6 +297,8 @@ module.exports.getFilePathItemFromFileUpdate = (fileUpdate) => {
  */
 module.exports.downloadFile = (url, filepath, contentTypeFilter = null) => {
   return new Promise(async (resolve, reject) => {
+    const DOWNLOAD_STALL_TIMEOUT_MS = 60000
+
     Logger.debug(`[fileUtils] Downloading file to ${filepath}`)
     axios({
       url,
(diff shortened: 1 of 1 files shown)


nichwall:
I believe this should also fix https://github.com/advplyr/audiobookshelf/issues/4116

Vito0912:
I just took a very short look at the linked issue, but I don't think so. This is realted to ffmpeg which should have printed progress logs. I didn't see any in the linked issue.

I will add a fix later for that too.

Vito0912:
Now it should also apply to the downloads mentioned in the issue

Issue #5114, opened 2026-03-10
[Bug]: Cannot play mka+opus file on desktop client


rktjmp:
### What happened?

Trying to play a `.mka` file with `opus` streams inside doesn't play anything and logs an error to the console.

### What did you expect to happen?

The file should play as normal.

### Steps to reproduce the issue

1. Add a `mka` container file with `opus` streams inside it ([Public domain sample file, 9mb](https://files.catbox.moe/71hpeh.mka)) (cant directly add .mka files to this issue).
2. Play the file on the desktop client
3. Observe the play button stuck spinning, see the following error in console:

```
[HLS] Error mediaError fragParsingError 
Object { type: "mediaError", details: "fragParsingError", fatal: true, error: Error, frag: {…}, reason: 'Found no media in msn 8 of level "https://domain/audiobookshelf/hls/decf9195-38ed-406a-af06-b75f800f851d/output.m3u8"', errorAction: {…} }
​
details: "fragParsingError"
​
error: Error: Found no media in fragment 8 of level 0 resetting transmuxer to fallback to playlist timing
​
errorAction: Object { action: 2, flags: 1, retryCount: 8, … }
​
fatal: true
​
frag: Object { _byteRange: Getter & Setter, _url: Getter & Setter, baseurl: Getter & Setter, … }
​
reason: 'Found no media in msn 8 of level "https://domain/audiobookshelf/hls/decf9195-38ed-406a-af06-b75f800f851d/output.m3u8"'
​
type: "mediaError"
​
<prototype>: Object { … }
5ac40cc.js:1:92627
```

### Audiobookshelf version

v2.32.1

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell

```

mihailim:
Oddly enough, mp4+opus works _perfectly_, both with the browser client on Linux (Firefox and Chromium-derived) and with the Android client.
(Please, please don't force transcoding for mp4+opus :D)

Vito0912:
> Oddly enough, mp4+opus works perfectly

That's not really oddly at all. Opus is supported in mp4/m4a/m4b and all these mp4 extensions.

[...]

PR #5115, opened 2026-03-10
Force AAC transcode when streaming mka+opus to desktop client

rktjmp:
## Brief summary

Support streaming Matroska audio containers with Opus streams inside them to the desktop client.

## Which issue is fixed?
https://github.com/advplyr/audiobookshelf/issues/5114
## In-depth Description

Matroska audio containers (aka `mka` files) with Opus codec streams inside were unplayable on the desktop client because hls.js was unable to decode the stream, resulting in an infinitely "spinning" play button.

When configuring a stream, we now check for the `opus` codec and force AAC transcoding.

Matroska containers support other codecs besides Opus, eg: mp3, which do not require transcoding and work fine before this patch, which is why we check for `opus` in `codecsToForceAAC` instead of `AudioMimeType.MKA` in `mimeTypesToForceAAC`.

The `AudioMimeType.OPUS` mimetype is already marked as requiring transcoding but since its inside a container this check does not evaluate to true, we must check the codec explicitly.

## How have you tested this?

By using these sample files created from a public domain recording:

-  [mka+opus, 9mb](https://files.catbox.moe/71hpeh.mka)
- [mka+mp3, 35mb](https://files.catbox.moe/poa4ag.mka)

### DIFF
--- a/client/players/LocalAudioPlayer.js
+++ b/client/players/LocalAudioPlayer.js
@@ -47,12 +47,18 @@ export default class LocalAudioPlayer extends EventEmitter {
     this.player.addEventListener('timeupdate', this.evtTimeupdate.bind(this))
 
     var mimeTypes = [
-      'audio/flac', 'audio/mpeg', 'audio/mp4', 'audio/ogg', 'audio/aac',
-      'audio/x-ms-wma', 'audio/x-aiff', 'audio/webm',
-      // `audio/matroska` is the correct mimetype, but at least as of 2026-04-09,
-      // the detected mimetype for matroska files by the server is `audio/x-matroska`.
+      'audio/flac',
+      'audio/mpeg',
+      'audio/mp4',
+      'audio/ogg',
+      'audio/aac',
+      'audio/x-ms-wma',
+      'audio/x-aiff',
+      'audio/webm',
+      // `audio/matroska` is the correct mimetype, but the server still uses `audio/x-matroska`
       // ref: https://www.iana.org/assignments/media-types/media-types.xhtml
-      'audio/matroska', 'audio/x-matroska'
+      'audio/matroska',
+      'audio/x-matroska'
     ]
     var mimeTypeCanPlayMap = {}
     mimeTypes.forEach((mt) => {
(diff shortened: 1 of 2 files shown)


Vito0912:
I mentioned this in the linked issue and while I think the added change should be kept as opus is not supported in hls, that does not solve the problem.

opus in mka is supported, so there shouldn't be any need to have transcoding. If you send

```json
{"deviceInfo":{"clientName":"Abs Web","deviceId":"xxx"},"supportedMimeTypes":["audio/flac","audio/mpeg","audio/mp4","audio/ogg","audio/aac","audio/webm","audio/x-matroska"],"mediaPlayer":"html5","forceTranscode":false,"forceDirectPlay":false}
```

See that `audio/x-matroska` was added. With that it does serve the correct file as opus is supported on many device inside mka
https://en.wikipedia.org/wiki/Opus_(audio_format)

JFYI so maybe you could add advertisement of x-matroska in https://github.com/advplyr/audiobookshelf/blob/64cbf596094e4d51fc47cb75af7f0c35ede76816/client/players/LocalAudioPlayer.js#L49 for this PR too.

I am not sure if that does work as `x-` suggests a non standard mimeType, so testing would be needed, but this would really fix the underlying issue.

<sub>I am just a contributor, no maintainer</sub>

rktjmp:
I have added the update to `LocalAudioPlayer`. `audio/matroska` is the correct type but the server returns `audio/x-matroska`, I think because it uses an old ffmpeg implementation? I added both types for future-proofing I guess.

Playback works fine in Firefox and Chrome falls back to transcoding as it does not support matroska.

When playing the `mka+mp3` file (matroska container with mp3 stream), both browsers fall back to HLS streaming via ffmpeg `copy`. This likely a pretty rare combination though and it seems *fine*.

Vito0912:
@advplyr Can you check? It's a relatively simple fix

advplyr:
The `x-matroska` comes from Abs in `utils/constants.js`.`AudioMimeType`.

I set that up 3 years ago before `audio/matroska` was registered. I'm not sure if it would be a breaking change if I switch that over.

Both of those sample mka files you provided direct play in chrome & firefox on Windows 11. 
Matroska would always transcode before adding it in the mimeTypes array in LocalAudioPlayer. That was an oversight.

Thanks!

PR #5126, opened 2026-03-16
Add CAMB AI TTS integration for ebook-to-audiobook synthesis

neilruaro-camb:
## Summary

- Integrate [CAMB AI](https://camb.ai) text-to-speech API to synthesize audiobooks from ebook content
- Add `TtsManager` for orchestrating chapter-level TTS jobs (polling, retries, progress tracking) and `TtsController` for the REST endpoints
- Add `textExtractor` utility to pull chapter text from EPUB files
- Wire up three new API routes: start a TTS task, check status, and list available voices
- Include unit tests for all new modules and a manual testing guide

## New dependencies

- `epub2` – EPUB parsing for chapter text extraction

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "2.33.0",
       "license": "GPL-3.0",
       "dependencies": {
+        "@camb-ai/sdk": "^1.0.0",
         "axios": "^0.27.2",
         "cookie-parser": "^1.4.6",
         "express": "^4.17.1",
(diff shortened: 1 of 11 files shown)


Vito0912:
The holy grail of AI slop packaged as self-advertisement

nichwall:
This is unlikely to be merged. We are not reviewing AI-generated PRs due to an increasing number of low-quality PRs.

Something like this will be better integrated as a separate service that then provides the complete media file to the server as a persistent file.

neilruaro-camb:
Thanks for the feedback @nichwall! 

By separate service, do you mean a standalone CLI tool / service that we maintain on our end? Something that synthesizes audio from ebooks via our API and drops the files into an audiobookshelf library folder for the watcher to pick up?
                           
Also, is there any kind of integrations page or community ecosystem where third-party tools like that get listed? Would love to make it easy for audiobookshelf users to discover it if we build it

Issue #3865, opened 2025-01-20
[Bug]: Missing subpath in ebook url


baal000:
### What happened?

If I read an ebook in the browser it loads the ebook with the wrong url into the browser:
http://192.168.178.78:13378/api/items/16ae1df2-f091-4bc4-be32-5b52dcc33ea5/ebook instead of http://192.168.178.78:13378/audiobookshelf/api/items/16ae1df2-f091-4bc4-be32-5b52dcc33ea5/ebook

This leads to issues if you are using reverse proxies like nginx as the url for the ebook will not be available then.

### What did you expect to happen?

It uses http://192.168.178.78:13378/audiobookshelf/api/items/16ae1df2-f091-4bc4-be32-5b52dcc33ea5/ebook to load the ebook into the browser

### Steps to reproduce the issue

1. open browser to audiobookshelf
2. read ebook and check network traffic

### Audiobookshelf version

v2.18.0

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Logs

```shell

```

nickpascucci:
Nice, thanks for the quick fix!

PR #3864, opened 2025-01-20
Fix missing texture image & epub ebook url for subdirectory support

mikiher:
## Brief summary

Fixes a missing texture image in the nuxt build.

## Which issue is fixed?

This addresses [this comment](https://github.com/advplyr/audiobookshelf/issues/385#issuecomment-2601167818) in #385

Fixes #3865

## In-depth Description

`assets/app.css` contained a direct url link to `/textures/wood_default.jpg`. This breaks subdirectory support. Changed this to `~static/textures/wood_default.jpg` so it is picked and served by _nuxt (like other urls in the .css files)

## How have you tested this?

Ran against a dev server. The texture image is now fetched successfully from `/audiobookshelf/_nuxt/img/wood_default.6d366e7.jpg` instead of  `/textures/wood_default.jpg`.

PR #5132, opened 2026-03-19
Fix bug in readers where web browser is unable to load the ebook when…

### LINKED ISSUE #5131: [Bug]: PDF Ebook fails to load in web-browser behind reverse proxy at /audiobookshelf
### What happened?

Trying to read an ebook using the web-interface does not work. Looking at the network a request is made to 
```
/api/items/83dd64e7-2ca0-4e44-8e14-80edbe7c84ee/ebook
```
but in the use-case of a reverse proxy serving audiobookshelf on /audiobookshelf, this path fails.

I expected the ebook to appear in the reader. Instead I only get a blank page.

Image

1. Reverse proxy audiobookshelf at /audiobookshelf. e.g. for nginx
```
location /audiobookshelf {                                                                                                                
    proxy_pass http://127.0.0.1:80;                                                                                                     
}
```
2. Add a pdf ebook to the library
3. Go to 127.0.0.1/audiobookshelf/library
4. Click on the book and click on read

### Audiobookshelf version

v2.33.0

Docker

Linux

Firefox

### Logs

```shell

```

### Additional Notes

The browser view works if I do not go through the reverse proxy. It is only with being served behind a reverse proxy at /audiobookshelf

kareemalgalaly:
… running behind a reverse proxy at /audiobookshelf

## Brief summary

Fixes issue wherein the browser client is unable to display ebooks when running audiobookshelf behind a reverse proxy at /audiobookshelf

## Which issue is fixed?

Fixes #5131 

## In-depth Description

The bug is simply that the client reaches for the ebooks at /api/... instead of /audiobookshelf/api which is expected as of version 2.18

I am not certain (seeing it had been previously fixed at https://github.com/advplyr/audiobookshelf/issues/3865) that this is the ideal solution. 

## How have you tested this?

I spun up a development docker container with my changes and connected to it both directly and through an nginx reverse proxy at /audiobookshelf

## Screenshots
Before
Screenshot From 2026-03-18 21-19-44

After
Screenshot From 2026-03-18 22-30-04


client/components/readers/ComicReader.vue
```diff
@@ -109,9 +109,9 @@ export default {
     },
     ebookUrl() {
       if (this.fileId) {
-        return `/api/items/${this.libraryItemId}/ebook/${this.fileId}`
+        return `/audiobookshelf/api/items/${this.libraryItemId}/ebook/${this.fileId}`
```
nichwall: Instead of adding `/audiobookshelf` to the start of every API call, I think we want to add `${this.$config.routerBasePath}`, similar to #3864

Issue #3803, opened 2025-01-06
[Bug]: [LibraryItemController] Failed to download file "{FILE}" Error: Request aborted


AthenaShikata:
### What happened?

When the error occurs a popup appears in the browser saying "Progress is not being synced, restart playback." The chapter finishes, but won't continue. In the flappa mobile app, the chapter finishes, but won't continue. It may say it is unable to connect to the server, but usually not (I think it only says that after the issue resolves itself, but not 100%). The docker logs show this:

`[2025-01-05 19:26:54.939] ERROR: [LibraryItemController] Failed to download file "{FILE}" Error: Request aborted
    at onaborted (/node_modules/express/lib/response.js:1052:15)
    at Immediate.<anonymous> (/node_modules/express/lib/response.js:1094:9)
    at process.processImmediate (node:internal/timers:483:21) {
  code: 'ECONNABORTED'
}`

When I reload the page, I get stuck on the my.domain.com/login, but no login appears. When I try to log in on another device, the login gives an unknown error. Restarting the container or compose fixes the issue.

This issue persists for an arbitrary amount of time before fixing itself and doesn't happen for every chapter. Last night I had this happen for a couple chapters, but later that night I got through several chapters without issue. Today I've had it for every chapter so far. If I restart the container after the issue appears, the issue will reappear after a minute or two. 

### What did you expect to happen?

Continuous playback

### Steps to reproduce the issue

1. It happens at random

### Audiobookshelf version

v2.17.6 - v2.17.7

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
[2025-01-05 18:56:58.344] INFO: [LibraryItemController] User "{USER}" requested download for item "{TITLE}" file at "{FILE}"
[2025-01-05 18:56:59.755] INFO: [LibraryItemController] Downloaded file "{FILE}"
[2025-01-05 19:26:54.929] INFO: [LibraryItemController] User "{USER}" requested download for item "{TITLE}" file at "{FILE}"
[2025-01-05 19:26:54.939] ERROR: [LibraryItemController] Failed to download file "{FILE}" Error: Request aborted
    at onaborted (/node_modules/express/lib/response.js:1052:15)
    at Immediate.<anonymous> (/node_modules/express/lib/response.js:1094:9)
    at process.processImmediate (node:internal/timers:483:21) {
[...]

PR #5143, opened 2026-03-22
ignore "ECONNABORTED" in 'downloadLibraryFile' function

takov751:
## Brief summary

Downloading library item via the apps are failing early mostly when delay too high between client and server.
My assumption is that very likely the default value for timeout for connections is low.

## Which issue is fixed?

fixing #3803

## In-depth Description

Very small change. Basically making sure that "ECONNABORTED" error is ignored  by resolve() that given Promise. This way the downloading continues.
I used github copilot Claude model to help me patch it quickly. It is awful , i don't like it and only sharing to inspire some more reasonable solution. BUT IT SEEMS TO WORK.

## How have you tested this?

Friends of mine downloaded a larger 600+ MB audiobook from a different country with larger delay. He was patient Zero where the issue surfaced first.
Above that I had issue with audiobooth downloading books over slow mobile data connection

### DIFF
--- a/client/players/LocalAudioPlayer.js
+++ b/client/players/LocalAudioPlayer.js
@@ -46,7 +46,20 @@ export default class LocalAudioPlayer extends EventEmitter {
     this.player.addEventListener('loadedmetadata', this.evtLoadedMetadata.bind(this))
     this.player.addEventListener('timeupdate', this.evtTimeupdate.bind(this))
 
-    var mimeTypes = ['audio/flac', 'audio/mpeg', 'audio/mp4', 'audio/ogg', 'audio/aac', 'audio/x-ms-wma', 'audio/x-aiff', 'audio/webm']
+    var mimeTypes = [
+      'audio/flac',
+      'audio/mpeg',
+      'audio/mp4',
+      'audio/ogg',
+      'audio/aac',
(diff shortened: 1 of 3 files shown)


takov751:
Greetings @Vito0912 . I have done further tests and it seems ignoring that connection error made the downloading stable even on slow connection.

Vito0912:
I didn't test it, but I also don't understand how that should really be a solution. 
We just move the problem statistically down the line. 
The unstable connection, of course, gets "more stable" because if there is an issue, we now just send down the whole file again (?), but that should be a retry logic in the clients imho. But I didn't look deep into it, so maybe I misunderstand the solution.

So yeah, I see how that helps, but that doesn't really fix the problem.

takov751:
> I didn't test it, but I also don't understand how that should really be a solution. 
> 
> We just move the problem statistically down the line. 
> 
> The unstable connection, of course, gets "more stable" because if there is an issue, we now just send down the whole file again (?), but that should be a retry logic in the clients imho. But I didn't look deep into it, so maybe I misunderstand the solution.
> 
> 
> 
> So yeah, I see how that helps, but that doesn't really fix the problem.

I 100% agree. I am probably going to mention it in audiobooth client dc to test some possible solution in testflight
version.

PR #5158, opened 2026-03-30
Emit proper author_updated/added events when updating book media

mikiher:
## Brief summary

When Adding/Removing an author in Edit Book Details, the proper author_updated/author_added events are not properly emitted by the server, so the authors page open on a different client is not properly updated.

## Which issue is fixed?

I don't think there's an issue open for this.

## In-depth Description

When adding or removing an author in Book Details and submitting, the server updates/adds/removes the author correctly, and creates/removes the proper bookAuthors table rows. However, it doesn't emit `author_added`/`author_updated` socket events (it does emit `author_removed`). Consequently, listeners to those events (e.g. on the authors page) are not getting called, and the UI doesn't get updated (except `author_removed` which does get sent correctly).

The fix properly sends the missing socket events.

## How have you tested this?

- Had two browser windows opened side-by-side, one on the homepage, and one on the authors page.
- Opened Edit Book Details on of the books on the homepage
  - added a new (non-existing) author, submitted
  - added an existing author, submitted
  - removed existing author, submitted
  - removed existing author with 1 book, submitted
  - added 2 authors, submitted
  - removed 2 authors, submitted
- After each of the submissions, the relevant author card/s on the authors page was automatically added/removed/num-books-updated properly as expected.

### DIFF
--- a/server/models/Author.js
+++ b/server/models/Author.js
@@ -111,7 +111,7 @@ class Author extends Model {
    *
    * @param {string} name
    * @param {string} libraryId
-   * @returns {Promise<Author>}
+   * @returns {Promise<{ author: Author, created: boolean }>}
    */
   static async findOrCreateByNameAndLibrary(name, libraryId) {
     const author = await this.getByNameAndLibrary(name, libraryId)


advplyr:
Thanks!

PR #5160, opened 2026-03-31
Fix item_removed payload to include libraryId

mikiher:
## Brief summary

Up until now, item_remolved was emitted with just the library item's id as payload. This makes it impossible to ignore the event if the client receiving the event is not displaying the same library as the item's.

## Which issue is fixed?

No issue.

## In-depth Description

To fix the issue, the libraryId is added to the item_removed payload. This has no effect on existing clients since it just adds to the payload, not changing existing fields.

## How have you tested this?

Removed library item on one window. Made sure that library item is still properly removed from the library page in a different window.


advplyr:
Thanks!

PR #5163, opened 2026-04-02
The timestamp in the share URL should override the saved position for the user.

pjkottke:
## Brief summary
When a share link includes a ?t= query parameter specifying a start time, it should seek to that position rather than resuming from the listener's previously cached position.
## Which issue is fixed?
No existing issue, this is a new bug fix.
## In-depth Description
Share links support a ?t= query parameter for specifying a start time. However, when a returning visitor has an active share session (via share_session_id cookie), the server returns the cached currentTime from that session and ignores the ?t parameter entirely.
The fix is in ShareController.getMediaItemShareBySlug: when an existing session is found and returned early, ?t is now checked and applied to playbackSession.currentTime before the response is sent. The client (_slug.vue) is also updated to prefer ?t over playbackSession.currentTime as a belt-and-suspenders guard.
This is useful any time someone wants to share a link to a specific moment (a particular chapter or scene) with someone who has already listened to part of the same share.
## How have you tested this?
`npm test`

Created a share link for an audiobook
Opened the link and listened for ~30 seconds (establishing a cached session)
Reloaded the page - confirmed it resumed from the cached position (expected baseline behavior)
Appended ?t=120 to the URL and reloaded - confirmed playback started at 2:00 rather than the cached position
Removed ?t and reloaded - confirmed it returned to resuming from cached position

### DIFF
--- a/client/pages/share/_slug.vue
+++ b/client/pages/share/_slug.vue
@@ -363,7 +363,9 @@ export default {
       console.log('Loaded media item share', this.mediaItemShare)
     }
 
-    const startTime = this.playbackSession.currentTime || 0
+    const startTime = this.$route.query.t && !isNaN(this.$route.query.t)
+      ? parseFloat(this.$route.query.t)
+      : (this.playbackSession.currentTime || 0)
     this.localAudioPlayer.set(null, this.audioTracks, false, startTime, false)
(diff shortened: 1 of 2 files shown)


advplyr:
thanks, i think we should rely on the value that comes from the server since we are checking it is valid.

PR #5183, opened 2026-04-13
Draft: Added initial usage of `parserDate`

### LINKED ISSUE #5192: [Enhancement]: Allow usage of YYYYMMDD date
### Type of Enhancement

Server Backend

For my use case, AudiobookShelf is my podcast central, and the Podcast entry is a mp3 file downloaded from other sources (Youtube, SoundCloud...) these files receive the date in the date metadata, with the format of YYYYMMDD. Right now the parser uses a plain new Date() which fails since this isn't a ISO standard.

So, my idea is to use a parserDate to parse the date from this weird edge cases. I think this should be enough. From my reading on the codebase, this could be solved in the AudioFileScanner file. Not sure if this is everything that is needed to be honest.

To be able to use other tools like `yt-dlp` to pick up essays and similar to add on the aubiobookshelf.

N/A

2.33.1

N/A

shinspiegel:
## Brief summary

For my use case, AudiobookShelf is my podcast central, and the Podcast entry is a mp3 file downloaded from other sources (Youtube, SoundCloud...) these files receive the date in the `date` metadata, with the format of `YYYYMMDD`. Right now the parser uses a plain `new Date()` which fails since this isn't a ISO standard.

This should use a `parserDate` to also attempt to parse these dates.

## Which issue is fixed?

Fixes #5192 

## In-depth Description

On the `AudioFileScanner` I'm using the newly created `parseDate` to parse the date from these edge cases.

## How have you tested this?

Unit tested and manual tested.

## Test Images

Screenshot 2026-04-15 at 08 59 29

Screenshot 2026-04-15 at 09 30 27

### DIFF
--- a/server/scanner/AudioFileScanner.js
+++ b/server/scanner/AudioFileScanner.js
@@ -5,6 +5,7 @@ const { LogLevel } = require('../utils/constants')
 const { parseOverdriveMediaMarkersAsChapters } = require('../utils/parsers/parseOverdriveMediaMarkers')
 const parseNameString = require('../utils/parsers/parseNameString')
 const parseSeriesString = require('../utils/parsers/parseSeriesString')
+const parseDate = require('../utils/parsers/parseDate')
 const LibraryItem = require('../models/LibraryItem')
 const AudioFile = require('../objects/files/AudioFile')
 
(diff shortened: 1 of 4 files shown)


shinspiegel:
Not sure if this is the only place that would benefit for this type of date parsing.

shinspiegel:
Updated the MR for it.
A naive question, everywhere I checked the date appears to be right, except on the UI. 
This is due to the fetch from metadata online? Or what... I'm trying to figure this out for the last couple of hours and I couldn't pin-point where the issue is :(

nichwall:
> Updated the MR for it.
> A naive question, everywhere I checked the date appears to be right, except on the UI. 
> This is due to the fetch from metadata online? Or what... I'm trying to figure this out for the last couple of hours and I couldn't pin-point where the issue is :(

Where in the interface are you talking about?

shinspiegel:
On the `/item/:id` for the podcast. The `Published {date}` it's showing the wrong timing. One month earlier.
I also looked into my instance on the `latest`, I manually changed the `date` property in the `mp3` files to have the release date in the `ISO` standard, and also added the `pubDate` property in the metadata with the same value. And this is showing the right date. So it's for sure my change that broke...

I'll dive more, not exactly sure what/where it broke.

PR #5191, opened 2026-04-14
Add Download by Chapters tool for audiobooks

vdaluz:
## Summary

- Adds a **Download by Chapters** tool in the item tools tab for audiobooks with chapters. The main motivation for this feature is to improve the audiobook user experience in limited players such as the Shokz OpenSwim which are intended for music and have limited or non existent seek/resume functionality.
- Splits the selected audio file into per-chapter files using `ffmpeg -c copy` (no re-encoding) and packages them as a ZIP download
- When a book has multiple audio files (e.g. both `.m4b` and `.mp3`), a dropdown lets the user choose which file to split

## How it works

**Backend (`GET /api/items/:id/download-chapters`)**

- Accepts an optional `?fileIno=` query param to identify which audio file to split; defaults to the first included audio file
- The endpoint computes each file's start offset (sum of durations of preceding files, ordered by `index`), filters chapters to those belonging to the selected file, and converts timestamps to file-relative before splitting
- Splits all chapters in parallel using `ffmpeg -c copy` (one process per chapter — no re-encoding, fast) into a temporary directory under `MetadataPath/cache/items/:id/chapter-split-:timestamp`
- The timestamp suffix prevents collisions between concurrent requests
- Streams the result as a ZIP (`archiver`, no compression — audio files are already compressed)
- Cleans up temp files in a `finally` block regardless of success or failure
- Chapter files are named `01 - Chapter Title.ext`, `02 - Chapter Title.ext`, etc., zero-padded based on total chapter count

**Frontend (`Tools.vue`)**

- New card visible when the book has chapters, the user has download permission, and at least one audio file is present
- Dropdown to select audio file when more than one is available
- Button shows "Preparing..." state while the request is in-flight; uses `$nextTick` + `requestAnimationFrame` to ensure the state is painted before the (blocking) anchor click

## Testing

1. Open any audiobook with chapters
2. Go to item detail → Tools tab
3. "Download by Chapters" card should appear below "Embed Metadata"
4. If the book has multiple audio files, a dropdown allows selecting which to split
5. Click "Download ZIP" — button shows "Preparing..." immediately
6. ZIP downloads containing one file per chapter, named and ordered correctly

## Screenshots

image

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nichwall:
We are not reviewing AI generated PRs at this time due to an increasing number of low-quality PRs.

I don't know if we need to have a new download function for splitting files back up when users can already download individual files from a book page (or just put a multi-file book there in the first place).

PR #5193, opened 2026-04-15
Add ntfy.sh as notification provider

mikkelmi2:
## Summary
- Add [ntfy.sh](https://ntfy.sh) as an alternative notification provider alongside the existing Apprise integration
- Each notification can be configured to use either Apprise or ntfy via a provider dropdown in the create/edit modal
- Supports self-hosted ntfy servers and optional token-based authentication
- No new dependencies — uses the existing axios library

## Motivation
While Apprise supports ntfy as one of its many backends, it requires running a separate Apprise API server as a middleman. With native ntfy support, users can send notifications directly to ntfy.sh (or a self-hosted instance) with just a topic name — no additional infrastructure needed.

## Changes

**Server:**
- `NotificationSettings`: Add `ntfyUrl` and `ntfyToken` fields, update `isUseable` to check either provider, add `undefined` guards in `update()` to prevent partial PATCH from resetting unrelated fields
- `Notification`: Add `ntfyTopic` field and `getNtfyPayload()` method
- `NotificationManager`: Route to Apprise or ntfy based on `ntfyTopic` presence, add null guard for `appriseApiUrl`
- `NotificationController`: Update error message to be provider-agnostic

**Client:**
- Add ntfy.sh settings section (server URL with placeholder, auth token) on the notifications config page
- Add provider dropdown (Apprise / ntfy.sh) in the notification edit modal
- Extract shared `saveSettings()` helper to deduplicate form submissions
- All user-visible strings use i18n (en-us.json only — other locales fall back to English)

## Test plan
- [ ] Existing Apprise notifications still work
- [ ] Create and test an ntfy notification
- [ ] Saving one provider's settings does not clear the other


nichwall:
Thanks for opening a PR, but we will not be reviewing it at this time

We are not planning to add support for more notification providers. Apprise already handles ntfy.sh notifications.

It also looks like the PR description is AI generated and not using our PR template. AI generated PRs are not being reviewed due to an increasing number of low quality PRs and limited reviewer time.

PR #5206, opened 2026-04-22
Add OpenAI-powered series tools, scan inference, directory grouping, and book dedupe

korjik:
## Brief summary

Adds OpenAI-assisted organization workflows for book libraries, including:

- server-side OpenAI configuration and UI settings
- library tools for AI series detection and AI duplicate-book cleanup
- series-page AI story ordering
- scan-time OpenAI metadata inference from messy paths and filenames
- optional OpenAI directory-tree interpretation during scans for poorly structured libraries
- improved non-AI scanner grouping for mixed parent-folder/direct-file book layouts

## Which issue is fixed?

N/A

## In-depth Description

This PR introduces an OpenAI integration for book-library organization and metadata cleanup.

Configuration:
- Adds OpenAI settings to the server and web UI
- Supports configuration from either the UI or environment variables:
  - `OPENAI_API_KEY`
  - `OPENAI_MODEL`
  - `OPENAI_BASE_URL`

Library tools:
- Adds `Detect Missing Series With AI`
- Adds `Re-evaluate All Series`
- Adds `Dedupe Books With AI`

Series tools:
- Adds `Organize Story Order With AI` on series pages to update series sequence values using AI

Scanner improvements:
- Adds `OpenAI path and filename inference` as an optional scan metadata source
- Adds `Use OpenAI to interpret poor directory trees during library scans` as an optional library scanner setting
- This allows scans to recover more cleanly from messy folder structures and infer metadata when standard folder parsing is weak

Scanner grouping fixes:
- Fixes a scanner edge case where a direct media file inside a parent series folder could cause nested book folders to collapse into a single detected item
- Sidecar files like `.nfo`, `.cue`, and cover images are now attached to the correct logical item in that scenario

## OpenAI response handling:
- Adds defensive validation for AI responses
- Invalid or partial AI rows are skipped instead of aborting the whole operation
- Duplicate, missing, or unknown ids in AI responses are tolerated where appropriate so long-running library operations can continue

## This change is aimed at users with poorly normalized audiobook libraries, especially libraries where:
- series membership is incomplete or inconsistent
- directory structure does not follow the expected author/series/title layout
- standalone files and nested book folders are mixed together
- duplicate copies of the same book exist with different metadata quality

## How have you tested this?

I tested the changes with focused server-side validation and scanner regression tests.

## Reproducible checks run locally:
- `node --check server/providers/OpenAI.js`
- `node --check server/scanner/BookScanner.js`
- `node --check server/scanner/LibraryScanner.js`
- `node --check server/controllers/LibraryController.js`
- `node --check server/routers/ApiRouter.js`

## Targeted test suites run locally:
- `./node_modules/.bin/mocha test/server/providers/OpenAI.test.js`
- `./node_modules/.bin/mocha test/server/utils/scandir.test.js`
- `./node_modules/.bin/mocha test/server/scanner/LibraryScanner.test.js`
- Combined runs of the focused suites also passed during development

## Specific behaviors covered:
- OpenAI response parsing and validation
- tolerance of malformed AI outputs
- duplicate/unknown/missing AI ids
- scan metadata inference payload validation
- directory-grouping payload validation
- duplicate-book decision payload validation
- scanner grouping regression for poor parent-folder/direct-file layouts
- scanner regrouping helper behavior for mixed standalone and nested book items

## Manual verification performed in the app/container workflow:
1. Build and run the local Docker image instead of the upstream image
1. Configure OpenAI through the UI
1. Run AI series detection from library tools
1. Run AI series re-evaluation from library tools
1 Run AI story ordering from a series page
1. Enable OpenAI scan metadata inference and AI directory grouping in library scanner settings

### Settings
image

### Library Tools
image

### Series menu update
image

### Library scanning setting update (off by default)
image

### DIFF
--- a/client/components/app/BookShelfToolbar.vue
+++ b/client/components/app/BookShelfToolbar.vue
@@ -134,6 +134,13 @@ export default {
         }
       ]
 
+      if (this.userCanUpdate && this.openAIConfigured) {
+        items.push({
+          text: 'Organize Story Order With AI',
+          action: 'organize-story-order-with-ai'
+        })
+      }
+
       if (this.userIsAdminOrUp || this.selectedSeries.rssFeed) {
         items.push({
           text: this.$strings.LabelOpenRSSFeed,
@@ -220,6 +227,9 @@ export default {
     },
     userCanDownload() {
       return this.$store.getters['user/getUserCanDownload']
+    },
+    openAIConfigured() {
+      return !!this.$store.getters['getServerSetting']('openAIConfigured')
     },
     currentLibraryId() {
       return this.$store.state.libraries.currentLibraryId
@@ -427,6 +437,12 @@ export default {
     seriesContextMenuAction({ action }) {
       if (action === 'open-rss-feed') {
         this.showOpenSeriesRSSFeed()
+      } else if (action === 'organize-story-order-with-ai') {
+        if (this.processingSeries) {
+          console.warn('Already processing series')
+          return
+        }
+        this.organizeStoryOrderWithAI()
       } else if (action === 're-add-to-continue-listening') {
         if (this.processingSeries) {
           console.warn('Already processing series')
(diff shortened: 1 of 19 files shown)


nichwall:
I don't think we want to add any AI capabilites or settings to the server. You are welcome to build something that interfaces with the server through the server API. Not sure if anyone else has other opinions.

korjik:
> I don't think we want to add any AI capabilites or settings to the server. You are welcome to build something that interfaces with the server through the server API. Not sure if anyone else has other opinions.

I agree for post-scan organization features. Those can reasonably live outside the server and use the existing API.

The scanner case is different. In this case the problem is that poorly structured directories are not being recognized as books at all, so no library items are created. Once that happens, there is nothing available through the API to fix afterward.

An external tool could reorganize files on disk and trigger a rescan, but that is a different and much more invasive workflow. What this change was trying to address is scan-time interpretation, not post-processing.

If that is still out of scope for core, I can split out the non-AI scanner fixes and drop the server-side AI pieces.

nichwall:
> > I don't think we want to add any AI capabilites or settings to the server. You are welcome to build something that interfaces with the server through the server API. Not sure if anyone else has other opinions.
> 
> I agree for post-scan organization features. Those can reasonably live outside the server and use the existing API.
> 
> The scanner case is different. In this case the problem is that poorly structured directories are not being recognized as books at all, so no library items are created. Once that happens, there is nothing available through the API to fix afterward.

Books are defined as a single folder, adding an external AI API call to it doesn't change whether a book is added. The changes would extend to a lot more than just updating the scanner.

I don't think we want to change the directory structure assumptions of "one book per folder" as of right now due to the web client being rewritten and migrating to a new framework.

> If that is still out of scope for core, I can split out the non-AI scanner fixes and drop the server-side AI pieces.

Yes. I would make a scanner specific update PR. As it is this is a massive PR that makes it difficult to read through. Also be aware that we currently have a large backlog of PRs so it may be several weeks/months before it would be reviewed (more than me just reading the description).

jeeftor:
I wrote an external tool that can kind of help with this -> it doesn't use AI - it uses either `metadata.json` as created by audio bookshelf - or embedded data in the files:

https://github.com/jeeftor/audiobook-organizer

Not overly trying to self-promote - but it may solve your needs.

Alternatively you could maintain your own fork! :)

PR #5210, opened 2026-04-26
Draft: Add Audible preorder integration with personalized shelf carousel

michaeldvinci:
## Brief summary

Adds a full Audible preorder integration that lets users connect their Audible account and display
upcoming preorder books in a dedicated "Audible Preorders" carousel on the library home page. 
## Which issue is fixed?

N/A

## In-depth Description

**Audible Preorder Integration**

Users can now link their Audible account (via an access token obtained from `audible-cli`) and sync
their preorders into Audiobookshelf. Once connected, a new "Audible Preorders" shelf row appears on
the library home page showing cover art, title, author, and release date for each upcoming title.

How it works:
- A new `/config/audible` settings page lets users add/remove accounts, select which library the
  shelf should appear in, and set a shelf position for the carousel row.
- On sync, the server calls Audible's `/1.0/orders` endpoint, filters for orders with
  `is_preorder: true` and `release_date > now`, then fetches full metadata for each ASIN via
  individual `/1.0/catalog/products/{asin}` lookups (the batch endpoint silently ignores the
  `asin` param).
- Sync is atomic — all books are destroyed and re-inserted inside a single Sequelize transaction;
  `lastSync` is only committed on success.
- Access tokens are stored AES-256-CBC encrypted using `AUDIBLE_ENCRYPTION_KEY` (falls back to
  `TOKEN_SECRET`; throws if neither is set).
- Endpoints are user-scoped only (no admin bypass), always return 403 for missing or unowned
  accounts to avoid enumeration, and are rate-limited to 10 syncs per 5 minutes per user.
- The carousel shelf position is the minimum `shelfPosition` across all accounts linked to a given
  library; books are sorted by release date ascending.

New files: `AudibleAccount` and `AudibleBook` Sequelize models, `AudibleController`,
`AudibleLibrary` provider, migration `v2.33.2-setup-audible.js`, and all associated client
components.

## How have you tested this?

Tested on a self-hosted Audiobookshelf instance running in Docker:

1. Obtained an Audible access token using `audible-cli` and copied it into the new
   `/config/audible` settings page.
2. Selected a library and clicked Sync. Verified 9 preorder books were fetched and appeared
   in the database.
3. Navigated to the library home page and confirmed the "Audible Preorders" carousel row
   rendered with correct cover art, titles, and authors.
4. Adjusted shelf position and re-synced; confirmed the carousel row moved to the correct
   position relative to other shelves.

## Screenshots

**Audible Preorders carousel on library home page:**
Audible Preorders shelf

**Audible account settings page (`/config/audible`):**
Audible settings page

### DIFF
--- a/client/components/app/BookShelfCategorized.vue
+++ b/client/components/app/BookShelfCategorized.vue
@@ -52,7 +52,7 @@ export default {
   },
   computed: {
     supportedShelves() {
-      return this.shelves.filter((shelf) => ['book', 'podcast', 'episode', 'series', 'authors', 'narrators'].includes(shelf.type))
+      return this.shelves.filter((shelf) => ['book', 'podcast', 'episode', 'series', 'authors', 'narrators', 'audible-preorder'].includes(shelf.type))
     },
     userIsAdminOrUp() {
       return this.$store.getters['user/getIsAdminOrUp']
--- a/client/components/app/BookShelfRow.vue
+++ b/client/components/app/BookShelfRow.vue
@@ -32,12 +32,17 @@
             <cards-narrator-card :key="entity.name" :narrator="entity" @hook:updated="updatedBookCard" class="mx-2e" />
           </template>
         </div>
+        <div v-if="shelf.type === 'audible-preorder'" class="flex items-center">
+          <template v-for="entity in shelf.entities">
+            <cards-audible-preorder-card :key="entity.id" :book="entity" class="mx-2e" />
+          </template>
+        </div>
       </div>
     </div>
     <div class="relative">
       <div class="relative text-center categoryPlacard transform z-30 top-0 left-4e md:left-8e w-44e rounded-md">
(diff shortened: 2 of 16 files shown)


michaeldvinci:
sidebar: This is still a WIP: just wanted to see how it validated against your workflows + merge rules

nichwall:
Thanks for opening a PR, but I don't think we will merge something like this in. The Audiobookshelf server is intended as a server for your local media files, not tracking against media files that are not released or available on your server. There are feature requests around tracking unreleased books, but not sure if/when that would be supported by the main server.

You can always make a placeholder book file, which I realize is a bit of a workaround. This would probably be something that can be a plugin (whenever the plugin system is done, but that will likely still be quite a while).

(I realize you have already closed this PR, but making sure to let you know before you spend a bunch of time working on a new PR)

michaeldvinci:
hey @nichwall -- thanks for the heads up! I kind of figured something like this would be way outside the norm -- I might maintain my own fork for a bit while I play with this functionality.

Let me know if you ever decide to look into any of that functionality 👍

PR #5214, opened 2026-04-29
Preserve user-cleared metadata fields on rescan

### LINKED ISSUE #5213: [Bug]: User-cleared metadata fields are overwritten on rescan
## Describe the bug

When a user clears a scalar metadata field (e.g. subtitle, description, publisher) via the UI, the cleared value is not preserved when the library is rescanned. Instead, the value from a lower-priority source - typically embedded audio file tags - is restored, undoing the user's edit.

## Steps to reproduce

1. Open a book whose audio files have a subtitle embedded in their tags.
2. Clear the subtitle field via Edit Book → Details → Subtitle and save.
3. Trigger a library rescan (manual or scheduled).
4. The subtitle is repopulated from the audio file tag.

## Root cause

`AbsMetadataFileScanner.scanBookMetadataFile` skips all `null` values from `metadata.json` unconditionally (see the `// TODO: When to override with null or empty arrays?` comment in the source). Since `absMetadata` is the highest-priority metadata source, a `null` written to `metadata.json` when the user clears a field should override lower-priority sources — but because it is skipped, the value from `audioMetatags` or another lower-priority source survives into `bookMetadata` and gets written back to the DB.

## Expected behavior

A field explicitly cleared by the user should remain blank after a rescan.

## Proposed fix

Propagate `null` from `metadata.json` into `bookMetadata` for clearable scalar string fields (`subtitle`, `publishedYear`, `publisher`, `description`, `isbn`, `asin`, `language`). This resolves the `TODO` comment and makes `absMetadata` behave consistently for null values the same way it does for non-null ones (i.e. it wins over lower-priority sources).

I have a one-file fix ready if this approach is acceptable.

## Version

Latest (`master`)

nickwolf:
## Brief summary

Clearing a scalar metadata field in the UI does not survive a rescan. Audiobookshelf writes the cleared field into `metadata.json` as `null`, but `AbsMetadataFileScanner` skips every `null` it reads back, so the audio file tags refill the field on the next rescan and the edit is gone.

## Which issue is fixed?

Fixes #5213

## In-depth Description

`absMetadata` is last in the precedence chain, so it already wins over `audioMetatags` for non-null values. It does not win for `null`, because the loop treats "the user cleared this" and "this key is not in the file" as the same thing. Since 1bad2d90 `parseJson` tells them apart, the scanner only has to stop discarding the `null` value.

Only keys the generator types as `string` get cleared. `title` is excluded because `PATCH /api/items/:id/media` accepts an empty title and writes `"title": null`, which would leave the item unnamed instead of recovering the title from the tags. Booleans are excluded because a null `explicit` lands in the column as SQL `NULL`, which `getUserPermissionBookWhereQuery` does not match with `explicit: false`, while `checkCanAccessLibraryItem` coerces the same value with `!!`. The book would be missing from a restricted user's listing but still reachable directly.

Empty arrays are unchanged. `BookScanner.updateBook` skips them itself, so propagating them from here would do nothing until that line changes too. The `TODO` stays, narrowed to arrays. `scanPodcastMetadataFile` has the same null skip and I left it alone, since `feedURL` wants its own decision rather than the same blanket rule.

## How have you tested this?

A throwaway instance of stock `ghcr.io/advplyr/audiobookshelf:2.36.0`, default settings, one generated mp3 whose `TIT3` tag is `Subtitle From Tags`. Both files this touches are byte identical between that image and master, so the diff is the only variable.

Clear the subtitle and `metadata.json` gets `"subtitle": null`. On stock 2.36.0, `POST /api/items/:id/scan` returns `{"result":"UPDATED"}` and the subtitle is back from the tag. With the patch bind mounted over those two files, same volumes and same item, the rescan returns `{"result":"UPTODATE"}` and a forced library scan logs `0 Added | 0 Updated | 0 Missing`. Going back to the unpatched container from that same cleared state, a forced scan logs `0 Added | 1 Updated | 0 Missing` and refills it.

For the boolean case, with `"explicit": null` written into `metadata.json` by hand, the patched build leaves `books.explicit` as `integer|1`. Without the boolean exclusion the same scan writes SQL `NULL`.

Six cases in a new `test/server/scanner/AbsMetadataFileScanner.test.js`. Run against master's version of the two source files, exactly one of them fails.

### DIFF
--- a/server/scanner/AbsMetadataFileScanner.js
+++ b/server/scanner/AbsMetadataFileScanner.js
@@ -34,8 +34,18 @@ class AbsMetadataFileScanner {
       libraryScan.addLog(LogLevel.INFO, `Found metadata file "${metadataFilePath}"`)
       const abMetadata = abmetadataGenerator.parseJson(metadataText) || {}
       for (const key in abMetadata) {
-        // TODO: When to override with null or empty arrays?
-        if (abMetadata[key] === undefined || abMetadata[key] === null) continue
+        if (abMetadata[key] === undefined) continue
+        if (abMetadata[key] === null) {
+          // Propagate null for scalar string fields so that values intentionally cleared
+          // by the user are preserved when the library is rescanned. absMetadata has the
+          // highest metadata precedence, so a null here means the user explicitly cleared
+          // the field and lower-priority sources (e.g. audio file tags) should not refill it.
+          const clearableStringFields = ['subtitle', 'publishedYear', 'publisher', 'description', 'isbn', 'asin', 'language']
+          if (clearableStringFields.includes(key)) {
+            bookMetadata[key] = null
+          }
+          continue
+        }
         if (key === 'authors' && !abMetadata.authors?.length) continue
         if (key === 'genres' && !abMetadata.genres?.length) continue
         if (key === 'tags' && !abMetadata.tags?.length) continue


nickwolf:
I've been running this fix in my library for about a week and haven't run into any issues so far.

PR #5219, opened 2026-05-02
Add `audible.br` (Audible Brasil) as a book metadata provider

danilopopeye:
## Brief summary

 Adds `audible.br` to the list of available book metadata providers, enabling users to search and match books using Audible's Brazilian catalogue (`audible.com.br`).

## Which issue is fixed?

 N/A

## In-depth Description

 The `Audible.js` provider already had the `br` region mapped to  `.com.br` in its `regionMap`, but `BookFinder.js` was not exposing `audible.br` in its providers array — so the option never appeared in the UI dropdown.

> Note: This PR depends on laxamentumtech/audnexus#861, which adds Brazil region support to the Audnexus API used for ASIN lookups.

## How have you tested this?

On my mahchine, see the screenshots bellow.

## Screenshots

Screenshot 2026-05-01 at 21 07 47
Screenshot 2026-05-01 at 21 08 05
Screenshot 2026-05-01 at 21 09 50
Screenshot 2026-05-01 at 21 10 07
Screenshot 2026-05-01 at 21 10 12

### DIFF
--- a/client/pages/audiobook/_id/chapters.vue
+++ b/client/pages/audiobook/_id/chapters.vue
@@ -347,7 +347,7 @@ export default {
       asinError: null,
       removeBranding: false,
       showSecondInputs: false,
-      audibleRegions: ['US', 'CA', 'UK', 'AU', 'FR', 'DE', 'JP', 'IT', 'IN', 'ES'],
+      audibleRegions: ['US', 'CA', 'UK', 'AU', 'FR', 'DE', 'JP', 'IT', 'IN', 'ES', 'BR'],
       hasChanges: false,
       timeIncrementAmount: 1,
       elapsedTime: 0,
--- a/server/controllers/SearchController.js
+++ b/server/controllers/SearchController.js
@@ -25,6 +25,7 @@ const providerMap = {
   'audible.it': 'Audible.it',
   'audible.in': 'Audible.in',
   'audible.es': 'Audible.es',
+  'audible.br': 'Audible.com.br',
   audnexus: 'Audnexus'
 }
 
--- a/server/finders/BookFinder.js
+++ b/server/finders/BookFinder.js
@@ -23,7 +23,7 @@ class BookFinder {
     this.audiobookCovers = new AudiobookCovers()
     this.customProviderAdapter = new CustomProviderAdapter()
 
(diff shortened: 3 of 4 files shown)


advplyr:
> Note: This PR depends on [laxamentumtech/audnexus#861](https://github.com/laxamentumtech/audnexus/pull/861), which adds Brazil region support to the Audnexus API used for ASIN lookups.

so this is not currently supported in audnexus?

danilopopeye:
> so this is not currently supported in audnexus?

not yet, still waiting for their review. do you want me to close this one untill (if) the other is merged?

Vito0912:
@advplyr On that topic, would you merge a pull request that removes Audnexus, or would you prefer to wait until after the client rewrite? The pull request would not change anything feature wise, it would only remove the unnecessary dependency on Audnexus, which would also allow to integrate br directly.

advplyr:
> @advplyr On that topic, would you merge a pull request that removes Audnexus, or would you prefer to wait until after the client rewrite? The pull request would not change anything feature wise, it would only remove the unnecessary dependency on Audnexus, which would also allow to integrate br directly.

Yeah I'm up for that. I know we've discussed this before, but is there anything audnexus offers from their API that isn't in the audible api? I remember I originally thought chapters weren't available but it turned out they were. What about author information?

Vito0912:
> Yeah I'm up for that. I know we've discussed this before, but is there anything audnexus offers from their API that isn't in the audible api? I remember I originally thought chapters weren't available but it turned out they were. What about author information?

Everything is doable via the API directly (even without big changes). Authors currently would still need to depend on Audnexus, because they do it in the DB. With my archived metadata bridge I had very, very good results, using the search endpoint. But this would in theory alter the author search behavior, so it would be your call whether we change that or not. Getting the author information itself is not a problem

Vito0912:
I pushed a branch here: https://github.com/advplyr/audiobookshelf/compare/master...Vito0912:audiobookshelf:feat/replaceAudnexus

The fetching of an item and chapters it should be identically.

For authors it worked pretty great for me, but my authors are also not that "small" (in terms of reach) and the smaller ones do not have an Audible page. So it's not really fair to compare

advplyr:
I didn't run your branch but skimmed it over and I think that'll work. We could keep the authors as they are with Audnexus but still build in the audible version for testing

danilopopeye:
@advplyr trying to not let this one dies as I would really like to have this source. 😆 
How do you see it going forward? Is it something else also required? 
I can try to drop a PR or two if you let me know where or what. 😉

PR #5221, opened 2026-05-02
Fix: RSS feed serves m4b files with correct Content-Type: audio/mp4

### LINKED ISSUE #5041: [Bug]: RSS feed serves m4b files with Content-Type: application/octet-stream instead of audio/mp4
### What happened?

When serving audiobook files via the RSS feed endpoint (`/feed/:slug/item/:episodeId/*`), `.m4b` files are served with `Content-Type: application/octet-stream` instead of `audio/mp4`.

This is because `RssFeedManager.getFeedItem()` in `server/managers/RssFeedManager.js` uses `res.sendFile(episodePath)` without setting the Content-Type header. Express's `sendFile` relies on the `mime` package for content-type lookup, and `.m4b` is not in the standard MIME database, so it falls back to `application/octet-stream`.

This causes **slow playback start (10-20+ seconds)** in podcast clients like Apple Podcasts when consuming audiobook RSS feeds, because the client cannot identify the audio format from the response headers and must sniff the file content instead.

**This is the same underlying issue that was fixed for the download endpoint in PR #3319 (issue #3310)**, but the RSS feed file-serving codepath was not included in that fix.

#### Root cause

In [`server/managers/RssFeedManager.js` line ~192](https://github.com/advplyr/audiobookshelf/blob/master/server/managers/RssFeedManager.js#L192):

```javascript
async getFeedItem(req, res) {
    // ...
    res.sendFile(episodePath)  // No Content-Type set — falls back to application/octet-stream for .m4b
}
```

Compare with the download endpoint in `server/controllers/LibraryItemController.js` which correctly uses `getAudioMimeTypeFromExtname()`:

```javascript
const audioMimeType = getAudioMimeTypeFromExtname(Path.extname(libraryFile.metadata.path))
if (audioMimeType) {
    res.setHeader('Content-Type', audioMimeType)
}
```

Also compare with `getFeedCover()` in the same `RssFeedManager.js` file, which correctly sets the content type:

```javascript
const extname = Path.extname(feed.coverPath).toLowerCase().slice(1)
res.type(`image/${extname}`)
```

#### Suggested fix

Use `getAudioMimeTypeFromExtname()` (from `server/utils/fileUtils.js`) in `getFeedItem()` to set the correct Content-Type before calling `res.sendFile()`:

```javascript
const { getAudioMimeTypeFromExtname } = require('../utils/fileUtils')

async getFeedItem(req, res) {
    // ... existing feed/episode lookup ...
    const audioMimeType = getAudioMimeTypeFromExtname(Path.extname(episodePath))
    if (audioMimeType) {
        res.setHeader('Content-Type', audioMimeType)
    }
    res.sendFile(episodePath)
}
```

The `AudioMimeType` map in `server/utils/constants.js` already correctly maps `.m4b` → `audio/mp4`.

RSS feed episode files should be served with the correct `Content-Type: audio/mp4` header for `.m4b` files (and correct MIME types for other audio formats), matching the behavior of the download endpoint.

1. Create an audiobook library with `.m4b` files
2. Open an RSS feed for the audiobook
3. Subscribe to the feed in Apple Podcasts (or any podcast client)
4. Observe slow playback start times
5. Inspect the HTTP response headers for a feed episode URL — `Content-Type` will be `application/octet-stream`

Verify with curl:
```bash
curl -sI "https://your-server/feed/your-slug/item/episode-id/media.m4b" | grep -i content-type
# Returns: Content-Type: application/octet-stream
# Expected: Content-Type: audio/mp4
```

### Audiobookshelf version

v2.32.1

Docker

Linux

### Logs

_No response_

### Additional Notes

**Workaround:** If using a reverse proxy (e.g., Caddy, nginx), you can rewrite the Content-Type header at the proxy level:

Caddy example:
```
reverse_proxy audiobookshelf:80 {
    header_down Content-Type application/octet-stream audio/mp4
}
```

Note: this workaround replaces `application/octet-stream` with `audio/mp4` for all responses from ABS, which works if your ABS instance primarily serves audiobooks. A server-side fix is more appropriate.

This issue affects all audio formats not in the standard `mime` package lookup table, but `.m4b` is the most common case since it's the standard audiobook format.

brandonfhall:
## Brief summary

Fixes RSS feed episode file serving so that `.m4b` files are returned with `Content-Type: audio/mp4` instead of `Content-Type: application/octet-stream`.

## Which issue is fixed?

Fixes #5041

## In-depth Description

Express's bundled `mime` package does not have an entry for `.m4b`, so `res.sendFile()` falls back to `application/octet-stream` for audiobook files. This causes podcast clients (notably Apple Podcasts) to content-sniff the file before starting playback, resulting in 10–20+ second delays.

The fix reuses `getAudioMimeTypeFromExtname()` from `server/utils/fileUtils.js`, which looks up extensions against the `AudioMimeType` map in `server/utils/constants.js` (where `.m4b → audio/mp4` is already defined). The header is set explicitly before calling `res.sendFile()`.

This is the same pattern already applied to the download endpoint in `server/controllers/LibraryItemController.js` (added in PR #3319 for issue #3310) — the RSS feed code path was simply missed in that fix.

The `if (audioMimeType)` guard ensures no regression for formats not in the map: those fall through and Express handles them as before.

## How have you tested this?

Added unit tests in `test/server/managers/RssFeedManager.test.js` covering:

- `.m4b` files receive `Content-Type: audio/mp4`
- Unknown extensions do not have a `Content-Type` header set (guard branch)
- 404 is returned when the feed slug is not found
- 404 is returned when the episode ID is not found in the feed

All four tests pass (`npm test`).

Manual verification approach: subscribe to an audiobook RSS feed in a podcast client and inspect the HTTP response headers for a feed episode URL — `Content-Type` should now be `audio/mp4` for `.m4b` files.

## Screenshots

N/A — no web client changes.

### DIFF
--- a/server/managers/RssFeedManager.js
+++ b/server/managers/RssFeedManager.js
@@ -2,6 +2,7 @@ const { Request, Response } = require('express')
 const Path = require('path')
 
 const Logger = require('../Logger')
+const { getAudioMimeTypeFromExtname } = require('../utils/fileUtils')
 const SocketAuthority = require('../SocketAuthority')
 const Database = require('../Database')
 
@@ -215,6 +216,11 @@ class RssFeedManager {
       Logger.error(`[RssFeedManager] Feed episode not found ${req.params.episodeId}`)
       res.sendStatus(404)
       return
+    }
+    // Express does not set the correct mimetype for m4b files so use our defined mimetypes if available
+    const audioMimeType = getAudioMimeTypeFromExtname(Path.extname(episodePath))
+    if (audioMimeType) {
+      res.setHeader('Content-Type', audioMimeType)
     }
     res.sendFile(episodePath)
   }


advplyr:
this is good but please remove the unit test, it's unnecessary. also please amend your commits to remove claude as a co-author and force push

brandonfhall:
@advplyr test removed, and commits cleaned up. 

Thank you!

advplyr:
Thanks!

PR #5239, opened 2026-05-13
finetune opus encoding

takov751:
## Brief summary

Just minor improvement to the opus encoding adding `vbr on` `compression-level 10` and `application voip`
to the ffmpeg command

## Which issue is fixed?

Before opus converted audiobooks quality was questionable with the default settings

## How have you tested this?

I have converted multiple audiobooks with different original codecs (multiple mp3, flac, aac) to 
`m4b format with opus audio channel`

### DIFF
--- a/server/utils/ffmpegHelpers.js
+++ b/server/utils/ffmpegHelpers.js
@@ -471,7 +471,11 @@ async function mergeAudioFiles(audioTracks, duration, itemCachePath, outputFileP
   ffmpeg.outputOptions(['-f mp4'])
 
   if (audioRequiresEncode) {
-    ffmpeg.outputOptions(['-map 0:a', `-acodec ${audioCodec}`, `-ac ${audioChannels}`, `-b:a ${audioBitrate}`])
+    if (audioCodec === 'opus') {
+      ffmpeg.outputOptions(['-map 0:a', '-c:a libopus', `-b:a ${audioBitrate}`, '-vbr on', '-compression_level 10', '-application voip', `-ac ${audioChannels}`])
+    } else {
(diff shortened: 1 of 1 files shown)


Sapd:
I think you should add as comment what `compression-level 10` does exactly. A bit of research on my side it seems to be the default actually.

VBR definitely makes sense and is self explanatory. 

Then what does `application voip` do exactly? Could be a problem on Radioplays (very popular in Germany especially "Hörspiele") like the latest Harry Potter full cast edition.

takov751:
> I think you should add as comment what `compression-level 10` does exactly. A bit of research on my side it seems to be the default actually.
> 
> VBR definitely makes sense and is self explanatory.
> 
> Then what does `application voip` do exactly? Could be a problem on Radioplays (very popular in Germany especially "Hörspiele") like the latest Harry Potter full cast edition.

Thank you @Sapd . Exactly the feedback i was hoping for. Yes indeed compression level 10 is redundant as the official ffmpeg codec documentation shows. application voip is meant to be optimizing for speech itself, however it does not seems to impact Radioplay style audiobooks above 50+ kbps bitrate target. However I might rethink this solution and maybe make that optional. I will turn this PR into a draft and ponder these questions further.

Vito0912:
> VBR definitely makes sense and is self explanatory.

Also want to add that VBR is also enabled by default. So probably comes down to the effect of voip

takov751:
@Vito0912 as we discussed on discord this change had been almost redundant. `application voip` is only beneficial with regular audiobook with radioplay style audiobooks it does alters the output in a negative way. However setting frame_duration to 60 does make a lot of difference quality wise

PR #5243, opened 2026-05-16
Fix slow downloads: larger buffers, Content-Length headers, keep-alive tuning

trevren11:
## Brief summary

Fixes slow audiobook downloads by increasing stream buffer sizes and switching to `res.sendFile()` which automatically provides `Content-Length` and `Accept-Ranges` headers for better download progress and resumability.

## In-depth Description

Node.js defaults to 16 KiB stream buffers, which causes many small write syscalls when downloading large audio files and limits throughput. The `res.download()` API also does not set `Content-Length`, so clients cannot show accurate progress bars or resume interrupted downloads.

**Changes:**
- Switched from `res.download()` to `res.sendFile()` in both single-file download paths. The `send` module used by `sendFile` automatically sets `Content-Length` and `Accept-Ranges`, enabling resumable downloads and accurate client progress reporting.
- Increased write-stream and archiver pipeline `highWaterMark` from 16 KiB to 512 KiB to reduce syscall overhead on large audio files.
- Raised `keepAliveTimeout` to 120s so clients downloading multiple files sequentially (e.g. individual chapters) reuse the existing TCP connection rather than opening a new one for each file.

## How have you tested this?

- All 359 existing + new tests pass (`npm test`)
- New unit tests added for `LibraryItemController.download`, `fileUtils.downloadFile` (real HTTP server integration test), and `zipHelpers.zipDirectoryPipe`
- Manual download of single audiobook files and multi-file zip archives

## Screenshots

No UI changes.

---

### `server/Server.js`
- `keepAliveTimeout = 120000` — extends the idle keep-alive window from Node's default 5s to 120s
- `headersTimeout = 125000` — must exceed `keepAliveTimeout` per Node.js requirements

### `server/controllers/LibraryItemController.js`
- Both single-file download paths (`download` for library items and `downloadLibraryFile` for individual files) replaced `res.download(path, filename, cb)` with:
  - Manual `Content-Disposition: attachment; filename="<url-encoded>"` header
  - `res.sendFile(path, cb)` — delegates to the `send` module which stat()s the file once and sets `Content-Length`, `ETag`, `Last-Modified`, and `Accept-Ranges`

### `server/utils/fileUtils.js`
- `fs.createWriteStream(filepath, { highWaterMark: 524288 })` — 512 KiB write buffer instead of 16 KiB default

### `server/utils/zipHelpers.js`
- `archiver('zip', { zlib: { level: 0 }, highWaterMark: 524288 })` — 512 KiB pipeline buffer on both `zipDirectoryPipe` and `zipDirectoriesPipe`

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -301,6 +301,11 @@ class Server {
     app.disable('x-powered-by')
 
     this.server = http.createServer(app)
+    // Keep connections alive for 120s so multi-file downloads don't pay a new
+    // TCP handshake per file. Default Node.js keepAliveTimeout is only 5s,
+    // which causes connections to drop mid-download for files > 5s.
+    this.server.keepAliveTimeout = 120000
+    this.server.headersTimeout = 125000
 
     router.use(
       fileUpload({
--- a/server/Server.js
+++ b/server/Server.js
@@ -301,9 +301,9 @@ class Server {
     app.disable('x-powered-by')
 
     this.server = http.createServer(app)
-    // Keep connections alive for 120s so multi-file downloads don't pay a new
-    // TCP handshake per file. Default Node.js keepAliveTimeout is only 5s,
-    // which causes connections to drop mid-download for files > 5s.
+    // Extend keep-alive timeout to 120s. Node's default (5s) closes idle
+    // connections quickly, forcing a new TCP handshake for each file when a
+    // client downloads multiple files in sequence (e.g. individual chapters).
     this.server.keepAliveTimeout = 120000
     this.server.headersTimeout = 125000
 
--- a/server/controllers/LibraryItemController.js
+++ b/server/controllers/LibraryItemController.js
@@ -187,7 +187,7 @@ class LibraryItemController {
         // Use sendFile so the `send` module sets Content-Length, Accept-Ranges,
         // and handles range requests natively — no extra stat() call needed.
         res.setHeader('Content-Disposition', `attachment; filename="${encodeURIComponent(Path.basename(req.libraryItem.relPath))}"`)
-        await new Promise((resolve, reject) => res.sendFile(libraryItemPath, { dotfiles: 'allow' }, (error) => (error ? reject(error) : resolve())))
+        await new Promise((resolve, reject) => res.sendFile(libraryItemPath, (error) => (error ? reject(error) : resolve())))
       } else {
         const filename = `${itemTitle}.zip`
         await zipHelpers.zipDirectoryPipe(libraryItemPath, filename, res)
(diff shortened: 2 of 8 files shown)


server/Server.js
```diff
@@ -301,6 +301,11 @@ class Server {
     app.disable('x-powered-by')
 
     this.server = http.createServer(app)
+    // Keep connections alive for 120s so multi-file downloads don't pay a new
```
nichwall: Does this actually change anything? My understanding is the 5 second timeout only triggers after the download has completed (connection is not used), not every 5 seconds of a download when the connection is still active.

server/controllers/LibraryItemController.js
```diff
@@ -184,7 +184,10 @@ class LibraryItemController {
         if (audioMimeType) {
           res.setHeader('Content-Type', audioMimeType)
         }
-        await new Promise((resolve, reject) => res.download(libraryItemPath, req.libraryItem.relPath, (error) => (error ? reject(error) : resolve())))
+        // Use sendFile so the `send` module sets Content-Length, Accept-Ranges,
+        // and handles range requests natively — no extra stat() call needed.
+        res.setHeader('Content-Disposition', `attachment; filename="${encodeURIComponent(Path.basename(req.libraryItem.relPath))}"`)
+        await new Promise((resolve, reject) => res.sendFile(libraryItemPath, { dotfiles: 'allow' }, (error) => (error ? reject(error) : resolve())))
```
nichwall: I don't think we want to add dotfiles to the download, unless there is a specific feature request for it.

nichwall (commented):
Thanks for opening a PR. We have not been reviewing AI-generated PRs due to an increasing number of low-quality PRs and limited reviewer time, but this one seems pretty straightforward.

Can you update the PR to use the PR template (for example, part of the template is to check if any open issues or feature requests are resolved so it is easier for us to review and close issues)?

I have not ran the server or any tests, but some initial comments attached.

server/Server.js
trevren11: keepAliveTimeout only applies to idle connections, not active transfers, so it doesn't help mid-download. I've updated the comment to clarify: the benefit is for clients that download multiple files sequentially (e.g. individual chapters), where the idle gap between requests could otherwise exceed 5s and force a new TCP handshake for each file.

server/controllers/LibraryItemController.js
trevren11: Removed

server/Server.js
nichwall: Was this included in real-world testing? Are there other side effects or benefits to increasing the idle TCP connection lifetime?
Not changing this and having a TCP handshake for a client that takes more than 5 seconds to request the next file seems reasonable, not sure why we need to wait up to 2 minutes for it to request the next file.

trevren11: Fair point. The TCP handshake overhead is minimal on modern connections, and 5s between sequential requests is already generous for most clients. I can remove the keepAliveTimeout change entirely if you prefer keeping the default behavior. The other changes (larger stream buffers and sendFile for Content-Length headers) are the more impactful improvements.

PR #5255, opened 2026-05-22
Add unique UUID to access and refresh tokens

### LINKED ISSUE #5253: [Bug]: Duplicate refresh tokens across sessions can cause unexpected logout
### What happened?

Opening a separate issue as requested from #4630.

I can reproduce this on v2.35.0.

It looks like two existing sessions for the same user can receive the same new refresh token if they call `/auth/refresh` within the same second. This does not require the two logins to happen at the same time. The sessions can be created at different times and start with different refresh tokens.

Session A and session B should keep independent refresh tokens. Refreshing or logging out one session should not invalidate another session for the same user.

1. Log in as the same user from session A.
2. Wait at least one second.
3. Log in as the same user from session B.
4. Confirm the initial refresh tokens are different.
5. Call `/auth/refresh` for both sessions at the same time, using each session's own `x-refresh-token`.
6. Both responses return the same new `user.refreshToken`.
7. Call `/logout` with session A's refreshed token.
8. Try `/auth/refresh` with session B's refreshed token.

Observed:

```text
device A login: refresh=e657984fd670
device B login: refresh=2abffdb3cf0c
initial tokens same: false

device A simultaneous refresh: returnedRefresh=e78ebcbc506d
device B simultaneous refresh: returnedRefresh=e78ebcbc506d
refreshed tokens same: true

logout device A refreshed token: 200
refresh device B refreshed token after A logout: 401 Invalid refresh token

### Audiobookshelf version

2.35.0

Docker

Linux

None

### Logs

```shell
[TokenManager] Failed to refresh token. Session not found
```

### Additional Notes

Possible cause:

The generated refresh token does not appear to include a per-session or per-token unique value, such as sessionId, jti, or a random nonce. Since JWT timestamps are second-based, two refreshes for the same user within the same second can produce the same token.

nichwall:
## Brief summary

Add guaranteed unique field to Access and Refresh tokens to ensure generating new tokens during the same second do not result in a collision for the same user.

## Which issue is fixed?

Fixes https://github.com/advplyr/audiobookshelf/issues/5253

## In-depth Description

The referenced issue should only occur when a user tries to refresh two separate tokens within 1 second, due to the `jsonwebtoken` library removing milliseconds from the expiration time and signing identical information with the same secret.

This PR adds a new JTI field (consisting of a UUIDv4) to all tokens to ensure new token uniqueness. I chose UUIDv4 because we already use v4 elsewhere in the server to be independent of generation time or ordering (including v1, v2, v6, v7). We cannot easily use the session ID to ensure token uniqueness (one suggestion included from #5253 ) because the token is used to create the session in the database, not the other way around.

This new JTI is only included on new tokens, so existing tokens will *not* force regeneration before they expire. I chose to not force the regeneration because this adds some complexity to validation to see if the JTI exists, and would prevent this from being an invisible update. The default access token expiration of 1 hour would automatically update nearly all user tokens almost immediately. Based on conversations on GitHub, Discord, and Reddit, pretty much every user who changed the default access token expiration did it because of the prior logout issue https://github.com/advplyr/audiobookshelf/issues/4630, but they will probably just continue using a long access time token unless there is a breaking change in the future.

## How have you tested this?

- Added temporary debug print to dump contents of JWT, verified UUID is present during verification.
- Ensured new tokens replaced old tokens from before the change with no user intervention

## Screenshots

N/A

### DIFF
--- a/server/auth/TokenManager.js
+++ b/server/auth/TokenManager.js
@@ -1,4 +1,5 @@
 const { Op } = require('sequelize')
+const uuid = require('uuid')
 
 const Database = require('../Database')
 const Logger = require('../Logger')
@@ -115,6 +116,7 @@ class TokenManager {
     const payload = {
       userId: user.id,
       username: user.username,
+      jti: uuid.v4(),
       type: 'access'
     }
     const options = {
@@ -138,6 +140,7 @@ class TokenManager {
     const payload = {
       userId: user.id,
       username: user.username,
+      jti: uuid.v4(),
       type: 'refresh'
     }
     const options = {


advplyr:
Makes sense, thanks!

PR #5280, opened 2026-06-01
fix: collapsed series with filters

### LINKED ISSUE #2807: [Bug]: Series do not collapse for User when tag restrictions are set for them
### Describe the issue

When a user has (positive) tag restrictions (ie. only allowed to see content with certain tags), Collapse Series does not work.

1. Untick "Can Access All Tags" for a user in user's settings
2. Add desired tag(s) to "Tags Accessible to User"
3. Make sure "Invert" is NOT ticked.
7. Login as said user (if not already)
8. Go to audiobook library
9. BUG: Observe that "Collapse Series" checkbox does nothing.

### Audiobookshelf version

v2.8.1

Docker

### LINKED ISSUE #3049: [Bug]: Collapse Series Option in the Library View  Doesn't Work When a Filter Is Enabled
### What happened?

I had a tag filter enabled and the "Collapse Series" box checked. The tag filter was applied correctly but all titles in a series were still listed separately.

I expected series to be collapsed after filtering.

### Prerequisites

* Having a library that has multiple titles in the series

### Steps

1. Navigate to the "Library" view.
2. Select the "Collapse Series" box.
3. Add a filter (I tested both with a tag filter and a genre filter)

Notice that series are not collapsed.

### Audiobookshelf version

v2.10.1

Docker

Linux

Safari

### Logs

_No response_

### Additional Notes

I also tried this on Chrome and Firefox but got the same results.

schummar:
## Brief summary

The "Collapse Series" feature in the Library view doesn't work when certain filters are active - for example a Tag filter.
And it never worked for users whose access is restricted by tags, because they always use a Tag filter implicitly.
This PR fixes the broken query.

## Which issue is fixed?

Fixes #2807
Fixes #3049

## In-depth Description

The filter results in a `(SELECT count(*) FROM json_each(tags) WHERE json_valid(tags) AND json_each.value = :filterValue)` query but the `replacements` were not included before, so `:filterValue` could not be resolved and the query returned empty.

There remains a related, but independent issue for users that are restricted by tags, which should be tackled in a separate PR: The count badge on the series tile includes all books in the series. That can be confusing for restricted users because the count does not match what they see once the open the series up and see less books.

## How have you tested this?

I started a fresh instance of ABS, created two users - a root user and a user restricted by a tgs.
Then I uploaded a few books and tagged some of them with the same tag.
I tested both the root users's Library view with an active tag filter and the restricted user's Library view without filters.
In both cases the "Collapse Series" option did nothing before the fix and worked after the fix.

## Screenshots

Before:
image

After:
image

### DIFF
--- a/server/utils/queries/libraryItemsBookFilters.js
+++ b/server/utils/queries/libraryItemsBookFilters.js
@@ -311,7 +311,7 @@ module.exports = {
    */
   async getCollapseSeriesBooksToExclude(bookFindOptions, seriesWhere, replacements = {}) {
     const allSeries = await Database.seriesModel.findAll({
-      attributes: ['id', 'name'],
+      attributes: ['id', 'name', [Sequelize.literal('(SELECT count(*) FROM bookSeries bs WHERE bs.seriesId = series.id)'), 'numBooks']],
       distinct: true,
       subQuery: false,
       replacements,
@@ -339,7 +339,7 @@ module.exports = {
           booksToInclude.push(book.id)
           bookSeriesToInclude.push({
             id: book.bookSeries.id,
-            numBooks: s.books.length,
+            numBooks: s.dataValues.numBooks,
             libraryItemIds: s.books?.map((b) => b.libraryItem.id) || []
           })
           booksToExclude = booksToExclude.filter((bid) => bid !== book.id)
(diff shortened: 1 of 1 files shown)


advplyr:
Nice catch, thanks!

PR #5282, opened 2026-06-02
fix: write metadata.json for podcast libraries when storeMetadataWithItem is enabled

### LINKED ISSUE #5093: [Bug]: Podcast Metadata Not Written to Disk
### What happened?

Global toggle for writing metadata next to the library item is not applied to podcast libraries. 
There is no metadata.json written alongside files for 'podcast' type libraries so any changes are lost when moving the library around.

A metadata.json file to be written next to the podcast episode so metadata is persistent.

1. Settings -> Store metadata with item is toggled ON
2. Add a podcast library and some podcasts
3. Create/update episode/title/description/etc
4. no metadata.json is written

### Audiobookshelf version

2.32.1

Docker

Linux

None

### Logs

```shell
audiobookshelf  | [2026-02-28 10:02:50.269] INFO: [PodcastController] Updated episode "Death in the Stars" keys [ 'description' ]
```

### Additional Notes

#1488 #4959 #4594 and #4304 are centered around embedding this in the mp3 tags directly, and so don't really address this inconsistency.
Having the data stored next to the library files helps with portability since restoring backups also overwrites system settings that may not be applicable. 

Since podcast libraries are structured differently (flat, compared to books) it might work best to write it as {{episodeFile}}.json but I can also see a benefit to writing a single metadata.json for the whole podcast and just appending episodes to that as they're added.

AruneshDwivedi:
## What is changed

BookScanner has the check 'storeMetadataWithItem && !existingLibraryItem.isFile' when setting libraryItemUpdated after writing metadata. PodcastScanner was missing the !isFile guard so file-type podcast items would still try to set libraryItemUpdated even though there is no directory to write metadata.json into.

This aligns PodcastScanner with how BookScanner already works.

Fixes #5093

## Review response

Fair point on the original code structure. I removed the else-if branch and the actual fix is just adding the missing !isFile check that BookScanner already has.

### DIFF
--- a/server/scanner/PodcastScanner.js
+++ b/server/scanner/PodcastScanner.js
@@ -240,11 +240,7 @@ class PodcastScanner {
     if (hasMediaChanges) {
       await media.save()
       await this.saveMetadataFile(existingLibraryItem, libraryScan)
-      libraryItemUpdated = global.ServerSettings.storeMetadataWithItem
-    } else if (global.ServerSettings.storeMetadataWithItem) {
-      // Always save metadata file when setting is enabled, even without media changes
-      await this.saveMetadataFile(existingLibraryItem, libraryScan)
-      libraryItemUpdated = true
+      libraryItemUpdated = global.ServerSettings.storeMetadataWithItem && !existingLibraryItem.isFile
     }
 
     if (libraryItemUpdated) {
--- a/server/scanner/PodcastScanner.js
+++ b/server/scanner/PodcastScanner.js
@@ -241,6 +241,10 @@ class PodcastScanner {
       await media.save()
       await this.saveMetadataFile(existingLibraryItem, libraryScan)
       libraryItemUpdated = global.ServerSettings.storeMetadataWithItem
+    } else if (global.ServerSettings.storeMetadataWithItem) {
+      // Always save metadata file when setting is enabled, even without media changes
+      await this.saveMetadataFile(existingLibraryItem, libraryScan)
+      libraryItemUpdated = true
     }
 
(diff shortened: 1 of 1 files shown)


nichwall:
What issue does this actually solve? The `metadata.json` is only written when there is an update to the library item (same as for books), there is no need to write it every time you scan the library.

The `metadata.json` does not include podcast episode data, so scanning and adding new episodes does not require updating the metadata file either.

AruneshDwivedi:
Fair point — the original code structure of only writing metadata on update is correct. I removed the else-if branch entirely.

The actual fix is just adding the missing `!existingLibraryItem.isFile` guard. BookScanner already has this check (`storeMetadataWithItem && !existingLibraryItem.isFile`) but PodcastScanner was missing it. That inconsistency was the real bug — file-type podcast items would set libraryItemUpdated true even though there is no directory to write metadata.json into.

Whole change is now just 1 line. Should be clean.

nichwall:
This still doesn't fix the referenced issue. The referenced issue is about changing the metadata structure stored in the metadata file, which will be a much more involved change.

AruneshDwivedi:
You're right, sorry about that. I misread the issue — it's about the metadata structure for podcasts being fundamentally different from books (flat vs directory-based), not just the isFile guard.

The real fix would need to handle writing metadata per-episode (episodeFile.json) or a single metadata.json for the whole podcast, which is a bigger change than I initially thought.

I'll close this PR since it doesn't address the actual issue. If I get time to look at the full podcast metadata structure I'll open a new one.

PR #5283, opened 2026-06-02
fix: case-insensitive author merge check on rename

### LINKED ISSUE #5278: [Bug]: Replacing author with Author: No updates were necessary
### What happened?

If you change the author from non capitalized to capitalized it won't update it, saying "No updates were necessary" - though they show up as distinct authors.

Thus changing a Book with author `max` to `Max` is impossible, though you will have Max and max as two different authors. Same issue present when mass-editing.

Changing the author works.

1. Create a book with author max, a second with Max.
2. Try changing the authors name to Max, click save, see the "No updates were necessary" popup.

Same with mass-editing.

### Audiobookshelf version

v2.35.1

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

AruneshDwivedi:
When renaming an author from e.g. max to Max, the merge check now uses case-insensitive comparison (Op.iLike) instead of exact match. Fixes #5278

### DIFF
--- a/server/controllers/AuthorController.js
+++ b/server/controllers/AuthorController.js
@@ -121,7 +121,9 @@ class AuthorController {
           id: {
             [sequelize.Op.not]: req.author.id
           },
-          name: payload.name,
+          name: {
+            [sequelize.Op.iLike]: payload.name
+          },
           libraryId: req.author.libraryId
         }
       })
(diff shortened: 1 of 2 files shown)


nichwall:
This PR has an unrelated commit. Please make sure your PR only includes relevant commits. Please also make sure you use the PR description template.

PR #5031, opened 2026-02-05
Revamp OIDC auth and add Back-Channel Logout

Sapd:
This PR revamps the OIDC system. Passport is removed. Support for groups mapping provided, and scopes can now be configured.

#### Passport was removed for OIDC

It clashed how we OIDC used. We are not only an OIDC relying party but also an OAuth2 proxy for mobile clients. Thats a use case the passport provider does not handle well, we did some hacks before to support that. The new code is much simpler

#### Add Server-Driven (or also called Schema-Driven) UI for the OIDC settings
 
Instead of defining the settings (again) in the frontend, the server provides a schema and the frontend automatically renders all settings. Minimizes code (less duplication) and bugs. I would also recommend that for all other settings, however I handled here only OIDC.

#### OIDC Mappings

Added group mappings. A group from the identity provider can be now directly mapped to a ABS group. Also scopes can now be configured. Fixes #2878 and Fixes #3006
Fixed also some edge cases of validation (Fixes #4744 )

#### Add verified email enforcement option

Every idP provider handles the email_verified field differently (some provide hardcoded true or false or do not provide it). The Admin can now configure how ABS should handle the field. Fixes #4832 
If the setting is turned on, it is expected that the IdP sends the value with and that its true.

#### Store OIDCToken in session table instead of cookie

The token is better placed server side. Also it can be quite long exceeding maximum allowed cookies length. 

#### Add Back-Channel Logout support

Authentik now supports Back-Channel logout. So I implemented it here, too.
There is a POST endpoint on ABS side. When the configuration is turned on, it accepts a signed JWT from Authentik and can cancel sessions.  
When a user logs out on Authentik, Authentik will use it to log out a user on ABS too. (The user's existing access token remains valid until it expires).

#### Other

Jose was added as explicit dependency for Backchannel Logout (there for checking the JWT). It was a dependency anyway from node-openid-client.

## How have you tested this?

Simply add a mapper for one on the groups and use as value an Authentik Group.

For Backchannel logout. Make sure to configure the URL in Authentik. Then check the sessions in ABS sqliteDB. Then log out in authentik. You should see in the ABS Console a message that there was a Back Channel logout. The session will be gone in the ABS sqliteDB.

Also added extensive unit tests.

## Screenshots

image

PR #5298, opened 2026-06-06
Add configurable OIDC group-to-role mapping (admin groups + default role)

cyberb:
## Problem

OIDC group-based role assignment currently only recognizes groups named **`admin`**, **`user`**, or **`guest`** (case-insensitive), and **denies login** when none of a user's groups match one of those names (`OidcAuthStrategy.setUserGroup`).

This makes the group claim feature unusable with identity providers whose group names don't happen to match the application's role names, and there's no way to admit such users with a sensible default role — their only options are to rename groups in the IdP (often not possible) or leave group mapping disabled entirely.

## Changes

Two optional server settings, both empty by default (so existing behavior is unchanged):

- **`authOpenIDAdminGroups`** — comma-separated list of group names from the group claim that should be mapped to the `admin` role, *in addition to* the built-in `admin`/`user`/`guest` names. Lets you grant admin to an IdP group like `my-org-admins` without renaming it.
- **`authOpenIDGroupDefaultRole`** — role assigned when none of the user's groups match a known role, instead of denying login. Set it to `user` or `guest` to admit unmatched users with that role; leave it empty to keep the current deny-on-no-match behavior.

`setUserGroup` now folds any configured admin groups into the `admin` role before the existing priority match, then falls back to the default role when nothing matches.

Included:
- Server settings (`ServerSettings`) wiring (persist + `authenticationSettings`).
- Auth settings UI fields on the OpenID config page + `en-us` strings.
- Unit tests for `setUserGroup` covering admin-group mapping, built-in names still working, default-role fallback, deny-when-no-default, and the no-group-claim no-op.

## Backward compatibility

With both settings empty, role resolution is identical to today: only `admin`/`user`/`guest` are recognized and an unmatched user is denied.


Sapd:
Thats a duplicate of https://github.com/advplyr/audiobookshelf/pull/5031

The other PR is already finished but it waits for the audiobookshelfs switch to react.

PR #5304, opened 2026-06-10
Add finishedDates history to keep multiple finished dates per book

pounat:
## Summary

Currently a media progress record only keeps a single `finishedAt` timestamp. When a user re-listens to a book and finishes it again (or restarts it), the previous finished date is lost. This PR keeps a history of every finish event.

## Changes

**Server**
- New `finishedDates` JSON column on `mediaProgresses` (array of ms-since-epoch timestamps), added by migration `v2.36.0-add-finished-dates` which backfills it from the existing `finishedAt` value. Both `up` and `down` are idempotent, and unit tests are included.
- `MediaProgress.applyProgressUpdate()`:
  - Marking an item finished (manually or via auto-finish thresholds) appends to `finishedDates`.
  - Explicitly un-marking as finished removes the most recent entry (treated as an undo of an accidental mark).
  - Re-listening to a finished item (auto reset of `isFinished` when progress drops below the threshold) keeps the history, so the previous finish date is preserved.
- `finishedAt` keeps its existing behavior (most recent finish) so existing API consumers are unaffected; `finishedDates` is additionally exposed on media progress API responses.
- Year in review stats (`getBookMediaProgressFinishedForYear`) now also match finish events from the history, so a book finished in 2025 and re-finished in 2026 counts in both years instead of disappearing from 2025.

**Client**
- The item page progress box shows a "Previously finished" line (with a tooltip listing all previous finish dates) when a finish history exists.

## Notes for review

- The migration is named `v2.36.0` assuming that is the next release version — happy to rename if the next release will be versioned differently.
- `finishedDates` is server-managed; values sent by clients in progress payloads are ignored.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/pages/item/_id/index.vue
+++ b/client/pages/item/_id/index.vue
@@ -73,6 +73,9 @@
             <p v-else class="text-xs">{{ $strings.LabelFinished }} {{ $formatDate(userProgressFinishedAt, dateFormat) }}</p>
             <p v-if="progressPercent < 1 && !useEBookProgress" class="text-gray-200 text-xs">{{ $getString('LabelTimeRemaining', [$elapsedPretty(userTimeRemaining)]) }}</p>
             <p class="text-gray-400 text-xs pt-1">{{ $strings.LabelStarted }} {{ $formatDate(userProgressStartedAt, dateFormat) }}</p>
+            <ui-tooltip v-if="finishedDateHistory.length" :text="finishedDateHistoryTooltip" direction="top" class="inline-block">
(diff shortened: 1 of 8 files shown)


nichwall:
Thanks for opening a PR, but we are not reviewing AI generated PRs at this time due to an increasing number of low-quality PRs and limited reviewer time.

Additionally, the front end is currently being migrated away from Vue and no changes to the existing frontend will be merged because they will be obsolete within a few weeks/months.

Vito0912:
I strongly advise against this change as long as progress is only part of the user object. Adding another JSON field and more return values will make the user response even larger. 

Especially since this is not strictly needed in the progress object because the session already tracks past finishes. You can see when and how often a book was finished by looking at the books sessions.

pounat:
Understood, thanks for the replies. I'll close this out.

Issue #3953, opened 2025-02-08
[Bug]: Cannot upload any audiobook


ambrusnemeth:
### What happened?

Unable to upload anything.

### What did you expect to happen?

I tried to upload a folder with 35 mp3 files.
For several minutes, i have seen this:

![Image](https://github.com/user-attachments/assets/85feda95-0574-4798-8081-1a20e608fadc)

Then, this:

![Image](https://github.com/user-attachments/assets/6b58a42c-0a50-423c-806c-dd29898c7184)

There was an error message about "Socket disconnected" or something like that.

### Steps to reproduce the issue

1. install audiobookshelf on ubuntu server
2. try to upload any folder

### Audiobookshelf version

v2.19.0

### How are you running audiobookshelf?

Debian/PPA

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

### Logs

```shell

```

advplyr:
The socket disconnect is usually because the server is crashing.
Please check the logs.

ambrusnemeth:
There was a problem with the teleipt, audiobookshelf was not allowed to create a new folder.
I fixed it, and now it's working.

Issue #3746, opened 2024-12-23
[Bug]: Upload Large file fails


Silther:
### What happened?

Often when uploading a new books via remote connection (Wireguard  or reverse proxy caddy) the upload fails on large books.

But if i try to upload the same book via Wireguard and ssh (scp local server) it works without any issue.
It takes around 5-10 min but no connection error or something else occurs.

I am not completely sure if it only happens with additionally an slow internet connection, but it may be so.

### What did you expect to happen?

Upload of large books (with slow internet) should work without any problems.

### Steps to reproduce the issue

0. (have slow internet connection
![image](https://github.com/user-attachments/assets/939ea245-3a11-4fc8-8208-b203620c17d3)

1. upload book with around 1gb size
2. upload will take for ever (30+ min) and then fail

### Audiobookshelf version

2.17.5

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
2024-12-23 14:33:18.067

INFO

[SocketAuthority] Socket Connected to /socket.io PAWRwANhuhUCS51SAAAx

2024-12-23 14:33:29.063

INFO

[SocketAuthority] Socket Connected to /socket.io QFKjlmjecIuku7c2AAAz

2024-12-23 14:35:10.046

INFO

[SocketAuthority] Socket Connected to /socket.io 6A-I8exjzx0YFZBcAAA1

2024-12-23 14:36:59.546

INFO

[SocketAuthority] Socket Connected to /socket.io Oqn5XDKuhzlJyQErAAA3

2024-12-23 14:37:44.512

INFO

[SocketAuthority] Socket Oqn5XDKuhzlJyQErAAA3 disconnected from client "yyy" after 44966ms (Reason: ping timeout)

2024-12-23 14:46:47.728

INFO

[SocketAuthority] Socket Connected to /socket.io xn21jhOWgeRm25l1AAA5

2024-12-23 14:57:29.297

INFO

[SocketAuthority] Socket Connected to /socket.io Gd7GqOz0huaQ53rRAAA7

2024-12-23 15:03:17.175

INFO

[SocketAuthority] Socket Gd7GqOz0huaQ53rRAAA7 disconnected from client "yyy" after 347878ms (Reason: ping timeout)

2024-12-23 15:03:28.481

ERROR

Invalid request, no files

2024-12-23 15:08:58.481

[...]

Issue #3733, opened 2024-12-19
[Bug]: upload issues


nathanapples:
### What happened?

When I attempt to upload a book format that already has a different format of the same book in the library, a toast message appears stating: Filepath "/[path]/[author]/[book]" already exists on server

### What did you expect to happen?

I expected the book to be added as another format option to the original upload.

### Steps to reproduce the issue

1. Pick a book you already have in a library (e.g. mp3)
2. upload a different format of the same book (e.g. epub)

### Audiobookshelf version

v2.17.5

### How are you running audiobookshelf?

Docker

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

Firefox

nichwall:
Multiple audio formats for the same book are not supported. https://github.com/advplyr/audiobookshelf/issues/2315

Uploading an ebook file to an existing audiobook: https://github.com/advplyr/audiobookshelf/issues/1894

advplyr:
Duplicate of #2315 and #1894

Issue #862, opened 2022-07-27
[Bug]: Crash uploading a file > 4 GiB


MM2-0:
### Describe the issue

Using the web interface to upload a file larger than 4 GiB will crash the server.
```
/server/libs/njodb/index.js:103
         throw error;
         ^
 
 RangeError [ERR_OUT_OF_RANGE]: The value of "length" is out of range. It must be >= 0 && <= 4294967296. Received 4_676_987_052
     at validateOffset (node:buffer:113:3)
     at Function.concat (node:buffer:550:5)
     at getBuffer (/server/libs/expressFileupload/memHandler.js:17:34)
     at complete (/server/libs/expressFileupload/memHandler.js:37:14)
     at FileStream.<anonymous> (/server/libs/expressFileupload/processMultipart.js:118:17)
     at FileStream.emit (node:events:539:35)
     at endReadableNT (node:internal/streams/readable:1345:12)
     at processTicksAndRejections (node:internal/process/task_queues:83:21) {
   code: 'ERR_OUT_OF_RANGE'
 }
```

Copying the file manually to the library directory works fine as a workaround.

I realize that 4 GiB is quite large and that it would probably be a good idea to split that file into multiple parts (or use better encoding settings). I think it'd be fine if the web interface just displayed an error message when a user attempts to upload a file that large.

### Steps to reproduce the issue

1. Use the web interface to upload a new audiobook
2. Select a file larger than 4294967296 bytes
3. The server will crash during the upload (not immediately, presumably after 4294967296 bytes have been uploaded) and the upload fails (no file exists at the target location)

### Audiobookshelf version

v2.1.0

### How are you running audiobookshelf?

Docker

advplyr:
Fixed in [v2.2.23](https://github.com/advplyr/audiobookshelf/releases/tag/v2.2.23)

PR #5305, opened 2026-06-10
Add chunked, resumable uploads

### LINKED ISSUE #5108: [Enhancement]: Chunked Upload
### Type of Enhancement

Server Backend

I would like to request support for chunked uploads when uploading files to Audiobookshelf. Instead of sending the entire file(s) in a single request, the upload could be split into smaller chunks that are uploaded sequentially and then reassembled on the server.

Chunked uploads would improve reliability when uploading large audiobook files, especially over slow or unstable connections. It would also allow interrupted uploads to resume instead of restarting from the beginning, making the process more robust and user-friendly.

Additionally, this would help users running Audiobookshelf behind services like Cloudflare or certain application security tools that enforce strict request size limits. For example, Cloudflare has a 100 MB upload limit per request. By splitting uploads into smaller chunks, these limits could be worked around more easily.

Image

[Image source](https://stackoverflow.com/questions/60989400/how-can-i-detect-a-file-change-during-a-chunking-upload)

v2.32.1

_No response_

cyberb:
## Brief summary

Adds chunked, resumable uploading to the web client. Instead of sending an entire item (a book can be dozens of files and several GB) as one multipart request, the client now splits each file into 5 MB chunks, uploads them sequentially with per-chunk retry, and the server reassembles them. If a file's chunks are already present (e.g. after a dropped connection), the client skips them and resumes where it left off.

## Which issue is fixed?

Fixes #5108

This also addresses the failure mode behind a number of older upload reports where a single large/long upload over a slow or unstable connection would fail and have to be restarted from zero (e.g. #862, #3733, #3746, #3953).

## In-depth Description

The current `submit()` flow posts one `multipart/form-data` request per item containing every file. On a slow or flaky connection that single request can run for many minutes; any interruption (proxy/body timeout, dropped Wi-Fi, mobile handover) aborts the whole upload and all progress is lost. Reverse proxies in front of Audiobookshelf also commonly cap request body time/size, which makes large multi-file uploads fail outright.

This PR keeps the existing item model but changes how the bytes move:

**Client (`client/pages/upload/index.vue`)**
- For each file, slice it into 5 MB chunks and `POST /api/upload/chunk` sequentially. Each chunk carries `uploadId`, `fileIndex`, `chunkIndex`, `numChunks`.
- Before uploading a file, `GET /api/upload/chunk/:uploadId/:fileIndex` to learn which chunks the server already has and skip them — this is what makes a retried upload resume instead of restart.
- Per-chunk retry with backoff (5 attempts) so a single dropped chunk doesn't fail the item.
- Progress is aggregated across chunks so the existing per-item progress bar is unchanged for the user.
- After all files are sent, `POST /api/upload/finalize` with the item metadata and per-file chunk counts.

**Server (`server/controllers/MiscController.js`, `server/routers/ApiRouter.js`)**
- `POST /api/upload/chunk` — stores one chunk under `<MetadataPath>/uploads/<uploadId>/<fileIndex>/<chunkIndex>`.
- `GET /api/upload/chunk/:uploadId/:fileIndex` — lists received chunk indices for resume.
- `POST /api/upload/finalize` — resolves the library/folder and output directory exactly as the existing `/api/upload` does, then streams each file's chunks together in order into the destination and removes the temp session.
- All path components are validated/sanitized: `uploadId` is restricted to a safe character set, `fileIndex`/`chunkIndex` must be integers, and file names go through the same `sanitizeFilename` as the legacy path, so a client cannot escape the uploads/library directories. Every endpoint enforces the same `canUpload` / library-access checks as `/api/upload`.

The legacy `POST /api/upload` endpoint is left untouched so the mobile apps and any API consumers keep working.

Why this approach: it reuses the existing item/library/folder resolution and the watcher-based scan (the assembled files simply land in the library folder, same as before), so it's a focused change to the transport rather than to how items are created. 5 MB keeps each request short enough to stay under typical proxy body timeouts while keeping the request count reasonable.

For context on motivation: this came out of packaging Audiobookshelf for [Syncloud](https://github.com/syncloud) (a self-hosting platform), where people run it on home connections and hit these large multi-file upload failures often. Happy to iterate on anything you'd like changed.

## How have you tested this?

- New server unit test `test/server/controllers/MiscController.test.js` (mocha/chai/sinon), covering: chunk storage, the resume listing endpoint, in-order assembly of a multi-chunk file, finalize failing when a chunk is missing, rejection of a path-traversal `uploadId`, and the permission check. `npm test` passes.
- End-to-end on a real deployment behind nginx: uploaded a single small file (1 chunk) and a 35-file folder (~321 MB, ~70 chunks) via the web UI; both upload, assemble, scan into the library, and play correctly. Manually interrupting the network mid-upload and retrying resumes from the already received chunks rather than restarting.

## Screenshots

The upload UI is unchanged (same per-item progress bar), so there is no visual difference to show; the change is in how the bytes are transferred.

### DIFF
--- a/client/pages/upload/index.vue
+++ b/client/pages/upload/index.vue
@@ -307,37 +307,87 @@ export default {
       }
     },
     async uploadItem(item) {
-      var form = new FormData()
-      form.set('title', item.title)
-      if (!this.selectedLibraryIsPodcast) {
-        form.set('author', item.author || '')
-        form.set('series', item.series || '')
-      }
-      form.set('library', this.selectedLibraryId)
-      form.set('folder', this.selectedFolderId)
+      const chunkSize = 5 * 1024 * 1024
+      const uploadId = `${Date.now()}-${Math.random().toString(36).slice(2)}`
+      const totalBytes = item.files.reduce((sum, file) => sum + file.size, 0)
+      let sentBytes = 0
 
-      var index = 0
-      item.files.forEach((file) => {
-        form.set(`${index++}`, file)
-      })
+      for (let fileIndex = 0; fileIndex < item.files.length; fileIndex++) {
+        const file = item.files[fileIndex]
+        const numChunks = Math.max(1, Math.ceil(file.size / chunkSize))
 
-      const config = {
-        onUploadProgress: (progressEvent) => {
-          if (progressEvent.lengthComputable) {
-            const progress = {
-              loaded: progressEvent.loaded,
-              total: progressEvent.total
-            }
-            this.updateItemCardProgress(item.index, progress)
+        let uploadedChunks = []
+        try {
+          const existing = await this.$axios.$get(`/api/upload/chunk/${uploadId}/${fileIndex}`)
+          uploadedChunks = existing.chunks || []
+        } catch (error) {
+          uploadedChunks = []
+        }
+
+        for (let chunkIndex = 0; chunkIndex < numChunks; chunkIndex++) {
+          const start = chunkIndex * chunkSize
+          const end = Math.min(file.size, start + chunkSize)
+          if (uploadedChunks.includes(chunkIndex)) {
+            sentBytes += end - start
+            this.updateItemCardProgress(item.index, { loaded: sentBytes, total: totalBytes })
+            continue
           }
+
+          const form = new FormData()
+          form.set('uploadId', uploadId)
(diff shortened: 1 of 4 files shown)


mclausing:
This would be really helpful for me - getting around cloudflare uplaod limits.

PR #5318, opened 2026-06-18
Enhance PodcastController to handle enclosure updates

### LINKED ISSUE #5317: [Bug]:  Manual episode match doesn't persist enclosure
### What happened?

When matching a podcast episode manually (Edit Episode → Match tab in the web client, or equivalent per-episode RSS search + apply), metadata such as title, subtitle, and description is saved, but the episode remains unlinked from the RSS feed. The Details tab continues to show **"Episode not linked to RSS feed"** instead of the RSS enclosure URL.

The web client sends an `enclosure` object in the `PATCH` body (see `client/components/modals/podcast/tabs/EpisodeMatch.vue`), but the server ignores it.

Quick match all episodes (`POST /api/podcasts/:id/match-episodes`) **does** persist enclosure via `Scanner.updateEpisodeWithMatch`, which sets `enclosureURL`, `enclosureSize`, and `enclosureType` on the episode model. Manual per-episode match uses a different code path and does not.

After selecting an RSS feed result in the episode Match UI, the episode should be linked to that RSS item: `enclosureURL` (and related fields) should be stored, and the episode edit UI should show the read-only **"Episode URL from RSS feed"** field — the same outcome as when quick match successfully links an episode.

1. Add or use a podcast library item with a valid RSS feed URL
2. Download a few episodes
3. Open episode edit → **Match** tab (or per-episode RSS search in a client that PATCHes on apply).
4. Search for an episode title that exists in the feed and select a matching RSS entry that has an enclosure URL.
5. Confirm the success toast / metadata update.
6. Open the episode **Details** tab (or re-fetch via `GET /api/podcasts/:id/episode/:episodeId`).
7. Observe `enclosure` is still `null` / UI still shows "Episode not linked to RSS feed".

### Audiobookshelf version

v2.35.1

Built from source

Windows

Chrome

### Logs

```shell

```

### Root cause

`PodcastController.updateEpisode` (`server/controllers/PodcastController.js`) only accepts:

- String keys: `title`, `subtitle`, `description`, `pubDate`, `episode`, `season`, `episodeType`
- `publishedAt` (number)
- `chapters` (array)

It does **not** handle `enclosure` or the flat model fields `enclosureURL`, `enclosureSize`, `enclosureType`.

mikiher:
## Brief summary

This updates enclosure values in `updateEpisode` if they exist in the request payload.

## Which issue is fixed?

Fixes #5317 

## In-depth Description

This is a small change that updates the enclosure values from the `updateEpisode` (`PATCH: /api/podcasts/:id/episode/:episodeId`) request payload. The web client already sends those, but they're currently ignored by the server.

## How have you tested this?

Manually matched an episode with no enclosure. 
After selecting the match, the details edit view shows the enclosure url, as expeceted.

### DIFF
--- a/server/controllers/PodcastController.js
+++ b/server/controllers/PodcastController.js
@@ -437,6 +437,17 @@ class PodcastController {
         }
 
         updatePayload[key] = req.body[key]
+      } else if (key === 'enclosure') {
+        const enclosure = req.body.enclosure
+        if (enclosure === null) {
+          updatePayload.enclosureURL = null
+          updatePayload.enclosureSize = null
+          updatePayload.enclosureType = null
+        } else if (typeof enclosure === 'object' && typeof enclosure.url === 'string') {
+          updatePayload.enclosureURL = enclosure.url
+          updatePayload.enclosureType = typeof enclosure.type === 'string' ? enclosure.type : null
+          updatePayload.enclosureSize = enclosure.length !== undefined && enclosure.length !== null ? enclosure.length : null
+        }
       } else if (key === 'chapters' && Array.isArray(req.body[key]) && req.body[key].every((ch) => typeof ch === 'object' && ch.title && ch.start)) {
         updatePayload[key] = req.body[key]
       } else if (key === 'publishedAt' && typeof req.body[key] === 'number') {


advplyr:
Thanks!

Issue #121, opened 2021-10-22
Enhancement: Keyboard control (Play/Pause with spacebar, etc.)


juriroemer:
Would love to be able to control AudioBookshelf via the usual keyboard inputs - some suggestions:

- space: play/pause
- arrow left/right: jump 10 seconds
- arrow up/down: raise/lower volume
- l: open/close playlist view
- escape: close player
- (m: mute/unmute)

advplyr:
I just added some of these in `v1.4.14`, but put it in experimental features because I haven't tested it enough yet.

If you have experimental features on it should work like this:

[Space] : Play/Pause
[ArrowLeft] : Jump backward 10s
[ArrowRight] : Jump forward 10s
[ArrowUp] : Increase volume 10%
[ArrowDown] : Decrease volume 10%
[M] : Mute/Unmute

If you have an input field focused OR any dialog is open it will ignore the hotkeys.

I just realized as I'm writing this that I forgot to check if the e-reader is open, so they will still work if the e-reader is open but I will fix that tomorrow.

I plan on mapping the escape key to close the player or any open dialogs, which is a bit more involved so I held off on that one for now.

I'm not sure what you meant by the playlist view.

juriroemer:
Wow, that was fast, exciting!
Mapping the escape key like that sound great. Mapping something like [Shift] + [Up/Down] to increase/decrease playback speed would also be nice.
Just had a super quick look at the code, what I meant to trigger/toggle with [L] is the showChapters() function.

advplyr:
I added these hotkeys and included in release [v1.5.0](https://github.com/advplyr/audiobookshelf/releases/tag/v1.5.0)

This is no longer experimental. I will need to put these somewhere people can find.

PR #5325, opened 2026-06-23
Let next.js handle file uploads through internal-api routes

mikiher:
## Brief summary

This passthrough is required for letting Next.js handle file upload routes on its own.

## Which issue is fixed?

No issue. 

This is required for [audiobokshelf-client-react PR #121](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/121)

## In-depth Description

In order for the new internal-api cover upload route in audiobokshelf-client-react to work correctly, We have to let it consume the multipart request body, so we have to exclude /internal-api/ requests from the server's express-fileupload middleware

## How have you tested this?

Large cover file uploads (~4MB) now work properly (before this change they ended in error)

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -302,7 +302,9 @@ class Server {
 
     this.server = http.createServer(app)
 
+    // Skip file upload parsing for internal-api routes (Next.js proxies read multipart bodies).
     router.use(
+      /^(?!\/internal-api).*/,
       fileUpload({
         defCharset: 'utf8',
         defParamCharset: 'utf8',


advplyr:
does this make the web client less secure in any way?

I forgot I also bypassed the JSON parsing when using internal-api which could also be a security issue. Not sure.

https://github.com/advplyr/audiobookshelf/blob/9b92b5de34b56f869b4f9eb348f98e8189864993/server/Server.js#L315-L316

For uploading backups and the upload page we worked around the limitation using this method:

https://github.com/audiobookshelf/audiobookshelf-client-react/blob/09a7bcdb3591ef1adbe8fce1dd7a65be0849b6c1/src/app/(main)/upload/UploadHelper.ts#L254-L299

If we decide that this PR is the proper solution for large files, then we should also update the backups and upload page to use the internal-api also.

mikiher:
I'm not a security expert, but I asked the AI to review this from a security perspective, and it says:

> Skipping fileUpload for /internal-api does not weaken auth or upload authorization. Cover uploads still require a valid session (httpOnly cookies) and are proxied to /api/items/:id/cover, where JWT auth, canUpload, and CoverManager validation all still apply. The change only prevents Express from consuming the multipart body before the Next.js proxy can read and forward it — the same reason we already skip express.json for /internal-api. The backend upload path and its security checks are unchanged.

It also mentions some minor considerations

> ### Minor considerations (worth mentioning, not blockers)
> 1. DoS / large bodies — On the browser → Next hop, the body is buffered in memory (arrayBuffer()). express-fileupload isn't applying its limits there anymore. In practice, the same authenticated user could already POST large multipart payloads to /api/... (Vue client, library upload, etc.), so this isn't a new capability — it's a slightly different code path. ***A body size cap in the Next route would be defense-in-depth if desired***.
> 2. CSRF — Cookie-authenticated POSTs to internal-api have the same CSRF posture as other internal-api POSTs (login, refresh). This change doesn't alter that; it's pre-existing for the React client architecture.
>3. Broader /internal-api skip — Skipping fileUpload for all /internal-api routes is slightly broader than cover upload alone, but no other route currently accepts multipart. Other POSTs expect JSON and would fail parsing if sent as multipart — same as before with the express.json skip.

Bottom-line, there does not seem to be any glaring new security issue. 
- If you want, we can put a limit on uploaded file size as 1 suggests.
- Also, if you want, I can reduce the scope of the skip as 3 suggests.
- Finally, if we want to use more idiomatic Next code, we can use 'request.formData()` instead of `request.arrayBuffer()`, though I don't think that makes a lot of difference as the route mostly just acts as a proxy.

Another advantage of this approach is that it retries with refreshed tokens (similar to other route handlers), whereas the other upload method seem to bypass Next altogether and will (likely) fail if the access token expires.

I can certainly move the other uploads to use the same approach. Do you want me to do that in the same PR?

advplyr:
One benefit of the `XMLHttpRequest` is that we can easily get upload progress. For cover image uploads and backup uploads it doesn't matter so much, but we should keep it for the upload page. I'm not sure how we would get progress feedback any other way.

If we are keeping the XMLHttpRequest for the upload page to get progress feedback, then maybe we create a separate better helper function for those that supports token refresh? Then use that for all file uploading.

I'm not sure what the best approach is. It would be nice to have the uploads handled the same way.

mikiher:
> One benefit of the `XMLHttpRequest` is that we can easily get upload progress. For cover image uploads and backup uploads it doesn't matter so much, but we should keep it for the upload page. I'm not sure how we would get progress feedback any other way.

I haven't checked yet, but can't we get that through websocket updates?

> If we are keeping the XMLHttpRequest for the upload page to get progress feedback, then maybe we create a separate better helper function for those that supports token refresh? Then use that for all file uploading.

Sorry, I keep forgetting all the time, so please remind me again - if we can automatically refresh tokens and bypass server actions by going to the server api directly, why aren't we doing this for all API calls?

> I'm not sure what the best approach is. It would be nice to have the uploads handled the same way.

advplyr:
> Sorry, I keep forgetting all the time, so please remind me again - if we can automatically refresh tokens and bypass server actions by going to the server api directly, why aren't we doing this for all API calls?

Most of the API calls are done in `api.ts` since those are requested server side.

The ones where we need to make client side requests we use the `/internal-api` primarily for convenience. In nextjs we are storing the tokens in cookies specifically so that those tokens are available when server side rendering. So when we do a token refresh we have to update those cookies, and that can only be done server side (specifically the nextjs server, not the abs server). Using the nextjs routes makes for an easier token refresh.

The file download routes are because we can put those `/internal-api/` URLs directly in  and <a> tags.

The refresh route is necessary for handling refreshes in the proxy.ts (previously middleware)

mikiher:
OK, thank you, I think I understand (again, until I forget once more 😄 ). 

Anyway, what's the verdict on this PR? As explained above, I don't think it counts as a security risk, and I think it may be possible to convert the other upload requests to use the same method (including websocket based progress updates, which will likely require another server change).

advplyr:
Thanks!

PR #5338, opened 2026-06-29
Add infinite scroll + load-more to podcast "Latest" page

### LINKED ISSUE #3343: [Enhancement]: Show more episodes on the "Latest" podcast library page
### Type of Enhancement

Web Interface/Frontend

As traditional podcatchers/podcasts clients, would be nice to show on "last" section all new pordcasts from all subscriptions and don't reset them over the last 24h.

Sometimes i just forgot to play an episode, and with some days those episodes disappear from the "latest" section, making me to go manually to the podcast profile under "Library" section to find it. Thats how AntennaPod/Pocket Cast and other podcast clients work.

Well, instead a 24h limit to reset the "Latest" section, if posible, it could be resetted within a week or "infinite". Would be nice to have a setting for delimite it.

2.12.3

_No response_

wakamex:
## Brief summary
Add infinite scroll to Latest page.
Fixes #3343.

## Which issue is fixed?
Addresses the request to see more episodes in the Latest page in #3343.

## In-depth Description
`latest.vue` fetched only the first page (`limit=50`) once and never loaded more. This adds a scroll listener that appends the next page near the bottom, with a "Load more" button as fallback, stopping when a page returns fewer than the page size. The endpoint already accepts `limit`/`page`, so it's client-only. The button reuses existing strings.

## How have you tested this?
I ran both the current release and this build locally. With the change, the Latest page keeps extending the list as I scroll. I couldn't run the Cypress component tests on my headless box, but they only cover isolated components. I only modified the latest.vue page, and the client builds clean.

## Screenshots
Before: oldest episode is from 3 days ago
Screenshot 2026-06-29 at 4 24 11 PM

After: I can keep scrolling to show episodes from 7d+ ago
Screenshot 2026-06-29 at 4 20 18 PM

### DIFF
--- a/client/pages/library/_library/podcast/latest.vue
+++ b/client/pages/library/_library/podcast/latest.vue
@@ -72,6 +72,11 @@
           </div>
           <div :key="index" v-if="index !== recentEpisodes.length" class="w-full h-px bg-white/10" />
         </template>
+        <div v-if="recentEpisodes.length && hasMore" class="flex justify-center pt-6">
+          <button class="h-9 px-5 border border-white/20 hover:bg-white/10 rounded-full text-sm font-semibold focus:outline-hidden" :disabled="processing" @click="loadMoreEpisodes">
+            {{ processing ? $strings.MessageLoading : $strings.LabelMore }}
+          </button>
+        </div>
       </div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


nichwall:
Thanks for opening a PR, but frontend changes are not being reviewed or merged due to the web client being fully rewritten and migrated to React.

PR #5341, opened 2026-07-01
Add all minified fields to expanded library item JSON

mikiher:
## Brief summary

Makes `toOldJSONExpanded()` a **strict superset** of `toOldJSONMinified()` for library items, books, and podcasts. Socket `item_updated` payloads and expanded API responses now include minified count/format fields (`numTracks`, `numEpisodes`, `numFiles`, etc.) that were previously missing, so clients can replace shelf entities from socket events without losing data.

This is an **additive-only** change: no fields removed or renamed on existing shapes.

## Which issue is fixed?

No issue

## In-depth Description

Home shelves and list endpoints load **minified** items (`toOldJSONMinified()`), but socket updates send **expanded** items (`toOldJSONExpanded()`). Expanded metadata was already a superset of minified metadata, but at the **media** and **library-item** levels several minified-only fields were absent from expanded:

| Model | Fields missing from expanded |
|-------|------------------------------|
| `Book` | `numTracks`, `numAudioFiles`, `numChapters`, `ebookFormat` |
| `Podcast` | `numEpisodes` |
| `LibraryItem` | `numFiles` |

When clients merged socket updates into minified shelf state, those fields disappeared and UI could regress (track counts, ebook format badges, etc.).

## Solution

Refactored `toOldJSONExpanded()` on `Book`, `Podcast`, and `LibraryItem` to **spread `toOldJSONMinified()` first**, then add expanded-only fields (`tracks`, `audioFiles`, `episodes`, `libraryFiles`, etc.). New minified fields added in the future are inherited by expanded unless explicitly overridden.

Documented the contract in JSDoc on each minified/expanded pair.

## How have you tested this?

Added `test/server/models/oldJsonSerialization.test.js` with book, podcast, and library-item fixtures asserting every minified key exists on expanded with matching values, plus consistency checks (`numTracks === tracks.length`, `numFiles === libraryFiles.length`, etc.).

On the React client:
- Edit an epub book from the library homepage and save. After saving, the epub overlay badge should still appear (before this fix it disappeared because expanded item didn't include `media.ebookFormat`)

### DIFF
--- a/server/models/Book.js
+++ b/server/models/Book.js
@@ -647,6 +647,11 @@ class Book extends Model {
     }
   }
 
+  /**
+   * Minified book JSON for list/shelf endpoints.
+   * `toOldJSONExpanded()` must be a strict superset: every key here must exist in expanded
+   * with the same value semantics. Only additive changes to expanded; never remove or rename keys.
+   */
   toOldJSONMinified() {
     if (!this.authors) {
       throw new Error(`[Book] Cannot convert to old JSON because authors are not loaded`)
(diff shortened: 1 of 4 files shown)


advplyr:
Thanks!

Issue #134, opened 2021-10-24
Issue with multiple versions of the same book


jarrodCoombes:
Folder structure looks like this:

```
Asimov, Isaac
├───1950 - I, Robot (Askey)
├───1950 - I, Robot (Brick)
├───1950 - I, Robot (Hagon)
├───1950 - Pebble in the Sky (Fass)
├───1950 - Pebble in the Sky (Overton)
├───1951 - Foundation (Brick)
├───1951 - Foundation (Donley)
├───1951 - Foundation (Mckeever)
├───1951 - The Stars, Like Dust (Tenny)
├───1951 - The Stars, Like Dust (Thorne)
├───1952 - Foundation & Empire (Brick)
├───1952 - Foundation & Empire (Lazar)
├───1952 - Foundation & Empire (Leclercq)
├───1952 - The Currents of Space (Collins)
├───1952 - The Currents of Space (Polk)
├───1953 - Second Foundation (Brick)
├───1953 - Second Foundation (Lazar)
├───1953 - Second Foundation (Metzler)
├───1954 - The Caves of Steel (Dufris)
├───1954 - The Caves of Steel (Kliban)
├───1955 - The End of Eternity (Boehmer)
├───1957 - The Naked Sun (Dufris)
├───1957 - The Naked Sun (Fox)
├───1959 - Nine Tomorrows (Fox)
├───1966 - Fantastic Voyage (Avers)
├───1968 - Asimov's Mysteries (Lazar)
├───1969 - Nightfall and Other Stories (Major)
├───1972 - The Gods Themselves (Brick)
├───1972 - The Gods Themselves (LeClaire)
├───1973 - The Best of Isaac Asimov (Lazar)
├───1976 - Murder at the ABA (Grace)
├───1982 - Foundation's Edge (Brick)
├───1982 - Foundation's Edge (Cartwright)
├───1982 - Foundation's Edge (DeLotel)
├───1982 - Foundation's Edge (Mckeever)
├───1982 - The Complete Robot (Mckeever)
├───1983 - The Robots of Dawn (Askey)
├───1983 - The Robots of Dawn (Dufris)
├───1983 - The Winds of Change and Other Stories (Johnson)
├───1985 - Robots & Empire (Ward)
├───1986 - Foundation & Earth (DeLotel)
├───1986 - Foundation & Earth (Mckeever)
├───1986 - Robot Dreams (Russotto)
├───1986 - Science Fiction Favorites (Asimov)
[...]

PR #5354, opened 2026-07-09
Emit author socket events when book scans create or link authors

mikiher:
## Brief summary

Emit `author_added` and `author_updated` socket events from `BookScanner` when library scans create or link authors.

## Which issue is fixed?

No issue.

## In-depth Description

Book scans were creating and linking authors without emitting socket events. Metadata edits via `Book.updateAuthorNames()` already emitted these events, but the scan paths in `BookScanner` did not.

This adds emissions for:
- **New book scan** — `author_added` for newly created authors, `author_updated` for linked existing authors (with `numBooks`)
- **Rescan / update existing book** — same events when authors are added during metadata reconciliation

Payload shapes match existing emitters (`author.toOldJSON()` / `author.toOldJSONExpanded(numBooks)`), so this is backward compatible.

### `author_added` (new author created during scan)

1. Remove a book whose author has no other books (soft delete)
2. On a different window, open the library Authors page
3. Rescan the library
4. Confirm the author **reappears on the Authors page** without a full page refresh

### `author_updated` (existing author re-linked during scan)

1. Remove a book whose author has **two books** in the library (soft delete)
2. On a different window, open the library Authors page
3. Rescan the library
4. Confirm the author’s **book count updates to 2** on the Authors page without a full page refresh

### DIFF
--- a/server/scanner/BookScanner.js
+++ b/server/scanner/BookScanner.js
@@ -228,6 +228,11 @@ class BookScanner {
                 bookId: media.id,
                 authorId: existingAuthorId
               })
+              const author = await Database.authorModel.findByPk(existingAuthorId)
+              if (author) {
+                const numBooks = await Database.bookAuthorModel.getCountForAuthor(existingAuthorId)
+                SocketAuthority.emitter('author_updated', author.toOldJSONExpanded(numBooks))
+              }
               libraryScan.addLog(LogLevel.DEBUG, `Updating book "${bookMetadata.title}" added author "${authorName}"`)
               authorsUpdated = true
             } else {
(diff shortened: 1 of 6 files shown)


server/scanner/BookScanner.js
```diff
@@ -228,6 +228,11 @@ class BookScanner {
                 bookId: media.id,
                 authorId: existingAuthorId
               })
+              const author = await Database.authorModel.findByPk(existingAuthorId)
+              if (author) {
+                const numBooks = await Database.bookAuthorModel.getCountForAuthor(existingAuthorId)
+                SocketAuthority.emitter('author_updated', author.toOldJSONExpanded(numBooks))
+              }
```
advplyr: I'm not sure about this one. This would be a lot of additional database calls for a full library scan.

Can we reduce this?

mikiher: You're right, this was sloppy.

Since all that is changing for existing authors inside a scan is the number of books, I added a separate new socket event `authors_num_books_updated` which sends an array of updated { authorId, numBooks }. 

I'm also calculating the payload and seding the new event at the end of the scan with only *one database call*. I profiled the call and it is very efficient even when handling thousands of author ids (taking about 13ms for all the ~3000 authors in my big library)

The Vue implementation obviously doesn't listen to this new event, so author updates will remain slightly broken (no more than before).

I'll also fix [audiobookshelf-client-react #134](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/134) to listen to this new event.

advplyr:
Thanks!

PR #5363, opened 2026-07-13
add progress and bookmarks specific endpoints

Vito0912:
## Brief summary

This adds dedicated endpoints for progress, bookmarks and getting bookmarks of a specific books.

Tests have been written by AI. They can be deleted though.

A note on this: I added a check to make sure a user has access when getting a single bookmark to that item. These are not currently used when getting all of them, but I think it is good practice to start doing so.

## Which issue is fixed?

None directly. Just to have the option to make less and smaller requests.
Related to some requests for bookmarks and progress per library item.

## In-depth Description

See above and question in Discord https://discord.com/channels/942908292873723984/954760207131615264/1526111010446905415

## How have you tested this?

Checked if results matches with user response


server/controllers/MeController.js
```diff
@@ -26,6 +26,52 @@ class MeController {
...
 
+  /**
+   * GET: /api/me/progress
+   *
+   * @param {RequestWithUser} req
+   * @param {Response} res
+   */
+  getAllMediaProgress(req, res) {
+    const mediaProgress = req.user.mediaProgresses?.map((mp) => mp.getOldMediaProgress()) || []
+    res.json(mediaProgress)
```
advplyr: I think these should be wrapped so we aren't returning an array as the root. There are still some API endpoints I set up early on that are returning arrays but since then I've tried to stay consistent with returning an object with a property that has the array.

Besides consistency, I remember with Kotlin, JSON arrays at the root had to be parsed differently than JSON objects. I'm not sure if that is still the case though.

```suggestion
    res.json({ mediaProgress })
```

I was going back and forth on whether this should be `mediaProgresses`. Not sure about that.

advplyr: I'll update the rest

Vito0912: I can do the rest tomorrow too. Just was not at the laptop to fix this yet.
Your suggestion was mergeable via th GH app

advplyr:
Thanks!

Vito0912:
@advplyr Is there a reason why the MeControler.test file was deleted entirely. I understand if the tests, added by this PR are not being added, but deleting the whole file, seems a bit extreme?

Also is there a reason the Podcast comment was deleted. Nothing major. But as podcasts allow bookmarks at API level it would be nice to have this documented somewhere?

And thanks for the quick merge!

advplyr:
While it is technically possible with the API to use a podcast library item id for a bookmark, that isn't supported anywhere in the UI. We need to be able to store the episode ID for it to be useful.

I assumed you didn't look at the test file since you said it was AI generated and could be deleted, so I didn't look at it either. Do you think it adds value?

PR #5369, opened 2026-07-15
fix(auth): prevent admin users from deleting the root account

mikiher:
## Summary

- Fix broken root-user delete guard in `UserController.delete()` that compared `req.params.id` to the literal `"root"` instead of checking `req.reqUser.isRoot`
- Add `User` model `beforeDestroy` hook as defense-in-depth against direct `user.destroy()` calls
- Add regression tests covering admin→root (403), root→root (400), admin→user (200), and self-delete (400)
- Add `SECURITY.md` with vulnerability reporting instructions

## Security impact

Any `admin` account could permanently delete the superuser (`type === 'root'`) via `DELETE /api/users/{root_uuid}` because root's primary key is a UUID, not the string `"root"`. CWE-863.

## Test plan

- [x] `npm test -- --grep "UserController|beforeDestroy root"` (6 passing)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,26 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please **do not** open a public GitHub issue for security vulnerabilities.
+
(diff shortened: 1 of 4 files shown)


mikiher:
Closing to allow local review before resubmitting.

PR #5370, opened 2026-07-15
Prevent admin users from deleting the root account

mikiher:
## Brief summary

This fixes an auth issue allowing admin users to delete the root account via API

## Which issue is fixed?

No issue.

## In-depth Description

- Fix broken root-user delete guard in `UserController.delete()` that compared `req.params.id` to the literal `"root"` instead of checking `req.reqUser.isRoot`
- Add `User` model `beforeDestroy` hook as defense-in-depth against direct `user.destroy()` calls
- Add regression tests covering admin→root (403), root→root (400), admin→user (200), and self-delete (400)

## How have you tested this?

`npx mocha test/server/controllers/UserController.test.js`

Also tested with the "REST Client" VSCode extension by sending actual root and admin API requests to a real running ABS server and verifying expected results.

### DIFF
--- a/server/controllers/UserController.js
+++ b/server/controllers/UserController.js
@@ -365,17 +365,13 @@ class UserController {
   async delete(req, res) {
     const user = req.reqUser
 
-    if (user.isRoot && !req.user.isRoot) {
-      Logger.error(`[UserController] Admin user "${req.user.username}" attempted to delete root user`)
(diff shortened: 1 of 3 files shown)


advplyr:
I thought spinning up a database for this test was too much so I simplified it down to testing that endpoint specifically.

Thanks!

PR #5374, opened 2026-07-18
Allow uploading author images from a file in the edit modal

zJeremiah:
## Brief summary

Added the ability to upload an image for an Author.
This is a feature I would personally like, since I've noticed that when using a URL for the Author it doesn't usually work.
Maybe others would appreciate it as well.

## In-depth Description

- Add a file upload option to the Update Author modal (alongside the existing Image URL), matching the book cover UI
- Extend `POST /api/authors/:id/image` to accept a multipart image file as well as a URL
- Clean up the previous author image file when replacing with a different extension (URL or upload)

## How have you tested this?

Manual test when running the container locally with my changes. I was able to upload the image of an Author when selecting a file after clicking the new Upload button.

## Screenshots
Before:
image

After:
image
image
image

### DIFF
--- a/client/components/modals/authors/EditModal.vue
+++ b/client/components/modals/authors/EditModal.vue
@@ -16,10 +16,19 @@
           </div>
         </div>
         <div class="grow">
-          <form @submit.prevent="submitUploadCover" class="flex grow mb-2 p-2">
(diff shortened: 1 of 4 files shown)


nichwall:

PR #5376, opened 2026-07-19
Make refresh-token grace period configurable (default 10 min)

### LINKED ISSUE #5281: [Bug]: /auth/refresh can make a session unrecoverable if the response is lost after token rotation
### What happened?

I think there is a small edge case in the refresh token rotation flow.

When `/auth/refresh` receives a valid refresh token, the server rotates it and returns the new refresh token. If that request succeeds on the server, but the client never receives the response, the client still only has the old refresh token.

After the grace period expires, retrying with the old refresh token returns `401 Invalid refresh token`, so the client has no way to recover even though the original refresh request was valid.

I would expect the refresh flow to tolerate this case for a short bounded window.

For example, if the old refresh token is retried shortly after rotation, the server could return the already-rotated current refresh token for that session, instead of treating the session as unrecoverable.

It should still reject old tokens outside the recovery window.

1. Log in and get refresh token A.
2. Call `/auth/refresh` with token A.
3. Let the server process the request, but discard the response with refresh token B.
4. Wait until the grace period expires.
5. Call `/auth/refresh` again with token A.
6. The server returns `401 Invalid refresh token`.

### Audiobookshelf version

v2.35.1

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

This can happen on mobile clients when the request reaches the server but the response is lost because of a network interruption or the app being suspended.

Refresh token rotation itself makes sense. The issue is just that this specific lost-response case leaves the client unable to recover the session.

DanielAshley:
## Brief summary

When `/auth/refresh` rotates the refresh token and the client never receives the response (dropped network, iOS backgrounding, VPN/proxy timeout), a retry with the old token only succeeds during a short grace window. That window was hardcoded to **60 seconds**, which is too short for real mobile conditions and permanently locks the session out. This makes the grace period configurable and raises the default to **10 minutes**, matching the range discussed on the issue.

## Which issue is fixed?

Fixes #5281

## In-depth Description

`TokenManager` already keeps the previous refresh token valid for a short period after rotation and, on an old-token retry inside that window, returns the already-rotated current token instead of failing. The mechanism is correct for the lost-response case in #5281; only the duration was too short.

Changes:

1. **`REFRESH_TOKEN_GRACE_PERIOD` env var** (seconds) — same pattern as existing `REFRESH_TOKEN_EXPIRY` / `ACCESS_TOKEN_EXPIRY`.
2. **Default raised from 60s → 600s (10 minutes)** — aligns with the 5–10 minute range suggested in the issue thread for iOS backgrounding / network handoff without going to hour-scale windows.
3. Logs when the env override is set (same style as the other token expiry overrides).

This deliberately does **not** redesign rotation (invalidate-on-use, opaque tokens, etc.) — those are larger design questions discussed on the issue. It only addresses the practical failure mode: a client that still holds the previous token after a lost rotation response can recover within a bounded, configurable window.

Operators who want the previous behavior can set `REFRESH_TOKEN_GRACE_PERIOD=60`.

## How have you tested this?

- **Unit tests** (`test/server/auth/TokenManager.test.js`): default is 600s; positive env values are applied and logged; non-positive/invalid env falls back to the default; no override log when env is unset. Full server suite green (**349** passing, including the 4 new tests).
- **Live reproduction of #5281** on a running server with the new default:
  1. Login to obtain refresh token A.
  2. Call `/auth/refresh` with A (discard / lose the response so the client only has A).
  3. Retry `/auth/refresh` with A after the old 60s window would have expired but still within 10 minutes → **HTTP 200** with the session's current rotated refresh token + a fresh access token (previously this returned 401).
  4. Confirmed the session row's grace window is ~599s ≈ 10 minutes.

## Screenshots

N/A — server-only change (no web client UI).


advplyr:
Thanks!

advplyr:
I'm wondering now if we should be making the default grace period the access token duration. I saw that Shopify is doing it this way for apps. They have access tokens last an hour, refresh tokens last 90 days, and a grace period of an hour.

Vito0912:
If you make the grace period longer, which weakens the security gained from it significantly, like a lot, (1 to 10 minutes already made the target attack windows way way bigger) there should be at minimum be the option to invalidate all/single sessions and see the sessions. If not having a rotating refresh token is kinda not usefull at all.

See my comment here: https://github.com/advplyr/audiobookshelf/issues/5281#issuecomment-4785693887 (but you also commented that already, so just a heads up)

At a refresh window of 1 hour it might make more sense to have it set at max 1 hour, but the first time an access token, issued by this refresh token has been used.

(Or probably the option, that would keep security in tact, is that the client needs to acknowledge that it gained x access token. It is invalid until the client sent it's access token. Is this unrealistic to change now, yes. But having a rotating refresh token that does not rotate for an hour is not really that kind of a rotating refresh token)

PR #5378, opened 2026-07-19
Fix narrator/author name lists mangled by non-breaking spaces (U+202F)

### LINKED ISSUE #5367: [Bug]: For a book with many narrators listed (full cast books, etc), narrators are grouped incorrectly when separated by a comma.
### What happened?

When an audiobook with only a few narrators is added to the library - those narrators being specified in the composer tag field and separated by a comma and space (, ) - Audiobookshelf is able to correctly parse out the narrators and display them correctly. If there are a great many narrators, such as 23, SOMETIMES audiobookshelf will join together two narrators at a time.

For example, this is the string of narrators for the book:
Joanna Wyatt, Paul Panting, Stephen Webb, Julian Glover, Tim Bentinck, Doulas Blackwell, Nigel Carrington, Emma D'Inverno, Teresa Gallagher, Eve Karpf, Kate Lock, Arthur Mitchell, Gordon Reid, William Roberts, Anne Rosenfeld, Stephen Thorne, Andrew Branch, Alison Dowling, Susan Sheridan, Stephen Grief, Garrick Hagon, Liza Ross, Hayward Morse

The attached photo shows how audiobookshelf interpreted the narrators. Also, because the number of narrators was odd, Hayward Morse did not have another narrator to join with and so audiobookshelf simply left him out.

This can be fixed by using semicolons instead of commas but I thought you might like to know.

Image

The narrators to be parsed correctly.

Add an audiobook to the library with a lot of narrators in the composer field and separate them by comma with space. I do not know exactly how many narrators it takes for this to happen, and it does not always happen. I have another book with 23 narrators, separated by comma and space, and audiobookshelf was able to parse them correctly.

These are the names used for that book and the photo below shows how audiobookshelf interpreted it:
Douglas Gresham, Paul Scofield, David Suchet, Jonny Scott, Nick Burnell, Jody Peach, Wendy Craig, Martin Jarvis, Robert Benfield, Sally Ann Burnett, Katy Glassborow, John Bluthal, Russell Boulter, Justin Butcher, Michael Haughey, Paul Jones, Peter Jones, Katherine Kellgren, Nathaniel Lippiett, Robin Meredith, Peter Moreton, David Oakley, Philip Sherlock

Image

### Audiobookshelf version

v2.35.1

Windows Tray App

Windows

None

### Logs

```shell

```

### Additional Notes

Here's another list that did not parse correctly, 36 narrators:
Douglas Gresham, Paul Scofield, David Suchet, Victor Spinetti, Andrew Sachs, Gwynn Beech, Philip Sully, Steven Webb, Jade Williams, Howard Attfield, Robert Benfield, Charlie Burnell, Nick Burnell, Elizabeth Counsell, Tenniel Evans, Paul Ewing, Katy Glassborow, Paul Grunert, Michael Haughey, Heathe Heyer, Katy Kellgren, Stash Kirkbride, Sallyanne Law, Polly March, Robert Meadwell, Robin Meredith, Joan Moon, Peter Moreton, David Oakley, Daniel Philpott, Gyuri Sarossy, Philip Sherlock, Katherine Suchet, Mervyn Stutter, Paul Webster, John Woodnut

Image

DanielAshley:
## Brief summary

Narrator/author lists are mangled when the names are separated **inside** each name by a non-breaking space (e.g. `U+202F` NARROW NO-BREAK SPACE) instead of a regular space. A list like `First Last, First Last, ...` then gets treated as `Last, First` pairs, merged two-at-a-time in swapped order, with the odd trailing name dropped — a 23-narrator list collapses to 11 mangled names.

## Which issue is fixed?

Fixes #5367

## Root cause

`parseNameString.parse()` decides whether a comma-separated list is `Last, First` or `First Last` by asking `checkIsALastName()` whether a chunk is a bare last name:

```js
function checkIsALastName(name) {
  if (!name.includes(' ')) return true // No spaces must be a Last name
  ...
}
```

That `' '` is a regular space (`U+0020`). In the reported files the first and last names are joined by **`U+202F` NARROW NO-BREAK SPACE**, not a regular space. So `"Joanna[U+202F]Wyatt".includes(' ')` is `false`, the chunk reads as a single space-less token — i.e. a "last name only" — and `firstChunkIsALastName` becomes `true`. The pairing loop then runs over the whole list:

```
"Joanna Wyatt, Paul Panting, Stephen Webb, Julian Glover, ..."   (first/last joined by U+202F)
  → ["Paul Panting Joanna Wyatt", "Julian Glover Stephen Webb", ...]
```

which is exactly the failure in the reporter's screenshots.

### Why it only happens *sometimes*

The format is decided from the **first chunk alone** (`checkIsALastName(splitNames[0])`). So whether a list is mangled depends entirely on whether its *first* name happens to use a regular space or a narrow no-break space:

- The **first** list in the issue (`Joanna[U+202F]Wyatt, ...`) starts with `U+202F` → mangled.
- The **second** list (`Douglas Gresham, Paul Scofield, David[U+202F]Suchet, ...`) starts with a **regular** space (only the later names use `U+202F`) → parsed correctly, as the reporter observed.
- The **third** list (`Douglas[U+202F]Gresham, ...`) starts with `U+202F` → mangled.

This matches the reporter's note that it "does not always happen."

> **Correction to an earlier revision of this PR:** a previous version of this description blamed a leading *mononym* / particle surname and changed the `Last, First` vs `First Last` decision heuristic. That was wrong — the reporter's list does start with a normal two-word name (`Joanna Wyatt`), and that change did not fix the reported input. Thanks to @gravethief13 for pushing back; their guess that `parseFullName` wasn't seeing a first name was correct, and the reason is the `U+202F` separator.

## The fix

Normalize non-breaking / other Unicode space separators to a regular space before parsing, at both entry points (`parse` and `nameToLastFirst`, which had the same latent bug):

```js
function normalizeNameSpaces(name) {
  return name.replace(/[\u00A0\u1680\u2000-\u200A\u202F\u205F]/g, ' ')
}
```

The `Last, First` vs `First Last` decision logic is left exactly as it was on `master`.

### 1. Confirm the separator really is `U+202F` (the invisible part)

The whole reason this is confusing is that `U+202F` renders identically to a normal space. To see it:

1. Open **https://leafo.net/characters/** (a per-character Unicode inspector).
2. Copy the narrator list straight out of issue #5367 — the `Joanna Wyatt, Paul Panting, Stephen Webb, Julian Glover, ...` list — and paste it into the input box.
3. In the character table, the separator between `Joanna` and `Wyatt` (and inside every other name) is listed as **`U+202F` — NARROW NO-BREAK SPACE**, while the separators after each comma are ordinary **`U+0020` — SPACE**.

That single difference is the entire bug: `checkIsALastName` only recognises `U+0020`, so every `First[U+202F]Last` chunk looks like a bare last name.

### 2. Before / after on the parser

```
Input (first/last joined by U+202F):
  "Joanna Wyatt, Paul Panting, Stephen Webb, Julian Glover"   (4 narrators)

master   → ["Paul Panting Joanna Wyatt", "Julian Glover Stephen Webb"]   (2, swapped)
this PR  → ["Joanna Wyatt", "Paul Panting", "Stephen Webb", "Julian Glover"]   (4, in order)
```

The full 23-name list from the issue goes from 11 mangled names on `master` to all 23 preserved with this change.

## Tests

Extended `test/server/utils/parsers/parseNameString.test.js` with cases built from the real issue strings (using explicit `\u202f` / `\u00a0` escapes):

- a `First Last` list joined by `U+202F` is no longer paired;
- the same for `U+00A0`;
- a large **odd-length** `U+202F` list keeps **every** name (nothing dropped);
- a genuine `Last, First` list that contains `U+202F` inside a compound surname still pairs **and** is normalized.

All four fail on `master` and pass with the fix. Existing `parseNameString` tests are unchanged and still pass (22 total). Files are Prettier-clean.

## Screenshots

N/A — server-side parsing change (no web client UI).

### DIFF
--- a/server/utils/parsers/parseNameString.js
+++ b/server/utils/parsers/parseNameString.js
@@ -5,6 +5,17 @@
 //
 const parseFullName = require('./parseFullName')
 
+// Some taggers join a first and last name with a non-breaking or other Unicode
+// space (e.g. U+202F NARROW NO-BREAK SPACE, U+00A0 NO-BREAK SPACE) instead of a
+// regular space. These are not matched by ' ', so a name like "First Last" reads
+// as a single space-less token - i.e. a last name only - which can flip an entire
+// comma-separated list into "Last, First" pairing mode and mangle it. Normalize
+// such spaces to a regular space before parsing. U+3000 (ideographic space) is
+// intentionally left alone so CJK names are returned untouched.
+function normalizeNameSpaces(name) {
+  return name.replace(/[\u00A0\u1680\u2000-\u200A\u202F\u205F]/g, ' ')
+}
+
 function parseName(name) {
   var parts = parseFullName(name)
   var firstName = parts.first
@@ -29,7 +40,7 @@ function checkIsALastName(name) {
 
 // Handle name already in First Last format and return Last, First
 module.exports.nameToLastFirst = (firstLast) => {
-  var nameObj = parseName(firstLast)
+  var nameObj = parseName(normalizeNameSpaces(firstLast))
   if (!nameObj.last_name) return nameObj.first_name
   else if (!nameObj.first_name) return nameObj.last_name
   return `${nameObj.last_name}, ${nameObj.first_name}`
@@ -43,6 +54,8 @@ module.exports.nameToLastFirst = (firstLast) => {
  */
 module.exports.parse = (nameString) => {
   if (!nameString) return null
+
+  nameString = normalizeNameSpaces(nameString)
 
   let splitNames = []
   const isCommaSeparated = nameString.includes(',')
--- a/test/server/utils/parsers/parseNameString.test.js
+++ b/test/server/utils/parsers/parseNameString.test.js
@@ -69,6 +69,33 @@ describe('parseNameString', () => {
       expect(result.names).to.deep.equal(['张三', '李四'])
     })
 
+    // Regression: https://github.com/advplyr/audiobookshelf/issues/5367
+    // Some taggers join first/last names with U+202F (narrow no-break space) or
+    // U+00A0 (no-break space) instead of a regular space. Those are not matched
+    // by ' ', so every chunk read as a space-less "last name" and the whole
+    // First Last list was flipped into "Last, First" pairing mode.
+    it('does not pair a First Last list joined by narrow no-break spaces (U+202F)', () => {
+      const result = parse('Joanna\u202fWyatt, Paul\u202fPanting, Stephen\u202fWebb, Julian\u202fGlover')
+      expect(result.names).to.deep.equal(['Joanna Wyatt', 'Paul Panting', 'Stephen Webb', 'Julian Glover'])
+    })
+
+    it('does not pair a First Last list joined by no-break spaces (U+00A0)', () => {
+      const result = parse('Jane\u00a0Doe, John\u00a0Smith')
+      expect(result.names).to.deep.equal(['Jane Doe', 'John Smith'])
+    })
+
+    it('keeps every name in a large odd-length First Last list joined by narrow no-break spaces', () => {
+      const expected = ['Joanna Wyatt', 'Paul Panting', 'Stephen Webb', 'Julian Glover', 'Tim Bentinck']
+      const input = expected.map((n) => n.replace(' ', '\u202f')).join(', ')
+      const result = parse(input)
+      expect(result.names).to.deep.equal(expected)
+    })
+
+    it('still pairs a genuine Last, First list that uses narrow no-break spaces inside first/last parts', () => {
+      const result = parse('von\u202fMises, Ludwig, Smith, John')
+      expect(result.names).to.deep.equal(['Ludwig von Mises', 'John Smith'])
+    })
+
     it('removes duplicate names', () => {
       const result = parse('John Smith & John Smith')
       expect(result.names).to.deep.equal(['John Smith'])


gravethief13:
I'm confused as to what you mean by idealized, the list of narrators used in my example started with Joanna Wyatt - a chunk with a space. Maybe the parseFullName didn't recognize it as having a first name? I've no idea its criteria of of course. Here's a screenshot from mp3tag with the exact information read by audiobookshelf:

Screenshot_1

If this change does fix the issue I'm wondering if it wouldn't be for the reason written here, but I suppose that wouldn't matter.

DanielAshley:
You're right, and thanks for pushing back — the "idealized" comment was wrong, and your instinct was correct: `parseFullName` was not seeing a first name in `Joanna Wyatt`.

The reason is subtle. The character between `Joanna` and `Wyatt` in your file isn't a regular space (`U+0020`) — it's `U+202F` **NARROW NO-BREAK SPACE**. It renders identically to a space, but the parser's `checkIsALastName()` only tests for a regular space (`name.includes(' ')`), so `Joanna[U+202F]Wyatt` looks like a single space-less token, i.e. one name with no space. That flips the whole list into `Last, First` pairing mode and merges the names two-at-a-time.

You can paste the narrator list from your issue into https://leafo.net/characters/ — it lists every character, and the separators inside the names show up as `U+202F NARROW NO-BREAK SPACE` while the ones after each comma are ordinary `U+0020 SPACE`.

That also explains the "sometimes": the format is decided from the **first name only**. Your second list (`Douglas Gresham, Paul Scofield, David...`) happens to start with a regular space, so it parsed fine even though the later names use `U+202F`; the first and third lists start with `U+202F` and got mangled. And switching to semicolons worked because that takes a different code path that never enters the pairing loop.

I've reworked this PR accordingly: instead of touching the `Last, First` heuristic, it now normalizes non-breaking / other Unicode spaces to a regular space before parsing, so `U+202F` names are handled correctly regardless of where they sit in the list. Updated description and tests (built from your actual strings) are above.

RyanGu-NZ:
Tested this locally on main first — reproduced the U+202F issue, and noticed U+2009 (thin space) was breaking things in a similar way too.
Tried out this PR and all 22 parseNameString tests passed without issues. I also tested the U+2009 case, and it passed too. Also gave nameToLastFirst() a quick run with U+202F and it worked fine.
Thanks for fixing this!

PR #5380, opened 2026-07-20
Match folder library items by media file inodes when the folder inode changes

### LINKED ISSUE #5379: [Bug]: Listening progress and finished status lost when a book folder is renamed, even though the audio files inside are unchanged
### What happened?

I reorganized part of my audiobook library this week, standardizing folder names into an Author/Series/Title layout. The files themselves were untouched, every book folder still holds the same m4b it always had. After the next scan several books came back with no progress at all: books I finished a year ago show as not started, and a family member who was ten minutes from the end of one lost their position.

In the database, each renamed book got a brand new library item with a new id, and the mediaProgresses rows still point at the old book id, so in the UI the progress just looks wiped.

The scanner had what it needed to connect them. `stat` confirms the audio files kept their inodes through the rename, since they were moved into the new folders rather than copied. Only the folder inode changed, because the tool created fresh directories.

In `server/scanner/LibraryScanner.js` the fallback matching tries exact path, folder inode (`ItemToItemInoMatch`), then file inode (`ItemToFileInoMatch`). The file inode check is gated on `libraryItem1.isFile`, so it only ever applies to single file items sitting directly in the library root. For a normal folder based book the file inodes are collected during the scan and then never used for matching, so a book folder that gets recreated can't be matched no matter how intact its files are.

A renamed book folder whose audio files are unchanged should be recognized as the same library item, keeping its id and everyone's progress.

1. Have a folder based audiobook, e.g. `Library/Author Name/Book Title/book.m4b`, scanned and with listening progress (or marked finished) for any user.
2. On the server, recreate the folder and move the file into it, so the folder inode changes and the file inode does not:
```
mkdir "Library/Author Name/Book Title (fixed)"
mv "Library/Author Name/Book Title/book.m4b" "Library/Author Name/Book Title (fixed)/book.m4b"
rmdir "Library/Author Name/Book Title"
```
3. Run a library scan.
4. The old item is marked missing and a new item appears for the same book with zero progress for every user. The mediaProgresses rows still reference the old item's book id.

Deterministic here, every time. Multi-file books behave the same way.

### Audiobookshelf version

v2.35.1

Docker

Linux

Chrome

### Logs

```shell
[LibraryScanner] Library Item "/media/Audiobooks/Author Name/Book Title" (inode: 2807619) is missing
```
(with the corresponding new item added at the new path in the same scan)

### Additional Notes

Official Docker image on a Linux host (Docker Swarm on Ubuntu Server 24.04), media on an NFS mount from a NAS. inotify events don't propagate over NFS, so the watcher never sees these renames live and everything goes through periodic scans. The same gap is in the watcher's matching path though, it isn't specific to either one.

I have a patch for this and opened it as #5380.

mels0n:
## Brief summary

Folder based library items lose their identity when their folder is renamed or recreated, and every user's progress goes with it, even when the audio files inside kept their inodes. This extends the scanner's inode fallback matching to consult the media files' inodes for folder items, on both the full scan and watcher paths, so those items are updated in place instead of being duplicated.

This is a bug fix rather than new behavior. Rename detection by inode is existing, intended scanner functionality, and the `isFile` gate makes it unreachable for folder based items, which are the most common library shape.

## Which issue is fixed?

Fixes #5379

## In-depth Description

The fallback matching in `LibraryScanner` tries exact path, folder inode (`ItemToItemInoMatch`), then file inode (`ItemToFileInoMatch`). The file inode check is gated on `libraryItem1.isFile`, so it only ever applies to single file items sitting directly in the library root. For a normal folder based book the file inodes are collected during the scan and then never used for matching, which means the most common library shape has the weakest rename detection: recreate the folder, which any external reorganization tool does, and the item can't be matched no matter how intact its files are.

What this adds:

- Full scan: after the existing matchers miss, and only for an item whose own folder no longer exists, `ItemToItemFileInoMatch` matches the item against the scanned folders that hold its media files. It has to match exactly one scanned folder, otherwise the item is left alone and logged.
- Watcher scan (`scanFolderUpdates`): `findLibraryItemByFilesToItemFilesInoMatch` does the same from the other direction, selecting folder items with any overlapping library file inode through a `json_each` query, then applying the same rule in JS and requiring a single surviving candidate.
- The shared rule is `scanUtils.checkItemFilesMatchByIno`: at least half of the item's media files, audio or ebook by the library's media type, have to be found by inode, with sizes compared when both sides know them. Sidecar files are ignored since covers, opf and desc.txt are routinely regenerated.
- Both fallbacks on the watcher path now share one pass of `getFileTimestampsWithIno`, which returns the inode and size together, instead of each matcher statting the same files itself.

Why the constraints are what they are:

- Requiring the item's folder to be gone is what keeps hardlinks honest. If a tool hardlinks a book into a second location while the original is still in place, both paths are real and should remain two items, and the file inodes are identical on both sides. Neither path will match an item whose folder still exists on disk.
- The majority threshold keeps a stray disc from claiming a large multi-file book, while tolerating a folder that was only partly reorganized.
- Sizes are compared because inode numbers are only unique per filesystem and are reused after a file is deleted. A move never changes a file's size, so this costs the real scenario nothing and rules out a recycled inode taking over an item. Library files scanned by older versions may have no size recorded, in which case the inode alone decides.
- Requiring an unambiguous match means a library that already contains hardlinked duplicates falls back to today's behavior instead of guessing.
- Items with no media files, or with no stored inodes, never match.

Matched items go through the existing update in place flow (`rescanLibraryItemMedia`), so the item id and book id are preserved and mediaProgresses rows keep working. No migration needed.

## How have you tested this?

- Unit tests: 11 cases for `checkItemFilesMatchByIno` in `test/server/utils/scandir.test.js`, covering single file and majority matches, minority rejection, sidecar immunity, ebooks, ebooks not counting as media in podcast libraries, an inode reused by a file of a different size, an unknown size falling back to the inode alone, and empty or missing input. Full `npm test` passes: 356 tests, 0 failures.
- Runtime, on a scratch server built from this branch with a throwaway library: a single m4b book and a three file mp3 book, one marked finished and one mid listen. Recreated both folders and moved the files in, so the folder inodes changed and the file inodes did not, then rescanned. Both items kept their ids, paths updated in place, the finished flag and the 45.5s in progress position survived, nothing was marked missing. Hardlinking the m4b into a second folder and rescanning left the original intact and created a separate second item for the copy, as intended. That run was against the first revision of this branch. The rules have since been tightened (the folder gone gate, size comparison, the unambiguous match requirement), which narrows what can match rather than widening it, and the unit tests cover the added rules.
- The bug itself was observed on a production v2.35.1 server: folder recreated, audio file inodes intact, item duplicated and progress stranded.

## Screenshots

Server side only, no web client changes.

### DIFF
--- a/server/scanner/LibraryScanner.js
+++ b/server/scanner/LibraryScanner.js
@@ -173,7 +173,7 @@ class LibraryScanner {
       let libraryItemData = libraryItemDataFound.find((lid) => lid.path === existingLibraryItem.path)
       if (!libraryItemData) {
         // Fallback to finding matching library item with matching inode value
-        libraryItemData = libraryItemDataFound.find((lid) => ItemToItemInoMatch(lid, existingLibraryItem) || ItemToFileInoMatch(lid, existingLibraryItem) || ItemToFileInoMatch(existingLibraryItem, lid))
+        libraryItemData = libraryItemDataFound.find((lid) => ItemToItemInoMatch(lid, existingLibraryItem) || ItemToFileInoMatch(lid, existingLibraryItem) || ItemToFileInoMatch(existingLibraryItem, lid) || ItemToItemFileInoMatch(lid, existingLibraryItem))
         if (libraryItemData) {
           libraryScan.addLog(LogLevel.INFO, `Library item with path "${existingLibraryItem.path}" was not found, but library item inode "${existingLibraryItem.ino}" was found at path "${libraryItemData.path}"`)
         }
@@ -576,7 +576,7 @@ class LibraryScanner {
       let updatedLibraryItemDetails = {}
       if (!existingLibraryItem) {
         const isSingleMedia = isSingleMediaFile(fileUpdateGroup, itemDir)
-        existingLibraryItem = (await findLibraryItemByItemToItemInoMatch(library.id, fullPath)) || (await findLibraryItemByItemToFileInoMatch(library.id, fullPath, isSingleMedia)) || (await findLibraryItemByFileToItemInoMatch(library.id, fullPath, isSingleMedia, fileUpdateGroup[itemDir]))
+        existingLibraryItem = (await findLibraryItemByItemToItemInoMatch(library.id, fullPath)) || (await findLibraryItemByItemToFileInoMatch(library.id, fullPath, isSingleMedia)) || (await findLibraryItemByFileToItemInoMatch(library.id, fullPath, isSingleMedia, fileUpdateGroup[itemDir])) || (await findLibraryItemByFilesToItemFilesInoMatch(library.id, fullPath, isSingleMedia, fileUpdateGroup[itemDir]))
         if (existingLibraryItem) {
           // Update library item paths for scan
           existingLibraryItem.path = fullPath
@@ -652,6 +652,17 @@ function ItemToItemInoMatch(libraryItem1, libraryItem2) {
   return libraryItem1.ino === libraryItem2.ino
 }
 
+function ItemToItemFileInoMatch(libraryItemScanData, existingLibraryItem) {
+  // Check if a scanned folder contains the media files of an existing folder item, matching by file inode.
+  //   Covers an item folder that was recreated (new folder inode) while the media files kept their inodes,
+  //   e.g. an external tool reorganizing the library by moving files into new folders
+  if (libraryItemScanData.isFile || existingLibraryItem.isFile) return false
+  return scanUtils.checkItemFilesMatchByIno(
+    existingLibraryItem.libraryFiles,
+    libraryItemScanData.libraryFiles.map((lf) => lf.ino)
+  )
+}
+
 function hasAudioFiles(fileUpdateGroup, itemDir) {
   return isSingleMediaFile(fileUpdateGroup, itemDir) ? scanUtils.checkFilepathIsAudioFile(fileUpdateGroup[itemDir]) : fileUpdateGroup[itemDir].some(scanUtils.checkFilepathIsAudioFile)
 }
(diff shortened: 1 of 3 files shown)


Kyomorie:
Tested the matching logic against a few edge cases. The inode fallback makes sense for filesystems where directory inodes are unstable, but I found an ambiguity in the full-scan path.

Right now, the full scan checks whether an old item matches one new folder, but not whether multiple missing items match the same folder. If two old items match the same new folder, whichever one is processed first can claim it and keep its Book.id. I reproduced this with two missing entries sharing the same media inode. The watcher path already bails out on `matches.length > 1`, so I think the full scan should enforce the same 1:1 relationship.

A few other edge cases I noticed:

- Threshold: `numMatchingMediaFiles * 2 >= existingMediaFiles.length` accepts exactly 50%. For a 2-part book, one matching file is enough. A strict majority (`>`) would be safer.
- Audio vs ebook: `isMediaFile()` treats ebooks and audio as media. For an item with `book.m4b + book.epub`, matching only the epub probably shouldn't establish the identity of the audiobook when audio is present.
- Context loss: `scannedFiles` is reduced to `{ ino, size }`, so the matcher loses the file extension/type.
- Leftover directories: `pathExists(existingLibraryItem.path)` causes a false negative if an external tool moves the files but leaves an empty source directory behind.
- Inode reuse: for single-file items with no stored size, matching on inode alone seems risky. Requiring a matching size in that case would be safer.

I'd also add tests for:

- 2 old items matching 1 new directory
- 1/2 file match (50% edge)
- m4b + epub where only the epub matches
- leftover empty source directory
- single-file item without a stored size

I'd keep this PR focused on the inode logic and leave the metadata/fingerprint matching to the follow-up PRs.

Issue #4062, opened 2025-02-27
[Enhancement]: Full Screen Player


Eroskuro:
### Type of Enhancement

None

### Describe the Feature/Enhancement

It would be nice if the Web UI had a full screen player much like the one in the android app.

### Why would this be helpful?

Mostly just a visual improvement, but also would make it easer to access controls with wireless mouse on HTPC

### Future Implementation (Screenshot)

similar if not the same as android full screen player

### Audiobookshelf Server Version

2.19.5

nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/2569 and https://github.com/advplyr/audiobookshelf/issues/3345

Issue #3345, opened 2024-08-29
[Enhancement]: Button for a car mode please


K3R0-42:
### Type of Enhancement

Web Interface/Frontend

### Describe the Feature/Enhancement

Larger buttons for operation in the car (Tesla Browser). Some other apps solve this by allowing you to click on the cover of the album you are listening to so that you have a player to operate. I have sent some pictures of a cell phone app to illustrate this. 

### Why would this be helpful?

I use the software in the Tesla, as there is no app for it I have to open it in the browser which only takes up half the screen when driving. Unfortunately, the buttons for operating the software are very small, which has often caused me to jump to the wrong chapter in the audio book instead of pressing pause as intended. It would therefore be good to have the option of larger control buttons. 

### Future Implementation (Screenshot)

![IMG_4966](https://github.com/user-attachments/assets/dafe33c2-51b4-4256-89db-845b580c9250)
![IMG_4969](https://github.com/user-attachments/assets/b6c2d923-124c-45fe-ba92-01cdf9f85757)
![IMG_4968](https://github.com/user-attachments/assets/657ab84d-7949-4b6d-a1be-1207153e0c6f)

### Audiobookshelf Server Version

v2.12.3

nichwall:
Related to https://github.com/advplyr/audiobookshelf/issues/2569

advplyr:
BookPlayer is a mobile app though and you are using the web client. The Abs mobile app also allows opening a full screen player.

I think a full screen player on the web client at full width might be strange.

K3R0-42:
When I click on the cover in the web player, a full screen of the cover also opens, so it would certainly be possible to place buttons there for easier operation.

suzhoum:
Unrelated to this issue, but when I use ABS on Tesla Browser, it can't seem to automatically start the next episode if I minimized the Browser App. It only starts the next episode when I reopen the Browser. Does it happen to any of you?

K3R0-42:
I have the “problem” too, but I'm glad that it works with a minimized browser. At least it's enough to watch something quickly without it stopping playback immediately. Before the update, minimizing didn't work at all.

Eroskuro:
> I think a full screen player on the web client at full width might be strange.

[...]

Issue #2569, opened 2024-01-30
[Enhancement]: Bring the mobile app playback UI to the PWA on mobile


ipcintron:
### Describe the feature/enhancement

The playback interface on mobile when using the PWA is the same as desktop which means it can be quite cluttered. I would like to see the interface brought over from the App Store release of ABS to the PWA when on mobile.

PWA Mobile (iOS):
![image](https://github.com/advplyr/audiobookshelf/assets/15165416/d4357127-2e21-4ab8-8b33-9be6c3851ccf)

Android App:
![Screenshot_20240129_155102_Audiobookshelf](https://github.com/advplyr/audiobookshelf/assets/15165416/fcddf9a0-6ae0-431a-9cf8-855226fd2ee1)

nigiriemoji:
The web player design is ELITE but it’s missing a sleep timer

PR #5385, opened 2026-07-20
feat(player): add fullscreen player to the web UI

sokol-matija:
## What

Adds an immersive fullscreen player to the web UI, opened from the expand button in the mini player. It aims to bring the web client closer to the mobile app's now-playing screen.

Opening as a **draft** — happy to adjust the visual direction, split this up, or gate it behind a setting if you'd prefer a different approach.

## Requested in

- #4062 — Full Screen Player (@Eroskuro)
- #2569 — Bring the mobile app playback UI to the PWA on mobile (@ipcintron)
- #3345 — Button for a car mode please (@K3R0-42) — larger, easier-to-hit controls

## Changes

**New: `PlayerFullscreen.vue`**

- Blurred cover backdrop behind large artwork
- Whole-book seek rendered as a ring around the artwork, concentric with the cover's corner radius; click or drag anywhere on it to seek, with a time readout while dragging
- Chapter progress bar with elapsed and remaining chapter time on either side
- Book elapsed / total under the title and author
- Volume, speed, sleep timer and bookmarks reveal on artwork hover; chapters button and an active sleep timer readout stay in the transport bar
- Chapter list opens inline and pushes the artwork so it stays centered, rather than covering the transport controls
- Keyboard: <kbd>Space</kbd> play/pause, <kbd>,</kbd> / <kbd>.</kbd> seek ±10s, <kbd>Esc</kbd> to exit

**Supporting changes to existing components** (all additive / opt-in, existing call sites unchanged):

- `PlayerPlaybackControls` — optional `no-tooltips` prop; play/pause icon now transitions on state change; all motion gated behind `prefers-reduced-motion`
- `PlaybackSpeedControl` — optional `menu-left-offset` prop so the menu can open away from neighboring content. Also collapses the two arrow-position branches into the one formula they already shared.
- `Modal` — `z-index` is now bound as a style rather than an interpolated Tailwind class (`` :class="`z-${zIndex}`" ``). Interpolated class names aren't seen by the Tailwind scanner, so those utilities were never generated and the prop had no effect. This surfaced as modals layering incorrectly over the fullscreen player.

## Notes

- Reuses the existing `playerIsFullscreen` store state
- No server-side changes, no new dependencies, no changes to existing player behavior when the fullscreen view isn't open


nichwall:

sokol-matija:
Understood on the freeze — leaving this open as reference. Adding a recording in
case it's useful for the React rewrite, or for anyone who wants to run the patch.

https://github.com/user-attachments/assets/1e532cd2-8a39-4bb6-ac67-5fb87652791e

The branch is up to date with everything shown. Since the initial draft it also
picked up:

- **Hotkey feedback** — arrow-key seeks spin the transport button with an
  expanding ring, and full screen shows a `−10s` / `+10s` burst over the artwork.
  Arrow-key volume raises a HUD beside the artwork.
- **Speed pill** in the transport bar, surfaced only when the rate is off 1×,
  with step controls and click-to-reset
- <kbd>,</kbd> / <kbd>.</kbd> now step playback speed instead of seeking — the
  arrow keys already seek, at the user's configured jump amount
- **Shortcut overlay** on <kbd>?</kbd>, generated from `$hotkeys.AudioPlayer`
  rather than hardcoded, so it can't drift from the actual bindings

---

Two bugs in existing code I ran into along the way. Both are in
`client/components/player/PlayerUi.vue` and independent of the full screen view,
so they affect the current frontend and may be worth carrying into the rewrite:

1. `increasePlaybackRate` / `decreasePlaybackRate` compute
   `rate + increment || 0.1`. `+` binds tighter than `||`, so this parses as
   `(rate + increment) || 0.1` — the fallback guards the sum rather than the
   increment. An undefined increment yields `NaN`, so the rate is set to `0.1`
   instead of being stepped. Latent in practice, since the store defaults the
   increment to `0.1`.

2. Those same two methods call `setPlaybackRate` but never
   `playbackRateChanged`, so a playback rate set by keyboard isn't persisted —
   while the identical change made from the speed menu is.

Fixes for both are in this branch (`11c1a8d4`) if useful.

sokol-matija:
@nichwall Check out the new video

PR #4635, opened 2025-08-30
Fix Invalid callback URL - must be same-origin for NPM users

Vito0912:
## Brief summary

NPM, even if it's not the spec, returns `'x-forwarded-proto': 'http, https',` as header. 

## Which issue is fixed?

Fixes #4609 

## In-depth Description

See Discord conversation. This is basically because NPM adds both protocols as forwarded.

## How have you tested this?

Local NPM setup

PR #5386, opened 2026-07-20
Fix oidc url validation

mikiher:
## Brief summary

This fixes an auth issue where OIDC post-login callback URLs could be bypassed via protocol-relative or encoded paths, allowing redirects to external origins.

## Which issue is fixed?

No issue.

## In-depth Description

- Refactor protocol and origin detection helper functions
  - Add `server/utils/requestUtils.js` with shared `isRequestSecure`, `getRequestProtocol`, and `getRequestOrigin` helpers for consistent HTTPS and origin detection
  - Use the shared helpers in `Server.js`, `TokenManager.js`, and `OidcAuthStrategy.js` instead of duplicated inline protocol/host logic
- Harden `OidcAuthStrategy.isValidWebCallbackUrl()` to reject protocol-relative (`//host`) and backslash-prefixed (`/\host`) callback URLs
- Resolve relative callback URLs against the server origin before validation, and reject encoded protocol-relative path segments (e.g. `/%2F%2Fevil.example/...`)
- Add regression tests for `requestUtils` and `isValidWebCallbackUrl`

## How have you tested this?

`npx mocha test/server/utils/requestUtils.test.js test/server/auth/OidcAuthStrategy.test.js`

Also tested with the "REST Client" VSCode extension by sending actual API requests to a real running ABS server configured with OIDC, verifying expected results.

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -14,6 +14,7 @@ const { version } = require('../package.json')
 const is = require('./libs/requestIp/isJs')
 const fileUtils = require('./utils/fileUtils')
 const { toNumber } = require('./utils/index')
+const { getRequestOrigin } = require('./utils/requestUtils')
 const Logger = require('./Logger')
 
 const Auth = require('./Auth')
@@ -288,10 +289,9 @@ class Server {
     // if RouterBasePath is set, modify all requests to include the base path
     app.use((req, res, next) => {
       const urlStartsWithRouterBasePath = req.url.startsWith(global.RouterBasePath)
-      const host = req.get('host')
-      const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
+      const { origin } = getRequestOrigin(req)
       const prefix = urlStartsWithRouterBasePath ? global.RouterBasePath : ''
-      req.originalHostPrefix = `${protocol}://${host}${prefix}`
+      req.originalHostPrefix = `${origin}${prefix}`
       if (!urlStartsWithRouterBasePath) {
         req.url = `${global.RouterBasePath}${req.url}`
       }
--- a/server/auth/OidcAuthStrategy.js
+++ b/server/auth/OidcAuthStrategy.js
@@ -4,6 +4,7 @@ const OpenIDClient = require('openid-client')
 const axios = require('axios')
 const Database = require('../Database')
 const Logger = require('../Logger')
+const { getRequestOrigin } = require('../utils/requestUtils')
 
 /**
  * OpenID Connect authentication strategy
@@ -289,8 +290,8 @@ class OidcAuthStrategy {
     const sessionKey = strategy._key
 
     try {
-      const protocol = req.secure || req.get('x-forwarded-proto') === 'https' ? 'https' : 'http'
-      const hostUrl = new URL(`${protocol}://${req.get('host')}`)
+      const { origin } = getRequestOrigin(req)
+      const hostUrl = new URL(origin)
       const isMobileFlow = req.query.response_type === 'code' || req.query.redirect_uri || req.query.code_challenge
 
       // Only allow code flow (for mobile clients)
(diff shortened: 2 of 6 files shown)


advplyr:
I'm not sure if this was intentional by you or not, but the x-forwarded-proto was found to get sent as `http, https` by NPM. Patched in https://github.com/advplyr/audiobookshelf/pull/4635
I made that change to prevent regression but let me know

advplyr:
Thanks!

mikiher:
Sorry about that - I didn't carefully read #4635 
- The check just looked weird to me and I found no evidence in NPM documentation that it was sending these values. It seemed to me like this was dealing with some specific user's reverse proxy misconfiguration.
- There were also other places in the code (before the refactor), that just checked `req.secure || req.get('x-forwarded-proto') === 'https'`

I really dislike that we are making all these finicky checks in abs codebase - they're easy to get wrong and should really be handled by express. Have we considered adding a server setting saying something like "My server is behind a reverse proxy", and when true, we set `trust proxy` on the express app? Then we can remove all those x-forwarded-* checks and only rely on req.secure (which is likely false in standard abs settings) and other standard req values.

mikiher:
Here's an example of a subtle difference between `trust proxy` and our codebase, which illustrates why we should use `trust proxy` and not write our own stuff.

Here's the express code for getting req.protocol (which also determines `req.secure`):

```
defineGetter(req, 'protocol', function protocol(){
  var proto = this.socket.encrypted
    ? 'https'
    : 'http';
  var trust = this.app.get('trust proxy fn');

  if (!trust(this.socket.remoteAddress, 0)) {
    return proto;
  }

  // Note: X-Forwarded-Proto is normally only ever a
  //       single value, but this is to be safe.
  var header = this.get('X-Forwarded-Proto') || proto
  var index = header.indexOf(',')

  return index !== -1
    ? header.substring(0, index).trim()
    : header.trim()
});
```

This is slightly different from our code. When `x-forwarded-proro` has multiple values:
- Express code only accepts: `x-forwarded-proto: https, http` as `secure`
  - this is likely because the value is (correctly) treated as a proxy chain, and therefore it is the first value in that chain that matters.
- Our code accepts both `x-forwarded-proto: https, http` and `x-forwarded-proto: http, https` as `secure`

It's not a big deal, but it illustrates the fact that the logic is subtle, and my point about using `trust proxy` instead of re-inventing the wheel.

mikiher:
As explained in https://expressjs.com/en/guide/behind-proxies/, `trust proxy` has a number of differenst settings (not just false/true), and if we want, we can design a good UI to expose some of those in a Trusted Proxies server setting (instead of just a simple checkbox).

I think that since so many of our server deployments need to run behind proxies, this will provide a more secure and stable way to do so. Users will still need to setup proxies on their own (as they should), but with the Trusted Proxies setting they'll be able to let ABS know what it needs to know about the proxy setup.

Vito0912:
I agree. This is also because a lot of this part is not documented well. For example, look at the documentation for X-Forwarded-Proto: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Proto>

It does not even say that it can have multiple values.

Only after searching specifically for why the order matters did I find a StackOverflow answer about how it appends. But this is also more of an universal agreement as it's a non-default Header unlike the actual Forwarded header would be.

advplyr:
I see what you mean but I think we should leave it for now unless this is a vulnerability you can think of. We haven't had issues reported since that patch.

mikiher:
This isn't about just about vulnerabilities. It's about simplifying our codebase, and separation of concerns.

advplyr:
Adding new settings is something we should generally try to avoid where we can. Particularly it seems we are talking about a breaking change on many servers, right?

mikiher:
>  Particularly it seems we are talking about a breaking change on many servers, right?

Not necessarily. It depends what the default policy would be. 

Practically, I think that today we're essentially implementing `app.set('trust proxy', true)` by ourselves implicitly (maybe with a number of minor changes).

I think this can designed so that we don't break most existing deployments.

At the minimum, I think we should consider starting with `app.set('trust proxy', true)` without any settings changes, and figure out how it changes server behavior, if at all.

PR #5387, opened 2026-07-20
 Reject refresh tokens on API and WebSocket authentication

mikiher:
## Brief summary

This fixes an auth issue allowing refresh tokens to authenticate API and WebSocket requests when sent as Bearer credentials.

## Which issue is fixed?

No issue.

Related discussion: https://github.com/advplyr/audiobookshelf/discussions/4460

## In-depth Description

- Add `TokenManager.isBearerAccessTokenPayload()` to distinguish bearer credentials from refresh-only JWTs (`type: 'refresh'`)
- Reject refresh tokens in `validateAccessToken()` (WebSocket auth) and `jwtAuthCheck()` (API auth); refresh tokens remain valid only at `POST /auth/refresh`
- Preserve compatibility for legacy access tokens without a `type` claim and for API key tokens
- Add regression tests covering payload classification, `validateAccessToken`, and `jwtAuthCheck` (refresh denied, access/legacy allowed, API keys unchanged)

## How have you tested this?

`npx mocha test/server/auth/TokenManager.test.js`

Also tested with the "REST Client" VSCode extension by sending actual API requests to a running ABS server and verifying expected results.

### DIFF
--- a/server/auth/TokenManager.js
+++ b/server/auth/TokenManager.js
@@ -67,6 +67,18 @@ class TokenManager {
   }
 
   /**
+   * Whether a decoded JWT payload may authenticate API/socket requests (not refresh-only credentials).
+   *
+   * @param {Object} decoded
+   * @returns {boolean}
+   */
+  static isBearerAccessTokenPayload(decoded) {
+    if (!decoded?.userId) return false
+    if (decoded.type === 'refresh') return false
+    return true
+  }
+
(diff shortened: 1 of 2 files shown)


advplyr:
I slimmed down the test to use a stub for the Database instead of building all the models, and removed the first few tests because they are covered in the below tests.

Thanks!

PR #5388, opened 2026-07-20
Enforce per-item access checks on bulk library download

mikiher:
## Brief summary

This fixes a permissions bypass where users could download library items via the bulk download endpoint (`GET /api/libraries/:id/download`) even when those items were blocked by explicit-content or tag restrictions.

## Which issue is fixed?

No issue.

## In-depth Description

- Fix missing per-item access enforcement in `LibraryController.downloadMultiple()` - the endpoint only verified library membership, not whether the user could access each requested item
- Load expanded library items via `findAllExpandedWhere()` so media metadata (explicit flag, tags) is available for permission checks
  - *Note: I chose to load the fully expanded items even though most of the expanded item is not needed. The main reason being to make it future-proof, and since the query seems not to be a very expensive one. I can change to a slimmer query if you think that makes more sense.*
- Call `req.user.checkCanAccessLibraryItem()` for every requested item before starting the zip; return 403 if any item is forbidden (including mixed allowed + forbidden requests)
- Add regression tests covering explicit items, tag-restricted items, mixed bulk requests (403), and allowed-only downloads (200)

## How have you tested this?

`npx mocha test/server/controllers/LibraryController.test.js`

### DIFF
--- a/server/controllers/LibraryController.js
+++ b/server/controllers/LibraryController.js
@@ -1432,13 +1432,16 @@ class LibraryController {
 
     const itemIds = req.query.ids.split(',')
 
-    const libraryItems = await Database.libraryItemModel.findAll({
-      attributes: ['id', 'libraryId', 'path', 'isFile'],
-      where: {
-        id: itemIds,
-        libraryId: req.library.id
-      }
+    const libraryItems = await Database.libraryItemModel.findAllExpandedWhere({
+      id: itemIds,
+      libraryId: req.library.id
     })
+
+    for (const libraryItem of libraryItems) {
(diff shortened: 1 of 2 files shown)


advplyr:
> * Note: I chose to load the fully expanded items even though most of the expanded item is not needed. The main reason being to make it future-proof, and since the query seems not to be a very expensive one. I can change to a slimmer query if you think that makes more sense.

I think it's okay for now. We limit downloads to 40 in the UI. Only issue may be podcasts since some podcasts can have thousands of episodes. But that is an issue we have to address in a lot of other places.

Thanks!

PR #5391, opened 2026-07-22
fix: prioritize exact book title matches in library search

### LINKED ISSUE #4484: [Bug]: Searching for two-letter book titles fails
### What happened?

When searching for "It" and "UR" by Stephen King, neither appear in the search results.

The book should have been the first result, since it is an exact match to the search term.

1. Add a book with a two-character title to an audiobook library (e.g. "It", "UR")
2. Search for the exact title of the book
3. The book is not in the results.

### Audiobookshelf version

v2.26.0

Docker

Linux

Firefox

### Logs

```shell

```

### Additional Notes

_No response_

Neon1201:
## Brief summary

Prioritises exact book title matches in library search so short titles such as "It" and "UR" are not excluded by the result limit.

## Which issue is fixed?

Fixes #4484

## In-depth Description

The existing search uses substring matching with a limited number of results, but exact title matches were not ranked first. This meant partial matches could fill the result list before the exact book title appeared.

This change adds a small SQL ordering expression so exact title matches are returned before partial matches. The existing substring search and permission filters remain unchanged.

This is currently a work in progress while I complete further validation and respond to feedback.

## How have you tested this?

- New library search regression tests: 6 passing
- Relevant existing tests: 107 passing
- `git diff --check`: passed

### DIFF
--- a/server/utils/queries/libraryItemsBookFilters.js
+++ b/server/utils/queries/libraryItemsBookFilters.js
@@ -1096,6 +1096,7 @@ module.exports = {
 
     const matchTitle = textSearchQuery.matchExpression('book.title')
     const matchSubtitle = textSearchQuery.matchExpression('book.subtitle')
+    const exactTitleMatch = `book.title COLLATE NOCASE = ${Database.sequelize.escape(query)}`
 
     // Search title, subtitle, asin, isbn
     const books = await Database.bookModel.findAll({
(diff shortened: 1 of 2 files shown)


server/utils/queries/libraryItemsBookFilters.js
```diff
@@ -1143,6 +1144,7 @@ module.exports = {
       ],
       subQuery: false,
       distinct: true,
+      order: [[Sequelize.literal(exactTitleMatch), 'DESC']],
```
nichwall: Can we update this to not use `Sequelize.literal`?
We currently only support SQLite databases, but want to support more databases in the future. There's still work to remove existing raw SQL queries throughout the database but we want to avoid adding additional raw SQL queries.

PR #5398, opened 2026-07-25
Service only change: Improve cover image cache-control headers

avanderhoorn:
## Brief summary

Cover images don't seem like they are caching well in the browser. The server capped the cache lifetime at 24 hours, and a couple of code paths sent no caching headers at all, so cover art kept getting downloaded again even when nothing about the book had changed. That's most noticeable in the library view: leave the tab open for a few hours or overnight, refresh, and every cover reloads from scratch. This fixes that.

## Which issue is fixed?

No linked issue. I found this myself while looking into slow loading covers, it wasn't filed as a separate GitHub issue first.

## In-depth Description

Every cover request already includes a `ts` query parameter set to the library item's `updatedAt` timestamp. That means a given URL only ever points to one specific image, since any real change to the cover or metadata updates `updatedAt` and changes the URL. Given that, there wasn't much reason to be conservative with caching.

Before this change, the server set `Cache-Control` to `private, max-age=86400` (24 hours) whenever `ts` was present, and set nothing at all when it was missing. A few places in the client build the cover URL without a timestamp (the podcast download queue and the player queue rows), so those covers never got cached in the browser at all.

Now, when `ts` is present the response is cached as `public, max-age=31536000, immutable`, a full year, since the URL can't ever point to stale content. `public` is safe here because this route is already unauthenticated (see the ignore patterns in `Auth.js`, added for mobile app cover and ebook access). When `ts` is missing, the response still gets a short `public` cache of one hour instead of nothing.

Net effect: once a cover has loaded once, it shouldn't need to be fetched again unless the book's metadata actually changes.

## How have you tested this?

Ran the existing `LibraryItemController` test suite under Node 20 and everything still passes. I also called `getCover` directly with a couple of fake requests, one with a `ts` param and one without, and confirmed the `Cache-Control` header comes back with the right value in each case.

## Screenshots

Not applicable, this is a backend header change only, nothing changed visually.

### DIFF
--- a/server/controllers/LibraryItemController.js
+++ b/server/controllers/LibraryItemController.js
@@ -399,7 +399,16 @@ class LibraryItemController {
       query: { width, height, format, raw }
     } = req
 
-    if (req.query.ts) res.set('Cache-Control', 'private, max-age=86400')
+    if (req.query.ts) {
+      // The ts query param is the library item's updatedAt timestamp, so this exact URL will
+      // only ever resolve to this exact image. It's safe (and unauthenticated/public) to cache
+      // it for a long time.
+      res.set('Cache-Control', 'public, max-age=31536000, immutable')
+    } else {
+      // No cache-busting timestamp was provided, so this same URL could resolve to a different
+      // image later on. Still cache briefly instead of not caching at all.
+      res.set('Cache-Control', 'public, max-age=3600')
+    }
 
     const libraryItemId = req.params.id
     if (!libraryItemId) {
--- a/server/controllers/LibraryItemController.js
+++ b/server/controllers/LibraryItemController.js
@@ -399,16 +399,8 @@ class LibraryItemController {
       query: { width, height, format, raw }
     } = req
 
-    if (req.query.ts) {
-      // The ts query param is the library item's updatedAt timestamp, so this exact URL will
-      // only ever resolve to this exact image. It's safe (and unauthenticated/public) to cache
-      // it for a long time.
-      res.set('Cache-Control', 'public, max-age=31536000, immutable')
-    } else {
-      // No cache-busting timestamp was provided, so this same URL could resolve to a different
-      // image later on. Still cache briefly instead of not caching at all.
-      res.set('Cache-Control', 'public, max-age=3600')
-    }
+    // ts is the library item's updatedAt, so this URL only ever resolves to one image - cache it for a long time
+    if (req.query.ts) res.set('Cache-Control', 'public, max-age=31536000, immutable')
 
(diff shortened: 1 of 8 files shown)


server/controllers/LibraryItemController.js
```diff
@@ -399,7 +399,16 @@ class LibraryItemController {
...
-    if (req.query.ts) res.set('Cache-Control', 'private, max-age=86400')
+    if (req.query.ts) {
+      // The ts query param is the library item's updatedAt timestamp, so this exact URL will
+      // only ever resolve to this exact image. It's safe (and unauthenticated/public) to cache
+      // it for a long time.
+      res.set('Cache-Control', 'public, max-age=31536000, immutable')
+    } else {
+      // No cache-busting timestamp was provided, so this same URL could resolve to a different
+      // image later on. Still cache briefly instead of not caching at all.
+      res.set('Cache-Control', 'public, max-age=3600')
```
Vito0912: I don't think this is a good way to handle that. If there are cover changes it will not show to the user, because it's cached. Especially if third party (web) clients do not use ts for some reason.

The year above I think is a very good idea.

Edit: Also these comments are overly redundant. This information should be in the PR description not permanently in the code.

Vito0912 (commented):
I am just a contributor no maintainer

server/controllers/LibraryItemController.js
avanderhoorn: Agreed... fixing.

avanderhoorn: Pushed a bigger change, but I believe this funnels through the `ts` where required so that the caching can work as intended. Let me know what you think

Vito0912:
The frontend gets currently rewritten and no frontend changes are being merged afaik. So while that probably are good changes, it will prevent this from being merged

avanderhoorn:
Well I can change the PR to not pass through the `ts` where it should and only optimize things when it is present???

Vito0912:
You can leave that. The following message is normally posted under a PR (by maintainer/core team) if it contains fronted changes. This was just a heads up (I didn't mean to say these changes are bad in anyway or anything):

> Thanks for opening a PR, but no frontend changes are being reviewed or merged at this time due to the frontend being migrated and fully rewritten from Vue2 to React.
>
> You are welcome to leave the PR open for inspiration or if someone wants to use your patch.

avanderhoorn:
In this case, if they could let me know which direction they would like... a) keep PR as is and merge, b) remove the ui change and merge, or c) not merging any of it... I'm happy to update, merge, or close. Just trying to be helpful, give back and fix something that has been bugging me a lot.

PR #5399, opened 2026-07-25
Add built-in Docker health check

VoltKraft:
## Brief summary

Adds a built-in Docker health check to the Audiobookshelf image, removing the need to define one separately in Docker Compose.

## Which issue is fixed?

No associated issue.

## In-depth Description

The Docker image now periodically requests the existing `/healthcheck` endpoint from inside the container.

  The health check:

  - runs every 30 seconds
  - uses a 5-second timeout
  - allows 30 seconds for application startup
  - marks the container as unhealthy after three consecutive failures
  - respects the configured internal `PORT`
  - uses Node.js directly, avoiding additional runtime dependencies such as `curl`

  This verifies that the Audiobookshelf HTTP server is responding successfully instead of only checking whether the Node.js process is running.

  Because the health check is defined directly in the Docker image, it works automatically with `docker run` and Docker Compose without requiring users to duplicate the configuration.

## How have you tested this?

 - Ran `docker build --check .` successfully with no warnings.
  - Built the complete Docker image successfully.
  - Started an Audiobookshelf container and confirmed that its health status changed to `healthy`.
  - Started a test container that returned HTTP 503 and confirmed that its health status changed to `unhealthy` after the configured number of retries.

## Screenshots

 Not applicable. This change does not affect the web client.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,5 +69,8 @@ ENV SOURCE="docker"
 ENV NUSQLITE3_DIR=${NUSQLITE3_DIR}
 ENV NUSQLITE3_PATH=${NUSQLITE3_PATH}
 
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
+  CMD ["node", "-e", "const http = require('http'); const request = http.get({ hostname: '127.0.0.1', port: process.env.PORT || 80, path: '/healthcheck' }, (response) => { response.resume(); process.exit(response.statusCode === 200 ? 0 : 1) }); request.on('error', () => process.exit(1))"]
+
 ENTRYPOINT ["tini", "--"]
 CMD ["node", "index.js"]


VoltKraft:
@Vito0912  Could you please explain in more detail what you don't like? Just giving it a thumbs-down isn't very helpful.

Vito0912:
Health checks only are needed if one actually does something with them. For this they can added. So there is no "duplication". ABS depends on nothing that would make them needed.
Also using node itself, when a proper, wget package is available, is "unique".

It's just my opinion and as I suspected AI written PR desc, not worth the time to spend on these imho.

Edit: So this comment does not contribute more to the fact than a thumbs down

VoltKraft:
Thanks for explaining.

My main motivation for the health check is monitoring, independent of whether other services depend on ABS. Monitoring tools can use the container health status to detect when the application stops responding.

As far as I understand, having it built into the image has no negative impact on users who don't use the health status, while providing a useful default for those who do.

I used Node.js because it's already available in the image, but I'm happy to switch to `wget` if that's preferred.

nichwall:
I don't think we want to add a health check to the docker file since almost nobody using ABS uses health checks, and if they are it's just a copy-paste of the compose from various websites/video guides people have made and they don't do anything with the health check.

This is something that can be configured by users on their own if they are actually going to use it. I think this would be better as a documentation update on the website.

Edit to add: related to https://github.com/advplyr/audiobookshelf/pull/4358

PR #5402, opened 2026-07-26
Fix loose files merging sibling book subfolders during scan

StPetr1:
groupFileItemsIntoLibraryItemDirs decided a file's library item group by walking from the shallowest directory down and merging into the first ancestor path that already had a registered group. That works for the intended case (a loose track sitting next to CD/Disc split subfolders belongs to the same book), but breaks when a loose file sits next to real book subfolders that are not CD splits: whichever file got processed first claims the shared parent directory as its group, and every subsequent file under a deeper sibling subfolder gets merged into that same group too, because the check only looks for "is some ancestor already a group" rather than "does this file's own directory deserve to be its own group."

Concretely: a series folder containing one stray audio file plus 30+ book subfolders got scanned as a single 1700+ file "book" instead of 30+ separate books, because the stray file's directory (the series folder itself) happened to get registered as a group before the subfolders were processed.

Fixed by keying each file's group off its own directory (walking up only for the CD/Disc-split case, which is now detected directly from the last path segment rather than inferred from a previous ancestor registration), so group assignment no longer depends on processing order and sibling subfolders can't be pulled into an unrelated loose file's book. Also fixed Step 4 (attaching non-media files like covers) to prefer the deepest matching group instead of the shallowest, which matters once Step 3 can produce nested groups.

Added a regression test for the loose-file/sibling-subfolder case.

Claude-Session: https://claude.ai/code/session_01GGgGjy1s9F3ETzDTvmzpad

### DIFF
--- a/server/utils/scandir.js
+++ b/server/utils/scandir.js
@@ -65,54 +65,44 @@ function groupFileItemsIntoLibraryItemDirs(mediaType, fileItems, audiobooksOnly,
     }
   })
 
-  // Step 3: Group media files (or non-media files if includeNonMediaFiles is true) in library items
+  // Step 3: Group media files (or non-media files if includeNonMediaFiles is true) into library items.
+  //
+  // Each file's own directory is its natural group (so every subfolder becomes its own book), except
+  // when the deepest path segment is a CD/Disc split, in which case the file belongs to the book one
+  // level up. Deciding the group from the file's own path - instead of walking up and merging into
+  // whichever ancestor happens to already be registered - keeps a loose file sitting next to book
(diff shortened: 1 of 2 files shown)


Vito0912:
To my knowledge thisnis intended behaviour and also documented in the ABS docs.
See here https://audiobookshelf.org/docs/documentation/libraries/book-library/directory-structure

Please check the docs on how to organise your library

nichwall:
Yes, the current behavior is intended functionality. Please review the documentation, every book must be in its own folder. This will not work with server settings that store metadata or other sidecar files in the same directory.

I think this can be closed.

PR #5409, opened 2026-07-28
Fix podcast rescan emitting stale episodes on item_updated

mikiher:
## Brief summary

After a podcast library item rescan, sync `media.podcastEpisodes` with the post-scan episode list so `item_updated` socket payloads include the current episodes (adds, removals, and updates).

## Which issue is fixed?

No Issue I know of.

## In-depth Description

During `rescanExistingPodcastLibraryItem`, episode changes were applied on a local `existingPodcastEpisodes` array (filter for removals, push for new episodes, save for modified audio). That array was never written back to `media.podcastEpisodes`.

`item_updated` is built from `libraryItem.toOldJSONExpanded()`, which serializes episodes from `media.podcastEpisodes`. Clients therefore kept the pre-rescan episode list until a full reload fetched fresh data from the API.

This change:

- Reassigns `media.podcastEpisodes = existingPodcastEpisodes` after episode processing so the socket payload matches the database.
- Tracks `hasEpisodeChanges` when episodes are added, removed, or updated so `wasUpdated` is true when only episodes changed (and `item_updated` is emitted when appropriate).

No API or socket schema changes—only correct data on the existing contract.

## How have you tested this?

1. Open a podcast item page in the web client (Vue or React).
2. Remove some existing episodes without hard delete
3. Run **Re-scan** from the item context menu in React, or the edit details tab in Vue.
4. Confirm the deleted episodes return without a full page reload (via `item_updated`).
5. Confirm a full page reload still shows the same episode list.

### DIFF
--- a/server/scanner/PodcastScanner.js
+++ b/server/scanner/PodcastScanner.js
@@ -58,8 +58,12 @@ class PodcastScanner {
     /** @type {AudioFile[]} */
     let newAudioFiles = []
 
+    // Track episode add/remove/update so item_updated includes the current episode list
+    let hasEpisodeChanges = false
+
     if (libraryItemData.hasAudioFileChanges || libraryItemData.audioLibraryFiles.length !== existingPodcastEpisodes.length) {
-      // Filter out and destroy episodes that were removed
+      // Filter out and destroy episodes that were removed.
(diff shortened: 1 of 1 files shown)


advplyr:
Makes sense, thanks!

PR #5417, opened 2026-07-30
Fix arrow keys hijacking player hotkeys while editing chapter timestamps (#2708)

### LINKED ISSUE #2708: [Bug]: While playing an audiobook and editing it's chapters, moving cursor left/right in timestamp alters current progress, up/down alters volume
### Describe the issue

While updating the timestamp of a chapter, using the cursor left/right, to access relevant digit, the progress time changes up/down in relation to left/right. Also the player volume alters up/down when using up/down to cycle through values of timestamp - seems like a mouse focus issue. :mouse: :crying_cat_face: 

1. Start playing an audiobook (also happens when paused)
2. Select "Edit chapters" dialog for the playing audiobook
3. Use mouse to select timestamp
4. Use left/right to navigate to digit
5. Progress time alters for each left/right action (also happens when editing a different audiobook)
6. Use up/down to cycle through timestamp digits
7. Player volume alters for each up/down action (also happens when editing a different audiobook)

### Audiobookshelf version

v2.8.0

Docker

SharonZhu0122:
Fixes #2708

## What was wrong
While editing a chapter's timestamp with the arrow keys, the same key presses also controlled the audio player at the same time: left/right seeked playback, up/down changed the volume.

## Root cause
The chapter timestamp editor (`TimePicker.vue`) is built from a plain `<div>`, not a native `<input>` element. `checkActiveElementIsInput()` in `layouts/default.vue` only recognises real input tags (`input`, `select`, `textarea`, etc.), so it never skips the global player hotkeys when the timestamp editor has focus. Both the picker's own key handling and the global hotkey handler ran on every key press.

## Fix
- Added a `data-hotkey-ignore` marker attribute to the timestamp editor's container.
- Updated `checkActiveElementIsInput()` to also treat elements with that marker as inputs, so global hotkeys are skipped while the timestamp editor is focused.
- Added `evt.preventDefault()` in the timestamp editor's own arrow-key handling, since the browser's default scroll behaviour was previously (accidentally) suppressed as a side effect of the bug, and became visible once the bug was fixed.

## Testing
Reproduced locally with a real audiobook and real chapters: confirmed the player's progress and volume changed while editing a chapter timestamp before the fix. After the fix, repeated the same steps and confirmed the player no longer reacts and the page no longer scrolls.

## AI disclosure
I used Claude (Anthropic) to help me navigate the codebase, debug my local dev setup, and review my changes, as I'm still learning full-stack development. All code was written and tested by me.

### DIFF
--- a/client/components/ui/TimePicker.vue
+++ b/client/components/ui/TimePicker.vue
@@ -167,12 +167,16 @@ export default {
       if (!this.focusedDigit || !evt.key) return
 
       if (evt.key === 'ArrowLeft') {
+        evt.preventDefault()
         return this.shiftFocusLeft()
       } else if (evt.key === 'ArrowRight') {
+        evt.preventDefault()
         return this.shiftFocusRight()
       } else if (evt.key === 'ArrowUp') {
+        evt.preventDefault()
         return this.increaseFocused()
       } else if (evt.key === 'ArrowDown') {
+        evt.preventDefault()
         return this.decreaseFocused()
       } else if (evt.key === 'Enter' || evt.key === 'Escape' || evt.key === 'Tab') {
         return this.removeFocus()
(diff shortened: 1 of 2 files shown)


nichwall:

PR #5441, opened 2026-08-08
Normalize typographic and fullwidth punctuation before search matching (#5436)

### LINKED ISSUE #5436: [Bug]: Search string and/or target strings are not normalized prior search matching
### What happened?

My whole library makes extensive use of the proper (smart) typographic quotes and Apostrophes (“double quotes”, ‘single quotes’, I’m), ellipses (…), em-dashes (—) in both embedded audio tags, metadata files (json and opf), _and_ in the whole folder structure path (only the m4b files are using straight quotes and placeholder symbols).

In addition I also use the following two symbols ？and ꞉ my folder paths for windows compatibility (but metadata tags and files use regular colon and question marks)

I've noticed that searching for titles by using the standard keyboard straight quote(s) does not find any of my books.

Having the search string and the target strings being normalized and/or stripped of symbols prior matching.

E.g., searching for **Now I'm a Demon Lord!** should match the book(s) **Now I’m a Demon Lord!** but it does not do this at the moment. As soon as i use **Now I’m a Demon Lord!** in search it does find it.

Normally quotes and apostrophes and other symbols are stripped and any dielectrics are normalized before matching. I do not know in what extend any normalizing is done, but I noticed the search issues with quotes and apostrophes specifically.

1. Have a book added with a title containing a quote or apostrophe and have all related paths and metadata make use of the (smart) typographic quotes or apostrophes.
2. Try searching for the the book specifically using part containing said apostrophe or quote using but using standard keyboard straight quotes or apostrophes.
3. Search result will fail
4. Try searching using the (smart) typographic quotes or apostrophes
5. Search will match

### Audiobookshelf version

v2.36.0

Docker

Linux

_No response_

### Logs

```shell

```

### Additional Notes

_No response_

SharonZhu0122:
Fixes #5436

## What was wrong
Search matching only normalized accented characters (via `unaccent()`). A search using a plain keyboard apostrophe or punctuation mark would not match a title stored with typographic ("smart") quotes, since the two are different Unicode characters. While testing, I also confirmed the same problem happens with fullwidth punctuation typed via a CJK input method (e.g. fullwidth `！` vs ASCII `!`), which matches a detail mentioned in the original issue as well.

## Fix
Added a small table of punctuation equivalents (typographic quotes and common fullwidth punctuation mapped to their plain ASCII form) and two helpers in `TextSearchQuery`:
- `normalizePunctuationJs()` normalizes the incoming search query in JS.
- `normalizePunctuationSql()` wraps the column expression with `REPLACE()` calls so the stored value is normalized the same way.

Both sides of the `LIKE` comparison are normalized consistently before matching, the same way accents are already handled.

## Testing
Reproduced locally with a test book titled using a typographic apostrophe: searching with a plain keyboard apostrophe returned no results before the fix, and returned the book after the fix. Also tested with a fullwidth exclamation mark from a CJK input method against a title using a plain `!`, with the same before/after result. Confirmed a normal lowercase search still works correctly (no regression).


ocgineer:
There is also full-width tilde and wave dash that might be used around japanese literature, not sure if that would ever be used in titles but you never know what people might use in romanized titles.

`～` -> `~` 
`〜` -> `~`

Also some other normal punctuations that i personally use in my titles

`…` -> `...`
`—` -> `--`

But honestly, wouldn't it be better to actually drop all of these non-letter symbols and then match a cleaned string against a cleaned database string? I believe that there should be some ready to use functions already that can normalize dialectics and drop all other symbols (or replace them with spaces), so you don't have to keep updating or building a remapper.

e.g. 
`Sòmé Title: with subtitle—and subsubtitle` ->`Some Title with subtitle and subsubtitle`

then it doesn't matter anymore either if an user doesn't use punctuations or symbols like forgetting the apostrophe in `it's`.

SharonZhu0122:
Thanks for the detailed feedback!

I've added both fullwidth tilde variants (`～` U+FF5E and `〜` U+301C → `~`), and updated the ellipsis/em dash rules to use single characters (`…` → `...`, `—` → `--`) instead of doubled ones, so they normalize correctly whether they appear once or repeated — matches your example. Pushed in e16434c8.

Re: dropping all non-letter symbols and matching against a cleaned string — that's a good point, and probably the more maintainable long-term approach. For this PR I'll keep the scope to the specific normalization cases you flagged, since a full redesign of the matching strategy would touch both the JS and SQL normalization paths and deserves its own separate discussion/testing.

ocgineer:
another possibility is to replace all symbols with spaces and then clean the strings to trim off double spaces to single spaces (simple regex replace i believe) this will also work for when users accidentally use 'two spaces' when searching, I actually encountered this just now and it even a double space would fail to match the book I am looking for!

For example i used `Under the  oak tree` (double space between the and oak) and it wouldn't find the book, when i removed the double space it found the books, but honestly these user input errors should be taken into account. I am not a programming wizard so i can only make assumptions, if you replace symbols with spaces and then replace \s+ with a ' ' space it should work much better in all kind of situations.

SharonZhu0122:
Thanks for catching this with a concrete example! You're right, this was a real gap — I've updated normalizePunctuationJs and normalizePunctuationSql to collapse repeated whitespace down to a single space (and trim the ends) after the punctuation substitutions, so "Under the  oak tree" now normalizes the same way as "Under the oak tree" on both the search query and the stored text side. Pushed in fc4b3b5b.

PR #5443, opened 2026-08-08
Add resumable uploads

keeper1978:
## What changed

- add an authenticated resumable-upload initialization endpoint
- accept file chunks with an explicit `Upload-Offset`
- persist upload sessions and partial files under the metadata temporary directory
- finalize completed uploads through a staging directory before moving them into the library
- serialize concurrent writes to the same partial file
- add manager tests covering resume, stale offsets, oversized chunks, and concurrent requests

## Why

Large audiobook uploads currently use one multipart request. If the connection drops, the client must restart the entire upload. These endpoints allow a client to query the saved offsets and continue from the last byte stored by the server.

The existing `/api/upload` endpoint is unchanged.

## API outline

- `POST /api/upload/resumable` initializes or resumes a manifest and returns per-file offsets
- `PATCH /api/upload/resumable/:uploadId/:fileId` appends a chunk at `Upload-Offset`
- `POST /api/upload/resumable/:uploadId/complete` validates sizes and atomically exposes the completed directory

## Validation

- `node --check server/managers/ResumableUploadManager.js`
- `node --check server/controllers/MiscController.js`
- `mocha test/server/managers/ResumableUploadManager.test.js` (4 passing)

This is opened as a draft to get maintainer feedback on the API shape before adding a client implementation, particularly whether a project-specific protocol or a standard such as tus is preferred.


keeper1978:
This PR addresses resumable client-to-server uploads, but that was not the problem we intended to solve. The actual issue is recovery of interrupted server-to-device audiobook downloads in the Android app, including stale Range offsets and files stuck at 100%. That client-side fix is proposed in advplyr/audiobookshelf-app#1953, so I am closing this unrelated server proposal.

PR #5445, opened 2026-08-09
Update jwa for node 25+ compat

slagiewka:
## Brief summary
This updates both the node_module as well as the vendored module.

Node 25 removes SlowBuffer (deprecated in v24). The upstream has a fix to stop using `buffer-equal-constant-time` [^1].

Closes #5433

[^1]:https://github.com/auth0/node-jwa/pull/52

## How have you tested this?

Running dev with node 26 no longer fails on the removed SlowBuffer.


silentnoodlemaster:
The tests didn't run for me (author of #5433) without updating mocha. Did you not need that or is it better to split into a separate PR?

slagiewka:
> The tests didn't run for me (author of #5433) without updating mocha. Did you not need that or is it better to split into a separate PR?

IIRC I focused on making the app running and I didn't run tests in this case. But of course you're right, `mocha` needs an upgrade for tests to pass, too.

Last time I took a look, all the dev dependencies were upgradeable (besides `nodemon` which be removed #5444). `chai` is already ESM-only, but if the project's baseline (assuming from Dockerfile) is node 20, then it supports `require(esm)` and all the test I run passed.
@nichwall given you've weighed in on #5460, would it be OK to bump these dev deps and remove `nodemon`? Of course this should affect the packaging the same way updating `axios` would.

nichwall:
I don't have an answer right now. I am not very familiar with how the packaging is done for the Debian and Windows binaries are done, but there have been problems in the past with upgrading dependencies.

I hesitate to say "yep we're ready to upgrade libraries or node" because I'm not sure if the dependency issues were due to the server or web client side.

slagiewka:
Lodged #5466 to take care of `mocha` more holistically.

PR #5454, opened 2026-08-11
Fix cron range with step expressions running at wrong times

### LINKED ISSUE #5453: [Bug]: Incorrect cron parsing/scheduling of expression combining ranges and steps, e.g. 0 1-23/2 * * *
### What happened?

When scheduling automatic podcast downloads, certain cron expressions are parsed incorrectly, leading to tasks being scheduled at the wrong times.
For example, the cron expression `0 1-23/2 * * *` is supposed to schedule a task at the start of every **odd** hour (01:00, 03:00, 05:00, ..., 23:00). However, it currently runs as if it was specified as `2-23/2`, i.e. at every **even** hour starting at 2 (02:00, 04:00, 06:00, ...).
The "next scheduled run" shown in the UI is, however, correct.

Cron jobs being scheduled correctly and being consistent with the UI preview of the next scheduled run.

1. Manually schedule podcast updates via cron expression, using a range and step specifier where the step size does not divide the start of the range, e.g. `0 1-23/2 * * *` or `0 4-10/3 * * *`
2. Observe the timing of runEpisodeCheck in the logs or the updated time in the "Look for new episodes after this date" field of the "Episodes" tab of the UI

### Audiobookshelf version

v2.36.0

Docker

Linux

None

### Additional Notes

I believe the bug stems from using/shipping code from an old version (around v3.0.0) of the `node-cron` package https://github.com/advplyr/audiobookshelf/commit/b7e546f2f5ee533974b23951638f3162f230f946, which first expands cron ranges to lists of possible values and then only keeps values divisible by the step size (e.g. `1-5/2` -> `1,2,3,4,5/2` -> `2,4`), while the correct behavior would be to begin at the start of the range and iterate in step size intervals until the end of the range is reached (`1-5/2` -> `1,3,5`).
This bug seems to be fixed in newer versions (v4.0.0+) of `node-cron`.

The mismatch with the UI is due to it using a different package to parse cron expressions, namely `cron-parser`, which interprets the expressions correctly.

Since newer versions of `node-cron` added the ability to preview the times of the next `n` scheduled runs via `task.getNextRuns(n)`, one might consider removing the separate UI parsing library and using `node-cron` exclusively to avoid any further mismatches between UI and backend.

lidonius1122:
## Brief summary

Cron expressions combining a range with a step (e.g. `0 1-23/2 * * *`, every odd hour) fire at the wrong times — the bundled cron library steps by absolute divisibility instead of from the range start. This affects podcast episode checks, automatic backups, and library scan schedules.

## Which issue is fixed?

Fixes #5453

## In-depth Description

In `server/libs/nodeCron/convert-expression/`, ranges are expanded to plain lists (`1-23/2` → `1,2,…,23/2`) before step values are applied. The step conversion then has no knowledge of the range start and filters by absolute divisibility (`value % step === 0`), keeping `2,4,…,22` instead of stepping from the start of the range (`1,3,…,23`) as standard cron does. The "next scheduled run" preview in the UI uses a different parser (`cron-parser`) and shows the correct time, so the UI and the actual job execution disagree.

The fix makes the range conversion step-aware: `first-last/step` expands directly to `first, first+step, … ≤ last`, consuming the step only when it terminates the list element. This matches vixie-cron semantics and node-cron ≥ 4.x, and also repairs related cases that were wrong before:

| Expression | Before | After |
|---|---|---|
| `0 1-23/2 * * *` | hours `2,4,…,22` | hours `1,3,…,23` |
| `0 4-10/3 * * *` | hours `6,9` | hours `4,7,10` |
| `5-59/10 * * * *` | minutes `10,20,…,50` | minutes `5,15,…,55` |
| `0 0 1-31/7 * *` | days `7,14,21,28` | days `1,8,15,22,29` |
| `0 0 * * Mon-Fri/2` | weekdays `2,4` | weekdays `1,3,5` |
| `0 1-5,10-20/2 * * *` | `2,4,10,12,…,20` | `1,2,3,4,5,10,12,…,20` (step binds to its range) |
| `0 7,1-5/2 * * *` | `2,4` (the `7` was lost) | `7,1,3,5` |
| `*/15 * * * *` | `0,15,30,45` | unchanged |
| `0 0-23/2 * * *` | `0,2,…,22` | unchanged |

Invalid steps (`/0`, `/2x`) are still rejected by `pattern-validation` exactly as before — the range is left unexpanded so the existing validation path catches it, which `cron.validate()` callers rely on. Expressions where the range start is divisible by the step (including all `*/n` forms) were already correct and are unchanged. Reversed ranges keep their existing no-wraparound behavior. This is not an edge case for one setup — it silently affects every schedule using a range whose start isn't divisible by the step.

## How have you tested this?

- New test file `test/server/libs/nodeCron/convertExpression.test.js` (11 tests): the cases above, invalid-step rejections, and `TimeMatcher` checks (for `1-23/2`, 01:00 fires and 02:00 doesn't). Removing the fix makes the new tests fail with exactly the buggy values.
- Full suite passes (365 tests).
- Differential check of several thousand generated expressions against the unpatched version: every behavior difference is either one of the intended fixes above or an invalid input that is still rejected identically.
- Reproducible by hand: `node -e "console.log(require('./server/libs/nodeCron/convert-expression')('0 1-23/2 * * *'))"` — hours field shows `2,4,…,22` before and `1,3,…,23` after.

## Screenshots

No web client changes.


nichwall:
Please use the PR template when opening PRs.

lidonius1122:
Sorry about that — both #5454 and #5455 are updated to follow the template now.

PR #5460, opened 2026-08-12
chore(deps): update axios

slagiewka:
## Brief summary

Updates the range as well as the locked version. `npm update axios` is ineffective because it's 0.x versioning (TIL actually).

Ideally `axios` gets replaces by native `fetch`. For now the update might mitigate some potential vulnerabilities.

npm audit before:
```
53 vulnerabilities (10 low, 8 moderate, 32 high, 3 critical)
```

npm audit after:
```
51 vulnerabilities (10 low, 8 moderate, 31 high, 2 critical)
``` 

## How have you tested this?

Tested a large podcast download in parts.


nichwall:
I don't think we want to focus on version bumps of dependencies right now. This has caused issues with packaging server releases in the past, and the reported axios vulnerabilities haven't affected how we use it. We are well aware that there are old library versions and need to make an update, which is a major reason why rewriting and migrating the web client rewrite is the priority (I realize this axios is in the server, but it's used in both parts and is just a general "libraries are out of date").

I did spend some time the past few days to entirely replace `axios` with `fetch`, but have not opened a PR for it yet. I will try and do a writeup of the changes and open a PR in the next few days, but either way it will still be a bit because there is a very large PR backlog we still need to get a handle on.

PR #5472, opened 2026-08-16
Refresh progress-filtered bookshelves on state changes

joshuapetersse:
## Summary

- refresh item and series bookshelves using progress filters when media progress crosses between not started, in progress, and finished states
- ignore playback updates that stay within the same progress classification to avoid unnecessary bookshelf resets
- discard stale lazy-load responses and fully reset rendered entity components when the filtered bookshelf is reloaded

## Problem

Media progress socket and user updates changed the Vuex state, but already-loaded progress-filtered bookshelf results remained cached. Items could therefore stay visible under the wrong progress filter, or fail to appear under the correct one, until the library view was manually reloaded.

## User impact

Progress-filtered item and series views now react immediately when playback or completion changes an item's filter membership.

## Validation

- `node_modules/.bin/mocha test/client/utils/mediaProgressState.test.js` — 4 passing
- `git diff --check origin/master...HEAD`
- manually verified locally against the reported library filtering workflow


nichwall:
Thanks for opening a PR, but no frontend PRs are being reviewed or merged at this time due to the web client migration and rewrite.

joshuapetersse:
Denied by admin.

PR #5480, opened 2026-08-18
Fix sync request duration overriding the library item duration

### LINKED ISSUE #5471: [Bug]: Providing custom duration in sync request behaves inconsistent
### What happened?

If you send the following payload in a session `sync` request:
```
{
  "currentTime": 10,
  "duration": 10
}
```

The item will be marked as finished no matter the actual duration. Having any value smaller than the configured time or percentage of marking an item as finished does result in the correct percentage being calculated. In this case the duration is omitted.

So sending 
```
{
  "currentTime": 10,
  "duration": 100
}
```

results in the actual percentage the book has. So e.g. around 0%, instead of 10%, which would be above.

I am not sure why duration is even accepted as part of the sync request as it is stored in ABS and in the session itself.

So in short: The markIsFInished gets calculated from the provided duration. The actual progress gets calculated from the provided duration saved in ABS

If a duration is present in ABS, omitting the duration

See above.

1. Open a session
2. Do a `/sync` request to that session with the payload mentioned above.

### Audiobookshelf version

v2.36.0

Docker

Linux

None

### Logs

```shell

```

### Additional Notes

_No response_

justadityaraj:
## Brief summary

The duration sent in a session `sync` request was taking priority over the duration the server already has, so a client could shorten it and trip the mark-as-finished check on an item that had barely started.

## Which issue is fixed?

Fixes #5471

## In-depth Description

`syncSession` built the progress payload like this:

```js
duration: syncData.duration || session.duration || 0,
currentTime: syncData.currentTime,
progress: session.progress,
```

`duration` came from the client, but `progress` on the very next line came from the session, which derives it from the library item. So the two disagreed whenever a client sent its own duration.

Mark-as-finished runs off `timeRemaining = duration - currentTime`, so the payload from the issue:

```json
{ "currentTime": 10, "duration": 10 }
```

leaves 0 seconds remaining and finishes an hour-long book at 10 seconds in, while the stored progress still reads ~0%.

Swapped the priority so the session duration wins, per the discussion in the issue. The sync payload is still used when the session has no duration, which keeps items that are no longer in the library syncing rather than reporting 0.

## How have you tested this?

Added `test/server/managers/PlaybackSessionManager.test.js` with three cases:

- a sync payload with `duration: 10` against a 3600s session sends 3600 to the progress update, not 10
- when the session has no duration, the sync payload is still used
- when neither has one, it reports 0

The first fails on master (`expected 10 to equal 3600`) and passes with the change. The other two pass either way, so they guard the fallback against regressing.

Full server suite goes from 354 to 357 passing, no failures before or after. Run on Node 20 to match CI. Prettier clean.

## Screenshots

No web client changes.


server/managers/PlaybackSessionManager.js
```diff
@@ -391,8 +391,11 @@ class PlaybackSessionManager {
     const updateResponse = await user.createUpdateMediaProgressFromPayload({
       libraryItemId: libraryItem.id,
       episodeId: session.episodeId,
-      // duration no longer required (v2.15.1) but used if available
-      duration: syncData.duration || session.duration || 0,
+      // The session duration comes from the library item, so prefer it over the
```
nichwall: This comment is unneeded and can be removed

justadityaraj: Removed, thanks.

PR #5507, opened 2026-08-27
Restore the mount prefix when handing requests to Next

mikiher:
## Brief Summary

This small change is in preparation of turning on dynamic base path for the new React client. It restores the configured base path before handling the request to Next.

## Which issue is fixed?

No issue

## In-depth Description

- Restore Express's mount prefix on `req.url` before handing Next the request, so Next sees the same base path the server was configured with
  - only do this if 
    - react client has basePath support (for backward compatibility) 
    - basePath is not empty
- Fail fast if `RouterBasePath` was never set, and log the path we are serving from (defensive)

## How have you tested this?
- 'npm run start-dev' works with ReactClientPath unset
- 'npm run start-dev' works with ReactClientPath set
- 'npm run start-dev' works with ReactClientPath set with basePath support ([client PR 249](https://github.com/audiobookshelf/audiobookshelf-client-react/pull/249))
  - with no `RouterBasePath` set
    - should serve under `/audiobookshelf` (like the Vue client)
  - with `RouterBasePath=''`
    - should serve under `/` (empty base path)
  - with `RouterBasePath='/abs'`
    - should serve under `/abs`

### DIFF
--- a/server/Server.js
+++ b/server/Server.js
@@ -158,6 +158,12 @@ class Server {
     }
 
     await Database.init(false)
+
+    if (typeof global.RouterBasePath !== 'string') {
+      throw new Error('[Server] RouterBasePath must be a string before serving requests')
+    }
+    Logger.info(`[Server] Serving from base path "${global.RouterBasePath || '/'}"`)
+
     // Create or set JWT secret in token manager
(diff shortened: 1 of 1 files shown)


advplyr:
Thanks!

PR #5508, opened 2026-08-27
Persist the playback position when the player pauses

### LINKED ISSUE #5492: [Bug]: Web player does not persist seek position when pausing immediately after seek
### What happened?

While an audiobook is playing, if I seek to a different position and immediately press Pause, the newly selected position is not persisted to the server.

The server continues to report the previous playback position. The new position is only persisted later after another progress sync, or if the player is closed.

I confirmed this by monitoring the active playback session and persisted media progress server-side.

When playback transitions from PLAYING to PAUSED, the web client should immediately persist the current playback position through the existing session progress sync endpoint.

Pausing should therefore be sufficient to save the position without requiring the player to be closed.

1. Start playing an audiobook in the web client.
2. While it is playing, seek to a clearly different position.
3. Immediately press Pause (before the normal periodic progress sync runs).
4. Check the saved/server-side playback position.
5. Observe that it remains at the pre-seek position.

Control case:
6. Keep the player paused.
7. Seek to another position.
8. Observe that this new position is persisted immediately.

### Audiobookshelf version

2.36.0

Docker

Linux

Chrome for Android

### Logs

```shell

```

### Additional Notes

I investigated this further and the two paths behave differently:

- Seek while PLAYING → immediately Pause: no session progress sync is sent and the server retains the old position.
- Seek while already PAUSED: the new position is immediately persisted through /api/session/:id/sync.

The current Vue client appears to stop the periodic progress interval on the PLAYING → PAUSED transition without first flushing the browser player's current position.

I tested a minimal local fix that calls the existing progress-sync path with the current player position on a genuine PLAYING → PAUSED transition before stopping the interval.

The focused regression tests passed, the production client/Docker build passed, and a locally patched 2.36.0 image passed the live reproduction test.

I could not verify whether the replacement React client already handles this differently. Would you prefer this to be addressed in the current Vue client, the replacement client, or both?

justadityaraj:
## Brief summary

Pausing after a seek does not save the new position. The server keeps reporting the pre-seek position until another sync happens or the player is closed.

## Which issue is fixed?

Fixes #5492

## In-depth Description

While the player is playing, `startPlayInterval` is the only thing writing progress to the server, and it syncs after 10 to 20 seconds of listening time. `playerStateChange` clears that interval on any state other than `PLAYING` without a final sync, so everything since the last sync is dropped, including a seek made just before pausing.

Seeking while already paused persists immediately, because `seek()` calls `sendProgressSync` when `!this.playerPlaying`. That is the reporter's control case and it is why the bug only shows up on the play-then-pause path.

The fix syncs on the transition out of `PLAYING`, reusing the same helper. It is guarded on the previous state so state changes that never involved playback (`LOADING`, `LOADED`) do not sync, and `sendProgressSync` already returns early below a second of drift, so pausing right after a periodic sync still issues no request. The listening time accumulated since the last sync is recorded rather than discarded, which also fixes a small under-count of listening time on every pause.

## How have you tested this?

The repository has no client-side test setup, so I drove `PlayerHandler` directly with a stub `ctx` and a stub player:

1. `playerStateChange('PLAYING')`, then move the stub player's `getCurrentTime()` to 4242 with 30s of accumulated listening time, then `playerStateChange('PAUSED')`.
2. On `master` no request is made. With this change a single `POST /api/session/<id>/sync` goes out with `currentTime: 4242`.
3. A following `playerStateChange('LOADED')` issues no further request, confirming the guard on the previous state.

`npm test` (mocha, `test/server`) is unaffected — the change is client-side only, and prettier reports the same pre-existing formatting warning on this file before and after.

## Screenshots

No visual changes.


nichwall:
Thanks for opening a PR, but no frontend changes are being reviewed or merged due to the [ongoing React rewrite and migration.](https://github.com/audiobookshelf/audiobookshelf-client-react)

justadityaraj:
Understood, thanks for the heads up — I'll leave it to you whether to close this.

One thing worth flagging while it's fresh, since the bug carries over: `audiobookshelf-client-react` has the same gap in `src/hooks/usePlayerHandler.ts`. The `stateChange` listener calls `stopSyncInterval()` on any non-`PLAYING` state without a final sync, so pausing still drops everything since the last interval tick.

It's actually a bit wider there — `seek()` in that file doesn't sync at all, where the Vue `seek()` at least persisted when the player was already paused. So the "seek while paused" case that works today would regress on the rewrite unless it's handled.

Happy to open this against the React client instead if that's useful.

PR #5509, opened 2026-08-28
Add client-settings endpoint

hypnotoad08:
## Brief summary

Adds a per-user `clientSettings` store so the React client can persist UI settings on the server instead of localStorage, where they aren't available at server render time.

## Which issue is fixed?

This is the server half of audiobookshelf/audiobookshelf-client-react#239, following the design @mikiher  proposed in https://github.com/audiobookshelf/audiobookshelf-client-react/pull/239#issuecomment-5423661949.

## In-depth Description

Settings live in the existing extraData JSON column at extraData.clientSettings[clientId]. They are written through PATCH /api/me/client-settings/:clientId and returned with the user object.

Updated from the original column-based approach after the discussion below.

Validation is structural rather than a key allowlist, so adding a setting does not require a server release.

An invalid key or value rejects the whole request instead of being skipped, so a client is never told a save succeeded when part of it was dropped. Sending null for a setting removes it.

Limits:

12 clients per user
64 settings per client
2048 bytes per client, 25600 bytes per user
Nesting depth 3
All sizes are measured in UTF-8 bytes.

## How have you tested this?

21 unit tests covering isolation from server-owned extraData, isolation between client bags, merge and removal, validation, UTF-8 byte accounting, prototype pollution, all 12 clients filling their bags within the total cap, and that the admin user route still rejects extraData.
Ran the server locally with the React client. Changed the bookshelf cover size, confirmed it held across a reload and a library switch, and checked the stored value in extraData.
Loaded the client on my phone and confirmed the mobile size persists separately from the desktop one.

## Screenshots

N/A


mikiher:
Hey I discussed this with @advplyr a bit. We prefer that you don't add a column. 
The user already has an extraData JSON field that you can use for storing client user data. 
You can extend the API with a PATCH that updates that field. Note, though, that once we add this API, other 3rd party client will likely start using it as well, so it requires proper validation and security.

hypnotoad08:
> Hey I discussed this with @advplyr a bit. We prefer that you don't add a column. The user already has an extraData JSON field that you can use for storing client user data. You can extend the API with a PATCH that updates that field. Note, though, that once we add this API, other 3rd party client will likely start using it as well, so it requires proper validation and security.

Ok I have updated the PR to use the extraData and added the requested validation.

server/models/User.js
```diff
@@ -912,6 +936,148 @@ class User extends Model {
...
+    }
+
+    const isArray = Array.isArray(value)
+    if (isArray || isJsonObject(value)) {
+      if (depth >= limits.maxDepth) return `exceeds maximum nesting depth of ${limits.maxDepth}`
+
+      if (!isArray) {
+        for (const key of Object.keys(value)) {
+          if (!User.clientSettingKeyPattern.test(key)) return `contains invalid key "${key}"`
+          if (User.unsafeObjectKeys.includes(key)) return `contains reserved key "${key}"`
```
mikiher: Better move this check before the line above it, since "__proto__" is also invalid, but you probably want the "contains reserved key" error. You'll also probably need to change the test.

mikiher (commented):
I checked this extensively (including all edge cases) using REST Client scripts.
Other than the small nit below, everything seems to be working as intended.
The limits make sense to me for now (we can always increase if needed)

server/models/User.js
hypnotoad08: Updated, thanks!

advplyr:
I think this is way overkill. I don't think we need to support passing in a client ID and having separate settings objects for each client. I think we should limit it to only allowing primitives, unless we have a specific need for supporting settings with object values?

I think it would suffice to have this endpoint accept an object that has settings keys and values that must be number, boolean or string. Is there a specific client setting that you think might need something more?

mikiher:
> I think this is way overkill. I don't think we need to support passing in a client ID and having separate settings objects for each client. I think we should limit it to only allowing primitives, unless we have a specific need for supporting settings with object values?
>
> I think it would suffice to have this endpoint accept an object that has settings keys and values that must be number, boolean or string. Is there a specific client setting that you think might need something more?

The client ids are IMO just a simple way to define namespaces to avoid potential name conflicts (in our case we also use them to loosely impose per-client storage limit, but I think that is less important)

Without them clients will likely add a client prefix/suffix to setting names to ensure conflicts don't arise, since we didn't provide that. That is ok, but I think not great.

As for object values - I don't think they add that much complexity, but sure, we can leave those out at least for now.

Vito0912:
I have several concerns with the current approach:

```js
static clientSettingsLimits = {
  maxBytesPerClient: 2048,
  maxBytes: 25600,
  maxKeys: 64,
  maxClients: 12,
  maxStringLength: 1024
}
```

### Limits

Why is `maxClients` limited to 12 at all?

Since `clientId` is self-asserted, a buggy or poorly behaved client could create multiple client IDs and consume the available namespaces, preventing another legitimate client from storing its settings.

Clients may also legitimately use different identifiers depending on platform, frontend, version, deployment, etc. A hard global limit on the number of client namespaces therefore seems unnecessarily restrictive.

I agree that limits in general are a good idea, but ABS already deals with significantly larger payloads in many other parts of the API. For client-specific preferences, I think substantially higher limits would be reasonable.

`2 KB` per client also seems quite small. It is certainly enough for a handful of basic preferences, but it becomes restrictive fairly quickly, especially because the limit applies to the serialized JSON object, meaning setting names and JSON overhead count toward it as well.

The combination of:

* 2,048 bytes per client
* 64 keys per client
* 1,024 bytes per string
* 12 clients per user

feels somewhat artificially constrained for data that is specifically intended to allow arbitrary third-party client settings.

### Including settings in every user response

My bigger concern is returning the entire `clientSettings` object as part of the normal user response.

I have mentioned this a few times on Discord already, but I think there should avoid continuously adding more data to the user object. That object is returned from a number of important and frequently used endpoints, and clients that have no interest in client settings will now receive the data anyway.

In the worst case this adds roughly 25 KB to the user payload.

25 KB by itself is obviously not enormous, but it is unnecessary overhead, particularly for endpoints such as `/api/me` and authentication-related responses that may be requested frequently.

This also creates an odd API model:

* A client that does not use server-side client settings receives all client settings anyway
* A client that does use them receives settings belonging to every other client as well
* There is no way to request only the namespace the caller actually needs.

Do we actually need the complete settings store to be included in every user response?

This somewhat contradicts my previous point about increasing the storage limits, but that is exactly why I think storage and retrieval should be separated. Increasing the amount a client can reasonably store should not mean increasing every user response by the same amount.

### Separate endpoint / API design

Since this is a new API, I think it would be better to introduce it in a way that provides a cleaner foundation for future changes.

For example, a dedicated endpoint for retrieving client settings:

```http
GET /api/me/client-settings/:clientId
```

and the existing:

```http
PATCH /api/me/client-settings/:clientId
```

would make the ownership of the data much clearer.

A client would then simply request the settings for its own namespace when needed.

For SSR, this should not fundamentally change much: the frontend can fetch the settings it needs explicitly. Regular clients can likewise fetch the user and client settings separately.

Alternatively, if ABS wants to continue using the `include` mechanism, `clientSettings` could be added or expanded on endpoints that return the user, ideally together with a `clientId`, so that only the requested namespace is returned.

For example, conceptually:

```http
GET /api/me?include=clientSettings&clientId=abs-web-react
```

The exact API does not matter as much

If retrieval is scoped to one client, the storage limits could also be reconsidered. The global response-size concern largely disappears, and limits such as `maxKeys` and `maxStringLength` could either be increased significantly or simplified. A reasonable per-client byte limit would probably be sufficient protection in most cases.

No limit would also open completely new options for clients of storing not only settings. If size is a concern, sessions as well as descriptions etc. also do not impose limits. So a DOS via that is already possible. At minimum it should be in the 100KB if not MB area imho.

### Overall

Sorry for the long comment. I understand that the larger API rewrite keeps getting pushed back, and that is completely reasonable.

That is also why I think it is important to be careful when extending the existing user response in the meantime. New features should ideally avoid making an already very broad user payload larger and harder to change later.

Ff the intention here is to provide generic storage for arbitrary third-party client settings, I think the API should treat those settings as a separate client-scoped resource rather than effectively making them another permanent part of the user object.

I appreciate the work that went into implementing this. My concern is mainly with the API model and the fairly restrictive limits around what is supposed to be generic client setting storage.

<sub>This was rewritten by an AI. Every meaning and the whole raw text was self written. I just have it rewrote based 1:1 on my input text (and yeah, also to make it a bit more friendly, because people say I sound a bit harsh and I do not want to say that). This should just be another view on that. Not saying current approach is wrong</sub>

---

Also on that:
> We prefer that you don't add a column.
The user already has an extraData JSON field that you can use for storing client user data.
You can extend the API with a PATCH that updates that field.

I think this is exactly going the opposite way again. Instead of normalizing the database model gets shaped into a NOSQL schema, with using a SQL database. Storing that as JSON blob in the user slows down the whole database again, everytime a user gets fetched and the extra data included as part of that.

advplyr:
Originally, when I was thinking that we should keep these in `extraData`, I was assuming it would be a few simple primitive key values. I can see why it's not that simple though.

I understand the choice to use client namespaces and it is logical. What I was hoping to avoid though is a lot of what Vito is highlighting. We're opening ourselves up to maintain a much larger feature.

If we are going to go this route of using client namespaces it no longer makes sense to use `extraData`. Mainly because it does not make sense to return all client settings with the user object.

This approach is more suitable for a new table, which I would like to avoid at this time. 

I'm back to my original thought of using cookies in the client

hypnotoad08:
@Vito0912, you're right about not returning the settings with every user response; that's a good call.

@advplyr I think that also addresses your concern. Dropping `clientSettings` from `toOldJSONForBrowser` and adding `GET /api/me/client-settings/:clientId` means the user object response stops growing and only the requested settings come back. `extraData` then works as storage without adding a new table.

@Vito0912 on the limits, though, I had AI run the numbers. With 20 users on SQLite, 200 reads averaged:

| | db size | user read |
|---|---|---|
| no client settings | 12KB | 0.305ms |
| current caps, worst case | 492KB | 0.306ms |
| 12 clients × 100KB | 23.8MB | 5.9ms |
| 50 clients × 100KB | 99.2MB | 23.2ms |

To be clear, I have not verified these against a real deployment. They are in-memory SQLite with synthetic settings data, so take the numbers with a grain of salt. The relative difference is what I would pay attention to.

At the current caps, the feature costs nothing measurable. The slowdown you were describing does show up, but it looks like it comes from larger limits rather than from using `extraData`. Scoping the get doesn't help there either, since Sequelize reads the whole column to serve any user query, so the read cost stays regardless of which namespace we return. The admin user list makes that worse because it loads every user.

On the client cap specifically, the idea is to keep the blob bounded. Even if you tried every vibe-coded iOS app out there, are you really going to have more than 12 clients installed that store settings on the server? The other risk is a client that generates a new id per version or per install, which writes forever and nothing reclaims it.

I did think about providing a list on the user profile page so people could remove settings for clients they no longer use, so they could free up space for other clients that write settings. That felt like too much scope for what this PR is trying to accomplish, but I'm happy to add it if that's the direction we want.

I'm also happy to raise the per-client cap if 2KB is too tight, but for primitives with no nesting it feels like plenty of space.

Vito0912:
Edit: advplyrs commented was not shown on the GitHub mobile app, only after sending (but now your comment is gone haha). Anyway I think exactly what advplyr said. Still leave the response below, so my points, if they got misunderstood have a other take on it.

But the comment by him summarise and points to the main problem way better than my text below. So ignore it :)

---

On the limits. Yeah, that was expressed poorly, and it will be marginal. The 100MB is ofc extreme (although in memory is ofc best case scenario), but it will creep trough every request (that fetches it).

The performance problem was more meant if the limit should be increased, which imho should be. I understand that the extraData is being used for storage, but wouldn't an extra table be more applicable?
As for scoping this was less for read performance and more for data transferred. The same as for performance, currently would be very small (although some clients call /API/me on sync so that still adds up, but the me response is already big, so yeah).
But anyway I do think that storing everything in the extraData is the wrong approach as described with the edit. 

If there are basically only strings allowed anyway, that's exactly what an SQL is for. Each setting is a row. Or at least a json row per clientId (probably would be faster, than rows and getting them into the json) column per setting in a setting table.

As for the general limit. 2kb is still a lot of data, yes. But (and this is also pure selfish) I can say that just for normal settings (and that would assume that I rename the settings keys to something like a, b, C), to be able to use the 64 key limit, I would hit it. And as ABS virtually limits nothing, I just wanted to raise the concern, why it is now limited and then so low that a normal client can hit it. And that would not include any stringified JSON values, which a client legitimately would need to save (e.g. shelf ordering).
But as mentioned a client setting would allow clients to add features that might need a few on of storage. The possibilities are endless there. This again is also selfish ofc.

And for a rough client, isn't that the issue now introduced? If you have one bad client it blocks every other client now, because of the 12 client limit. 12 should probably be enough, but it's 12 lifetime. And with that many client apps that is reachable very quickly. (Not even assuming one client takes it all)

As said, I just have that concern, because, don't get me wrong, but this whole new API is based on all the problems that exists. If extraData is used more in the future too, the problem just gets worse etc.

So TL;DR: Basically two concerns, why to build it upon the current database schema, that absolutely does not fit the use case and a second, why are limits as low (which are a result of the approach taken, which imho shows that this is an issue). But as said 2kb are better than nothing, so I will take it.

advplyr:
We're going to hold off on this for now and use client cookies as an interim solution until we switch focus to cleaning up the API. To do this properly we would need a new table for user client settings. We can revisit that later.

Thanks for digging into this @hypnotoad08. I didn't think this through all of the way when we originally landed on `extraData`, so that's my fault. It is helpful to work some of this out though.

PR #5510, opened 2026-08-28
Typescript initial migration framework

nichwall:
## Brief summary

This PR adds initial support for TypeScript.

## Which issue is fixed?

No specific issue.

## In-depth Description

This PR adds `tsc` and `@types/node` to the dev dependencies and builds all TS files into a new `dist-server/` directory for use by package targets.

The downside with this implementation is the dev dependencies need to be installed for any builds to be completed, which increases the build time and bandwidth usage slightly. The linux packaging only includes files directly referenced by the server and docker image explicitly removes the dev dependencies after performing the typescript compile, so no dev dependencies are included in the production build.

I experimented with having a separate package.json which handled these "build dependencies", but decided against it because this drives a much larger change and would be more difficult to maintain if the build process needs additional updates. I also considered having the typescript installed manually like yao-pkg is, but that also felt wrong from a DX standpoint when just working on the code and not running production builds.

No changes should be needed for https://github.com/mikiher/audiobookshelf-windows workflows because all dependencies are installed as part of the build process.

Other things to note:
- Removed `--unsafe-perm=true --allow-root` because both options are obsolete and print deprecation warnings (removed in npm v7)
- Readme updates
- `lazy_transform.js` uses the built in `buffer` instead of `crypto/util` to fix a new packaging warning introduced by this workflow change
- new warning `watcher/types.js` import of `ripstat` fixed

## How have you tested this?

I have done the following on my Linux machine after making just the build flow changes to make sure each server starts up correctly:
- ran the dev server
- ran `build-linux`
- ran `docker-amd64-local`

I then converted `parseNameString.js` to `parseNameString.ts` and confirmed the server still operates correctly for the same 3 builds with the new TS build step. I will open a follow-up PR tomorrow which includes this file conversion and a first draft of TS guidelines for the project.

## Screenshots

No screenshots, workflow change.


advplyr:
Is there a reason that we wouldn't be able to use the `allowJs` in tsconfig and avoid having the `buildServer.js` step?

https://www.typescriptlang.org/tsconfig/#allowJs

advplyr:
I did a bunch of tests and so far the `buildServer.js` script isn't necessary. There may be something you found that I haven't stumbled on yet though.

The `buildServer.js` script got ran on every file save when running the dev script. Which would delete `dist-server` folder and re-copy everything over. This change also adds `incremental` flag to tsconfig because we are still starting a new `tsc` on every file save, so that flag helps to not have to check every file again.

We can eventually switch to use `tsc --watch` at some point to be a little faster in dev, but it isn't too bad now in my testing. I think this is good to test on edge

advplyr:
I did some tests with typescript files and it works well. Will test on edge. Thanks!

nichwall:
Thanks! This is partially related to https://github.com/advplyr/audiobookshelf/pull/5550, so that one can also probably be merged or closed. I think we do want to start making the transition to ESM like it talks about either way and it's a pretty simple change to review.

PR #5519, opened 2026-09-01
Add external audiobook search via a configurable external service

danielkastner:
## Brief summary
Add external audiobook search via a configurable external service:
- Show BookFinder results in a new "External Audiobooks" section of the global search (max 5 entries)
- Clicking a result creates a task and sends the book data, user, library, server URL, task id and a callback token to the service
- The service updates the task through PATCH /public/external-search-tasks/:id; finished/failed end the task and raise a toast, unanswered tasks fail after 30 minutes
- Tasks with a userId are only emitted to and listed for that user
- Configure the service in the new admin section "External Services"
- All new texts use the i18n mechanism (en-us, de)

## Which issue is fixed?

Not related to an Issue

## In-depth Description

Add external audiobook search via a configurable external service

The global search box can now offer audiobooks that are not in the
library yet. The user input is checked against BookFinder and the
results are shown in a new "External Audiobooks" section, limited to
five entries.

Clicking a result creates a task for the requesting user and sends the
full BookFinder data, the user (id and username), the URL of this
Audiobookshelf instance, the active library (id and name), the task id
and a per-task callback token to the configured external service.

The service reports back through PATCH /public/external-search-tasks/:id
authenticated with that callback token: "running" updates the task
description, "finished" and "failed" end the task. A task that is never
updated is failed after 30 minutes so it cannot hang forever.

Task updates are private to the user that started the search. Tasks now
carry an optional userId; those tasks are emitted only to that user's
clients and are filtered out of GET /api/tasks for everyone else. A new
task_updated socket event reports progress of still running tasks, and
finished or failed external searches raise a toast with the task
description or error message.

The service is configured by admins in the new "External Services"
settings section: search enabled, service URL, authentication (none,
bearer token or basic auth), metadata provider and the Audiobookshelf
URL used for the callback. Token and credentials are only returned by
the admin-only endpoint, never to regular clients.

All user facing texts use the existing i18n mechanism (en-us and de).

## How have you tested this?

Added Unit-Tests


nichwall:
What is the goal of this PR? The Audiobookshelf server is intended as a server of local files. I don't think it make sense to have an additional online search for books that are not on the server. You can already fetch metadata from online sources for local books you have on the server.

Additionally, no front end changes are being reviewed or merged due to the ongoing React migration and rewrite.

danielkastner:
Yeah, I noticed the Hold on merging FE Changes. Fair Point @nichwall 
My Usecase is that I've some Data/Audiobooks stored on external Services/Storage and want to "pull them into ABS only upon Request". Hence the proposed Changes

nichwall:
You can already do this by making a remote file system available using something like rsync. I don't think it makes sense to introduce a larger change for what I'd consider a large change (how does the data model change, etc) for a niche use case.

The server is intended for local files, not requesting things outside of the database.

danielkastner:
All right, keeping this in my personal Fork instead

PR #5526, opened 2026-09-03
AudioMetadataManager: Fix logging error

mikiher:
## Brief summary

Fix erroneous logging of number of queued tasks in `AudioMetadataManager`

## Which issue is fixed?

No issue.

## In-depth Description

The modifed log message printed `[object object]` because it used the array object rather than it's length.

## How have you tested this?

Checked that the logger now prints the proper message.


advplyr:
Thanks!

PR #5545, opened 2026-09-09
Fix #4390: /api/me media progress missing displayTitle, coverPath and other display fields

### LINKED ISSUE #4390: [Bug]: User object returned from /api/me differes from User object returned from /api/users/userID
### What happened?

While writing a guide to using the API I found an inconsistency with the User object returned from the /api/me endpoint.

Expected Title and Subtitle info to be included with the User object from /api/me

1. Hit /api/me endpoint
2. Copare to /api/users/userID endpoint

/api/me
Image

/api/users/userID
Image

### Audiobookshelf version

v2.24.0

Windows Tray App

Windows

None

### Logs

```shell

```

### Additional Notes

Links from Vito to relevant controllers/methods.

`/api/me`
https://github.com/advplyr/audiobookshelf/blob/8ecc912c2db108d34f7e2a4158fdfdda0c78944f/server/controllers/MeController.js#L113

`/api/users/userID`
https://github.com/advplyr/audiobookshelf/blob/master/server/controllers/UserController.js#L86

Quaiyyum:
## What's the bug?

`/api/me` returns media progress items without `displayTitle`, `displaySubtitle`, `coverPath`, or `mediaUpdatedAt`. So if someone building something against the API using `/api/me`, can get progress entries with IDs and timestamps but no book title or cover.

The admin endpoint `/api/users/:id` already returns these fields correctly by doing a database join on the book and podcast models.

## What changed?

Made `getCurrentUser` in `MeController` async and added the same `mediaProgressModel.findAll` join that `UserController.findOne` already uses. Both endpoints now return consistent media progress objects.

One file changed: `server/controllers/MeController.js`

Fixes #4390


Vito0912:
Before this gets merged, there should be considered:
a) That this does make little sense. The user endpoint has it to easily show all books the user listened too. But for the user itself () it is doable either with using the user endpoint (admin) or the library items endpoint 
b) This will heavily increase data transmitted, data usage etc. for little gain.
c) This PR causes just more problems. All other endpoints still return the old data, including /progress. So this PR would bring in the real inconsistencies.

b is not even a problem of this PR specifically, but on how this data is transmitted.

I was mentioned in #4390, so idk why or what I said back then, but I so not think that is a bug, it's two different endpoints with two different responses

Quaiyyum:
Hi Vito, appreciate your response. I am closing in the PR in that case.

Vito0912:
Hey,

as written in the the other issue, I am not a maintainer. You can/should keep that open for an opinion from a maintainer, maybe they think it's a good idea. Above is just my personal opinion.

If that gets added, that should at least also apply to the progress endpoints too (again my opinion)

PR #5550, opened 2026-09-10
refactor: use extensionful requires

slagiewka:
## Brief summary
These are optional for CommonJS but are essential for ES modules. Having extensions explicit will make it easier to eventually use ESM syntax.

## In-depth Description

I saw #5510 and I wanted to propose a way to avoid the build step for TypeScript support.

node.js 22+ has TS support by means of type stripping. This means that w/o any extra dependencies, node can run server code written TS.

There's a catch - type stripped code needs to be compliant with node's expectations. TS stripped of types is ESM.

This is step 1 - comply with ESM's expectation for extensions in imports.

Next steps would be:
1. Convert `require`s to `import`s and convert other syntax to ESM
2. Start adding TS code in `.ts` files once audiobookshelf is running on node 22+ (could be soon).

## How have you tested this?

Running server, downloading and playing podcast episodes.


nichwall:
I'm not familiar with how node's TS support works. Does avoiding the build step still check the types? That is a big reason for switching to TS.

Vito0912:
It is no real/full TS support. It is basic support.
Node.js TypeScript support is quite limited. When running it (primarily) only strips TypeScript syntax and executes the JavaScript (more or less that is all what it does). It does not perform any type checking and it does not use any sort of tsconfig.json to my knowledge (but did not test again) and the docs [below] also state taht

That also means some TypeScript features are unavailable with plain type stripping, for example:

- enums (also a big one, although there are ways around that)
- parameter properties (convenience)
- import aliases
- decorators (Very big one e.g. for OpenAPI generators, most use that. Altough there is a bit story around decorators and decorators != decorators)
- tsconfig path aliases

For more information also see [here](https://nodejs.org/api/typescript.html) (especially Type stripping). If a full Typescript migration wants to be done, Node.js TS support is very inferior

(That does not mean that this PR has no ground)

slagiewka:
> I'm not familiar with how node's TS support works. Does avoiding the build step still check the types? That is a big reason for switching to TS.

It doesn't check types. You should still have lint step that runs `tsc` to confirm. You still have TS LSP at your disposal during development etc.

A project can still run either `tsc` as a build step for release (e.g. in `Dockerfile`) or `tsc --noEmit` to fail the pipeline.
Dev can run type stripping to be fast and not rely on external transformers.

Either way, node could still execute compliant code without the need to provide source maps etc in dev and prod.

---

On (slightly) unrelated note, should the project require external TS support, I think hooking projects like `tsx` or `@nubjs/loader` (if `nub` can't be used directly) might be better still.

Given invalid TS code:
```ts
function foo(id: number) {
  return
}

foo('bar')
```

tsc fails:
```
❯ npx typescript testts/testts.ts
testts/testts.ts:5:5 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'number'.

5 foo('bar')
      ~~~~~


Found 1 error in testts/testts.ts:5
```

However, no industry standard runtime or runtime-extender type checks on execution. `bun`, `deno`, `nub` and `tsx` all report no issues when executing the same code.

slagiewka:
> It is no real/full TS support. It is basic support. Node.js TypeScript support is quite limited. When running it (primarily) only strips TypeScript syntax and executes the JavaScript (more or less that is all what it does). It does not perform any type checking and it does not use any sort of tsconfig.json to my knowledge (but did not test again) and the docs [below] also state taht
> 
> That also means some TypeScript features are unavailable with plain type stripping, for example:
> 
>     * enums (also a big one, although there are ways around that)
> 
>     * parameter properties (convenience)
> 
>     * import aliases
> 
>     * decorators (Very big one e.g. for OpenAPI generators, most use that. Altough there is a bit story around decorators and decorators != decorators)
> 
>     * tsconfig path aliases
> 
> 
> For more information also see [here](https://nodejs.org/api/typescript.html) (especially Type stripping). If a full Typescript migration wants to be done, Node.js TS support is very inferior
> 
> (That does not mean that this PR has no ground)

Yeah, the support in node is not full TS feature set and it sometimes sucks. However, it can still take you a long way. Especially in cases like Audiobookshelf where the server has absolutely no TypeScript code and only new code will be subject to whether these features are necessary or not - the decision to have full TS support can be postponed and no build step will be required until that point.

nichwall:
Thanks! The reasons for a simple change make sense, but like I said earlier the big reason is to get type checking on everything (existing files and new files). We are also wanting to get OpenAPI specs generated from the TS (I have been using `tsoa` for this and it's been working really well), which doesn't work if we are still just using node.

It might make sense to at least start with converting to ESM like in this PR, but I would need to read up on this more before making any more decisions. I think the build step is still the correct way to go, since if we spend time getting everything converted without type checking and then enable type checking and other TS features, we might have to go through a lot of the process twice which just takes even longer on a project with limited time.

I've been working on a local branch for the last month or two to convert most everything to TS as a proof-of-concept for binary creation and full OpenAPI generation. I'm about 80% done with the code conversion (no behavioral or route changes, so the custom backend you were working on Vito would not be a drop in replacement).

PR #5557, opened 2026-09-11
Use node's built-in UUID v4 generator

slagiewka:
## Brief summary

Instead of relying on third-pary, node's stdlib can be used instead. `randomUUID` generates v4. `randomUUIDv7` would generate v7, but it requires node 24.16+.

This implementation acknowledges that the stdlib has entropy cache that's enabled by default and can be disabled with `disableEntropyCache`.

Unfortunately, there's no stdlib functionality to replace `uuid.version()` or `uuid.validate()` that are still crucial for some flows.

## How have you tested this?

Start server, create library, add and listen to podcasts.


nichwall:
I like the idea of removing a dependency, but we are already including the uuid dependency due to Sequelize, so not sure if this really buys us anything (and removes two functions that we use).

slagiewka:
> I like the idea of removing a dependency, but we are already including the uuid dependency due to Sequelize, so not sure if this really buys us anything (and removes two functions that we use).

To be clear, I'm not removing these two functions (`version` and `validate`) or the `uuid` dep from `package.json` ([it isn't there](https://github.com/advplyr/audiobookshelf/pull/5549)).

That's unfortunate to have this split, but at least we could move what ever we can off the supply chain. Now the runtime is our guarantee (given it's still supported). `uuid` is quite simple, the `8.3.2` package has [1 vulnerability](https://github.com/advisories/GHSA-w5hq-g745-h8pq) against no code in use here (at least not directly), so this is mostly noise until there are actual issues with the library.

nichwall:
I think we want to stick with the uuid package due to a lot of people just running ABS locally over HTTP, and the uuid package needs to be included due to Sequelize anyway.

Actual Budget had an issue a few months ago when they switched to using the built-in crypto library. https://actualbudget.org/blog/release-26.5.1
https://github.com/actualbudget/actual/issues/7715 is one of the main issue threads, but there are some duplicates

PR #5558, opened 2026-09-11
Add workflow to parse weblate commits

nichwall:
## Brief summary

This PR adds a workflow to make crediting translators easier for the release notes.

## Which issue is fixed?

No issue

## In-depth Description

This workflow uses the https://github.com/audiobookshelf/get-translator-credits action to make crediting translators in the release workflow much easier. It:
- Searches through commits between HEAD the most recent tag for the "Weblate commit" template
- Extracts the language and author information for the commit
- Filters out a unique mapping of author e-mails to language
- Uses the unauthenticated GitHub REST API to look up GitHub users against the commits and replaces author names with GitHub usernames, is present
- Prints the "More Strings Translated" to the action log in the `Show translator credits` for easy copying

The workflow can be ran as either a dispatch or will run automatically on commits to master which include `weblate` in the title. I have included the title on this PR so the output can be seen immediately.

## How have you tested this?

Tested in my fork of the server. My fork does not have the tags, so testing was done against the most recent 50 commits on `master`. The action output can be seen here https://github.com/nichwall/audiobookshelf/actions/runs/34640360980

## Screenshots

workflow update

### DIFF
--- a/.github/workflows/contribute-credits.yml
+++ b/.github/workflows/contribute-credits.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   release-notes:
+    if: github.event_name == 'workflow_dispatch' || contains(github.event.head_commit.message, 'weblate')
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -17,10 +18,8 @@ jobs:
           fetch-depth: 0
 
       - id: translator-credits
-        uses: audiobookshelf/get-translator-credits@v1
-        with:
-          token: ${{ github.token }}
+        uses: audiobookshelf/get-translator-credits@v1.0.0
 
       - name: Show translator credits
         run: |
-          printf '%s\\n' "${{ steps.translator-credits.outputs.credits }}"
+          printf '%s\n' "${{ steps.translator-credits.outputs.credits }}"
--- a/.github/workflows/contribute-credits.yml
+++ b/.github/workflows/contribute-credits.yml
@@ -1,0 +1,26 @@
+name: Prepare contributor credits
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  release-notes:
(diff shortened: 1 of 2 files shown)


advplyr:
Nice, the other thing I like to do is sort language alphabetical.
The other thing I came across when writing my own script a while back was when new languages are added. If you haven't already, you will want to look for a weblate commit of a new language since it has a different format than the others.

nichwall:
Thanks, I'll work on getting it updated to include "new language" added.

It does sort languages alphabetically and then sorts the contributors alphabetically by GitHub username, then alphabetically by author name.

advplyr:
I only put the new language added in the "Added" section of release notes when we have actually added it as a language option. Usually they are added in weblate before we have enough translations to put it as an option.

nichwall:
> I only put the new language added in the "Added" section of release notes when we have actually added it as a language option. Usually they are added in weblate before we have enough translations to put it as an option.

To clarify, are you wanting something different about the workflow for the "Added" section, or is it fine how it is once I add the extra checking for a brand new language in Weblate?

New languages in the settings usually have a dedicated PR, but the release notes have included additional strings even before it's available in the settings. For example, Slovak was added to the settings in [v2.33.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.33.0), but string additions show up in the release notes of [v2.30.0](https://github.com/advplyr/audiobookshelf/releases/tag/v2.30.0)

nichwall:
Okay, I have added a check for new languages from Weblate (the regex match is nearly identical, just needed an alternator on the first word).

I also added blacklist functionality for users, so advplyr and nichwall will not show up in the credits due to admin actions. I still print out the "blacklisted" commit information so it is easy to verify at a glance (see upper right of attached image).

image

advplyr:
I just meant that those weblate commits that are when languages are added have messed up my scripts previously because they are a different format.

PR #5565, opened 2026-09-14
Fix scroll listener leak in MultiSelect* components

### LINKED ISSUE #5564: [Bug]:  MultiSelect dropdown scroll listener not removed when component destroyed while open
### What happened?

When a MultiSelect / MultiSelectDropdown / MultiSelectQueryInput dropdown is open and its parent component gets destroyed (e.g. closing an edit modal via the X button without first clicking elsewhere to blur the input), the document-level scroll event listener added by setListener() is never removed. beforeDestroy() only removes the menu DOM element but never calls removeListener().

The scroll listener should be cleaned up whenever the component is destroyed, regardless of whether the dropdown was open or closed at that moment.

 1. Open any book's edit modal → Details tab.
 2. Click the Genres input to open its dropdown (this calls setListener())
 3. Immediately close the edit modal via the X button, without clicking elsewhere first
 4. beforeDestroy() fires while showMenu is still true, and removeListener() is never called — confirmed via console logging that document.removeEventListener('scroll', ...) is not invoked for this instance

### Audiobookshelf version

 v2.36.0

Built from source

macOS

None

### Logs

```shell

```

### Additional Notes

_No response_

zoujunhui1:
## Brief summary
Fixes a memory leak where MultiSelect / MultiSelectDropdown / MultiSelectQueryInput leave a document-level scroll listener attached after the component is destroyed while its dropdown is still open.

## Which issue is fixed?
Fixes #5564

## In-depth Description
The bug is in beforeDestroy() — it removes the menu DOM element but never calls removeListener(), so the scroll listener from setListener() just stays on document forever.

Dug into why: inputBlur() waits 50ms before setting isFocused to false. If the component gets destroyed before that timeout runs (closing the edit modal via the X button is fast enough to beat it), Vue kills the showMenu watcher along with everything else, so removeListener() never gets a chance to run.

Fix is just calling removeListener() directly in beforeDestroy() instead of relying on the watcher. removeEventListener on something that's not there is a no-op anyway so this is safe either way. Same copy-pasted pattern exists in all 3 files so fixed all of them.

## How have you tested this?
Opened a book's edit page, clicked into the Genres dropdown to open it, then immediately hit the X to close the whole modal without clicking anywhere else first.

Before the fix, beforeDestroy() ran while the menu was still open and removeListener() was never called - checked this with some temporary logging. After the fix it gets called every time regardless of whether the menu was open.

## Screenshots
 not a visual change


nichwall:
No updates to the front end are being reviewed or merged at this time. The react rewrite should be done in the next few weeks and the existing Vue frontend will be removed.

Issue #5535, opened 2026-09-07
[Enhancement]: Allow metadata providers to set the explicit flag on books


Chris-McQuiggan:
### Type of Enhancement

Server Backend

### Describe the Feature/Enhancement

Metadata providers currently have no way to set the explicit flag on books. The flag already exists and can be set manually, but it isn't populated during metadata lookups.

It would be great if the existing explicit flag on books could be populated automatically when fetching metadata. At least some providers already supply the necessary data — for example, the Audnexus API (https://api.audnex.us/books/), which powers the Amazon lookups, returns an isAdult field.

Ideally each provider mapping, including custom ones, would define how its adult content field maps to the explicit flag (like isAdult -> explicit), and just leave it unset when the provider doesn't report it.

### Why would this be helpful?

It would improve metadata quality, since providers already report this info and it would fill in flags that currently get left blank on import.
It would give parents of younger users some peace of mind, since they could rely on imported metadata instead of having to review and flag every adult title by hand.

### Future Implementation (Screenshot)

<img width="798" height="678" alt="Image" src="https://github.com/user-attachments/assets/57b577c0-31e2-4dbd-a29c-13603e507d80" />

### Audiobookshelf Server Version

v2.36.0

### Current Implementation (Screenshot)

<img width="803" height="687" alt="Image" src="https://github.com/user-attachments/assets/9e0e4c43-5aa7-43da-a882-903f66344f93" />

PR #5575, opened 2026-09-18
Set explicit flag from Audible match results

SharonZhu0122:
Related to #5535

Audible/Audnexus already returns whether a book is adult content via the `isAdult` field, but `Audible.js`'s `cleanResult` never read it, so the explicit flag was never populated when matching. Users had to set it manually every time even when the source clearly identified the book as explicit.

`explicit` was already in the list of fields the scanner applies from a match result (`detailKeysToUpdate` in Scanner.js), so this only needed one line added to `cleanResult`, following the same pattern already used for `abridged`.

**Testing:**
- Added unit tests covering isAdult=true, isAdult=false, and isAdult missing — all pass
- Ran the full existing test suite (357 tests) to confirm no regressions


advplyr:
Thanks!

PR #5576, opened 2026-09-18
Fix: enforce loggerScannerLogsToKeep so scan logs are pruned

jordanfelle:
## Brief summary

Scan logs in `metadata/logs/scans/` are never deleted. The `loggerScannerLogsToKeep` server setting (default 2) exists but nothing reads it, so the directory grows without bound.

## Which issue is fixed?

No existing issue that I could find.

## In-depth Description

`LibraryScan.saveLog()` writes a new `logs/scans/<date>_<uuid>.txt` after every library scan. `LogManager` only prunes daily logs (via `loggerDailyLogsToKeep`). `loggerScannerLogsToKeep` is defined, loaded and serialized in `ServerSettings.js` but is never consumed anywhere in `server/`.

On my instance an hourly scan produces a ~3.5MB log, and 16,000+ files (~35GB) had accumulated since Dec 2024.

This adds `LogManager.removeOldScanLogs()`, which keeps the newest `loggerScannerLogsToKeep` files (ordered by mtime, since the filename only has a date prefix) and unlinks the rest. It is called on startup (`init()`) and after each scan log is saved. Failures are logged and never throw, so they cannot break a scan.

## How have you tested this?

Added `test/server/managers/LogManager.test.js`, covering pruning down to the limit and doing nothing at or below it. I ran the test logic directly with a small describe/it shim because mocha would not start on Node 26 in my environment (yargs ESM error); I have not run the full `npm test` suite, so please rely on CI for that. I did not run a live server against the change.

## Screenshots

N/A (no client changes)


nichwall:
Hey, thanks for opening a PR, but this is a duplicate of https://github.com/advplyr/audiobookshelf/pull/5015

We have a really big backlog of PRs from focusing on getting the frontend rewrite completed.

PR #5512, opened 2026-08-28
Upgrade to node24

nichwall:
## Brief summary

This PR upgrades all workflows from node20 to node24

## Which issue is fixed?

No specific issue.

## In-depth Description

This PR depends on #5510 and is a single commit of 4db5816d6770f4a678170951b716c00ff7056541 based off of that branch. 

This PR changes all of the GH workflows from node20 to node24. GitHub actions are already running on node24 runners anyway, as the node 20 runner is deprecated. A warning appears on completed actions.

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

This PR will also need https://github.com/mikiher/audiobookshelf-windows/blob/master/.github/workflows/build-installer.yml to be updated to use node 24 @mikiher 

I think this is a good `edge` candidate to get more eyes on the major version upgrade to ensure nothing is broken. I believe @Vito0912 is already using node 24 on his fork and has not run into any specific issues, but not sure if that's only using Docker or also includes the Linux packaging.

## How have you tested this?

I confirmed the docker build, dev server, and linux package complete successfully on my local machine and did a cursory inspection of different pages.

Existing unit tests continue to run.

## Screenshots

N/A

PR #5578, opened 2026-09-18
Upgrade to node24

nichwall:
## Brief summary

This PR upgrades the server to use node 24.

## Which issue is fixed?

No issue.

## In-depth Description

This PR updates each of the build targets to use node 24 instead of the deprecated node 20. This PR supersedes https://github.com/advplyr/audiobookshelf/pull/5512 now that #5510 was merged. This PR changes all of the GH workflows from node20 to node24. GitHub actions are already running on node24 runners anyway, as the node 20 runner is deprecated. A warning appears on completed actions. https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

This PR also fixes the `DEP0187` warning in the build log, which was a result of the old Vue2 client trying to start a socket server. The socket server is already provided by the backend and is not needed in the web client.

## How have you tested this?

Successfully built the following package.json targets using Ubuntu on AMD:
- `build-linux`
- `docker-amd64-local`. Web client worked in local testing
- `build-win`
- `build-win-no-compress`

I did not test `docker-arm64-local` due to cross compilation. I do not currently have a Windows computer set up to test the server on.

## Screenshots

N/A, backend only


advplyr:
I tested on Windows. I also snuck in an update to run the linter in workflows.
Thanks!

PR #5586, opened 2026-09-19
fix: resolve ReferenceErrors in scandir.js under strict mode

binyaminyblatt:
## Description
Fixes `ReferenceError` crashes occurring during library scans and media directory parsing under strict mode (`"strict": true`). 

### Changes Made:
- **`getBookDataFromDir`**: Explicitly declared `series` and `author` using `let` to prevent them from leaking into the global scope / throwing reference errors in strict mode.
- **`getPublishedYear`**: Declared `pattern` with `const` to ensure proper block scoping.

## Related Issue / Context
Addresses strict mode regressions where implicit global variable assignments in legacy utility files caused unhandled rejections and fatal server crashes during background library scans.

sorry i used ai to write this message


Vito0912:
More info here: https://discord.com/channels/942908292873723984/1550959828044746792

The typescript changes added strict, which does not allow the ambiguous undeclared varaiables

nichwall (approved):
This was discussed in Discord prior to opening the PR, looks good.

server/utils/scandir.js
```diff
@@ -150,8 +150,8 @@ function getBookDataFromDir(relPath, parseSubtitle = false) {
   const splitDir = relPath.split('/')
 
   var folder = splitDir.pop() // Audio files will always be in the directory named for the title
-  series = splitDir.length > 1 ? splitDir.pop() : null // If there are at least 2 more directories, next furthest will be the series
-  author = splitDir.length > 0 ? splitDir.pop() : null // There could be many more directories, but only the top 3 are used for naming /author/series/title/
+  var series = splitDir.length > 1 ? splitDir.pop() : null // If there are at least 2 more directories, next furthest will be the series
```
Vito0912: I mean it doesn't really matter, but why var and not const?

var is outdated

Vito0912 (approved):

server/utils/scandir.js
binyaminyblatt: consistency

binyaminyblatt: im in middle of a pr for eslint and it found sum more issues like this

advplyr:
If you're still updating this PR let me know. I cleaned up scandir

binyaminyblatt:
im trying to address any additional vaeables like this but there are some i cant get like 
audiobookshelf\server\utils\podcastUtils.js
  72:27  error  'value' is not defined
 audiobookshelf\server\managers\BinaryManager.js
  379:62  error    'binaryPath' is not defined

binyaminyblatt:
other then that i fixed all the ones i could figger out

advplyr:
okay i will fix those, thanks

PR #5587, opened 2026-09-19
feat: add ESLint configuration and npm lint scripts

binyaminyblatt:
## Description
This PR introduces **ESLint (v9 Flat Config)** to the project to establish consistent code quality and catch common runtime errors and syntax bugs across the backend and test suites.

### Key Changes
- **ESLint Flat Configuration (`eslint.config.mjs`)**:
  - Configured `@eslint/js` recommended rules with Node.js CommonJS and ES2022 environment settings.
  - Added Mocha and Chai test globals for test files in `test/**/*.js`.
  - Integrated `eslint-config-prettier` to avoid conflicts with existing `.prettierrc` formatting rules.
  - Sane rule defaults (e.g., ignoring unused variables prefixed with `_`, allowing empty catch blocks).
- **Exclusions**:
  - **`client/**`**: Completely excluded from linting as the legacy client is planned for deprecation.
  - **`server/libs/**`**: Excluded vendored third-party libraries.
  - Excluded build output directories (`dist/**`, `dist-server/**`, `coverage/**`, `**/.tsbuildinfo`).
- **NPM Scripts (`package.json`)**:
  - Added `npm run lint` (`eslint .`)
  - Added `npm run lint:fix` (`eslint . --fix`)
---

---

## How to Test
1. Run `npm install`
2. Run `npm run lint` to inspect server and test code.
3. Run `npm run lint:fix` to automatically resolve fixable linting issues.


advplyr:
I decided to go a different direction and use a much simpler eslint config. https://github.com/advplyr/audiobookshelf/commit/ad8b2350585dd8f1b0ddec541bc734716cab3d1d

I couldn't see how `eslint-config-prettier` was doing anything useful. 
The `globals` dep seemed unnecessary overhead for the small number we actually need. Also, the [eslint docs](https://eslint.org/docs/latest/use/configure/language-options#predefined-global-variables) say that the ECMAScript standard built-in globals are automatically enabled.

Some of the `@eslint/js` recommended rules could be useful, but I'd prefer adding those as needed rather than including a dep for it.

Let me know if you think I missed anything in that commit. I wasn't sure if there was a particular reason you chose not to use the latest version of `eslint`.

Thanks for digging into this

binyaminyblatt:
The whole point of this was simply to help us eliminate the problem where we had undeclared variables and stuff like that so I just had AI spit out a standard configuration I wasn't really thinking that much about it the whole PR request was trying to drum up some support for adding static analysis and a linter so that we don't have the same issue as my previous issue again

Issue #3562, opened 2024-10-29
[Bug]: Users are able to see items they do not have access to in filter list


nichwall:
### What happened?

Users are able to see items in the filter list even if they don't have access to the list, such as "Authors", "Series", "Tag", etc. The function `getFilterData` creates the filter based on the library without taking user access into account.

I limited this user to only view the "Science Fiction" tag, which for example should not show "Fantasy" or "Brandon Sanderson".
![image](https://github.com/user-attachments/assets/b58200ed-c122-4a63-b0c7-7f3af1f74724)
![image](https://github.com/user-attachments/assets/e1a6596a-3e11-4b0e-92f2-0ca78e0ef484)

### What did you expect to happen?

Filter information by user similar to other parts of the API.

A potential solution is to add a user parameter to filter the data and have user specific caches for the filter data for each library.

### Steps to reproduce the issue

1. Create a user
2. Limit user access by tags
3. Use the "filter" dropdown to see tags they shouldn't be able to see, along with other fields

### Audiobookshelf version

v2.16.1

### How are you running audiobookshelf?

Built from source

### What OS is your Audiobookshelf server hosted from?

Linux

### If the issue is being seen in the UI, what browsers are you seeing the problem on?

None

### Logs

```shell
[2024-10-28 18:40:29.782] DEBUG: Filter data: {
  authors: [
    {
      id: 'b059d5e3-02a6-457e-be40-bb12a908fddd',
      name: 'Brandon Sanderson'
    },
    { id: '60841878-f8c7-4bd1-a69b-b4ad8c596d60', name: 'Comics' },
    {
      id: '4c5deaa4-5a16-46cd-8727-2cca67b9b764',
      name: 'Isaac Asimov'
    },
    {
      id: '87943f2e-dae8-40d3-abcd-540e0516c6f2',
      name: 'Isaac Asimov/Kevin T. Collins'
    },
    {
      id: 'cb568b56-004e-41c0-bd8b-bcaa81620934',
      name: 'Isaac Asimov/Larry McKeever'
    },
    {
      id: 'c08f18b7-5eb2-41af-b09e-89aad87e84e3',
      name: 'Isaac Asimov/Robert Fass'
    },
    {
      id: 'd031f7c3-775b-45da-9d4e-a8387a2deaae',
      name: 'Isaac Asimov/Scott Brick'
    }
  ],
  genres: [ 'Audiobook', 'Fiction', 'Science Fiction' ],
[...]

PR #5590, opened 2026-09-20
Fix/filterdata respect user permissions

Altonymous:
## Brief summary

Filter menus are built from the whole library, so a user restricted by tags sees authors, series, narrators, publishers and genres from items they can't open.

## Which issue is fixed?

None open that I could find. Same class of leak as the narrators PR.

## In-depth Description

`getFilterData` has no user parameter, and its cache is keyed by library, so it can't hold per-user results.

Rather than changing the cache key and multiplying cached copies, restricted users take a separate uncached pass using the existing `getUserPermissionBookWhereQuery` / `getUserPermissionPodcastWhereQuery` helpers. Everyone else keeps the current cached path untouched. Authors and series are collected from the filtered rows instead of queried per library, so hidden items can't contribute.

Both call sites pass the user: the standalone `/filterdata` endpoint, and `findOne` with `?include=filterdata`, which is the one the web client actually uses.

Worth noting this matters for families using tags to keep adult material away from kids. The items are hidden, but an author or series name in a dropdown is enough to send a curious kid looking elsewhere.

## How have you tested this?

On 2.36.1 with a 15 book library. 13 books tagged `restricted`, 2 untagged. Non-admin user set to block the `restricted` tag.

Before, that user's filter menus listed all 7 authors, all 3 series and all 4 publishers in the library.

After, they list only the 2 authors, 1 series and 1 publisher from books they can access. Admin view unchanged.

Timing on that library: 2.4-2.6ms for the uncached filtered pass versus 2.3ms cached, so no practical difference at this size. On a very large library the filtered pass would be slower, and could be cached under library id plus a permission signature if you'd prefer that.

## Screenshots

Can add before/after of the filter dropdown if useful. The menus are unchanged structurally, only their contents.


nichwall:
The uncached path can take a long time (some users have provided logs showing longer than 60 seconds) for large libraries due to this information being stored in JSON arrays. I think this needs to have caching enabled instead of relying on small library testing.

I haven't gone through this other than briefly reading the description.

Edit to add reference to https://github.com/advplyr/audiobookshelf/issues/3562

